PR #6405, opened 2022-07-19
Adds in UI-side validation for Ingress Form plus additional fixes.

### LINKED ISSUE #4575: Additional UI-side validation for Ingress form
Original description below. Per the comments, there is no backend issue, but the ask is for UI to have additional validation to prevent these requests from going through until data is valid.

**Rancher Server Setup**
- Rancher version: 2.6.2
- Installation option (Docker install/Helm Chart):  Docker install
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc):  RKE1
- Proxy/Cert Details:  self signed

- Kubernetes version:   v1.21.6
- Cluster Type (Local/Downstream): Local installation via docker, 1x Rancher, 2x Worker Nodes

**Describe the bug**
I can't create an ingress.  No matter what value I use for "Name" I get the message:

`Ingress.extensions "example.com" is invalid: [spec.rules[0].http.paths[0].path: Invalid value: "": must be an absolute path, spec.tls[0].hosts[0]: Invalid value: "": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')]`

**To Reproduce**
Try to create an ingress and use any name.  Click create and the error message appears. 

**Result**
The error `Ingress.extensions "whatever" is invalid` and the ingress is not created.

**Expected Result**
Create the ingress with the name given.

**Screenshots**
With example.com (like the error message even gives as a valid example):
image

With 'something-else':
image

With 'foo':
image

**Additional context**

### LINKED ISSUE #6319: [UI] - Removing Path in Ingress resets the page
Rancher v2.6.6

Step to reproduce : 
  1. Create an Ingress
  2. Click on "Add path"
  3. Input a path "/foo"
  4. Click on "Add path"
  5. Input a path "/bar"
  6. Click on "Add path"
  7. Input a path "/loo"
  8. Click on "Remove" on the "/bar" path

=> Page reload

Expected : Not to reload, just remove the Path

Please note it also happend with existing Ingress - the pattern to reproduce is to remove a "Non-last Path"

Sean-McQ:
### Summary
Fixes #4575
Fixes #6319 

- Add in validator injection to form-validation mixin 
- Add in "allTabsExtraContent" slot to "tabbed" for messaging that should span all tabs 
- Error component optionally displays as banner
- InputWithSelect can now show validationMessage on text-box without a label
- LabeledSelect can now show tooltip and validationMessage at the same time
- Select component can now show validation message
- Adding and removing rules and rulepaths displays correct content
- Changing "Target Service" in Default Backend tab would visually clear "Port" field but did not clear out the form's payload value. This is no longer the case.
- Removing ingress path does not reload page (fixes #6319)
- Ingress form shows proper validation messages for the following states: 
  - Invalid "name" field
  - Invalid request host 
  - Invalid rule path
  - Invalid rule target service
  - Invalid rule port
  - Invalid default backend target service
  - Invalid default backend port
  - Either rules or default backend must be present
- Ingress form is validated and "Create" button is disabled if payload is not validated as correct.

### Technical notes summary
It was possible by creating and removing both rules and rulepaths to make the form payload for rules and rulepaths to no longer be in-sync with the values shown in the form. This change should correct for that.


shell/edit/networking.k8s.io.ingress/Rule.vue
```diff
@@ -41,30 +50,55 @@ export default {
       } else {
         delete out.host;
       }
-      this.$emit('input', out);
+
+      this.$nextTick(() => {
+        if ((this.paths?.length === 1 && this.pathObjectIsEmpty(this.paths[0])) || this.path?.length === 0) {
```
jordojordo, line 55: For the condition - `this.path?.length === 0` - should it be `this.paths?.length === 0`?

jordojordo (changes requested):
A couple of changes, but otherwise looks good.

shell/assets/translations/en-us.yaml
```diff
@@ -4905,6 +4909,7 @@ validation:
   noSchema: No schema found to validate
   noType: No type to validate
   number:
+    requiredInt: '"{key} must be integer'
```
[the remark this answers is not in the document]
Sean-McQ: Nice catch, thanks!

shell/components/form/Error.vue
```diff
@@ -1,17 +1,23 @@
 <script lang="ts">
 import Vue, { PropType } from 'vue';
+import Banner from '@components/Banner/Banner.vue';
```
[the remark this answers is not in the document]
Sean-McQ: rebased, confirmed works on my machine! 👍

shell/edit/networking.k8s.io.ingress/Rule.vue
Sean-McQ: worst kind of broken, the kind that looks like it works... fixed!

shell/assets/translations/en-us.yaml
```diff
@@ -4909,6 +4913,7 @@ validation:
   noSchema: No schema found to validate
   noType: No type to validate
   number:
+    requiredInt: '"{key}"" must be integer'
```
jordojordo, line 4916: I suspect auto-complete is messing with you here?

jordojordo (changes requested):
One small thing, sorry!

jordojordo (approved):

jordojordo (approved):

PR #6408, opened 2022-07-19
delay showing and removing socket disconnects

### LINKED ISSUE #5093: Add growls to notify user when connection is lost
QA has noticed that when self-signed certs are used for Rancher, the UI doesn't update correctly. This is often caused because when you use self-signed certs, Google Chrome can decide that the cert is invalid after a period of time, causing connection problems.

Vince said the Ember UI had growls to notify a user when the connection is lost. A similar growl should be added to the dashboard so that users will be warned that the UI might not function correctly due to a connection problem.

mantis-toboggan-md:
### Summary
Fixes #5093 

The socket disconnect notification wont be dispatched until the socket is disconnected for 10 seconds. Notifications wont be removed until they have been visible for at least 3 seconds. These numbers are arbitrary; we can be even more conservative about sending growls if they still seem excessive.

I tested this by upgrading my rancher instance and watching both the locally-running and its actual ui side by side.

### DIFF
--- a/shell/plugins/steve/subscribe.js
+++ b/shell/plugins/steve/subscribe.js
@@ -16,6 +16,12 @@ import { escapeHtml } from '@shell/utils/string';
 export const NO_WATCH = 'NO_WATCH';
 export const NO_SCHEMA = 'NO_SCHEMA';
 
+// minimum length of time a disconnect notification is shown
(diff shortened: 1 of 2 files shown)


catherineluse (approved):
LGTM

Issue #6376, opened 2022-07-14
Cannot update to RKE2 1.21.12 for Rancher from migration purposes


gaktive:
Internal reference: SURE-4797
Reported in 2.6.5

Issue description:

The RKE2 release 1.21.12 is not available through Rancher.

Business impact: 
Cannot update to latest available RKE2 version

Troubleshooting steps:

```
$ curl -sL https://raw.githubusercontent.com/rancher/kontainer-driver-metadata/dev-v2.6/data/data.json | jq -r '.rke2.releases[] | select(.maxChannelServerVersion == "v2.6.4") | .version'
> v1.21.4+rke2r2
v1.21.5+rke2r1
v1.21.5+rke2r2
v1.21.6+rke2r1
v1.21.7+rke2r2
v1.21.8+rke2r2
v1.21.9+rke2r1
v1.21.10+rke2r2
v1.21.12+rke2r1
v1.21.12+rke2r2
```

```
curl -sL https://raw.githubusercontent.com/rancher/kontainer-driver-metadata/dev-v2.6/data/data.json | jq -r '.rke2.releases[] | select(.maxChannelServerVersion == "v2.6.5") | .version
> 
```
 
```
curl -sL https://raw.githubusercontent.com/rancher/kontainer-driver-metadata/release-v2.6/data/data.json | jq -r '.rke2.releases[] | select(.version == "v1.21.4+rke2r2") | .maxChannelServerVersion'
v2.6.4
curl -sL https://raw.githubusercontent.com/rancher/kontainer-driver-metadata/dev-v2.6/data/data.json | jq -r '.rke2.releases[] | select(.version == "v1.21.4+rke2r2") | .maxChannelServerVersion'
v2.6.4
``` 

Repro steps:
- Install RKE2 1.21.9
- Try to update through Rancher to RKE2 1.21.12

Workaround:
Use your own modified data.json from the settings.

Actual behavior:
Only default version are shown

Expected behavior:
Newest RKE2 version from the Matrix should be there

Additional notes:
Per https://github.com/rancher/rancher/issues/37547, this is expected but for people migrating from CaaSP to RKE2, this is a scenario that needs to be considered.

From a backend engineer:
> These versions restrictions are implemented in the Dashboard.  My suggestion would be to add an option to show deprecated versions in the ui that is disable by default.  Users should be able to edit the cluster as yaml and modify the k8s version to a deprecated version, no backend changes are required to support provisioning deprecated versions.

[...]

PR #6410, opened 2022-07-20
Allow user to select deprecated K8s patch versions

catherineluse:
This PR addresses https://github.com/rancher/dashboard/issues/6376 by allowing a user to reveal deprecated Kubernetes patch versions when provisioning RKE2/K3s clusters, which can be useful when migrating from another Kubernetes distro.

When you go to provision an RKE2/K3s cluster, the default Kubernetes version dropdown menu is unchanged:
Screen Shot 2022-07-19 at 5 11 31 PM

Below the dropdown, there is a new checkbox to reveal deprecated patch versions:
Screen Shot 2022-07-19 at 5 11 37 PM

There is a tooltip that explains that the deprecated versions are a bad idea unless you are migrating:
Screen Shot 2022-07-19 at 5 11 43 PM

When the box is checked, all but the most recent minor versions are marked as deprecated:
Screen Shot 2022-07-19 at 5 12 17 PM

I confirmed that I could successfully provision a v1.21.12 cluster, whereas our supported version is v1.21.14:
Screen Shot 2022-07-19 at 5 21 00 PM

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -1199,6 +1199,9 @@ cluster:
   kubernetesVersion:
     label: Kubernetes Version
     experimental: experimental
+    deprecated: deprecated
+    deprecatedPatches: Show deprecated Kubernetes patch versions
+    deprecatedPatchWarning: We recommend using the latest patch version for each minor Kubernetes version. Deprecated patch versions can be useful for migration purposes.
   toolsTip: Use the new Cluster Tools to manage and install Monitoring, Logging and other tools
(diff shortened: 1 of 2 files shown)


nwmac:
@catherineluse - the reviewer you assigned is on vacation - can you find another person to review.

rak-phillip (approved):
LGTM - works as advertised.

I also tested the case where we select a deprecated version from the list and uncheck the new checkbox. I was greeted with a clear warning, but we do seem to loose the (deprecated) annotation..

image

PR #6429, opened 2022-07-20
Edit actions for builtin roletemplates and global roles

### LINKED ISSUE #6212: Roles cannot be updated to inherit or grant resources
**Setup**
- Rancher version: 2.6.5
- Browser type & version: firefox nightly 102.0a1 (2022-05-17) (64-bit)

**Describe the bug**

I cannot update rbac for roles through the UI.

**To Reproduce**

 Logged in as full administrator I then go to
`Users & Authentication` -> `Roles` -> `Projects/Namespaces` -> `Project Member` -> `...` -> `Edit Config` -> `Inherit From` -> `Add Resource` -> `View Secrets` -> `Save`

**Result**
It says the save was successful, but the inherited role is not listed, only `Kubernetes Edit`, and view-secrets permission is not listed in the table. Project Members assigned to that namespace are not able to view secrets.

**Expected Result**

Inheritance and GrantResources are correctly saved to roles

**Screenshots**

**Additional context**

per instructions from slack I am able to edit the `roletemplates.management.cattle.io/project-member` resource manually like so
```
- apiGroups:
  - ""
  resources:
  - secrets
  verbs:
  - get
  - list
  - watch
```

jordojordo:
### Summary
Fixes #6212

The main issue of not being able to update roles was actually due to the role being `builtin`, as in this is a Rancher specific role that ***cannot*** be updated or deleted. In the old UI we actually prevent the user from being able to update or delete a role by disabling the inputs.

Instead of disabling all of the inputs, a quick fix here would be to disable the actions for editing/deleting from the resource itself.

### Technical notes summary

If the role has `builtin: true` we will disable the `goToEdit` and `promptRemove` actions. 

You should be able to create a role and still have the ability to edit/delete.

### Screenshot/Video

_`builtin` role has no edit/delete action_
removedaction

_The old UI for reference_
oldui

### DIFF
--- a/shell/models/management.cattle.io.roletemplate.js
+++ b/shell/models/management.cattle.io.roletemplate.js
@@ -66,7 +66,7 @@ export default class RoleTemplate extends SteveModel {
       }
     });
 
-    if ( editActions ) {
+    if ( editActions.length > 0 ) {
       editActions.forEach((a) => {
         a.enabled = !this.builtin;
       });
--- a/shell/models/management.cattle.io.roletemplate.js
+++ b/shell/models/management.cattle.io.roletemplate.js
@@ -56,6 +56,25 @@ export const VERBS = [
 ];
 
 export default class RoleTemplate extends SteveModel {
+  get availableActions() {
+    const out = super._availableActions;
+
+    const toFilter = ['goToEdit', 'promptRemove'];
+    const editActions = out.filter((a) => {
+      if ( toFilter.includes(a.action) ) {
+        return a;
+      }
+    });
+
+    if ( editActions ) {
+      editActions.forEach((a) => {
+        a.enabled = !this.builtin;
+      });
+    }
+
+    return out;
+  }
+
   get customValidationRules() {
     return Role.customValidationRules();
   }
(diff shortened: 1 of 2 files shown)


shell/models/management.cattle.io.roletemplate.js
```diff
@@ -56,6 +56,25 @@ export const VERBS = [
...
+    const out = super._availableActions;
+
+    const toFilter = ['goToEdit', 'promptRemove'];
+    const editActions = out.filter((a) => {
+      if ( toFilter.includes(a.action) ) {
+        return a;
+      }
+    });
+
+    if ( editActions ) {
```
catherineluse, line 69: The `if (editActions)` will always return true because an empty array is truthy.

Can it be changed to `if (editActions.length > 0)`?

catherineluse (changes requested):

shell/models/management.cattle.io.roletemplate.js
jordojordo: Ah nice catch, fixed!

catherineluse (approved):
LGTM

PR #6432, opened 2022-07-21
Performance: Unwatch some resource types on cluster dashboard and nodes pages

### LINKED ISSUE #6431: Performance: Cluster Dashboard and Nodes pages keep watching resources
The Cluster dashboard and nodes pages fetch resources - e.g. events, nodes - which sets up watches.

If you change page, these watches remain, even though the data is not (typically) of use to other pages.

We should stop watching these resources

nwmac:
Fixes #6431

This PR exposes forgetType as an action and adds in un-watching of the type as well.

It wires this into the cluster dashboard and nodes page, un-watching Steve types that are used in those pages.

In particular, the cluster dashboard page watches events - which can be changing quite regularly, so if you then change to another page, we still get event messages, even though  we're not showing them anywhere.

On the nodes page, we fetch pods, which can also be changing a lot. These will now also be un-watched. Other pages do use pods, so the downside is that if you are looking at deployments and go to nodes, when you go back, the pods will have to re-load - but this should not block the page, just the time taken for certain columns to have the pod data to be able to update.

Reducing the number of types being actively watched seems like a good thing to do - less load on the backend and less load on the frontend, which has to process the web socket messages for those updates and add then to the store.

To test:

- Open up the browser dev tools, go to networking, show web sockets only and select web socket 1 (or whichever is the Steve web scoket)
- Go to the cluster dashboard page for a cluster (local or downstream)
- Verify that the events table shows data correctly
- In dev tools, verify that you see messages to watch events and nodes - they look like `{ "resourceType": "event", "resourceVersion": "<NUMBER>"}`
- Go to the projects/namespaces view
- In dev tools, verify that you see messages to un-watch events and nodes - they look like `{ "resourceType": "event", "resourceVersion": "<NUMBER>", "stop": true}`
 - Go back to the cluster dashboard page and verify that the events table loads correctly
 - Repeat the above for the nodes table, checking that the `node` and `pod` resource types are correctly un-watched

### DIFF
--- a/shell/list/node.vue
+++ b/shell/list/node.vue
@@ -70,6 +70,13 @@ export default {
     };
   },
 
+  beforeDestroy() {
+    // Stop watching pods, nodes and node metrics
+    this.$store.dispatch('cluster/forgetType', POD);
+    this.$store.dispatch('cluster/forgetType', NODE);
+    this.$store.dispatch('cluster/forgetType', METRIC.NODE);
+  },
+
   computed: {
     hasWindowsNodes() {
       return (this.kubeNodes || []).some(node => node.status.nodeInfo.operatingSystem === 'windows');
--- a/shell/pages/c/_cluster/explorer/index.vue
+++ b/shell/pages/c/_cluster/explorer/index.vue
@@ -12,6 +12,7 @@ import {
   STATE,
 } from '@shell/config/table-headers';
 import {
+  EVENT,
   NAMESPACE,
   INGRESS,
   MANAGEMENT,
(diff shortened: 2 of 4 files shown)


aalves08 (approved):
LGTM

richard-cox (commented):
One observation, possibly ok.

When flicking between Cluster Dashboard and nodes pages i see intermittent `Resource error [cluster] pod : event watch error: an error on the server ("unable to decode an event from the watch stream: context canceled") has prevented the request from succeeding` warning logs originating from `subscribe.js` ```console.warn(`Resource error [${ getters.storeName }]`, msg.resourceType, ':', msg.data.error);```. Could this be something like the watch cancelling whilst an update is incoming, as is ok?

I've tested the Pods column in the Nodes table still updates automatically after seeing the above warning

Update: Interestingly am seeing a stack of similar but not exactly the same errors on the webworker PR when creating role bindings `Resource error [management] management.cattle.io.project : event watch error: an error on the server ("unable to decode an event from the watch stream: http2: response body closed") has prevented the request from succeeding`. Maybe the backend has improved their logging output?

nwmac:
@richard-cox Great find - I have updated the PR with a fix - the issue is that there may be resource updates in the pending queue for the type that is being forgotten - the flush timer then processes them after we've removed the type.

Fix is to remove these update messages from the pending queue.

richard-cox (approved):

PR #6435, opened 2022-07-21
Provide extra feedback to the user until form-validation can be added

### LINKED ISSUE #5924: [Monitoring] When installing `Alerting Monitors` chart, the `Install` button in step two is not greyed out if SMS or Microsoft Teams checkboxes are disabled
**Setup**
- Rancher version: `v2.6.5-rc10`

**Describe the bug**

When installing the chart `Alerting Monitors`, step two clearly states that you need to enable either SMS or Microsoft Teams. However, when you have neither option selected, you can still click the `Install` button as seen below:

image

From a user's perspective, this is misleading as it would prompt me to believe that I don't need to select either option to proceed forward. With that mindset, I can click Install, but immediately it fails and the terminal shows me the following:

```
helm install --namespace=cattle-monitoring-system --timeout=10m0s --values=/home/shell/helm/values-rancher-alerting-drivers-100.0.2.yaml --version=100.0.2 --wait=true rancher-alerting-drivers /home/shell/helm/rancher-alerting-drivers-100.0.2.tgz
Error: INSTALLATION FAILED: execution error at (rancher-alerting-drivers/templates/cluster-role.yaml:2:4): At least one Driver must be enabled to install the chart.
```
We can avoid this user error from occurring by simply greying out `Install` until either the Microsoft Teams or SMS checkbox is enabled. Likewise, if the user disables either box, then grey out `Install` again until they reenable one of the checkboxes.

**To Reproduce**

1. Setup Rancher and navigate to Rancher UI in a browser.
2. Create a downstream RKE2 cluster (node driver or custom doesn't matter, I chose custom in my testing).
3. After cluster creation, navigate to Cluster Explorer > cluster > Apps > Charts > Alerting Monitors.
4. Fill out step one and in step two, DO NOT select SMS or Microsoft Teams and click Install; should error in the terminal.

**Result**
User can click `Install`, but this will fail as expected when trying to install the chart.

**Expected Result**

If a user has not selected either SMS or Teams, they should be allowed to click `Install`; the button should be greyed out.

Sean-McQ:
Provide extra feedback to the user until form-validation can be added and also adds in a default value for "Enable SMS" so we're not loading the form in an invalid state.

### Summary
Fixes #5924 

Provide extra feedback to the user until form-validation can be added and also adds in a default value for "Enable SMS" so we're not loading the form in an invalid state.

### Technical notes summary
Greying out the "Install" button will have to be part of a larger effort after technical evaluation. Providing more clear user feedback should prevent the user from installing this chart with invalid values in the interim.

* Values step should load with "Enable SMS" preselected. When installed without touching the form, payload should include "sachet.enabled" as being true.
* Banner informing the user that they must select at least one of the options in the form should become red if the user deselects both options.

### DIFF
--- a/shell/chart/rancher-alerting-drivers.vue
+++ b/shell/chart/rancher-alerting-drivers.vue
@@ -11,12 +11,26 @@ export default {
         return {};
       }
     }
+  },
+  computed: {
+    mustSelectOne() {
+      if (!this.value.prom2teams.enabled && !this.value.sachet.enabled) {
+        return 'error';
+      }
+
+      return 'info';
+    }
+  },
+  mounted() {
+    if (this.mustSelectOne === 'error') {
+      this.value.sachet.enabled = true;
+    }
   }
 };
 </script>
 <template>
   <div>
-    <Banner color="info">
+    <Banner :color="mustSelectOne">
       {{ t('rancherAlertingDrivers.selectOne') }}
     </Banner>
     <div class="row">


jordojordo (approved):
Looks good!

PR #6445, opened 2022-07-21
Resolves issues with ServicePort validation

### LINKED ISSUE #4238: Add Validation to form fields going forward
**Summary:**

There are many fields in the Vue UI that we do not validate. I propose we programmatically try to validate fields as we implement/design new areas when/where reasonable.

-----

**Environment:**
Rancher version: v2.6-head `ba43472` pulled at 9/21/21 10:43am
Rancher cluster type: single-node docker install
Docker version: 20.10
 
Downstream cluster type: n/a
Downstream K8s version: n/a

**Details:**
- Various fields today don't validate input of the fields. The field may be wanting for just a number or alphanumeric charatcters but not actually validate. The field/form will allow the user thus to submit and cause issues which can break things.
- I propose going forward we try to validate most fields when reasonable. That is to say I suggest using best judgement and only validating fields when we are 100% certain what types of data should be passed.

**Specific Examples:**
It would require extensive work to try to manually identify all the areas we don't do validation. I'm sure if I had time I could find many more than are listed below.

Today we don't validate (as of 9/21/21) the following fields.

apps.Deployment > Security Context > Filesystem Group
image
image

- I should not be able to enter spaces
- This should only take integers (1-30 I think? not 100% sure)
- Note: the form will submit but no securityContext gets passed in the POST

apps.Deployment > Security Context > Add/Drop Capabilities drop downs
image
- Dropdowns let you enter text - is this something we could prevent? if text does not match entries/options then remove it when user clicks off the field? Not sure if this one is easy or a reasonable ask.

Sean-McQ:
### Summary
Fixes #4238
Some small issues arose during testing of serviceports
* Node Port and Listening Port now report errors when outside of port ranges
* Node Port and Target Port no longer allow extra characters (note that decimal points with no numbers and/or 0s following them are simply trimmed by the input instead of caught by validation

Reworked Serviceport validation. Potentially worth breaking up into several smaller validators in the future

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -4968,9 +4968,11 @@ validation:
         required: 'Port Rule [{position}] - Name is required.'
       nodePort:
         requiredInt: 'Port Rule [{position}] - Node Port must be integer values if included.'
+        between: 'Port Rule [{position}] - Node Port must be between 1 and 65535'
       port:
         required: 'Port Rule [{position}] - Port is required.'
         requiredInt: 'Port Rule [{position}] - Port must be integer values if included.'
+        between: 'Port Rule [{position}] - Port must be between 1 and 65535'
       targetPort:
         between: 'Port Rule [{position}] - Target Port must be between 1 and 65535'
         iana: 'Port Rule [{position}] - Target Port must be an IANA Service Name or Integer'
--- a/shell/components/form/ServicePorts.vue
+++ b/shell/components/form/ServicePorts.vue
@@ -189,7 +189,7 @@ export default {
           <span v-if="isView">{{ row.targetPort }}</span>
           <input
             v-else
-            v-model.number="row.targetPort"
+            v-model="row.targetPort"
             :placeholder="t('servicePorts.rules.target.placeholder')"
             @input="queueUpdate"
           />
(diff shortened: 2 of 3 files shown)


mantis-toboggan-md (approved):
lgtm

catherineluse:
Hm. For me the target input looks like it still allows letters. Screen Shot 2022-07-21 at 12 43 55 PM

I created a target port with letters in it and the backend didn't stop me either:
Screen Shot 2022-07-21 at 12 45 58 PM


A couple of minor nits -

- The node port name should have a red asterisk
- I think the red error message shouldn't appear until the form has been touched
Screen Shot 2022-07-21 at 12 40 41 PM
- I think the error message should be on its own row so it doesn't look like it only goes with the first column
Screen Shot 2022-07-21 at 12 42 43 PM

Sean-McQ:
@catherineluse TargetPort is a bit of a special case because it could be a port or a `dnsLabelIanaServiceName`. It should complain if you use illegal characters for that like periods or the like.

These servicePort validators are on my list of "refactor for UX" things. Right now, all the rules are tied up into a single validator because that's how the model expects the validator to work but it prevents me from tying the validation message to a single field like I do elsewhere. I need to extend the mixin to optionally ignore certain rules from the model so I can break the servicePort validator up into several smaller ones that'll work much like all the others.

Issue #5922, opened 2022-05-11
`config` view of custom rke2 cluster brings user to the `edit imported cluster` screen instead


slickwarren:
**Setup**
- Rancher version: v2.6.5-rc10
- Browser type & version: chrome

**Describe the bug**
with a custom cluster, when the user goes to `config` view, the user is instead taken to the `edit` page for the cluster. however the view is identical to the `edit` page of an imported cluster. 

**To Reproduce**
* create a custom rke2 cluster
* register at least 1 node with the cluster
* from `cluster management` -> select your cluster -> select `config` view

**Result**
when the user goes to `config` view, the user is instead taken to the `edit` page for the cluster. however the view is identical to the `edit` page of an imported cluster. 

**Expected Result**
user should be shown a read-only screen displaying values that the user entered while creating the cluster (this is the behavior for node driver clusters)

**Screenshots**

<img width="1425" alt="Screen Shot 2022-05-11 at 8 42 53 AM" src="https://user-images.githubusercontent.com/16691014/167891446-1915038d-0d64-4976-a8f2-e09373b194cf.png">
<img width="1000" alt="Screen Shot 2022-05-11 at 8 42 32 AM" src="https://user-images.githubusercontent.com/16691014/167891456-0a0dd088-344d-4552-89de-7fef990763dc.png">

**Additional context**
* this is not happening for node driver clusters at this time, only for custom clusters

catherineluse:
I couldn't reproduce this bug. For the backend I'm using v2.6.5 and for the front end I'm using the latest master branch.

Here's what I tried:

1. Provisioned a digital ocean droplet
2. Went to Cluster Manager and clicked Create
3. Clicked Custom
4. Ran the custom cluster command to install RKE2 on the digital ocean node
5. Went to Cluster Manager, went to the new RKE2 cluster and clicked **⋮ > Edit Config**
6. Confirmed the form is correct:
<img width="1488" alt="Screen Shot 2022-05-27 at 4 01 26 AM" src="https://user-images.githubusercontent.com/20599230/170687559-5ffa1654-aa3e-4904-b547-8f713d89aed1.png">

catherineluse:
In the issue description, I noticed that even for an imported cluster, that form would still be incorrect because there are fields missing. I wonder if the issue is that Rancher disconnected from the back end but didn't show it on the front end.

catherineluse:
I'm having trouble creating a custom cluster in v2.6-head:
[...]

PR #6467, opened 2022-07-23
Render the correct form for imported RKE2 clusters

catherineluse:
This PR addresses https://github.com/rancher/dashboard/issues/5922 by updating the conditional logic that renders the read-only config form for custom K3s/RKE2 clusters.

The route of this read-only form is `rancherServerUrl/c/c-m-4brtk95v/manager/provisioning.cattle.io.cluster/fleet-default/catherine-test-custom?as=config`. This PR makes it so that now we render the custom cluster form using the `as=config` query param.

Previously, you could still get to the correct form by going to the context menu and clicking Edit Config. This PR only affects the Config view that you can get to from the cluster detail page in Cluster Management:
Screen Shot 2022-07-23 at 1 39 25 PM

When the Config button is clicked, it now shows the correct config form in read-only mode:

Screen Shot 2022-07-23 at 1 43 58 PM


mantis-toboggan-md (approved):
lgtm

Issue #6465, opened 2022-07-22
Hide Rancher's automatically created RBAC namespaces from users


catherineluse:
Every time you add a namespace to a project, Rancher automatically creates a separate namespace in the `local` cluster to keep track of RBAC for the project. When you delete a namespace, Rancher also deletes the corresponding RBAC namespace on the local cluster.

From an end user's perspective, this behavior looks strange as it is not immediately obvious what the namespace is for, or why Rancher is auto-deleting namespaces.

Recently I created a PR to make the namespace results consistent in the Projects/Namespaces page and the top nav filter options in Cluster Explorer. https://github.com/rancher/dashboard/pull/6394 We made it consistent by taking the system namespace filters out of the list to make it consistent with the top nav. However, we received feedback from QA that when you can see these system namespaces being automatically created and deleted, it looks like a bug. It would also be bad if a user deleted these auto-created namespaces because they would lose role bindings for cluster and would need to restore the cluster from backup. So the new approach is to filter the system namespaces for RBAC from both the Projects/Namespaces list AND the top nav.

jameson-mcghee:
@catherineluse, good news and bad news lol
Good News:
In v2.6-head, I am no longer seeing the RBAC Namespaces being added to the "Not in a Project" project.

v2.6-head RBAC Namespaces not present after creating a new Project and Namespace:
![image.png](https://zube.io/files/rancher/6169481f372b15135f9e04f8ecc489c9-image.png)
![image.png](https://zube.io/files/rancher/72f39aeaa9e4608be935908a930a7466-image.png)
![image.png](https://zube.io/files/rancher/917d2a000c764a64005d9bd60ebd4681-image.png)


Bad News:
During testing I noticed what looks like a regression of the fix for ticket [6387](https://zube.io/rancher/dashboard-ui/c/6387). When I add a new Namespace to a project, I have to refresh the page before Namespace is present in the UI grid. Therefore I am Reopening this ticket

v2.6-head new Namespace not present in UI until after refresh:
![image.png](https://zube.io/files/rancher/e30ef9d541cdefe5030dda8654472d54-image.png)
![image.png](https://zube.io/files/rancher/42454af53e64d74d59f97b0f71409a8b-image.png)

jameson-mcghee:
[...]

PR #6468, opened 2022-07-23
Hide auto-created RBAC namespaces

catherineluse:
This PR addresses https://github.com/rancher/dashboard/issues/6465 by hiding auto-created RBAC namespaces starting with `p-`. After this change, you should only be able to see the namespaces beginning with `p-` by going to the user preferences and turning on the developer tools.

I've updated the description of the developer tools to explain that it changes what namespaces you see:
Screen Shot 2022-07-23 at 4 08 44 PM

This PR updates the namespace filtering logic in the top nav to make it consistent with the list view. The top nav now also shows the RBAC namespaces based on the user preference set in the developer tools.

With the developer tools turned off:

Screen Shot 2022-07-23 at 4 34 14 PM

With the developer tools turned on:
Screen Shot 2022-07-23 at 4 23 32 PM

To test this PR,

1. I created a project and namespace in a downstream cluster
2. Got the project ID from the project's YAML
3. Went to the user preferences and turned on the developer tools
4. Went to the local cluster in Cluster Explorer and filtered by "Not in a Project"
5. Confirmed that Rancher had automatically created a namespace with the new project ID, and it was listed under "Not in a Project"
6. Went to the user preferences and turned off the developer tools
7. Went back to "Not in a Project" and confirmed that the RBAC namespaces are hidden

Note: Didn't hide system namespaces because we still want to see the contents of the system project.

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -3632,7 +3632,7 @@ prefs:
     emacs: 'Emacs'
     vim: 'Vim'
   advanced: Advanced
-  advancedTooltip: Enables 'View in API' option and keyboard shortcut to toggle light or dark theme (shift+t)
+  advancedTooltip: Enables 'View in API' option and keyboard shortcut to toggle light or dark theme (shift+t). This setting also reveals namespaces that are managed by Rancher and are not meant to be manually edited or deleted.
   dev:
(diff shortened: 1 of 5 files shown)


mantis-toboggan-md (approved):

mantis-toboggan-md:
lgtm

PR #6469, opened 2022-07-25
[Epinio] Ops/GitHub deployment

scures:
### Summary
Fixes: https://github.com/epinio/ui/issues/63 part 1

Refactor of  `githubPicker` and githubstore to allow interacting with the API for fetching repos, commits and branches

### Test
1. Create a new App
2. Deployment, choose Github
3. Search for the desired `username`, `repo`, `branch`...
4. Should have a preview with what we're about to deploy.

App won't deploy at this moment, we need to deploy the latest Epinio Backend.

### DIFF
--- a/shell/components/form/GithubPicker.vue
+++ b/shell/components/form/GithubPicker.vue
@@ -45,7 +45,7 @@ export default {
       {
         name:        'date',
         width:        220,
-        label:       this.t('githubPicker.tableHeaders.message.label'),
+        label:       this.t('githubPicker.tableHeaders.date.label'),
         value:       'date',
         sort:        'date',
         formatter:   'Date',


shell/components/form/GithubPicker.vue
```diff
@@ -1,343 +1,355 @@
...
-        >
-        </LabeledSelect>
+      <div class="spacer archive">
+        <template>
+          <div class="spacer archive">
+            <LabeledInput
+              v-model="selectedUsername"
+              data-testid="epinio_app-source_git-username"
+              :tooltip="t('epinio.applications.steps.source.github.username.tooltip')"
+              :label="t('epinio.applications.steps.source.github.username.inputLabel')"
```
richard-cox, line 252: as this is a generic component the l10n should live in the root dashboard en-us file

shell/components/form/GithubPicker.vue
```diff
@@ -1,343 +1,355 @@
...
-        file: this.value.file || this.preferredFile
-      });
+        return acc;
+      }, []);
 
-      if ( !file ) {
+      return res;
+    },
+    trimCommit(commit) {
+      if (!commit) {
```
richard-cox, line 163: this can be simplified to `return !!commit ? commit.slice(0, 7) : undefined;`

shell/components/form/GithubPicker.vue
```diff
@@ -1,343 +1,355 @@
...
-    },
+      if (this.selectedUsername && this.selectedRepo && this.selectedCommit) {
+        const url = `https://github.com/${ this.selectedUsername }/${ this.selectedRepo }`;
 
-    selectBranch(branch) {
-      this.selectedFile = null;
-      this.selectedBranch = branch;
-      if ( branch ) {
-        this.fetchFiles(this.selectedRepo, branch);
+        this.$emit('generateUrl', url, this.selectedUsername, this.selectedCommit.sha);
```
richard-cox, line 175: Minor one on terminology, we're emitting values in this component and it's up to the parent component what happens to them (in this case generating a url). For this case would recommend using [v-model](https://v2.vuejs.org/v2/guide/components.html#Using-v-model-on-Components) instead of a `generateUrl` property

pkg/epinio/components/application/AppSource.vue
```diff
@@ -416,6 +446,9 @@ export default Vue.extend<Data, any, any, any>({
         />
       </div>
     </template>
+    <template v-else-if="type === APPLICATION_SOURCE_TYPE.GIT_HUB">
+      <GithubPicker @generateUrl="githubUrL" />
```
richard-cox, line 450: image

The size of the commits table shouldn't be limited to the size of the form fields. it should grow to more of the available space (whilst not making the form field's grow as well)

richard-cox (changes requested):
- On selecting github
  - image
  - There's a large space between the source type component and the username input field. The other's consume this by having a label for the input field. In our case this could use this as a general title for the process, something like `Github Commit` perhaps
  - `Username` should be `Organization/Username` (same for tooltip)
- On entering text in the org/username input it's highly likely to start off valid and then turn invalid
  - image
  - The `Repository` drop down should be disabled if the user/org is invalid and shouldn't be shown as an error
  - We should show a relevant error message if the user/org isn't found
  - Now Sean's new form validation feature is in we can use that to follow a standard pattern. It would be worth asking for a specific introduction where you can show him the github picker component
- Hit an exception when viewing `richard-cox` / `tmi.js` / `v2.0.x` / commit page three. It looks like the author of a commit isn't defined (possibly a deleted account) . The actual error is in the table where we try to use `author.avatar_url`, but we should safeguard this elsewhere in code

Note - This is round one of the review (github store and some of the githubpicker still to do).

shell/store/github.js
```diff
@@ -1,239 +1,95 @@
-import dayjs from 'dayjs';
-import { addParam, parseLinkHeader } from '@shell/utils/url';
-import { addObjects, isArray } from '@shell/utils/array';
-import { GITHUB_REPOS, GITHUB_SCOPES, _DATE } from '@shell/config/local-storage';
-
-const API_BASE = 'https://api.github.com/';
-
 export const EXTENDED_SCOPES = ['repo'];
```
richard-cox, line 1: This legacy property isn't used anywhere and can be removed

shell/store/github.js
```diff
@@ -1,239 +1,95 @@
...
-
-  // Add our proxy prefix
-  url = `/v1/github/${ url.replace(/^\/+/, '') }`;
-
-  // Less pages please
-  addParam(url, 'per_page', 100);
-
-  return url;
-}
+export const GITHUB_BASE_API = 'https://api.github.com';
```
richard-cox, line 2: This is only used within this file, so doesn't need to be exported

shell/store/github.js
richard-cox, line 1: Now that we're using this store, we should probably ensure that it's cleared on log out. Everything in there at the moment is public, but this may change in the future and it would be good to cover that case now. 
There's a auth store `logout` action where we can call a new `github/reset` action

shell/store/github.js
```diff
@@ -1,239 +1,95 @@
...
-    if ( allowCache && hasCached() ) {
-      const cached = getCachedRepos();
-
-      if ( cacheExpired() ) {
-        dispatch('fetchRecentRepos', { allowCache: false });
-      }
-
-      return cached;
+export const actions = {
+  // eslint-disable-next-line no-empty-pattern
```
richard-cox, line 31: We generally have a pattern to use `ctx` as the first arg even if it's not used. That removes the requirement to disable linting for `{ }`

shell/store/github.js
```diff
@@ -242,11 +98,15 @@ export const actions = {
 export const mutations = {
```
richard-cox, line 98: I'm a bit split on if we should include this. There are no `getters` so they're not used anywhere and the result of the action is just returned. Maybe we just remove them, plus the state. Then we won't have to worry about resetting the state when logging out

shell/store/github.js
richard-cox: Might not need to do this, see comment on `mutations`

richard-cox (changes requested):
- there's a lot of churn in `pkg/epinio/components/application/AppProgress.vue` (see inline comments) probably due to some linting. Can you double check you're not using custom rules in your IDE?

This is the last part of the review, no more for the moment.

shell/components/form/GithubPicker.vue
```diff
@@ -1,343 +1,355 @@
...
-    return {
-      repos:       [],
-      branches:    [],
-      files:       [],
+    const eventHeaders = [
+      {
+        name:  'index',
+        label: 'Chose',
+      },
+      {
```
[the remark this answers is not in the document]
richard-cox: There's a `Link` formatter (./shell/components/formatter/Link.vue) that can be used here instead of having a custom template within the list (there's an example in ./shell/edit/networking.k8s.io.ingress/Certificates.vue). If you need to pass in a `value` that has better properties for the formatter to use you can add a `getValue` property to the header definition to convert the `row` value into a nicer object

shell/components/form/GithubPicker.vue
```diff
@@ -1,343 +1,355 @@
...
+            <!-- Selection resume -->
+            <template v-if="showSelections && selectedCommit && selectedBranch">
+              <div class="separator">
+                <div class="resumed">
+                  <img ref="img" :src="selectedCommit.author.avatar_url" alt="" />
+                  <div class="resumed-details">
+                    <div class="label">
+                      {{ t('epinio.applications.steps.source.github.details.tooltip') }}:
+                    </div>
+                    <div class="resumed-details-source">
```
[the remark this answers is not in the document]
richard-cox: I don't think we need to be concerned too much with mobiles. We could make use of the card on a new 'summary' step, that would be a nice way to present the card

shell/components/form/GithubPicker.vue
```diff
@@ -1,343 +1,355 @@
...
 
   data() {
-    return {
-      repos:       [],
-      branches:    [],
-      files:       [],
+    const eventHeaders = [
+      {
+        name:  'index',
+        label: 'Chose',
```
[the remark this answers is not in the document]
richard-cox: `/shell/config/table-headers.js` contains a lot of good examples for header properties. For setting a fixed with there's just `width`

shell/components/form/GithubPicker.vue
```diff
@@ -1,343 +1,355 @@
...
+      },
+      {
+        name:  'sha',
+        label: 'SHA',
+        value: 'sha',
+      },
+      {
```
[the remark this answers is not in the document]
richard-cox: The links are just to `https://github.com/<value>`

pkg/epinio/components/application/AppSource.vue
richard-cox: Adding a max to the commit table is a reasonable step, but we shouldn't grow the input fields at the same time which leads to situations like this (width of entire screenshot is 1300px)
image

pkg/epinio/components/application/AppSource.vue
```diff
@@ -29,6 +29,12 @@ interface GitUrl {
   branch: string
 }
 
+interface GitHub {
+  username?: string,
```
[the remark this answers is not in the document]
richard-cox: `usernameOrOrg`

shell/components/form/GithubPicker.vue
```diff
@@ -1,343 +1,355 @@
...
+          // Reset selections once username changes
+          if (this.oldUsername !== this.selectedUsername) {
+            this.oldUsername = this.selectedUsername;
+
+            return this.reset();
+          }
         }
-      } else {
-        return 'Select a Repository First';
+      } finally {
```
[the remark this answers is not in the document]
richard-cox: Ok, slightly missunderstood.

The github store `fetchGithubAPI` has the catch block

```
  } catch (error) {
    if (error.response) {
      return error.response;
    }
  }
```
This will only return a string in some cases (if error.response is missing `undefined` will return). In github picker we only check for `res?.message`, which might have overlap with something returned by github. 

If would probably better to
- change `fetchGithubAPI` to find a suitable error message (error.response.message or undefined) and throw that
- in apiList catch this, if undefined throw a custom message ('failed to fetch x')
- in github picker catch the guaranteed message and set error etc

shell/store/github.js
richard-cox: sorry, wasn't quite clear, `ctx` is the whole object, so it would be `async apiList(ctx, {`

richard-cox (changes requested):
Looking good! I've updated the conversations from the previous review, almost all done, just a few left. Plus

- I missed this last time, sorry. The gap between github picker and advanced settings is pretty big compared to the other types
  image
- This was probably broken from before, there's no gap between the expanded advanced settings and bottom bar. Could you sneak in a fix in this PR? 
  - image

richard-cox (approved):
I've made a few tweaks, LGTM. If those last 5 commits are fine then this is good to merge (probably worth a squash merge)

Issue #6363, opened 2022-07-13
Error in console when user clicks on Namespace dropdown in Continuous Delivery --> Clusters window


sowmyav27:
**On 2.6head commit id: `2c21373`**

- Deploy RKE1 and RKE2 clusters
- Navigate to Continuous Delivery --> Clusters
- Click on namespace Dropdown on top right --> Error TypeError: Failed to execute 'getComputedStyle' seen in console.
- Error:
```
vue.runtime.esm.js:1897 TypeError: Failed to execute 'getComputedStyle' on 'Window': parameter 1 is not of type 'Element'.
    at c (WorkspaceSwitcher.vue?5709:30:21)
    at f.<anonymous> (WorkspaceSwitcher.vue?9674:158:27)
    at Array.<anonymous> (vue.runtime.esm.js:1989:12)
    at ce (vue.runtime.esm.js:1915:12)
```

jameson-mcghee:
@aalves08, On the plus side, I am not seeing ins errors in the console log while looking at this on v2.6-head. However, I noticed that the page is not updating properly when I chance from one fleet namespace to another.

e.g. when looking at fleet-local, its shoes my local cluster, but when I change the dropdown to fleet-default, the UI does not automatically update to show the correct clusters. It will display the correct clusters after refreshing the page.

Because of this issue, I am reopening this ticket.
cc @sowmyav27 @Shavindra 

Prior to changing namespace dropdown:
![image.png](https://zube.io/files/rancher/6b1f9ad0f090069179f15efc47e43d38-image.png)

After changing namespace dropdown:
![image.png](https://zube.io/files/rancher/0f376a70b03a7d1ef2dbde862d1599e8-image.png)

After refreshing the page:
![image.png](https://zube.io/files/rancher/4bc5472643a4e5a68becd440cf72b900-image.png)

aalves08:
@jameson-mcghee did this feature (workspace switcher connected properly to fleet resources) ever work on older versions? By looking into the code, I don't think it ever worked... 🤔

jameson-mcghee:
@aalves08, assuming we are talking about the same thing, then it appears to work fine for me on 2.6.5

![image.png](https://zube.io/files/rancher/1a86cc5571d191cca846e036a85bee1f-image.png)
![image.png](https://zube.io/files/rancher/9aaaaad9cf806b3f92059b873d1171b2-image.png)
[...]

PR #6474, opened 2022-07-25
add workspace as a filter to fleet clusters list

aalves08:
Addresses issue [6363](https://github.com/rancher/dashboard/issues/6363#issuecomment-1192906662) 

- add workspace as a filter to fleet clusters list so that list view becomes reactive to workspace change on header

@jameson-mcghee

### DIFF
--- a/shell/components/ResourceTable.vue
+++ b/shell/components/ResourceTable.vue
@@ -7,6 +7,7 @@ import SortableTable from '@shell/components/SortableTable';
 import { NAMESPACE } from '@shell/config/table-headers';
 import { findBy } from '@shell/utils/array';
 import { NAME as HARVESTER } from '@shell/config/product/harvester';
+import { NAME as FLEET_NAME } from '@shell/config/product/fleet';
 
 // Default group-by in the case the group stored in the preference does not apply
 const DEFAULT_GROUP = 'namespace';
@@ -126,6 +127,8 @@ export default {
 
   computed: {
     ...mapGetters(['isVirtualCluster']),
+    ...mapGetters(['workspace']),
+    ...mapGetters(['productId']),
     isNamespaced() {
       if ( this.namespaced !== null ) {
         return this.namespaced;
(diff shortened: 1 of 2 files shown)


mantis-toboggan-md:
This will fix filtering for the fleet cluster list view but not other workspaced resources (repos, cluster groups, bundles). I think the underlying issue is that the filtering that should happen here https://github.com/rancher/dashboard/blob/master/shell/components/ResourceTable.vue#L198 uses a getter that doesn't pick up on workspace changes. The `activeNamespaceCache` was added to reduce the need to iterate over many namespaces but that concern doesn't apply to workspaces (we're switching between single workspaces rather than creating a list of 'all user' 'all system' etc) I don't think we need to track `WorkspaceFilter` updates in `activeNamespaceCache`. I think it would be enough to update the `ResourceTable` `filteredRows` to check `currentProduct.showWorkspaceSwitcher` and use the `workspace` getter instead of `activeNamespaceCache` when relevant. I found another issue when switching between two workspaces with the same number of whatever resource is currently viewed but I think its a flaw in `SortableTable` and will file a separate issue.

mantis-toboggan-md (changes requested):
This will fix filtering for the fleet cluster list view but not other workspaced resources (repos, cluster groups, bundles). I think the underlying issue is that the filtering that should happen here https://github.com/rancher/dashboard/blob/master/shell/components/ResourceTable.vue#L198 uses a getter that doesn't pick up on workspace changes. The `activeNamespaceCache` was added to reduce the need to iterate over many namespaces but that concern doesn't apply to workspaces (we're switching between single workspaces rather than creating a list of 'all user' 'all system' etc) I don't think we need to track `WorkspaceFilter` updates in `activeNamespaceCache`. I think it would be enough to update the `ResourceTable` `filteredRows` to check `currentProduct.showWorkspaceSwitcher` and use the `workspace` getter instead of `activeNamespaceCache` when relevant. I found another issue when switching between two workspaces with the same number of whatever resource is currently viewed but I think its a flaw in `SortableTable` and will file a separate issue.

nwmac:
Closed in favour #6524

PR #6479, opened 2022-07-25
do not set driver type using isCustom on create

mantis-toboggan-md:
Fix related to changes here: https://github.com/rancher/dashboard/pull/6467
On create, `value.isCustom` is briefly true because no `machinePools` have been added to the spec. This causes `this.selectType('custom', false);` to be called, preventing any custom driver components from rendering.


catherineluse (approved):
LGTM

PR #6480, opened 2022-07-25
Passes rules to certificate host field, extends arraylist to use rules

### LINKED ISSUE #4575: Additional UI-side validation for Ingress form

Sean-McQ:
### Summary
Fixes #4575
Adds in rules for certificate host when present, passes down to certificate host input, extends arraylist to allow for rules

### DIFF
--- a/shell/components/form/ArrayList.vue
+++ b/shell/components/form/ArrayList.vue
@@ -4,11 +4,12 @@ import { _EDIT, _VIEW } from '@shell/config/query-params';
 import { removeAt } from '@shell/utils/array';
 import { TextAreaAutoGrow } from '@components/Form/TextArea';
 import { clone } from '@shell/utils/object';
+import { LabeledInput } from '@components/Form/LabeledInput';
 
 const DEFAULT_PROTIP = 'Tip: Paste lines into any list field for easy bulk entry';
 
 export default {
-  components: { TextAreaAutoGrow },
+  components: { TextAreaAutoGrow, LabeledInput },
 
   props: {
     value: {
@@ -85,6 +86,13 @@ export default {
     disabled: {
       type:    Boolean,
       default: false,
+    },
+
+    rules: {
+      default:   () => [],
+      type:      Array,
+      // we only want functions in the rules array
+      validator: rules => rules.every(rule => ['function'].includes(typeof rule))
     }
   },
 
@@ -263,6 +271,17 @@ export default {
                 :placeholder="valuePlaceholder"
                 :mode="mode"
                 :disabled="disabled"
+                @paste="onPaste(idx, $event)"
+                @input="queueUpdate"
+              />
+              <LabeledInput
+                v-else-if="rules.length > 0"
+                ref="value"
+                v-model="row.value"
+                :placeholder="valuePlaceholder"
+                :disabled="isView || disabled"
+                :rules="rules"
+                :compact="false"
                 @paste="onPaste(idx, $event)"
                 @input="queueUpdate"
               />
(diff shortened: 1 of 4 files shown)


shell/components/form/ArrayList.vue
```diff
@@ -266,6 +274,17 @@ export default {
                 @paste="onPaste(idx, $event)"
                 @input="queueUpdate"
               />
+              <LabeledInput
```
codyrancher, line 277: This only makes rules work for a small subset of the ArrayList functionality. I think rules need to be evaluated within arraylist itself otherwise it won't work with slots, multiline or regular input.

codyrancher (changes requested):

shell/components/form/ArrayList.vue
Sean-McQ: With slots, we'll be able to pass in rules directly to the input component where they're passed into the slot and the arraylist itself won't even need the rules passed into it. As far as multi-line input, it doesn't look like we actually use this option for arraylist anywhere in the codebase (I considered ripping it out actually but figured that'd be stepping outside of the scope of the ticket a bit much). For regular input, it should never actually be used if there are more than 0 rules passed in since the rules array length is what I'm actually evaluating to know if I should be using labeled-input or the standard input.

I _could_ refactor this component to add in the labeled tool-tip to the basic html input on arraylist that we're using here but that'd basically just be copy/pasta from labeled-input which already knows how to handle rules and really just needs the compact option turned off in order to fit into this form correctly. I could also just rip out the basic html input altogether but I'd run the risk of messing up any deep targeted CSS that might be using arraylist and potentially increase the QA overhead (I didn't do a full analysis to figure out how common that is though); instead I chose to basically gate it behind the presence of rules so that I know it's only used in one place for now.

codyrancher: I think you should file an issue to do the refactor so we can remove the dead code. Once the issue is filed so we can track it I'm fine to ship it.

Sean-McQ: https://github.com/rancher/dashboard/issues/6507

codyrancher (approved):

PR #6502, opened 2022-07-26
HARVESTER: Fix Harvester image issues

n313893254:
### Summary
Fixes Harvester image issues

- https://github.com/harvester/harvester/issues/2562
- https://github.com/harvester/harvester/issues/2563
- https://github.com/harvester/harvester/issues/2569

### Technical notes summary

- Fix clone image lack of labels
- Fix image can save the invalid name
- Fix some settings that the input is focused and then press enter keyboard will refresh the page. This should be the implicit behavior： https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#implicit-submission

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -5973,7 +5973,6 @@ harvester:
     fileName: File Name
     uploadFile: Upload File
     source: Source
-    displayNameValidate: Please input a valid Name.
     sourceType:
       download: URL
       upload: File
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -5973,6 +5973,7 @@ harvester:
     fileName: File Name
     uploadFile: Upload File
     source: Source
+    displayNameValidate: Please input a valid Name.
     sourceType:
       download: URL
       upload: File
--- a/shell/components/ResourceDetail/Masthead.vue
+++ b/shell/components/ResourceDetail/Masthead.vue
@@ -384,7 +384,7 @@ export default {
             </nuxt-link>
             <span v-else>{{ parent.displayName }}:</span>
             <span v-if="value.detailPageHeaderActionOverride && value.detailPageHeaderActionOverride(realMode)">{{ value.detailPageHeaderActionOverride(realMode) }}</span>
-            <t v-else :k="'resourceDetail.header.' + realMode" :subtype="resourceSubtype" :name="displayName" />
+            <t v-else :k="'resourceDetail.header.' + realMode" :subtype="resourceSubtype" :name="displayName" :escapehtml="false" />
             <BadgeState v-if="!isCreate && parent.showState" class="masthead-state" :value="value" />
           </h1>
         </div>
--- a/shell/edit/harvesterhci.io.setting/index.vue
+++ b/shell/edit/harvesterhci.io.setting/index.vue
@@ -159,6 +159,7 @@ export default {
           :localized-label="true"
           :mode="mode"
           :options="enumOptions"
+          @keydown.native.enter.prevent="()=>{}"
         />
       </div>
       <div v-else-if="setting.kind === 'boolean'">
@@ -179,6 +180,7 @@ export default {
         <LabeledInput
           v-model="value.value"
           :label="t('advancedSettings.edit.value')"
+          @keydown.native.enter.prevent="()=>{}"
         />
       </div>
     </div>
--- a/shell/edit/harvesterhci.io.setting/index.vue
+++ b/shell/edit/harvesterhci.io.setting/index.vue
@@ -159,7 +159,6 @@ export default {
           :localized-label="true"
           :mode="mode"
           :options="enumOptions"
-          @keydown.native.enter.prevent="()=>{}"
         />
       </div>
       <div v-else-if="setting.kind === 'boolean'">
@@ -180,7 +179,6 @@ export default {
         <LabeledInput
           v-model="value.value"
           :label="t('advancedSettings.edit.value')"
-          @keydown.native.enter.prevent="()=>{}"
         />
       </div>
     </div>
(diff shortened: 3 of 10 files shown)


shell/edit/harvesterhci.io.setting/index.vue
```diff
@@ -159,6 +159,7 @@ export default {
           :localized-label="true"
           :mode="mode"
           :options="enumOptions"
+          @keydown.native.enter.prevent="()=>{}"
```
mantis-toboggan-md, line 162: We should try to solve this in a more general way so we don't accidentally cause this bug on other pages in the future. We can just put `@submit.prevent` on the `form` element here: https://github.com/rancher/dashboard/blob/master/shell/components/CruResource.vue#L344

This only happens with `<form>` that have only one input element, so I do not expect this change to cause problems elsewhere. https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#implicit-submission

mantis-toboggan-md (changes requested):

shell/assets/translations/en-us.yaml
```diff
@@ -5973,6 +5973,7 @@ harvester:
     fileName: File Name
     uploadFile: Upload File
     source: Source
+    displayNameValidate: Please input a valid Name.
```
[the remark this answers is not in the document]
WuJun2016: There is no need to add displayName validator：  https://github.com/harvester/harvester/issues/2563#issuecomment-1196320939

shell/edit/harvesterhci.io.setting/index.vue
n313893254: Done, add the `@submit.prevent` for `CruResource` form.

n313893254:
@mantis-toboggan-md Thanks for your review. Updated the PR per the comments.

shell/components/ResourceDetail/Masthead.vue
```diff
@@ -384,7 +384,7 @@ export default {
             </nuxt-link>
             <span v-else>{{ parent.displayName }}:</span>
             <span v-if="value.detailPageHeaderActionOverride && value.detailPageHeaderActionOverride(realMode)">{{ value.detailPageHeaderActionOverride(realMode) }}</span>
-            <t v-else :k="'resourceDetail.header.' + realMode" :subtype="resourceSubtype" :name="displayName" />
+            <t v-else :k="'resourceDetail.header.' + realMode" :subtype="resourceSubtype" :name="displayName" :escapehtml="false" />
```
mantis-toboggan-md, line 387: I don't think you have to add a prop to the i18n mixin to render the name in a way that matches the list view, as required in the ticket https://github.com/harvester/harvester/issues/2563. This should render the same way:
```
<span v-else>
   {{ t(`resourceDetail.header.${realMode}`, {subtype: resourceSubtype, name:displayName}, {raw:true}) }}
</span>
```

mantis-toboggan-md (changes requested):

shell/components/ResourceDetail/Masthead.vue
WuJun2016: @mantis-toboggan-md  This is very strange.
1. The `raw` parameter indicates that we want it to convert the html correctly.   (But here we don't want it to convert, the html stays as it is)，   
2. `{raw: true} `should be changed to `true`， https://github.com/rancher/dashboard/blob/a71dee857effc74995b4638fa816f8af1138c17e/shell/plugins/i18n.js#L27

WuJun2016: The question here is, how should I interpret the `raw` parameter inside [Vue.prototype.t](https://github.com/rancher/dashboard/blob/a71dee857effc74995b4638fa816f8af1138c17e/shell/plugins/i18n.js#L27). 
If `raw` here is the same as `raw` in [Vue.component](https://github.com/rancher/dashboard/blob/a71dee857effc74995b4638fa816f8af1138c17e/shell/plugins/i18n.js#L89), then I think it makes sense to add `escapehtml` prop,

mantis-toboggan-md: > {raw: true} should be changed to true，

You're absolutely right, I apologize for being confusing. The way I wrote it is wrong but "worked" because of how this if statement is written:
https://github.com/rancher/dashboard/blob/a71dee857effc74995b4638fa816f8af1138c17e/shell/plugins/i18n.js#L20-L24

My solution works because [Vue interprets expressions in `{{}}` as plain text](https://v2.vuejs.org/v2/guide/syntax.html#Raw-HTML). So these two examples do not render the same way:
```
<t v-else :k="'resourceDetail.header.' + realMode" :subtype="resourceSubtype" :name="displayName" raw />
<span v-else>
  {{ t(`resourceDetail.header.${realMode}`, {subtype:resourceSubtype, name: displayName}, true) }}
</span>
  ```
` <t raw>` will not call `escapeHtml` on the translation string, and will set the element's innerHTML to include raw (aka unescaped) html. That would render the example text `<strong><em>something_interesting</em></strong>` as **_something_interesting_** https://github.com/rancher/dashboard/blob/a71dee857effc74995b4638fa816f8af1138c17e/shell/plugins/i18n.js#L89-L94
When we use `t(<translation key>, <translation args>, true)` inside `{{}}`, `t` returns a string with HTML unescaped, but the html is rendered as plain text by vue anyway. I think this is also why, in the original issue, we saw html tags as plain text in the list view:
https://github.com/rancher/dashboard/blob/a71dee857effc74995b4638fa816f8af1138c17e/shell/components/formatter/LinkDetail.vue#L45-L47

`value` in the issue example is `"<strong><em>something_interesting</em></strong>"` but then, because it is in `{{}}`, it is rendered as plain text.

I think needing to render a translation containing user input is rare, so I didn't want to add a prop to i18n mixin to handle this one situation. Having said that I understand that my solution relies on a detail of Vue implementation instead of something we have explicitly written in the codebase, so if it is not clear enough and you want to keep the prop I'm okay with that.

WuJun2016: Yes, I think it might be good to keep this prop.

mantis-toboggan-md (approved):

PR #6514, opened 2022-07-26
Fixed issue where willSave() was not detecting if targetPort was number

### LINKED ISSUE #6511: [BUG] Unable to set number value for Target Port when creating a service
**Setup**
- Rancher version: v2.6-head 8ffaf56

**Describe the bug**

When creating a service, I was unable to set the Target Port value to 8087, or any number. Rancher kept throwing the error `Service "dns" is invalid: spec.ports[0].targetPort: Invalid value: "8087": must contain at least one letter or number (a-z, 0-9)
`
I was actually able to create a service using alphabet letters, i.e. `a`
image

**To Reproduce**

1. Provision AKS 1.24.0 cluster on Rancher v2.6-head
2. Create service
3. Set value for Target Port to `8087`

**Result**
Rancher does not create service and throws the error, `Service "dns" is invalid: spec.ports[0].targetPort: Invalid value: "8087": must contain at least one letter or number (a-z, 0-9)` but then is able to take alphabet letters as a valid value.

**Expected Result**

Service target port value should be a number and should not accept alphabetic or any non-integer values.

Sean-McQ:
### Summary
Fixes #6511 
regex being used to test if string was a valid number was incorrect, refactors to use the same port class that validation uses to determine if a port is an integer or not and correct the payload accordingly.

### DIFF
--- a/shell/edit/service.vue
+++ b/shell/edit/service.vue
@@ -25,6 +25,7 @@ import { matching } from '@shell/utils/selector';
 import { NAME as HARVESTER } from '@shell/config/product/harvester';
 import { allHash } from '@shell/utils/promise';
 import { isHarvesterSatisfiesVersion } from '@shell/utils/cluster';
+import { Port } from '@shell/utils/validators/formRules';
 
 const SESSION_AFFINITY_ACTION_VALUES = {
   NONE:     'None',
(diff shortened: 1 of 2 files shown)


catherineluse (approved):
LGTM

PR #6521, opened 2022-07-27
Exclude namespaces from API calls using resource field selectors

catherineluse:
This is an experimental PR that addresses https://github.com/rancher/dashboard/issues/6515 by trying to reduce the amount of data fetched from the back end based on the filters selected in the top nav.

- Attempts to exclude namespaces when finding resources of that type (checks if resource is namespaced first)
- Affected pages:
  - Workload list page
  - Pod list page
  - Secrets, configmaps, services - any resource using the generic resource list

Note: This can't replace namespace filtering done in ResourceTable because most custom list pages would need to be manually updated to re-fetch their data when the top nav changes.

Will try updating it so that if the user only has one namespace, or a handful of namespaces selected, it will attempt to get the resources at the Steve endpoints for specific namespaces.

### DIFF
--- a/shell/components/ResourceList/index.vue
+++ b/shell/components/ResourceList/index.vue
@@ -16,10 +16,6 @@ export default {
 
     let hasFetch = false;
 
-    const inStore = store.getters['currentStore'](resource);
-
-    const schema = store.getters[`${ inStore }/schemaFor`](resource);
-
     if ( this.hasListComponent ) {
       // If you provide your own list then call its asyncData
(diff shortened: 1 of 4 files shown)


cnotv:
I would like to help with the PR and test this, but I do not know how 😬 
Could you write the steps please?

cnotv:
I can see there's a lot of changes which needs to be reviewed. We should probably also speed up the PR, so we would not face this again.

Issue #6526, opened 2022-07-27
[AlertmanagerConfig] When editing a saved receiver, the webhook type shows as generic even when configured otherwise


ronhorton:
**Setup**
- Rancher version: `2.6.7-rc5`
- Browser type & version: `chrome`

**Describe the bug**

When creating a webhook type as one creates a receiver, selecting 'sms' correctly allows configuration of the sms receiver. when viewing the confg after saving, the webhook type is listed as 'generic.' 

**To Reproduce**

1. create alertmanagerConfig
2. add receiver - webhook type `sms`
3. save
4. click receiver to view config
**Result**
webhook generic
**Expected Result**

webhook sms
**Screenshots**

<img width="1386" alt="2022-07-27_13-41-41" src="https://user-images.githubusercontent.com/89212856/181313969-98682dea-3645-4a1d-ab7e-36be9fb19616.png">

<img width="1161" alt="2022-07-27_13-42-19" src="https://user-images.githubusercontent.com/89212856/181318082-8e494ac3-479a-4fff-9021-6807ed1d3ba2.png">

**Additional context**

**Detailed Description**

**Context**

ronhorton:
**Pass** verified fixed in `2.6-head Commit ID ecbce2e`
1. create alertmanagerConfig
2. add receiver - webhook type `sms`
3. save
4. click receiver to view config

`webhook now shows as 'SMS'`
<img width="1422" alt="2022-08-01_16-41-42" src="https://user-images.githubusercontent.com/89212856/182243606-08dd7a3e-4a10-4a57-94fd-23f394b771ab.png">

PR #6532, opened 2022-07-28
Selected webhook type is persisted in UI

catherineluse:
This PR fixes https://github.com/rancher/dashboard/issues/6526 by making it so that the webhook type is correctly persisted after a receiver of type webhook is configured. To test, I did the same steps as the steps to reproduce in the issue description. Alibaba SMS, MS Team, and Generic types should all be persisted when coming back to the config form.


Shavindra (approved):
KGTM

PR #6536, opened 2022-07-28
Minor improvement to cluster dashboard page for alerts tab

nwmac:
This PR makes a minor improvement to the cluster dashboard page:

- Un-watches the ENDPOINT type that is used by the alerts table when v2 monitoring is installed
- Only renders the alerts table when its tab is active - prevents background polling when not active


aalves08 (approved):
LGTM

PR #6538, opened 2022-07-28
[Epinio] Limits the files format of the file allowed to be uploaded

scures:
### Summary
Fixes # https://github.com/epinio/ui/issues/143

Added a new prop to `FileSelector` component, 

### Technical notes summary

Added a new prop called `accept` to `FileSelector` component (default allows all) which takes file formats that we want to allow.

Uses HTML Attribute :accept , supported as Chrome 16 +, Safari 6 +, Firefox 9+ , IE 10 +,  Opera 11 +

### Screenshot/Video

image

### DIFF
--- a/pkg/epinio/components/application/AppSource.vue
+++ b/pkg/epinio/components/application/AppSource.vue
@@ -94,7 +94,6 @@ export default Vue.extend<Data, any, any, any>({
       archive: {
         tarball:             this.source?.archive.tarball || '',
         fileName:            this.source?.archive.fileName || '',
-        fileType:            this.source?.archive.fileType || '',
       },
 
       container: { url: this.source?.container.url },
--- a/pkg/epinio/components/application/AppSource.vue
+++ b/pkg/epinio/components/application/AppSource.vue
@@ -17,8 +17,7 @@ import { EpinioAppInfo } from './AppInfo.vue';
 
 interface Archive{
   tarball: string,
-  fileName: string,
-  fileType: 'zip' | 'tar' | 'tgz'| 'tbz' | 'txz',
+  fileName: string
 }
 
 interface Container {
--- a/pkg/epinio/components/application/AppSource.vue
+++ b/pkg/epinio/components/application/AppSource.vue
@@ -349,7 +349,7 @@ export default Vue.extend<Data, any, any, any>({
             :label="t('epinio.applications.steps.source.archive.file.button')"
             :mode="mode"
             :raw-data="true"
-            :accept="'.zip, .tar, .tgz, .tbz, .txz'"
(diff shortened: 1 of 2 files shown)


cnotv:
Please do not forget to rebase to run the tests/linters, although there should not be anything relevant.

shell/components/form/FileSelector.vue
```diff
@@ -58,6 +58,10 @@ export default {
     rawData: {
       type:    Boolean,
       default:    false
+    },
+    accept: {
```
cnotv, line 62: Let's add some jsdoc annotation, even if trivial.

cnotv (commented):
Functionality seems working fine.

Added some opinions but you may feel free to discard if disagree.

cnotv (approved):
LGTM

pkg/epinio/components/application/AppSource.vue
```diff
@@ -349,6 +349,7 @@ export default Vue.extend<Data, any, any, any>({
             :label="t('epinio.applications.steps.source.archive.file.button')"
             :mode="mode"
             :raw-data="true"
+            :accept="'.zip, .tar, .tgz, .tbz, .txz'"
```
richard-cox, line 352: Can we confirm these cover the same file extensions that the backend does? https://github.com/epinio/helm-charts/blob/3954c214de3d7b957cfc2054ba4fa4bfa140f5a3/chart/epinio/templates/stage-scripts.yaml#L27-L62 @jimmykarily

richard-cox: It looks like the tar compression extensions aren't quite correct given the comment at the top of that link. What happens when you provide a `tar.gz` file?

cnotv: I double checked and it seems not working with nested extensions 😢 

Screenshot 2022-08-05 at 11 15 53

cnotv: Also it seems like you'll have to use ` .gz, .bz2, .xz`

cnotv (approved):
LGTM 2

pkg/epinio/components/application/AppSource.vue
jimmykarily: For reference, this is the list of content types we check against in the "upload" endpoint: https://github.com/epinio/epinio/blob/83b68f9510a3f1681eec0ef558a9f26188e14087/internal/api/v1/application/upload.go#L22-L28

[We read the first 512 bytes of the file](https://github.com/epinio/epinio/blob/83b68f9510a3f1681eec0ef558a9f26188e14087/internal/api/v1/application/upload.go#L90) to get the type but that's obviously not possible at the front end.

richard-cox (approved):

PR #6539, opened 2022-07-28
Adds form-validation back in after deployments redesign

Sean-McQ:
### Summary
Form validation was lost during a rearrangement of the workloads edit form, this adds it back in. This is not an exhaustive validation of the form, just on a few key fields as part of the initial implementation of the form-validation mixin.

### DIFF
--- a/shell/edit/workload/mixins/workload.js
+++ b/shell/edit/workload/mixins/workload.js
@@ -42,6 +42,7 @@ import { UI_MANAGED } from '@shell/config/labels-annotations';
 import { removeObject } from '@shell/utils/array';
 import { BEFORE_SAVE_HOOKS } from '@shell/mixins/child-hook';
 import NameNsDescription from '@shell/components/form/NameNsDescription';
+import formRulesGenerator from '@shell/utils/validators/formRules';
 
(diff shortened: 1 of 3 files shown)


catherineluse:
> just on a few key fields

Which ones will QA need to test for this one?

Sean-McQ:
Name, Container Image, Replicas*, Schedule (if present),  Service Name* (if present)

These should trigger an error at the top of the form for right now since they are "unreported" (meaning not tied to a field in the form directly).

Shavindra:
Form validation itself need to be improved to detect `onblur` so warning/error/required fields icons are only shown after user interact with the field. Which is beyond the scope of this PR. 

`hasIcon` method get called repeatedly. Performance issue right now is negligible. However, it's something we need to keep in mind and revisit as a tech debt item.

Shavindra (approved):
Form validation itself need to be improved to detect `onblur` so warning/error/required fields icons are only shown after user interact with the field. Which is beyond the scope of this PR. 

`hasIcon` method get called repeatedly. Performance issue right now is negligible. However, it's something we need to keep in mind and revisit as a tech debt item.

PR #6554, opened 2022-07-30
Script to create standalone build for testing

nwmac:
This PR adds a script in `shell/scripts/standalone/package.sh` - this when run from its parent folder will generate `dashboard.tar.gz` - a standalone package which can then be installed on a machine and browsed to test out the latest UI against an existing backend.

You need to run this on the same target platform as you intend to run it on (e.g. linux etc).

This bundles in the node executable as well from the host machine.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,3 +106,7 @@ dist-pkg
 
 # Plugins - temporary
 shell/rancher-components
+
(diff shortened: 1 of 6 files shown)


nwmac:
Yeah sorry - you could remove that

PR #6575, opened 2022-08-02
[2.5] fix issue with force-reloading norman objects

### LINKED ISSUE #6516: Azure AD Authentication must be updated warning is seen on explorer UI after updating the endpoint in ember
Setup

Rancher version: v2.5.15 >> upgraded to 2.5.16-rc1

**Describe the bug**
Warning message seen on the explorer UI after updating end points on ember UI

**To Reproduce**
1. Install rancher v2.5.15 and enable Azure AD
2. Upgrade rancher server to v2.5.16-rc1
3. Update the end points for azure AD
4. Navigate to cluster explorer 

**Result**
Warning message is still seen `Azure AD Authentication must be updated: it is using the Azure Graph API, which will be deprecated at the end of 2022. Update it here.`

**Expected Result**
Expected the warning message to be gone.

**Screenshots**
25warning

mantis-toboggan-md:
Fixes #6516  
https://github.com/rancher/dashboard/issues/6516#issuecomment-1200473823

The AzureAD upgrade warning banner re-appears when a user visits the auth provider detail page. When the config view for a specific auth prover is loaded the authConfig mixin will always send an api request for the norman config object by using the `force` option.

If a resource being loaded from a network request is already in the store, the steve plugin `load` mutation overwrites the existing object in the store key by key. Because this overwrite data is never run through `proxyFor`, `remapSpecialKeys` is not called on it, and the norman keys that intersect with resource-class methods are not properly remapped. In this case, `value.annotations` needs to be mapped to `value._annotations`. 

As far as I can tell, the lack of remapping is only an issue if `force` is used to fetch a norman resource already in the store. I think this is a safe change (no re-mapping of something that shouldn't have been)  because `remapSpecialKeys` is called every time a resource is proxified initially, so `entry` in the context of this change is always something that was run through `remapSpecialKeys` itself.

### DIFF
--- a/plugins/steve/mutations.js
+++ b/plugins/steve/mutations.js
@@ -2,7 +2,7 @@ import Vue from 'vue';
 import { addObject, addObjects, clear, removeObject } from '@/utils/array';
 import { SCHEMA } from '@/config/types';
 import { normalizeType, KEY_FIELD_FOR } from './normalize';
-import { proxyFor } from './resource-proxy';
+import { proxyFor, remapSpecialKeys } from './resource-proxy';
 
 function registerType(state, type) {
   let cache = state.types[type];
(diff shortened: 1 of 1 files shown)


rak-phillip (approved):
LGTM - AFAICT this change resolves the issue at hand

PR #6606, opened 2022-08-04
do not attempt to filter by namespace for any unnamespaced resource

### LINKED ISSUE #6596: [Bug] Access Harvester project/namespace page hangs
**Setup**
- Rancher version: v2.6.7-rc6

**Describe the bug**

Harvester namespace page is stuck

**To Reproduce**

- Deploy Rancher and import a Harvester cluster
- Go to Harvester cluster
- Go to Projects/Namespaces

**Result**

- Projects/Namespaces page hangs

**Expected Result**

- Show namespace list

**Screenshots**

image

**Additional context**

```
[Vue warn]: You may have an infinite update loop in a component render function.
```

The issue is caused by `filteredRows` keep update, maybe introduced by this line https://github.com/rancher/dashboard/blob/55b05644900bec94cd454eae0d86a1be9ce0979a/shell/components/ResourceTable.vue#L206

If fix like this, the namespaces list works as expect

```js
const includedNamespaces = this.$store.getters['activeNamespaceCache'];
```

**Detailed Description**

**Context**

Harvester issue: 
- https://github.com/harvester/harvester/issues/2626

mantis-toboggan-md:
fixes #6596 
If this is not true, resource table filters its rows wrt the namespace filter (and calculates active namespaces to do so):
`if ( (!this.isNamespaced || isAll) && !isVirtualProduct)` 

The logic here is a little off. We don't want to attempt to filter by ns if the resource is not namespaced, regardless of whether it's a harvester resource or not. "Why does this cause an infinite loop in the project/namespace list specifically?" You may ask. Good question. I think this is because the project/namespace list component also retrieves namespaces via getter and these two calls are causing each other to re-calculate. Regardless, that `if` block is wrong and fixing it fixes the bug.

### DIFF
--- a/shell/components/ResourceTable.vue
+++ b/shell/components/ResourceTable.vue
@@ -199,7 +199,8 @@ export default {
       const isVirtualProduct = this.$store.getters['currentProduct'].name === HARVESTER;
 
       // If the resources isn't namespaced or we want ALL of them, there's nothing to do.
-      if ( (!this.isNamespaced || isAll) && !isVirtualProduct) {
+      // If this is a harvester list, 'all' must still be filtered
+      if ( !this.isNamespaced || (isAll && !isVirtualProduct)) {
         return this.rows || [];
       }
 


Sean-McQ (approved):
LGTM

PR #6607, opened 2022-08-04
Create github action for publishing @rancher/components to npm

### LINKED ISSUE #5934: [@rancher/components] Add CI/CD

rak-phillip:
### Summary
This introduces a github action for publishing @rancher/components to npm. I considered automating this process a little bit by automatically creating release when merging into a release branch, but I decided to walk things back a little bit and keep the release process relatively similar to what we already do today with our other releases. 

For this release, we explicitly choose to only lint, test, and build anything located under `pkg/rancher-components`. I think it might make sense to publish storybook along with releases of the Rancher Component framework, but we already publish changes to storybook on a much more frequent basis than we would with our releases. For this reason, it makes sense to follow up in a separate PR if we think this is needed.

It might make sense to publish a few alpha releases based off this PR, but I think I think a bulk of the testing will be to publish a few alpha/beta releases with this workflow after merging. 

### Release Workflow

Untitled

closes #5934

### DIFF
--- a/.github/workflows/release-rancher-components.yml
+++ b/.github/workflows/release-rancher-components.yml
@@ -39,5 +39,9 @@ jobs:
       with:
         fetch-depth: 1
         persist-credentials: false
+    - uses: actions/setup-node@v2
+      with:
+        node-version: '16.x'
+        registry-url: 'https://registry.npmjs.org'
     - name: Publish to npm
       run: yarn publish:lib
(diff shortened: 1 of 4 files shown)


richard-cox (approved):
PR looks good. The release workflow though needs to ensure the rancher-component's package.json version gets rev'd as well (to a semver acceptable to npm). 

We should also document that flow somewhere (might need to sync with @rishabhmsra to ensure it doesn't get lost with the docusaurus change)

PR #6651, opened 2022-08-12
Add a description slot to RadioButton

### LINKED ISSUE #6649: [@rancher/components] Add description slot for radio button

rak-phillip:
### Summary

This change upstreams the inclusion of the description slot for `RadioButton.vue`. 

This slot was a customization used by Rancher Desktop to provide arbitrary content to a radio button's description. For example, using code blocks in the description:

image

`descriptionKey` & `description` props should take priority over the slot in order to preserve backward compatibility. 

We need to look out for regressions in `RadioButton` usage. Performing a search for `RadioButton` should provide a wide array of examples to test. Since this slot is optional, there should be no change to the existing radio button implementations. Any observed change in output is to be considered a regression. 

closes #6649 

### DIFF
--- a/pkg/rancher-components/src/components/Form/Radio/RadioButton.vue
+++ b/pkg/rancher-components/src/components/Form/Radio/RadioButton.vue
@@ -88,7 +88,14 @@ export default Vue.extend({
     muteLabel(): boolean {
       // Don't mute the label if the mode is view and the button is checked
       return this.disabled && !(this.mode === _VIEW && this.isChecked);
-    }
+    },
+
+    /**
+     * Determines if the description slot is in use.
+     */
+    hasDescriptionSlot(): boolean {
+      return !!this.$slots.description;
+    },
   },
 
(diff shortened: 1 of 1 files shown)


pkg/rancher-components/src/components/Form/Radio/RadioButton.vue
```diff
@@ -155,6 +162,12 @@ export default Vue.extend({
           {{ description }}
         </template>
       </div>
+      <div
+        v-else-if="hasDescriptionSlot"
+        class="radio-button-outer-container-description"
+      >
+        <slot name="description" />
```
catherineluse, line 169: I'm not sure the conditional logic is needed - wouldn't it still work if line 169 was the only line added? I believe if you have a named slot that is not used, it just won't render.

catherineluse (changes requested):

catherineluse (approved):
LGTM

PR #6661, opened 2022-08-16
Add issues templates that set labels

nwmac:
Updates the issue template and splits out into 3 templates, each of which sets the appropriate `kind` label.


richard-cox (changes requested):
There's a `For feature requests` section at the bottom of the old isssue_template, now bug_report, that can be nuked. Otherwise LGTM

richard-cox (approved):

PR #6668, opened 2022-08-16
Port ToggleSwitch to @rancher/components

rak-phillip:
This ports the `ToggleSwitch` component `@rancher/components`. `ToggleSwitch` is planned to be used for upcoming features in Rancher Desktop and it would be nice to have it available in the next release of `@rancher/components` so we don't have to resort to duplicating this one.

To the best of my knowledge, `provisioning.cattle.io.cluster` is the only dashboard component that makes used of `ToggleSwitch`. We need to assert that it still renders and behaves as expected here. Any change in design or behavior is considered a regression. 


catherineluse (approved):
LGTM

PR #6670, opened 2022-08-17
Fix RBAC issues with PVC creation

### LINKED ISSUE #6444: Create PVC form doesn't show Storage Classes if user can't read Persistent Volumes
Internal reference: SURE-4826
Reported in 2.6.5

Issue description:
If a user doesn't have permission to list PVs, the Create PVC form doesn't show any Storage Classes, even though the user can list Storage Classes. This means certain users do not have access to view PVs, when they need to be able to create PVCs

Repro steps:

- Create a custom project role based on project-member, but with no persistentvolume permissions(example attached)
- Assign user access within project
- Log-in as user and confirm user can list storage classes and PVCs
- Attempt to create a PVC with source='Use a Storage Class to provision a new Persistent Volume'

Workaround:
The old /g/ interface works for this

Actual behavior:
The 'Create PVC' interface does not show any Storage Classes

Expected behavior:
User should be able to provision a PVC from a Storage Class{}

Screenshot

 (You can then compare what the user can see in the StorageClasses list)

nwmac:
Fixes #6444

This PR fixes the above issue.

The problem is that we don't check if the user has permission to list storage classes or persistent volumes and this causes the fetch code path to break.

This PR updates this to check permissions and to use input boxes rather than select components - the user still has permission to create a PVC, so we don't block, we just aren't able to show them the existing options.

It also adds a tooltip explaining this.

Issue #6444 was about persistent volumes, but the bug was also there if the user did not have permission to list storage classes - this PR fixes both.

The linked issue describes steps to reproduce, but to test:

- As Admin, create a new Standard User
- Create a new Cluster Role named 'Custom Storage' and add Verbs '*' and Resource 'persistentvolumeclaims'
- Add the new user as a Cluster Member and give them the 'Custom Storage' Role
- Log in as the new user
- Validate that when you go to Create a PVC, both the options show input boxes and not select options and that they both show a tooltip
- As Admin, edit the 'Custom Storage' and give the user permissions to list storage classes.
- Login as the user
- Validate that when creating a PVC, the storage class option now lists the storage classes (since the user now has access to do so)
- Repeat the above for persistent volumes

How it looks:

image

image

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -3593,6 +3593,9 @@ persistentVolumeClaim:
     storageClass: Storage Class
     requestStorage: Request Storage
     persistentVolume: Persistent Volume
+    tooltips:
+      noStorageClass: You don't have permission to list Storage Classes, enter a name manually
+      noPersistentVolume: You don't have permission to list Persistent Volumes, enter a name manually
   customize:
     label: Customize
     accessModes:
--- a/shell/edit/persistentvolumeclaim.vue
+++ b/shell/edit/persistentvolumeclaim.vue
@@ -5,6 +5,7 @@ import CruResource from '@shell/components/CruResource';
 import NameNsDescription from '@shell/components/form/NameNsDescription';
 import Tab from '@shell/components/Tabbed/Tab';
 import { RadioGroup } from '@components/Form/Radio';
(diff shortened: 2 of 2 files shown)


jordojordo (approved):
LGTM - tested and the inputs change as intended

PR #6676, opened 2022-08-17
Fix for 'Description' field for Rancher roles not editable or 'viewable'

### LINKED ISSUE #6585: Description field for Rancher roles not editable or 'viewable'
Internal reference: SURE-4824
Reported in 2.6.4 & 2.6.5

Issue description:
Rancher roles were created some time back. We are able to add and edit the resources available to these roles, but are not able to add to or update the Description field.

We add the text, click "Save" and it appears to take the changes. But after logging out, the changes are gone and other users are not able to see the change either. 

It does show up in YAML but if you 'create' a new one then you can 'see' the description. However, if you log out and log back in it disappears. 

It is 'similar' to the following 
https://github.com/rancher/dashboard/issues/5568

Troubleshooting steps:
Confirm that it was created by 'checking' YAML

Repro steps:

- Install Rancher 2.6.4 or 2.6.5 (Helm) 
- Create a 'new' Roles -> 'Create Project/Namespaces Role'
- Fill out the form 
- Log out and log back in

Workaround:
None

Actual behavior:
The 'description' field is NOT shown. 

Expected behavior:
The 'description' filed is SHOWN.

nwmac:
Fixes #6585

Resources that have a `description` stored at the top level of the resource object and that use the Steve model don't work - this PR adds a new Model that supports this pattern and uses it for the role template resource.

To Test:

- Create a new Role Template and set a description
- Reload the dashboard in the browser
- Click on the name of the new role template in the list and verify that the description is shown in the detail masthead
- From the action menu, edit the config and verify that the description is populated
- Change the description and save the resource
- View the resource again and verify that the description is correct
- Refresh the browser and verify that the description is still correct

### DIFF
--- a/shell/models/management.cattle.io.roletemplate.js
+++ b/shell/models/management.cattle.io.roletemplate.js
@@ -2,7 +2,7 @@ import Vue from 'vue';
 import { get } from '@shell/utils/object';
 import { DESCRIPTION } from '@shell/config/labels-annotations';
 import { NORMAN } from '@shell/config/types';
-import SteveModel from '@shell/plugins/steve/steve-class';
+import SteveDescriptionModel from '@shell/plugins/steve/steve-description-class';
 import Role from './rbac.authorization.k8s.io.role';
 
 export const CATTLE_API_GROUP = '.cattle.io';
@@ -55,7 +55,7 @@ export const VERBS = [
   'watch',
 ];
 
-export default class RoleTemplate extends SteveModel {
+export default class RoleTemplate extends SteveDescriptionModel {
   get availableActions() {
     const out = super._availableActions;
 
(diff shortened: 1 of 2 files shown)


richard-cox (approved):
I dug a bit more into the mystery of _name/_description issue. in 2.6.2 with the resource-proxy there was a `remapSpecialKeys` value that mapped all of the properties listed in (the new) `potentialNormanHoldovers` to their underscore equivalents. That was called when the proxy was created and in the store `load` function. Looks like the post 2.6.2 norman-class has workarounds for labels and annotations, but not name or description. description works somehow, but not name.

This PR still looks good though, as it's a steve resource with a non-standard description location

PR #6684, opened 2022-08-18
Remove window.v debugging code

nwmac:
Removes code that stores resources on the window object for debugging. The object being store won't get garbage collected.


richard-cox (approved):
We might want to nuke `shell/plugins/lookup.js` if no-one uses it (`window.$` is on there)

nwmac:
@richard-cox I resolved the conflict and removed lookup.js

richard-cox (approved):

PR #6700, opened 2022-08-19
allow for console.log and console.info to have correct traceability on dev mode

### LINKED ISSUE #6699: Recover stack trace for console.log and console.info on dev mode
- With changes introduced with the implementation of the diagnostics page, we have lost traceability for `console.log` and `console.info` when developing Rancher Dashboard.
The goal is to prevent the console plugin from capturing these events when app is on dev mode.

Test:
- make sure we get the correct line number when calling `console.log` and `console.info` when we are developing the application

aalves08:
Fixes #6699

- allow for `console.log` and `console.info` to have correct traceability on dev mode

### DIFF
--- a/shell/plugins/console.js
+++ b/shell/plugins/console.js
@@ -1,12 +1,17 @@
 /* eslint-disable no-console */
-export default (context) => {
-  const logTypes = ['log', 'error', 'info', 'warn'];
+export default () => {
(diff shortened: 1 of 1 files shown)


richard-cox (approved):
LGTM, though please ensure the `fixes` issue is updated before merging (it doesn't point to the right one)

PR #5234, opened 2022-02-28
Toggle Menu & Install Monitoring for Rancher Desktop build

rak-phillip:
This toggles UI elements related to the top level menu and the prompt to install monitoring for Rancher Desktop builds. This PR also groups the menu spacer and product sections into a generic grid area named `rd-header-left` to simplify the task of adapting the header for toggled elements.

There should be no discernible changes between this PR and current state unless the environment variable `RANCHER_ENV=desktop` is set at the time of build. If you see anything visual or behavioral changes, please consider them as a regression so that they can be fixed before merge.  

supports rancher-sandbox/rancher-desktop#1364

PR #6717, opened 2022-08-22
[Rancher Desktop] Deactivate Menu & Install Monitoring

rak-phillip:
### Summary

This toggles UI elements related to the top level menu and the prompt to install monitoring for Rancher Desktop builds. This PR also groups the menu spacer and product sections into a generic grid area named `rd-header-left` to simplify the task of adapting the header for toggled elements.

There should be no discernible changes between this PR and current state unless the environment variable `RANCHER_ENV=desktop` is set at the time of build. If you see anything visual or behavioral changes, please consider them as a regression so that they can be fixed before merge.  

supports rancher-sandbox/rancher-desktop#1364

* Navigation header should not render main menu when `RANCHER_ENV=desktop`
* Install Monitoring should not render when `RANCHER_ENV=desktop`
* Header and Navigation should render without any regressions when `RANCHER_ENV=web`

previous PR #5234


richard-cox (changes requested):
I'll leave @nwmac to run through the changes to `shell/components/nav/Header.vue`, regarding the other parts though we've done a lot of work in https://github.com/rancher/dashboard/pull/6729 to remove the `if harvester/virtual show/hide something` pattern. A lot of this was pushed in to the `isSingleProduct` object with generic flags and is used by both harvester and epinio. Once that PR merges (hopefully this week) it would be good to look at this again. Specifically if we can enable the `isSingleProduct` for desktop and beef it up with the required flags. If not we might need a general way to achieve the same results

rak-phillip:
Closing this PR as it has gone stale

PR #6718, opened 2022-08-23
fix(ingress): check if the ingress class is available before requesting it

ly5156:
### Summary
Fixes #

1. check if the ingress class is available before requesting it and give appropriate defaults when not available

https://github.com/rancher/rancher/issues/38695


mantis-toboggan-md (approved):
lgtm

PR #6723, opened 2022-08-23
Handle rests of the state when moving between steps

scures:
### Summary
Fixes  https://github.com/epinio/ui/issues/153

Small state resets when navigating between steps.
The issue was,  once we send data with `$emit('githubData')`, the state & step inside parent component `AppSource` will be valid. to fix this, when we reset an input field we also emit the event with the current `githubData` so the parent component can re-validate the step.  Added the revalidation for when the `GithubPicker` is mounted (so its resets on the `back` button) and for each input event for each step in the Picker.

The mentioned issue explains how to reproduce it, now it should be fixed.

### DIFF
--- a/pkg/epinio/components/application/AppSource.vue
+++ b/pkg/epinio/components/application/AppSource.vue
@@ -31,8 +31,8 @@ interface GitUrl {
 
 type GithubAPIData = {
   repos: any[],
-      branches: any[],
-      commits: any[]
+  branches: any[],
+  commits: any[]
 }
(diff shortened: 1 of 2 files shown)


richard-cox (changes requested):
If the user has supplied github values and then goes to step 2, we shouldn't lose those values when returning to step 1

richard-cox (approved):
I've rebased after updating `epinio-dev` (also fixed some minor alignment issues). That should get the PR passed gates

PR #6522, opened 2022-07-27
Documents hierarchy

bisht-richa:
Fixes #6334
- Created Documents hierarchy and updated some contents
- Added landing page for Devkit section
- Added storybook link

PR #6727, opened 2022-08-23
Added React reference in Docusaurus src files

bisht-richa:
### Summary
After merging https://github.com/rancher/dashboard/pull/6522 PR there was Docusaurus build issue.

Fixes https://github.com/rancher/dashboard/issues/6334

Fixed build issue for Docusaurus.

### DIFF
--- a/docusaurus/src/components/HomepageFeatures/index.js
+++ b/docusaurus/src/components/HomepageFeatures/index.js
@@ -1,3 +1,4 @@
+import React from 'react';
 import clsx from 'clsx';
 import styles from './styles.module.css';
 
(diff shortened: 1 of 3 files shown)


docusaurus/src/components/HomepageFeatures/index.js
```diff
@@ -1,3 +1,5 @@
+// eslint-disable-next-line no-use-before-define
```
cnotv, line 1: Please let's never add ignore to the linter unless is something planned or discussed.

cnotv (changes requested):
The issue seems caused by the preferences of ESLint extended to `/docusaurus/` path. 
To solve the false negative error you may extend the settings [at this line](https://github.com/rancher/dashboard/blob/master/.eslintrc.js#L183) with the following overriding:
```js
,
    {
      files:   ['docusaurus/**/*.{js,ts}'],
      rules:   { 'no-use-before-define': 'off' },
    }
```

We may need to review later on the definitions for the linter, so it will have and override rulesets for React.

cnotv (approved):
LGTM

Issue #6665, opened 2022-08-16
Change way supported version is detected


nwmac:
Currently we use the subscription ID setting.

Change this to use the metadata from the new `/rancherversion` endpoint - see: https://github.com/rancher/rancher/pull/38445/files

We should add a request to this API when the UI loads - before login and store that data. This will not change.

We can do this now and if we get a 404, we should populate empty data and store that the UI is not a supported deployment.

slickwarren:
I’m not sure what all should be tested besides testing that /rancherversion has a response. Could I get the QA template filled out for this one?
I currently get the following response: {"Version" : "<rancher_version>", "GitCommit" : "<gitcommit>", "RancherPrime" : "false"} is that all that is expected? Also, when using 2.7-head, the `Version` value is the same as GitCommit. 

here is my current test plan:
* test /rancherversion response on community version (see question above)
* test /rancherversion response on non-community version (tbt)
* possibly test upgrade from 2.6 (should have no response for rancherversion) -> 2.7

nwmac:
@slickwarren This was to make sure we wired up non-community detection to the new version endpoint.

To validate:

- As community, check that the browser makes a request to /rancherversion and the response indicates community. Verify on the UI that the home page shows the 'Commercial Support' link on the right hand panel

- As non-community, check that the browser makes a request to /rancherversion and the response indicates non-community. Verify on the UI that the home page does not show the 'Commercial Support' link on the right hand panel

Shavindra:
Further instructions: https://github.com/rancher/dashboard/pull/6736#issue-1349291691

@slickwarren 

Could also be resolved by:
PR
https://github.com/rancher/dashboard/pull/7131

Issues:
https://github.com/rancher/dashboard/issues/6663
https://github.com/rancher/dashboard/issues/6665

slickwarren:
Similar / related issue that may have been part of these changes: There is no longer an option on the support page to enter a subscription ID. Perhaps related to the way version is detected? Or in other words, perhaps another reason that commercial support option still appears (because subscription ID is not set)
![Screen Shot 2022-10-28 at 11 10 03 AM](https://user-images.githubusercontent.com/16691014/198704371-517f2d83-e5d8-48aa-b4cb-cf6d9bb68343.png)

nwmac:
@slickwarren 

[...]

PR #6736, opened 2022-08-24
Remove add subscription ID box #6663

### LINKED ISSUE #6663: Update Support Page
Changes required to the support page:

- Hide the button and text to add a subscription ID
- Remove the right-hand card 'Support'
- Move the 'Find out more ... or contact ..' link to the right hand area
- Ensure that the 'Get Support' box is only shown when commercial support is enabled

Shavindra:
### Summary
- #6663
  - Ensure that the 'Get Support' box is only shown when commercial support is enabled
  - Hide the button and text to add a subscription ID
- #6665

There should be a request to rancher with following endpoint
`/rancherversion` and should receive a response `{"Version":"b3c5e9f37","GitCommit":"b3c5e9f37","RancherPrime":"false"}
`

**1. Rancher community**
{"Version":"b3c5e9f37","GitCommit":"b3c5e9f37","RancherPrime":"false"}
  - Create Rancher with Rancher Prime enabled
  - Left hand navigation > Get Support (at the bottom)
  - User should not see "Get Support" box
  - User should see commercial links
  - Subscription ID box and 2 links shouldn't exist
image

**2. Rancher Prime**
`{"Version":"b3c5e9f37","GitCommit":"b3c5e9f37","RancherPrime":"true"}`
  - Create Rancher with Rancher Prime enabled
  - Left hand navigation > Get Support (at the bottom)
  - User should not see "Get Support" box
  - User should not see commercial links
  - Subscription ID box and 2 links shouldn't exist
image


nwmac:
@Shavindra This PR has conflicts - I assume the work to get the version number from the API will be part of this PR?

scripts/standalone/index.js
```diff
@@ -31,27 +31,28 @@ const dev = (process.env.NODE_ENV !== 'production');
...
+  '/k/':             proxyOpts('https://127.0.0.1:8000'),
+  '/g/':             proxyOpts('https://127.0.0.1:8000'),
+  '/n/':             proxyOpts('https://127.0.0.1:8000'),
+  '/p/':             proxyOpts('https://127.0.0.1:8000'),
+  '/assets':         proxyOpts('https://127.0.0.1:8000'),
+  '/translations':   proxyOpts('https://127.0.0.1:8000'),
+  '/engines-dist':   proxyOpts('https://127.0.0.1:8000'),
   // Plugin dev
-  '/verdaccio/':   proxyOpts('http://127.0.0.1:4873/-'),
+  '/verdaccio/':     proxyOpts('http://127.0.0.1:4873/-'),
```
nwmac, line 55: This needs rebasing

shell/components/CommunityLinks.vue
```diff
@@ -64,19 +68,20 @@ export default {
 
       // Custom links set from settings
       if (!!this.uiCustomLinks?.value) {
-        try {
-          const customLinks = JSON.parse(this.uiCustomLinks.value);
+        const val = JSON.parse(this.uiCustomLinks.value);
```
nwmac, line 71: The previous code looks more robust for parsing the value

shell/config/version.js
```diff
@@ -0,0 +1,9 @@
+let _isRancherPrime = false;
+
+export function isRancherPrime() {
+  return _isRancherPrime;
+}
+
+export function setIsRancherPrime(v) {
+  _isRancherPrime = v.isRancherPrime === 'True';
```
nwmac, line 8: This does not match what the AP returns - should be property `RancherPrime` and the value is a lowercase string (but can we please to a case insensitive check ... also should accept a boolean value as well)

shell/pages/c/_cluster/settings/links.vue
```diff
@@ -63,7 +64,12 @@ export default {
         uiCommunitySetting: this.$store.dispatch('management/find', { type: MANAGEMENT.SETTING, id: SETTING.COMMUNITY_LINKS })
       });
 
+      // const response = await version({ $axios: this.$axios });
```
nwmac, line 67: Remove this commented code

shell/pages/support/index.vue
```diff
@@ -102,8 +99,7 @@ export default {
     },
 
     hasSupport() {
-      // NB: This is temporary until API implemented
-      return false;
+      return isRancherPrime();
```
nwmac, line 102: I think we should use `hasSubscriptionSupport` everywhere and get rid of `hasSupport`

shell/plugins/version.js
```diff
@@ -0,0 +1,11 @@
+import { setIsRancherPrime } from '@shell/config/version';
+export default async function({ $axios }) {
+  try {
+    const response = await $axios.get('/rancherversion');
+
+    setIsRancherPrime(response.data);
+    // TODO: Set value using setIsRancherPrime
```
nwmac, line 7: Remove TODO

nwmac (changes requested):
Also notices that I can not get the links box to display on the home page -  I must have dismissed it previously and I can not get it back by using 'Restore hidden cards', so this seems broken and needs to be looked at.

nwmac (changes requested):
Please resolve conversations so I can see what you have fixed

scripts/standalone/index.js
Shavindra: It's already rebased. This is the standalone file. The changes in shell not applied here?

Shavindra:
> Also notices that I can not get the links box to display on the home page - I must have dismissed it previously and I can not get it back by using 'Restore hidden cards', so this seems broken and needs to be looked at.

This is because rancher/rancher is sending `{}` by default. So it treated as user didn't set any links by default. It should be fixed by this PR

https://github.com/rancher/rancher/pull/39186

cnotv:
I have a bunch of errors going on here when I try to access the page (left master, right this PR), but I could however login successfully.

Screenshot 2022-10-06 at 16 50 06

shell/plugins/version.js
```diff
@@ -0,0 +1,10 @@
+import { setIsRancherPrime } from '@shell/config/version';
+export default async function({ $axios }) {
+  try {
+    const response = await $axios.get('/rancherversion');
```
cnotv, line 4: This request seems to trigger the schema request from the error in the E2E.
There must be some race condition issues which are totally unknown from me. Please check it out with @nwmac or @richard-cox as they implemented most of it and may have an idea to avoid it.

cnotv: Just to keep updated also the PR, it seems like the issue was in somehow produced also by the E2E, as it pointed to `v2.6-head` generating schema errors.

cnotv:
It seems like the user is logged out for some reasons when he gets 404 in the index.json request as in the screenshot.
The error can be reproduced also manually on browser first load when login as in the screenshot, left PR and right master:
Screenshot 2022-10-07 at 18 27 48

Shavindra:
Closed in favour of  #7131

PR #6740, opened 2022-08-24
Use `@rancher/components` in imports

rak-phillip:
### Summary
Fixes #


rak-phillip:
Closing this PR as it has gone stale

PR #6751, opened 2022-08-25
Make `ToggleSwitch` state react to changes from parent

### LINKED ISSUE #6741: [@rancher/components] ToggleSwitch state
ToggleSwitch.vue accepts a `value` prop and uses it as an initial value to keep track of state locally in the `state` data prop. Using this pattern means that any updates to the `value` props will not be reflected in the rendered output. 

https://github.com/rancher/dashboard/blob/6ba2a191b04e9b9ef66635e12a2062ce46f6801d/pkg/rancher-components/src/components/Form/ToggleSwitch/ToggleSwitch.vue#L29-L31

rak-phillip:
### Summary
This adds a watch to the `value` prop so that changes will be reflected locally in the `state` data prop when changes come through.

The current behavior subverts expectations by only using the `value` prop to set an initial value. This pattern of managing state internally doesn't align with how I would expect a base component to behave.  

Fixes #6741 

* Makes `value` prop reactive
* Adds unit tests for `ToggleSwitch`

`provisioning.cattle.io.cluster` makes use of the toggle switch for switching between RKE and RKE2. Behavior on this page should not regress.

### DIFF
--- a/pkg/rancher-components/src/components/Form/ToggleSwitch/ToggleSwitch.test.ts
+++ b/pkg/rancher-components/src/components/Form/ToggleSwitch/ToggleSwitch.test.ts
@@ -36,7 +36,7 @@ describe('ToggleSwitch.vue', () => {
     expect(toggleInput.checked).toBe(true);
   });
 
-  it('emits an input event with default values when clicked', async () => {
+  it('emits an input event with a true value', async () => {
     const wrapper = shallowMount(ToggleSwitch);
 
     (wrapper.vm as any).toggle(true);
@@ -47,25 +47,35 @@ describe('ToggleSwitch.vue', () => {
     expect(wrapper.emitted().input?.length).toBe(1);
     expect(wrapper.emitted().input?.[0][0]).toBe(true);
 
+  });
+
+  it('emits an input event with a false value', async () => {
+    const wrapper = shallowMount(
+      ToggleSwitch,
+      {
+        propsData: {
+          value: true
+        }
+      }
+    );
+
     (wrapper.vm as any).toggle(false);
 
     await wrapper.vm.$nextTick();
 
     expect(wrapper.emitted('input')).toBeTruthy();
-    expect(wrapper.emitted().input?.length).toBe(2);
-    expect(wrapper.emitted().input?.[1][0]).toBe(false);
-  });
+    expect(wrapper.emitted().input?.length).toBe(1);
+    expect(wrapper.emitted().input?.[0][0]).toBe(false);
+  })
 
-  it('emits an input event with custom values when clicked', async () => {
+  it('emits an input event with a custom onValue', async () => {
     const onValue = 'THE TRUTH';
-    const offValue = 'NOT THE TRUTH';
 
     const wrapper = shallowMount(
       ToggleSwitch,
       {
         propsData: {
           onValue,
-          offValue,
         }
       });
 
@@ -76,13 +86,26 @@ describe('ToggleSwitch.vue', () => {
     expect(wrapper.emitted().input).toBeTruthy()
     expect(wrapper.emitted().input?.length).toBe(1);
     expect(wrapper.emitted().input?.[0][0]).toBe(onValue);
+  });
+
+  it('emits an input event with a custom offValue', async () => {
+    const offValue = 'NOT THE TRUTH';
+
+    const wrapper = shallowMount(
+      ToggleSwitch,
+      {
+        propsData: {
+          value: true,
+          offValue,
+        }
+      });
 
     (wrapper.vm as any).toggle(false);
 
     await wrapper.vm.$nextTick();
 
     expect(wrapper.emitted('input')).toBeTruthy();
-    expect(wrapper.emitted().input?.length).toBe(2);
-    expect(wrapper.emitted().input?.[1][0]).toBe(offValue);
-  });
+    expect(wrapper.emitted().input?.length).toBe(1);
+    expect(wrapper.emitted().input?.[0][0]).toBe(offValue);
+  })
 });
--- a/pkg/rancher-components/src/components/Form/ToggleSwitch/ToggleSwitch.test.ts
+++ b/pkg/rancher-components/src/components/Form/ToggleSwitch/ToggleSwitch.test.ts
@@ -10,7 +10,7 @@ describe('ToggleSwitch.vue', () => {
     expect(toggleInput.checked).toBeFalsy();
   });
 
-  it('renders truthy', () => {
+  it('renders a true value', () => {
     const wrapper = shallowMount(
       ToggleSwitch,
       {
@@ -21,7 +21,7 @@ describe('ToggleSwitch.vue', () => {
 
     const toggleInput = wrapper.find('input[type="checkbox"]').element as HTMLInputElement
 
-    expect(toggleInput.checked).toBeTruthy();
+    expect(toggleInput.checked).toBe(true);
   });
 
   it('updates from falsy to truthy when props change', async () => {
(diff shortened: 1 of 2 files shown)


pkg/rancher-components/src/components/Form/ToggleSwitch/ToggleSwitch.test.ts
```diff
@@ -0,0 +1,89 @@
...
+  });
+
+  it('emits an input event with default values when clicked', async () => {
+    const wrapper = shallowMount(ToggleSwitch);
+
+    (wrapper.vm as any).toggle(true);
+
+    await wrapper.vm.$nextTick();
+
+    expect(wrapper.emitted().input).toBeTruthy()
```
cnotv, line 46: We should have different tests for different assertions. Also the first assertion may not be required as the second and third may cover it.

pkg/rancher-components/src/components/Form/ToggleSwitch/ToggleSwitch.test.ts
```diff
@@ -0,0 +1,89 @@
...
+    (wrapper.vm as any).toggle(false);
+
+    await wrapper.vm.$nextTick();
+
+    expect(wrapper.emitted('input')).toBeTruthy();
+    expect(wrapper.emitted().input?.length).toBe(2);
+    expect(wrapper.emitted().input?.[1][0]).toBe(false);
+  });
+
+  it('emits an input event with custom values when clicked', async () => {
```
cnotv, line 59: I am not so sure why are we passing a string for a toggler.

pkg/rancher-components/src/components/Form/ToggleSwitch/ToggleSwitch.vue
```diff
@@ -27,14 +27,23 @@ export default {
...
     toggle(neu) {
       this.state = neu === null ? !this.state : neu;
       this.$emit('input', this.state ? this.onValue : this.offValue);
     }
+  },
+
+  watch: {
+    value: {
+      handler() {
+        this.state = this.value === this.onValue;
```
cnotv, line 43: Why do we have a duplication of the `value` as `state`? What is the benefit? We may probably omit code duplication.

cnotv (changes requested):

pkg/rancher-components/src/components/Form/ToggleSwitch/ToggleSwitch.vue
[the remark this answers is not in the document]
rak-phillip: These are good questions, but I cannot answer a lot of the `why` as I'm not the original author of this component..  

> Why do we have an `onValue` and `offValue` and why the type is `null` when it is a `Boolean`?

`onValue` and `offValue` seem to allow for other values than `boolean`. I'm guessing the type is `null` when the original author intended to allow for `any`. I see this usage of `null` as a type elsewhere throughout our components. 

> Why is it required to have a watch and a state, something which may be simply set as a prop and updated as emit?

I asked myself the same question but opted to keep the original implementation intact. With this original implementation in place, `state` uses the `value` prop as an initial value and nothing else. `state` is then solely managed internally by the `ToggleSwitch` after this initial. 

> The logic used from the parent should not be passed down to a component, that's against the separations of concerns.

I agree, but I don't understand why you're making this point..

> We may considering removing this logic or create an issue for it while explaining the purpose inside the component.

Much agreed. I think it makes sense to follow up with a refactor of this component. And getting tests in place will help us in this process. Please remember that we are incrementally improving what already exists in production.

rak-phillip: Once again, I agree. I'm not quite too sure what the benefits are for having this component manage its own state like this. On the other hand, I'm trying to manage the scope of changes to make reduce the change of inadvertently introducing regressions.

pkg/rancher-components/src/components/Form/ToggleSwitch/ToggleSwitch.test.ts
rak-phillip: We are passing a string because this is what the component supports. We need to test our public contracts as they are..

pkg/rancher-components/src/components/Form/ToggleSwitch/ToggleSwitch.test.ts
rak-phillip: These tests are simple enough that it's easy to split. 

I think it makes sense to remove the first assertion as well.

pkg/rancher-components/src/components/Form/ToggleSwitch/ToggleSwitch.test.ts
```diff
@@ -0,0 +1,89 @@
+import { shallowMount } from '@vue/test-utils';
+import { ToggleSwitch } from './index';
+
+describe('ToggleSwitch.vue', () => {
```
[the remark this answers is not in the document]
rak-phillip: Fair enough, fixed.

pkg/rancher-components/src/components/Form/ToggleSwitch/ToggleSwitch.vue
rak-phillip: With that said, I think this is a simple enough component with a limited use that it would be easy to regression test such a refactor. I just prefer to err on the side of caution and approach problems through extending existing implementations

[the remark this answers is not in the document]
rak-phillip: I think it's also good to mention that we already make use of constructs like `onValue` and `offValue` in `provisioning.cattle.io.cluster`: 

https://github.com/rancher/dashboard/blob/44fd817f5bfbbeac183aaa09e86cdde5b5cdbbec/shell/edit/provisioning.cattle.io.cluster/index.vue#L466-L474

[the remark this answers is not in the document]
cnotv: Sorry for having so many questions, just I usually tend to do it when it's not documented 😓 
I personally refactor small part of code when I work on it, as that's pretty much the reason we have to extend the code in this case too.
However, it's up to you to make a decision, as it just wanted to be a suggestion for simplify the logic.
Could we add a couple of notes, especially considering it has to be published as library and out of context may be confusing?

cnotv: Ideally the scope of the test is to have certainty of the result, which gives us freedom to refactor it.
Sorry for the pressure, I'll avoid next time 😅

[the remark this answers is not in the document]
cnotv: Since I've seen you added the type, it may probably suffice.

cnotv (approved):
LGTM

pkg/rancher-components/src/components/Form/ToggleSwitch/ToggleSwitch.vue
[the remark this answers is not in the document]
rak-phillip: No apologies needed. I said it before and I said it again, I think you're asking good questions 😁

There's always a need for documentation where it is lacking, which is unfortunately the case for a lot of our components. I've created #6759 to follow up with some more comments in our code.

There's definitely a strong case for refactoring here and perhaps I should have approached this from the angle of writing tests first and then following up with a separate PR to address behavior. Maybe this would give me more confidence to make a bigger change.. Anyhow, I've created #6760 to circle back and improve the watcher implementation.

rak-phillip: My apologies. It's no my intent to discourage discourse, I will try to check my written tone next time. Perhaps it was too late in the day and that was causing me to be too terse 😅

Please, ask questions 😁

cnotv: All good and also thanks 😄 🍻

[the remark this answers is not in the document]
cnotv: Fair enough for me. Thanks for writing it down in issues so we don't lose track 👍

PR #6756, opened 2022-08-26
Shows fetched value for the default builder

scures:
### Summary
Fixes #https://github.com/epinio/ui/issues/154

1. We now fetch the default paketo packager from the `/info` endpoint instead of hardcoding it.
2. There was an issue where if you manually set another package, it would not respect it and it will use the default value.

1. From the UI, create a new Application from git (`https://github.com/scures/epinio-sample-app` and `master` branch is a good example). 
2. In advance settings for the paketo image, select custom and type `paketobuildpacks/builder:tiny`
3. Follow the rest of the steps.

This app won't deploy with the tiny version, it's fine. We need to validate that the package used for the build was actually the `builder:tiny` and not `builder:full`. To do so, one handy command would be:
`kubectl  get apps.application.epinio.io -n {{NAMESPACE}} {{APP_NAME}} -o yaml`
The outcome should look as:

```
spec:
  // .......
  builderimage: paketobuildpacks/builder:tiny
  // .......
```

### DIFF
--- a/pkg/epinio/components/application/AppSource.vue
+++ b/pkg/epinio/components/application/AppSource.vue
@@ -67,7 +67,7 @@ interface FileWithRelativePath extends File {
    readonly webkitRelativePath: string;
 }
 
-const DEFAULT_BUILD_PACK = 'paketobuildpacks/builder:full';
+const DEFAULT_BUILD_PACK = 'paketobuildpacks/builder:test';
 
 // Data, Methods, Computed, Props
 export default Vue.extend<Data, any, any, any>({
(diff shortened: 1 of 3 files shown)


mmartin24:
I did a check on this branch and the custom Packeto Images seem to be recognized:

I tried with  `paketobuildpacks/builder:tiny` and `paketobuildpacks/builder:base` and they both seem to be captured
image

```
➜  epinio git:(main) ✗ kubectl  get apps.application.epinio.io -n workspace test-base-packeto -o yaml
apiVersion: application.epinio.io/v1
kind: App
metadata:
  annotations:
    epinio.io/created-by: admin
  creationTimestamp: "2022-08-26T10:27:37Z"
  generation: 4
  name: test-base-packeto
  namespace: workspace
  resourceVersion: "2392"
  uid: 982f7d56-41b2-4c7b-acdd-1ecddf999d9d
spec:
  blobuid: fea84488-e637-44c7-aaf8-40db8d0c6e59
  builderimage: paketobuildpacks/builder:base
  chartname: standard
  imageurl: registry.epinio.svc.cluster.local:5000/apps/workspace-test-base-packeto:2cdb9617f0cd73fc
  origin:
    git:
      repository: https://github.com/epinio/example-go
      revision: main
  routes:
  - test-base-packeto-a279f.172.19.0.3.omg.howdoi.website
  stageid: 2cdb9617f0cd73fc

```
@scures the app seems to work as well, it just needed some custom env vars during the app deployment you probably were not aware of on the `details` step: Name=`BP_KEEP_FILES`  and Value=`static/*` 
image

PR #6757, opened 2022-08-26
Display warning when custom role is deleted

### LINKED ISSUE #6578: Add UI/API warning when removing a custom role that is in use
Internal reference: SURE-4773
Reported in 2.6.3

Request description:
An admin had removed a project role that inadvertently orphaned access for a number of users. It was a manageable amount so they could resolve it with some corrective effort; however, the takeaway was that there might be an improvement to add a UI warning/guard rail -- some visual reference to indicate if it’s in use etc., -- or at least some documentation notes about this being a potentially destructive action.

Actual behavior:
Project role was deleted, orphaning access for users that were added as members. Some documentation around role lifecycle would also be ideal.

Expected behavior:
Visual indicator of custom roles that are being used, a warning should prevent deletion without explicit acceptance.

Workaround:
Maybe, perhaps not so easy: manually check role bindings?

aalves08:
Fixes #6578 

- Adds custom message with warning if a custom role being deleted is currently binded to any user

 **Screenshots**
_checking in progress_
Screenshot 2022-09-14 at 14 15 50

_binded users_
Screenshot 2022-09-14 at 14 16 51

__no users binded__
Screenshot 2022-09-14 at 14 30 27

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -3957,17 +3957,17 @@ rbac:
       noContext:
         label: No Context
   globalRoles:
-    noBinding: 'No user bindings found <i class="icon icon-checkmark" style="margin-left: 5px"></i>'
-    unableToCheck: Unable to check user bindings. Please try again.
+    notBound: 'No users bound <i class="icon icon-checkmark" style="margin-left: 5px"></i>'
+    unableToCheck: Unable to check if any user is bound to the role(s). Please try again.
     waiting: |-
       {count, plural,
-        =1 { Checking if there are any users binded to this role <i class="icon-spin icon icon-spinner" style="margin-left: 5px"></i> }
-        other { Checking if there are any users binded to these roles <i class="icon-spin icon icon-spinner" style="margin-left: 5px"></i> }
+        =1 { Checking if there are any users bound to this role <i class="icon-spin icon icon-spinner" style="margin-left: 5px"></i> }
+        other { Checking if there are any users bound to these roles <i class="icon-spin icon icon-spinner" style="margin-left: 5px"></i> }
       }
-    usersBinded: |-
+    usersBound: |-
       {count, plural,
-        =1 { Caution: There is 1 user binded to the roles about to be deleted. Do you want to proceed? }
-        other { Caution: There are {count} users binded to the roles about to be deleted. Do you want to proceed? }
+        =1 { Caution: There is 1 user bound to the role(s) about to be deleted. Do you want to proceed? }
+        other { Caution: There are {count} users bound to the role(s) about to be deleted. Do you want to proceed? }
       }
     types:
       global:
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -3957,6 +3957,11 @@ rbac:
       noContext:
         label: No Context
   globalRoles:
+    usersBinded: |-
+      {count, plural,
+        =1 { Caution: There is 1 user binded to this role! }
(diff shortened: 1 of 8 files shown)


nwmac:
My main concern with this is at scale we will retrieve a large number of bindings.

shell/promptRemove/mixin/roleDeletionCheck.js
```diff
@@ -0,0 +1,91 @@
...
+      switch (resourceType) {
+      case MANAGEMENT.GLOBAL_ROLE:
+        resourceToCheck = MANAGEMENT.GLOBAL_ROLE_BINDING;
+        propToMatch = 'globalRoleName';
+        break;
+      default:
+        resourceToCheck = MANAGEMENT.CLUSTER_ROLE_TEMPLATE_BINDING;
+        propToMatch = 'roleTemplateName';
+        break;
+      }
```
jordojordo, line 55: Should we be checking for `MANAGEMENT.PROJECT_ROLE_TEMPLATE_BINDING` as well?

shell/promptRemove/mixin/roleDeletionCheck.js
```diff
@@ -0,0 +1,91 @@
...
+
+      try {
+        const data = await this.$store.dispatch('management/request', {
```
jordojordo, line 58: This is a nitpick, but could we call this variable something other than `data`? I just have a rough time when trying to understand what `data.data` is referring to.

jordojordo (changes requested):
Looks mostly good to me, I had a couple of changes. Deleting Global and Cluster roles worked as expected, although I was running into some strangeness when deleting project roles.

Forgive me if I'm mistaken on the issue, but when deleting a project role ( as an admin ) I should be able to see the users that are bound to the project role. However, the check would appear to run and returned no users for project roles that I had bound to a couple of users.

**Project members**
proj-members

**Deleting role `my-project-role`**
role-1

**Deleting role `my-other-proj-role`**
role-2

shell/assets/translations/en-us.yaml
```diff
@@ -3957,6 +3957,18 @@ rbac:
...
+    unableToCheck: Unable to check user bindings. Please try again.
+    waiting: |-
+      {count, plural,
+        =1 { Checking if there are any users binded to this role <i class="icon-spin icon icon-spinner" style="margin-left: 5px"></i> }
+        other { Checking if there are any users binded to these roles <i class="icon-spin icon icon-spinner" style="margin-left: 5px"></i> }
+      }
+    usersBinded: |-
+      {count, plural,
+        =1 { Caution: There is 1 user binded to the roles about to be deleted. Do you want to proceed? }
+        other { Caution: There are {count} users binded to the roles about to be deleted. Do you want to proceed? }
```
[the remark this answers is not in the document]
aalves08: updated translation strings according to your feedback. Thanks!

shell/promptRemove/mixin/roleDeletionCheck.js
aalves08: true! I missed that! Thanks for the heads up 🙏 🙇  Fix has been added and you now should see the correct check when deleting `Project Members`

aalves08:
> Looks great! Performs as expected 🧙
> 
> One thing I noticed ( although it's nothing serious ) that may need to be accomplished in another issue - After deleting a project role, a Project Member that was assigned the deleted role will linger within the Project config page with an empty Role.
> 
> deleted

We should then create an issue out of this to be evaluated if we should remove the users as well. Thanks for the review @jordojordo

PR #6781, opened 2022-08-31
Checkbox to accept strings for valueWhenTrue - ec2 httpEndpoint, httpTokens to string values

### LINKED ISSUE #6573: [BUG] RKE2 `Use tokens for metadata` and `Allow access to EC2 metadata` options do not work
Root issue: https://github.com/rancher/rancher/issues/38405
**Setup**
- Rancher version: v2.6.7 and older
- Browser type & version: test on latest Chrome, but applicable to all browsers

**Describe the bug**
When creating RKE2 clusters which must adhere to a strict [IMDSv2 AWS policy](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html) both the `Use tokens for metadata` and `Allow access to EC2 metadata` check boxes must be selected for the instance to provision successfully. This issue represents two bugs regarding the selection of those check boxes 

1. The `Use tokens for metadata` check box is initially grayed out, implying that it is enabled by default, however I cannot see that field sent in the POST request when provisioning a cluster. When you select `Allow access to EC2 metadata` the `Use tokens for metadata` box will be selectable, but this leads to bug 2

2. When enabling `Allow access to EC2 metadata` or `Use tokens for metadata` the UI will report the following type error.
```
Amazonec2Config.rke-machine-config.cattle.io "nc-haffel-button-tests-pool1-jqxsr" is invalid: [httpTokens: Invalid value: "boolean": httpTokens in body must be of type string: "boolean", httpEndpoint: Invalid value: "boolean": httpEndpoint in body must be of type string: "boolean"]
```
This error occurs when one box is selected or when both boxes are selected (above error is when both are selected). When this error is encountered the cluster will not begin provisioning. 

**To Reproduce**
+ Open Cluster-management 
+ Click Create 
+ Enable `RKE2/K3s`
+ Select `Amazon EC2`
+ Create a cloud-credential if needed
+ Under `VPC/Subnet` click `Show Advanced`
+ Observe that `Use tokens for metadata` is grayed out
+ Select `Allow access to EC2 metadata`
   + Observe that `Use tokens for metadata` is no longer grayed out 
+ Select either `Allow access to EC2 metadata`, `Use tokens for metadata` or both
+ Add a Cluster Name / VPC and press Create

**Result**

Observe error message at the top of the page stating the error message mentioned above. 

**Expected Result**
The error message does not appear and the cluster begins provisioning

**Screenshots**

Screen Shot 2022-08-02 at 11 05 26 AM

**Additional context**
IMDSv2 is a security policy used by several of our customers, instances will fail to create if both the `httpToken` and `httpEndpoint` options are not set. An issue was discovered in RKE1 where these fields were not set when creating spot-instances. This error was discovered while testing similar functionality for RKE2.

jordojordo:
### Summary
Fixes #6573 

This will change the `Checkbox` component to accept strings for the `valueWhenTrue` property, and converts ec2 `httpEndpoint` and `httpTokens` checkboxes to string values.

According to [the spec](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-options.html) of an ec2 instance, `httpEndpoint` can have a value of `enabled` or `disabled`, and `httpTokens` can have a value of `required` or `optional`. In our case, if the checkbox for either is unchecked the value with return to `undefined`.

In order to supply a string for each value the `Checkbox` component needed some logic to allow this behavior, along with the added `valueWhenTrue` prop.

### Technical notes summary

`Checkbox` will now accept a string as the value when passed the `valueWhenTrue` prop. If a checkbox is unchecked it will revert the value to `undefined`.

- Steps for testing are in the issue listed above
- Other areas where the Checkbox component is used

There are only two other instances which the `valueWhenTrue` is used:
- https://github.com/rancher/dashboard/blob/master/shell/edit/provisioning.cattle.io.cluster/rke2.vue#L1791
- https://github.com/rancher/dashboard/blob/master/shell/components/GlobalRoleBindings.vue#L317

Where these two files are being used should be tested as well.

### DIFF
--- a/pkg/rancher-components/src/components/Form/Checkbox/Checkbox.test.ts
+++ b/pkg/rancher-components/src/components/Form/Checkbox/Checkbox.test.ts
@@ -2,11 +2,11 @@ import { shallowMount } from '@vue/test-utils';
 import { Checkbox } from './index';
 
 describe('Checkbox.vue', () => {
-  it('renders falsy by default', () => {
+  it('is unchecked by default', () => {
     const wrapper = shallowMount(Checkbox);
     const cbInput = wrapper.find('input[type="checkbox"]').element as HTMLInputElement;
 
-    expect(cbInput.checked).toBeFalsy();
+    expect(cbInput.checked).toBe(false);
   });
 
   it('renders a true value', () => {
@@ -56,7 +56,7 @@ describe('Checkbox.vue', () => {
     await wrapper.vm.$nextTick();
 
     expect(wrapper.emitted().input?.length).toBe(1);
-    expect(wrapper.emitted().input?.[0][0]).toBeTruthy();
+    expect(wrapper.emitted().input?.[0][0]).toBe(valueWhenTrue);
   });
 
   it('updates from valueWhenTrue to falsy', async () => {
@@ -72,6 +72,6 @@ describe('Checkbox.vue', () => {
     (wrapper.vm as any).clicked(event);
     await wrapper.vm.$nextTick();
 
-    expect(wrapper.emitted().input?.[0][0]).toBeFalsy();
+    expect(wrapper.emitted().input?.[0][0]).toBe(null);
   })
 });
--- a/pkg/rancher-components/src/components/Form/Checkbox/Checkbox.test.ts
+++ b/pkg/rancher-components/src/components/Form/Checkbox/Checkbox.test.ts
@@ -1,0 +1,77 @@
+import { shallowMount } from '@vue/test-utils';
+import { Checkbox } from './index';
+
+describe('Checkbox.vue', () => {
+  it('renders falsy by default', () => {
+    const wrapper = shallowMount(Checkbox);
+    const cbInput = wrapper.find('input[type="checkbox"]').element as HTMLInputElement;
+
+    expect(cbInput.checked).toBeFalsy();
+  });
+
+  it('renders a true value', () => {
+    const wrapper = shallowMount(Checkbox, { propsData: { value: true } });
+    const cbInput = wrapper.find('input[type="checkbox"]').element as HTMLInputElement;
+
+    expect(cbInput.checked).toBe(true);
+  });
+
+  it('updates from false to true when props change', async () => {
+    const wrapper = shallowMount(Checkbox);
+    const cbInput = wrapper.find('input[type="checkbox"]').element as HTMLInputElement;
+
+    expect(cbInput.checked).toBe(false);
+
+    await wrapper.setProps({ value: true });
+
+    expect(cbInput.checked).toBe(true);
+  });
+
+  it('emits an input event with a true value', async () => {
+    const wrapper = shallowMount(Checkbox);
+    const event = {
+      target: { tagName: 'input', href: null },
+      stopPropagation: () => { },
+      preventDefault: () => { }
+    };
+
+    (wrapper.vm as any).clicked(event);
+    await wrapper.vm.$nextTick();
+
+    expect(wrapper.emitted().input?.length).toBe(1);
+    expect(wrapper.emitted().input?.[0][0]).toBe(true);
+  });
+
+  it('emits an input event with a custom valueWhenTrue', async () => {
+    const valueWhenTrue = 'BIG IF TRUE';
+    const event = {
+      target: { tagName: 'input', href: null },
+      stopPropagation: () => { },
+      preventDefault: () => { }
+    };
+
+    const wrapper = shallowMount(Checkbox, { propsData: { value: false, valueWhenTrue } });
+
+    (wrapper.vm as any).clicked(event);
+    await wrapper.vm.$nextTick();
+
+    expect(wrapper.emitted().input?.length).toBe(1);
+    expect(wrapper.emitted().input?.[0][0]).toBeTruthy();
+  });
+
+  it('updates from valueWhenTrue to falsy', async () => {
+    const valueWhenTrue = 'REAL HUGE IF FALSE';
+    const event = {
+      target: { tagName: 'input', href: null },
+      stopPropagation: () => { },
+      preventDefault: () => { }
+    };
+
+    const wrapper = shallowMount(Checkbox, { propsData: { value: valueWhenTrue, valueWhenTrue } });
+
+    (wrapper.vm as any).clicked(event);
+    await wrapper.vm.$nextTick();
+
+    expect(wrapper.emitted().input?.[0][0]).toBeFalsy();
+  })
+});
--- a/pkg/rancher-components/src/components/Form/Checkbox/Checkbox.vue
+++ b/pkg/rancher-components/src/components/Form/Checkbox/Checkbox.vue
@@ -9,7 +9,7 @@ export default Vue.extend({
      * The checkbox value.
      */
     value: {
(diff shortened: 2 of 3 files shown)


pkg/rancher-components/src/components/Form/Checkbox/Checkbox.vue
```diff
@@ -169,6 +169,12 @@ export default Vue.extend({
           addObject(this.value, this.valueWhenTrue);
         }
         this.$emit('input', this.value);
+      } else if (this.isString(this.valueWhenTrue)) {
+        if (this.isChecked) {
+          this.$emit('input', undefined);
+        } else {
+          this.$emit('input', this.valueWhenTrue);
+        }
```
rak-phillip, line 177: note to self: This gave me some pause but it's, in fact, the correct implementation. The entire containing code block is intended to "flip the checkbox value". I think this can be simplified in a refactoring in the future.

pkg/rancher-components/src/components/Form/Checkbox/Checkbox.vue
```diff
@@ -169,6 +169,12 @@ export default Vue.extend({
           addObject(this.value, this.valueWhenTrue);
         }
         this.$emit('input', this.value);
+      } else if (this.isString(this.valueWhenTrue)) {
+        if (this.isChecked) {
+          this.$emit('input', undefined);
```
rak-phillip, line 174: I'm not sure how I feel about `undefined` to represent the absence of a value in this case.. maybe `null` or an empty string would be more appropriate. Neither feel great to me, what do you think?

We also don't have the concept of a `valueWhenFalse` for this component. Looking into these change makes me feel strongly about the need to revisit this component as a whole further down the line.

rak-phillip (commented):
LGTM - I'd like for use to consider if there's a better value than `undefined` for when the checkbox is unchecked.

I created issue #6787 so we can circle back in the future to see how we might be able to improve the overall design of this component.

Finally, consider writing a unit test for `Checkbox.vue` to cover the new functionality. I think our most robust example is `pkg/rancher-components/src/components/Form/ToggleSwitch/ToggleSwitch.test.ts` if you want to give it a try. Otherwise, I think we should at least create an issue to make sure we come back implement (since we don't have any tests as it is today 🙂)

pkg/rancher-components/src/components/Form/Checkbox/Checkbox.vue
jordojordo: I initially went with `null` for the absence of a value, but decided on `undefined` as it was the default value for the resources of a container. But I think `null` would be the best solution for now.
Having a `valueWhenFalse` would be beneficial for sure, especially in this case as the false values for `httpEndpoint` and `httpTokens` are intended to be `disabled` and `optional` respectively. It's not entirely necessary for these properties but definitely worth a revisit.

jordojordo:
@rak-phillip I added a few tests for the basic value and with a custom `valueWhenTrue`, let me know what you think :grin:

pkg/rancher-components/src/components/Form/Checkbox/Checkbox.test.ts
```diff
@@ -0,0 +1,77 @@
...
+      stopPropagation: () => { },
+      preventDefault: () => { }
+    };
+
+    const wrapper = shallowMount(Checkbox, { propsData: { value: valueWhenTrue, valueWhenTrue } });
+
+    (wrapper.vm as any).clicked(event);
+    await wrapper.vm.$nextTick();
+
+    expect(wrapper.emitted().input?.[0][0]).toBeFalsy();
```
rak-phillip, line 75: Similar to above, I think we would want to assert that the value is `null` for this test.

pkg/rancher-components/src/components/Form/Checkbox/Checkbox.test.ts
```diff
@@ -0,0 +1,77 @@
+import { shallowMount } from '@vue/test-utils';
+import { Checkbox } from './index';
+
+describe('Checkbox.vue', () => {
+  it('renders falsy by default', () => {
+    const wrapper = shallowMount(Checkbox);
+    const cbInput = wrapper.find('input[type="checkbox"]').element as HTMLInputElement;
+
+    expect(cbInput.checked).toBeFalsy();
```
rak-phillip, line 9: Would `toBe(false)` make more sense here? I think `cbInput.checked` should be a `boolean` type.

pkg/rancher-components/src/components/Form/Checkbox/Checkbox.test.ts
```diff
@@ -0,0 +1,77 @@
+import { shallowMount } from '@vue/test-utils';
+import { Checkbox } from './index';
+
+describe('Checkbox.vue', () => {
+  it('renders falsy by default', () => {
```
rak-phillip, line 5: This is really a nitpick, but I think a name like `it('is unchecked by default')` might be a little easier to read.

tbh I think I've named a test or two like this recently 🤐

rak-phillip (changes requested):
Thanks for circling back to write a few tests! I think your work here looks great 😁

I just left a few points to consider, otherwise I think we'll be good to merge.

pkg/rancher-components/src/components/Form/Checkbox/Checkbox.vue
rak-phillip: I can see how the lack of a `valueWhenFalse` might be problematic when certain data might expect a different value than 'null`.. We run the risk of overwriting expected values for fields that might expect a custom value for a `false` by only providing `null` when `valueWhenTrue` is in use.

I updated #6787 with some notes about this for us to consider when we look at a refactor.

pkg/rancher-components/src/components/Form/Checkbox/Checkbox.test.ts
```diff
@@ -0,0 +1,77 @@
...
+      preventDefault: () => { }
+    };
+
+    const wrapper = shallowMount(Checkbox, { propsData: { value: false, valueWhenTrue } });
+
+    (wrapper.vm as any).clicked(event);
+    await wrapper.vm.$nextTick();
+
+    expect(wrapper.emitted().input?.length).toBe(1);
+    expect(wrapper.emitted().input?.[0][0]).toBeTruthy();
```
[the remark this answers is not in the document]
jordojordo: Good point! I changed this to `toBe(valueWhenTrue)`.

pkg/rancher-components/src/components/Form/Checkbox/Checkbox.test.ts
jordojordo: Yeah that makes sense for this check, changed.

pkg/rancher-components/src/components/Form/Checkbox/Checkbox.test.ts
jordojordo: For the default value `toBe(false)` sounds right, fixed.

pkg/rancher-components/src/components/Form/Checkbox/Checkbox.test.ts
jordojordo: I may have acquired your naming schemes a bit  :disguised_face:

rak-phillip (approved):
LGTM 🚀

PR #6788, opened 2022-09-01
Add developer debug panel for watches and types info

nwmac:
When run in development environment, adds a keyboard shortcut that (CTRL+i, META+i) that shows/hides a panel showing the active watchers and type info from the store.

Useful for debugging performance issues.


nwmac:
Will re-do as an extension

PR #6795, opened 2022-09-01
Set initial endpoint radio group for azuread

### LINKED ISSUE #6428: Azure AD Edit Config always picks Standard endpoint in radio button
**Setup**

**Describe the bug**

**To Reproduce**

**Result**

**Expected Result**

**Screenshots**

**Additional context**

**Detailed Description**

**Context**

jordojordo:
### Summary
Fixes #6428 

There was no functionality that set the radio group button to the correct endpoint option, it would always default to `standard`.

Added a method that will find the appropriate endpoint option depending on the `graphEndpoint`, and if not found it will default to `custom`.

- Steps to reproduce are in the issue listed
- If no China endpoint is available you can switch the [key names](https://github.com/rancher/dashboard/blob/43d338fac2cc87bf9c0cda6ac5d38fb2c4c7d96d/shell/edit/auth/azuread.vue#L32-L51) in `ENDPOINT_MAPPING` to spoof a different endpoint

### DIFF
--- a/shell/edit/auth/azuread.vue
+++ b/shell/edit/auth/azuread.vue
@@ -67,6 +67,10 @@ export default {
 
   async fetch() {
     await this.reloadModel();
+
+    if ( this.value?.graphEndpoint ) {
+      this.setInitialEndpoint(this.value.graphEndpoint);
+    }
   },
 
   data() {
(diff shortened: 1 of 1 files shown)


mantis-toboggan-md (approved):
lgtm

PR #6796, opened 2022-09-01
move harvester translations to plugin

### LINKED ISSUE #6791: Move harvester l10n into plugins

mantis-toboggan-md:
### Summary
Fixes #6791 

 Translations under 'harvester' as well as harvester-specific type label and setting descriptions have been moved into the harvester plugin. There are references to harvester in the shell translation files still because they are either:
* part of harvester manager,
* part of cluster provisioning using the Harvester provider, or
* used in fleet (to explain why clusters are hidden)

Harvester UI broadly; Harvester manager w/o Harvester plugin loaded


richard-cox (approved):
LGTM. I've resolved merge conflicts and tweaked one thing (for harvester upgrader header used the l10n for the product name rather than `harvester` const string)

PR #6806, opened 2022-09-06
Add harvester feature

WuJun2016:
### Summary

Fixes #


shell/mixins/harvester-disk.js
```diff
@@ -0,0 +1,17 @@
+import { findBy } from '@shell/utils/array';
```
mantis-toboggan-md, line 1: Can this file be moved to pkg/harvester/mixins? Otherwise, the PR looks good.

mantis-toboggan-md (changes requested):

shell/mixins/harvester-disk.js
WuJun2016: Done,  PTAL.  thank

mantis-toboggan-md (approved):
lgtm

PR #6812, opened 2022-09-06
Update parsing for nodeTimeout durations

### LINKED ISSUE #6275: auto replace unit is displayed in minutes instead of seconds when creating node driver rke2 cluster, advanced settings
**Setup**
- Rancher version: 2.6.5
- Browser type & version: any

**Describe the bug**
The "auto replace" value added in the issue below is displayed as in minutes, but should be seconds : 
https://github.com/rancher/dashboard/issues/4449#issuecomment-1048230874

**To Reproduce**
    Start creating a node driver rke2 cluster
    go to the advanced settings and you will see the option in the screenshot below:

**Result**
The value is correctly accounted for in seconds in the yaml file, oi this is purely a ui issue

**Expected Result**
The ui should indicate the value being in seconds
**Screenshots**
image
**Additional context**

**Context**

It is misleading

by not being misleading ;-)

jordojordo:
### Summary
Fixes #6275

When a cluster has been defined with an `unhealthyNodeTimeout` greater than `1m` prior to this version, the value would appear as seconds (i.e. `1 seconds`).

Added some logic to the `parseDuration` method to return the full amount of Duration time in seconds.

- Creating/Editing clusters with a different `unhealthyNodeTimeout` duration.

### DIFF
--- a/shell/edit/provisioning.cattle.io.cluster/MachinePool.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/MachinePool.vue
@@ -64,11 +64,17 @@ export default {
 
   data() {
     const parseDuration = (duration) => {
-      // The back end stores the timeout in Duration format, for example, "10m".
-      // Here we convert that string to an integer.
-      const numberString = duration.split('m')[0];
+      // The back end stores the timeout in Duration format, for example, "10m30s".
+      // Here we convert that string to an integer and return the duration as seconds.
+      const splitStr = duration.split(/[m,s]/).filter(s => s);
+      const minutes = parseInt(splitStr[0], 10);
+      const seconds = parseInt(splitStr[1], 10);
 
-      return parseInt(numberString, 10);
+      if ( isNaN(seconds) ) {
+        return minutes;
+      }
+
+      return seconds + minutes * 60;
     };
 
     return {
--- a/shell/edit/provisioning.cattle.io.cluster/MachinePool.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/MachinePool.vue
@@ -1,5 +1,4 @@
 <script>
-
 import { LabeledInput } from '@components/Form/LabeledInput';
 import { Checkbox } from '@components/Form/Checkbox';
 import { _EDIT } from '@shell/config/query-params';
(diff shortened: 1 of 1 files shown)


shell/edit/provisioning.cattle.io.cluster/MachinePool.vue
```diff
@@ -64,11 +64,17 @@ export default {
 
   data() {
     const parseDuration = (duration) => {
-      // The back end stores the timeout in Duration format, for example, "10m".
-      // Here we convert that string to an integer.
-      const numberString = duration.split('m')[0];
+      // The back end stores the timeout in Duration format, for example, "10m30s".
+      // Here we convert that string to an integer and return the duration as seconds.
+      const splitStr = duration.split(/[m,s]/).filter(s => s);
```
catherineluse, line 69: This seems to return the same array as-is - can it be removed? 

```
.filter(s => s)
```

catherineluse (changes requested):
I looked up some docs on Go Duration formats (https://www.geeksforgeeks.org/time-durations-in-golang/) and now I'm thinking the logic may need to also account for hours and days. Might need to check with someone on the back end to see if the number could have hours or days in it.

shell/edit/provisioning.cattle.io.cluster/MachinePool.vue
jordojordo: I had this to remove the empty string at the end of the array, but it isn't really necessary and can be removed.

jordojordo:
@catherineluse If the possibility of having hours or days for a timeout should we continue with displaying it as seconds?

catherineluse:
I believe so. Maybe in a future release we could have a reasonable limit on the number of seconds, as you suggested, and/or change the display to account for longer intervals using a library such as luxon, as Phil suggested.

jordojordo:
I added parsing for days, hours, minutes, and seconds. I agree, we should revisit this in the future. There are some [planned updates](https://tc39.es/proposal-temporal/docs/duration.html) to javascript which would help us parse these without the need for a library.

catherineluse (approved):
LGTM

PR #6814, opened 2022-09-06
Add global setting to disable websocket notification growls

### LINKED ISSUE #6793: Add setting to disable growlers to allow users a chance to diagnose what's going on if they pop up too regularly.
Based on rancher/dashboard#6785 , we're impeding customers from troubleshooting network connectivity by popping up growlers all the time.

We need a setting to temporarily turn these off. Perhaps under Troubleshooting.

The thinking is that this behaviour crept in here:
https://github.com/rancher/dashboard/issues/5093#issuecomment-1188467756

jordojordo:
### Summary
Fixes #6793 

This will add a global setting in the Performance tab to disable the websocket disconnection growls/notifications from appearing.  Also moved the banner and descriptions around within the performance tab.

### Technical notes summary

For the sake of simplicity, this setting has been added to the json blob value of `/v1/management.cattle.io.settings/ui-performance`. As this is set by an admin, the setting will persist for all users.

To test you can block any websocket requests made to `v1/subscribe`.

Steps:

- Within the network tab of the browser dev tools - add `wss://<cluster-ip>/k8s/clusters/<cluster-id>/v1/subscribe` to the blocked requests list
blockedwss

- Navigate to your cluster and notice the blocked requests and the "Websocket Disconnected" growl
enabled

- Then navigate to the Performance tab within the Global Settings page, enable the "Disable websocket notifications" checkbox and hit Apply
perfsettings

- Navigate back to your cluster and within the browser console notice the websocket connection errors and also take note that a growl has not appeared
console

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -6780,24 +6780,27 @@ featureFlags:
 performance:
   label: UI Performance Settings
   settingName: Performance
-  description: |-
-    Performance settings allow the behavior of the UI to be configured when dealing with many resources in a cluster. These settings only apply to the following resources: Pods, Deployments, Cron Jobs, Daemon Sets, Jobs, Stateful Sets, Replica Sets, Replication Controllers, Workloads and Secrets.
-  banner: These settings are experimental and may be removed or updated in future versions
   incrementalLoad:
     label: Incremental Loading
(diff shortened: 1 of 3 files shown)


mantis-toboggan-md (approved):
LGTM - I applied this setting and websocket disconnect messages no longer appeared whent he connection was blocked. I disabled the setting, they appeared again.

PR #6824, opened 2022-09-07
Fixes chart filtering when chart has no description

### LINKED ISSUE #6823: Charts filtering doesn't work if an App doesn't have description
Internal Ref: SURE-5191
 
If you have Helm charts that don't have descriptions, when filtering, these match the filter, when they should not.

1. Explore a cluster
2. Go to 'Apps'
3. Go to 'Repositories'
4. Add the Helm Repository: https://charts.truecharts.org
5. Go to Charts
6. Type `test` into the filter, you'll see that charts with an empty description match the filter, when they should not - e.g. the `fsm` and `notarious` charts show

nwmac:
Fixes #6823 

See issue above for steps to reproduce.

Simple fix to ensure we don't match charts that do not have a description.


jordojordo (approved):
lgtm

PR #6828, opened 2022-09-08
Move locale user preferences

### LINKED ISSUE #6660: Move locale selector from slide-in to user preferences
Now that we have the locale selector on the login screen, we should move the locale selector from the slide-in menu to the user preferences page.

bisht-richa:
### Summary
Fixes #6660

- Moved local from Side nav to preference page 
- Local style should change as per the page location

### How to test

- The login page should have a Local selector
- After logging in side nav should not have a Local selector, it should be on the user preference page
- Style of the Local selector should change as per the page.


shell/pages/prefs.vue
```diff
@@ -29,6 +30,7 @@ export default {
     menuMaxClusters: mapPref(MENU_MAX_CLUSTERS),
 
     ...mapGetters(['isSingleProduct']),
+    ...mapGetters('i18n', ['selectedLocaleLabel', 'availableLocales']),
```
richard-cox, line 33: Not needed

shell/pages/prefs.vue
```diff
@@ -153,6 +155,13 @@ export default {
     <div class="mt-10">
       <t k="prefs.theme.autoDetail" :pm="pm" :am="am" />
     </div>
+    <hr />
+    <h4 v-t="'prefs.language'" class="mt-10" />
```
richard-cox, line 159: `mt-10` isn't needed

richard-cox, line 159: I'd vote we make this the first entry above `Theme` (if this does move, the `mt-10` should remain)

shell/pages/prefs.vue
```diff
@@ -153,6 +155,13 @@ export default {
...
+    <div class="row">
+      <div class="col span-4">
+        <LocaleSelector></LocaleSelector>
```
richard-cox, line 162: If there's no content for an element it's usually better to close it within the first tag `<LocaleSelector/>` (same goes for shell/pages/auth/login.vue)

richard-cox (changes requested):

shell/pages/prefs.vue
bisht-richa: Fixed

shell/pages/prefs.vue
bisht-richa: fixed

bisht-richa: @richard-cox You mean something like this?
Screenshot 2022-09-22 at 14 27 38

richard-cox (approved):

PR #6850, opened 2022-09-12
Add warning banner for zero desired nodes in eks

### LINKED ISSUE #6808: Add UI Enhancement to Warn Users of EKS 0 Nodes in Node Group Behavior

Per issue https://github.com/rancher/rancher/issues/38698, the AWS SDK was updated to v1.44 which allows EKS clusters to be provisioned with node groups with 0 nodes. As stated in [this comment](https://github.com/rancher/rancher/issues/38698#issuecomment-1227776792) and @Josh-Diamond's [test results](https://github.com/rancher/rancher/issues/38698#issuecomment-1233580417), an EKS cluster will not become/stay active in rancher if there are 0 nodes in all node groups. This is not a bug. If there is no node to run the downstream agent on, the cluster won't be able to connect back to the rancher server.

We want the UI to show a warning if total number of desired nodes across of all node groups is 0 to let the user know that the cluster won’t come up as active until at least one node is available. Ex: _There are 0 nodes available to run agent. Cluster will not become active until at least one node is available._

jordojordo:
### Summary
Fixes #6808 

Added a banner to the resource detail page that will warn the user that the Cluster will not become active until at least one node is available.

### Technical notes summary

Created a method in the `provisioning.cattle.cluster.io` model to return all of the `nodeGroups` within an `eksConfig`, then searching for a `desiredSize` equal to `0` across of the node groups.

- Steps for testing can be found in this issue: https://github.com/rancher/rancher/issues/38698

### Screenshot/Video

nodegroupwarning

### DIFF
--- a/shell/detail/provisioning.cattle.io.cluster.vue
+++ b/shell/detail/provisioning.cattle.io.cluster.vue
@@ -321,6 +321,14 @@ export default {
       return false;
     },
 
+    showEksNodeGroupWarning() {
+      if ( this.value.isEKS && this.nodes?.length < 1 ) {
+        return !!this.value.eksNodeGroups.find(g => g.desiredSize === 0);
+      }
+
+      return false;
+    },
+
     machineHeaders() {
       return [
         STATE,
@@ -585,6 +593,7 @@ export default {
   <Loading v-if="$fetchState.pending" />
   <div v-else>
     <Banner v-if="showWindowsWarning" color="error" :label="t('cluster.banner.os', { newOS: 'Windows', existingOS: 'Linux' })" />
+    <Banner v-if="showEksNodeGroupWarning" color="error" :label="t('cluster.banner.desiredNodeGroupWarning')" />
 
     <Banner v-if="$fetchState.error" color="error" :label="$fetchState.error" />
     <ResourceTabs v-model="value" :default-tab="defaultTab">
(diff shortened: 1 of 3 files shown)


shell/detail/provisioning.cattle.io.cluster.vue
```diff
@@ -321,6 +321,14 @@ export default {
       return false;
     },
 
+    showEksNodeGroupWarning() {
+      if ( this.value.isEKS && this.nodes?.length < 1 ) {
+        return !!this.value.eksNodeGroups.find(g => g.desiredSize === 0);
```
mantis-toboggan-md, line 326: It looks like this is returning `true` (showing the banner) if there are 0 nodes in _any_ node group but (assuming I've understood the issue correctly; I didn't read the entire thread in the original issue) it should only show if there are 0 nodes available in _all_ node groups.

mantis-toboggan-md (changes requested):

shell/detail/provisioning.cattle.io.cluster.vue
jordojordo: Good catch, I updated the method to filter for _all_ node groups and check against that total number.

mantis-toboggan-md (approved):
LGTM

Issue #6743, opened 2022-08-25
Can't create resource when using 'Create Namespace' if there is an error first time


nwmac:
Fo to a resource, like secrets and click create and choose 'custom'.

Choose 'Create a new Namespace' in the namespace dropdown and enter a new namespace name.

Enter a name for the resource and for the type choose the service-acount-token option.

Click 'Create'.

You'll get an error - because for the secret type we chose, there are required key/values that we did not provide.

Change the secret type to 'Opaque'.

Click Create.

Form errors:  Could not create the new namespace. namespaces "dsfsdf" already exists

Looks like we create the new namespace, but then if we get an error from the backend because of bad data, you can't correct this and try again, because we try to create the namespace a second time and we already created it.

catherineluse:
# QA Template
## Root cause
The UI did not check to see if a namespace already existed before creating it. So upon the second attempt to create the same namespace, it threw an error.
 
## What was fixed, or what changes have occurred
This PR https://github.com/rancher/dashboard/pull/6853 changes the logic for namespace creation in the resource creation form so that if the namespace already exists, it does not attempt to create the namespace.
 
## Areas or cases that should be tested

To test this PR, I did the same steps as in the above issue description:

- Created a namespace `test1`.
- In the secret creation form, created a Custom secret and selected the service account token as the secret type.
- In the namespace dropdown, I selected Create a new namespace and entered `test1`.
- Attempted to create the secret multiple times.
- Confirmed that the error message was about the secret lacking required fields, not that the namespace already exists. The error was as expected: `Secret "test6" is invalid: metadata.annotations[kubernetes.io/service-account.name]: Required value`

I also tested the positive test case in which the secret is created successfully on the second attempt:

- In the secret creation form, created a Custom secret and selected the service account token as the secret type.
- In the namespace dropdown, I selected Create a new namespace and entered `test1`. (Although at this point, that namespace already exists)
- Entered a name for the secret
- Created a valid token for a valid service account (Note: If the secret is invalid, the backend won't throw errors. In order to see it show up successfully, the secret must use a valid service account token.) There was already a service account named `default` in the test namespace that I used, so I created the token by running `kubectl create token -n test1 default` to create a token for the service account called `default`. (Another note: Kubectl 1.24 is required to create the token)
- In the secret form, in the data tab, added a key-value pair `token`: and pasted the new token.
[...]

PR #6853, opened 2022-09-13
Only attempt to create namespace if it does not already exist

catherineluse:
This PR addresses https://github.com/rancher/dashboard/issues/6743 by changing the logic for namespace creation in the resource creation form so that if the namespace already exists, it does not attempt to create the namespace.

To test this PR, I did the same steps as in https://github.com/rancher/dashboard/issues/6743:

- Created a namespace `test1`.
- In the secret creation form, created a Custom secret and selected the service account token as the secret type.
- In the namespace dropdown, I selected Create a new namespace and entered `test1`.
- Attempted to create the secret multiple times.
- Confirmed that the error message was about the secret lacking required fields, not that the namespace already exists. The error was as expected: `Secret "test6" is invalid: metadata.annotations[kubernetes.io/service-account.name]: Required value`

I also tested the positive test case in which the secret is created successfully on the second attempt:

- In the secret creation form, created a Custom secret and selected the service account token as the secret type.
- In the namespace dropdown, I selected Create a new namespace and entered `test1`. (Although at this point, that namespace already exists)
- Entered a name for the secret
- Created a valid token for a valid service account (Note: If the secret is invalid, the backend won't throw errors. In order to see it show up successfully, the secret must use a valid service account token.) There was already a service account named `default` in the test namespace that I used, so I created the token by running `kubectl create token -n test1 default` to create a token for the service account called `default`. (Another note: Kubectl 1.24 is required to create the token)
- In the secret form, in the data tab, added a key-value pair `token`: and pasted the new token.
- In the secrets form, in the labels and annotations tab, added the key-value pair `kubernetes.io/service-account.name`: `default` to link the token to the service account named `default`.
- Confirmed that the token was successfully created in the `test1` namespace

I also tried the positive test case where a new namespace is successfully created at the same time as the service account token secret is successfully created.

### DIFF
--- a/shell/components/CruResource.vue
+++ b/shell/components/CruResource.vue
@@ -2,7 +2,7 @@
 import isEmpty from 'lodash/isEmpty';
 import { createYaml } from '@shell/utils/create-yaml';
 import { clone, get } from '@shell/utils/object';
-import { SCHEMA } from '@shell/config/types';
+import { SCHEMA, NAMESPACE } from '@shell/config/types';
 import ResourceYaml from '@shell/components/ResourceYaml';
 import { Banner } from '@components/Banner';
 import AsyncButton from '@shell/components/AsyncButton';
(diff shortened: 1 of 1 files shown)


Shavindra (approved):
LGTM

PR #6865, opened 2022-09-13
Saves configuration when adding new applications

scures:
### Summary
Fixes # https://github.com/epinio/ui/issues/155

Adds a new step in the `validationPassed` that checks if a new application binding occured.

Follow https://github.com/epinio/ui/issues/155 to reproduce, it should be working now.

### Screenshot/Video

image


richard-cox (approved):
I've rebased with epinio-dev and pushed a small improvement

PR #6886, opened 2022-09-14
Workload storage mounts should be on containers - #6800

### LINKED ISSUE #6800: Workload storage mount point for volumes should be on containers
> Need to remove the mount points config from the the pod.
> Need to add a storage tab to the container to configure volume mounts - this should allow multiple - same fields as are 
> currently in the UI above - but you will need a drop-down to allow the user to choose the volume to mount.

**Describe the bug**
REF: https://github.com/rancher/dashboard/pull/6632#issuecomment-1235287918
REF: https://github.com/rancher/dashboard/pull/6632#issuecomment-1233479509

**To Reproduce**
Navigate to Cluster Explorer > Cluster > Workload > Deployment
On the Pod tab there is a sub tab called Storage. 
Mount point should be on the container

**Expected Result**
Should have mount volume path on the container

image
image

Shavindra:
### Summary
Fixes #6800 

### Technical notes summary
Storage component was moved under Pods tab. When user is updating these volumes get applied to all containers.
So this component is moved to Container tab. 
Now when user add volume to container it updates correctly. 

1. Navigate to Cluster > Workload > Deployment
2. In Pod > Storage ad a volume
3. From Container tab > Storage. Select a Volume from previous step and add a path
4. Using `add container` to create a second container tab
5. Click on newly created container tab and add a different storage. 
6. Save

Inspecting the YAML and deployment should show correct options.

Workload Create/Update


catherineluse (changes requested):
This PR moves the storage tab from the pod tab to the container tabs, but storage needs to be configured at both the pod level AND the container level.

So basically, the pod tab needs a storage section that allows you to create volumes or select existing volumes. Basically:

- At the pod level, you can create a list of volumes that will be used by any container(s) within the pod. Each volume needs a name and a type. If a volume with that particular name doesn't already exist, the form will attempt to create it before creating the workload. (This functionality already exists)
- At the container level, there is a storage tab as well. In this container-level storage section, you can configure a) the name of the volume that the container needs, and b) the mount path. This is so that a certain directory in the container will lead to the volume and any program running in the container will be able to access everything in the volume from there.

To test this PR, the following will need to be done:

- Make sure you can still create or select volumes from the pod section, as usual
- Make sure that in the container section, you can select the name(s) of volumes that were configured in the pod section, and configure the mount paths for each of them
- Make sure that it is possible to create pods where two different containers can access the same volume using different mount paths (same as this example https://kubernetes.io/docs/tasks/access-application-cluster/communicate-containers-same-pod-shared-volume/#creating-a-pod-that-runs-two-containers)

Shavindra:
The UI is now split. 
- User can add volumes from pod tab only.
- On each container tab > storage, user can select a volume from a drop down and define a mount path and subpath.

NB: This is a minimal implementation. Had a discussion with @gaktive there would be much thought through implementation later.

CC: @catherineluse

catherineluse:
I think there should be a few changes:

1. In the container tab, if the storage options are empty, it should say "Volumes will appear here after they are added in the Pod tab"
Screen Shot 2022-09-22 at 1 46 56 PM

2. The names of the volumes need to be displayed on the Pod tab. Because currently the UI gives them a hidden ID as the name, but you need to select the ID from the container tab. So you have to choose between a confusing set of numbers. If I create multiple volumes on the Pod tab:

Screen Shot 2022-09-22 at 1 47 47 PM

Then if I go to the container tab, I have no idea which ID corresponds to which volume:
Screen Shot 2022-09-22 at 1 45 40 PM

There are no hints to help me out after I select them since the mount path config is the same for all volume types, so it would be easy to fill out the path for the wrong volume:

Screen Shot 2022-09-22 at 1 47 38 PM

3. I'm sorry for this, because in the other PR I had drawn a red box around the read/write setting that made it look like it needed to go in the container tab. But that's not right. The Kubernetes documentation on volumes (https://kubernetes.io/docs/concepts/storage/volumes/) shows the read/write setting configured at the volume level, not in the container. Here's an example for NFS:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: test-pd
spec:
  containers:
  - image: registry.k8s.io/test-webserver
    name: test-container
    volumeMounts:
    - mountPath: /my-nfs-data
      name: test-volume
  volumes:
  - name: test-volume
    nfs:
      server: my-nfs-server.example.com
      path: /my-nfs-volume
      readonly: true
```

Shavindra:
> 1. In the container tab, if the storage options are empty, it should say "Volumes will appear here after they are added in the Pod tab"

Fixed

> 2. The names of the volumes need to be displayed on the Pod tab. Because currently the UI gives them a hidden ID as the name, but you need to select the ID from the container tab. So you have to choose between a confusing set of numbers. If I create multiple volumes on the Pod tab:

I agree! one of the reasons I wasn't sure splitting is a good idea. This is not the ID. This is the automatically generated name. if user change the name it would be reflected. I've defined the `type` in the label.

image


> 3. I'm sorry for this, because in the other PR I had drawn a red box around the read/write setting that made it look like it needed to go in the container tab. But that's not right. The Kubernetes documentation on volumes (https://kubernetes.io/docs/concepts/storage/volumes/) shows the read/write setting configured at the volume level, not in the container. Here's an example for NFS:

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#volumemount-v1-core

It's possible to have `volumeMount.readOnly` of a container. It's also possible on the volume as well and already exists
image

catherineluse:
Something strange is happening - I tried creating a pod with 12 volumes, one of each type, and got unexpected results. The volume same displays for some types but not others. Only 6 out of the 12 volume types had their names exposed in the pod tab.

The 6 types of volumes where the names were not displayed are:

- Amazon EBS disk
- Azure disk
- Azure file
- CSI
- Google persistent disk
- VMWare vSphere volume

Then after going to the container tab, I realized that the list of volumes wasn't updating when a volume was added on the pod tab. After creating 3 volumes on the pod tab, I saw 3 volume options on the container tab, as expected. But after adding 9 more volumes to the pod tab, the container tab still only had the 3 options when I went back to it:

Screen Shot 2022-09-23 at 4 30 39 PM

nwmac:
@catherineluse Can you take another look now this has been updated?

Shavindra:
> Something strange is happening - I tried creating a pod with 12 volumes, one of each type, and got unexpected results. The volume same displays for some types but not others. Only 6 out of the 12 volume types had their names exposed in the pod tab.

This seemed intermittent. I've applied a fix. Hopefully it won't be replicable anymore.

catherineluse:
I checked out your PR locally. These are the steps I used:

1. Go to the deployments list
2. Go to a deployment. In the Pod tab, click **Storage > Add Volume > Open EBS Disk**
3. The name of the volume is not shown

Screen Shot 2022-09-26 at 9 33 16 PM

Shavindra:
> but I'm still seeing the issue with volume names not showing. 
I missed that part. I've checked the Kubernetes API for specific volume types. They don't have name field?

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#volume-v1-core

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#awselasticblockstorevolumesource-v1-core

I also don't see it in the current master in those volumes types except some of them.

e.g. EBS 
https://github.com/rancher/dashboard/blob/release-2.6.8/shell/edit/workload/storage/awsElasticBlockStore.vue
vs PVC
https://github.com/rancher/dashboard/blob/release-2.6.8/shell/edit/workload/storage/persistentVolumeClaim/index.vue#L121

If it really need to be there I can create a follow up and look into it right away.

@catherineluse

catherineluse:
Yeah, it wasn't in master because it didn't need to be. The new structure of the form is the reason to expose the name of all volumes.

catherineluse (approved):

PR #6887, opened 2022-09-14
Add support for building and serving non-minified plugins

nwmac:
Useful for debugging plugin load issues.

Allows the`build-pkg` script to take a `-d` arg which outputs non-minified files.


richard-cox (changes requested):
I couldn't get this to work with Harvester
- The build succeeds and is served & available at http://127.0.0.1:4500/harvester-1.1.0/harvester-1.1.0.umd.js 
- Going to the cluster though errors in `shell/core/plugins.js`
  - ```
     if (!window[id]) {
              return reject(new Error('Could not load plugin code'));
            } 
    ```
  - Note - the same happens when loading from the plugins page
- The browser is showing a 200 for the request to fetch the file
- Noticed shell/scripts/pkgfile pkg.main references the minified extension, but that won't be used in this context right as we directly access the file?

richard-cox (approved):
It was as your thought, there was something wrong my env. 

I've since tried again with the harvester cluster and both load approaches work (via clicking on a harvester cluster and in the plugins page)

PR #6893, opened 2022-09-14
Correctly shows bound services in applications view

scures:
### Summary
Fixes # https://github.com/epinio/ui/issues/156

Follow the steps at https://github.com/epinio/ui/issues/156 to reproduce. 
Now an app with a single bounded service should show only that one and not all available services.

> Note: Binding apps to services from UI won't work before merging  https://github.com/rancher/dashboard/pull/6865, you can use epinio-cli for it.

### Screenshot/Video

image

### DIFF
--- a/pkg/epinio/pages/c/_cluster/applications/index.vue
+++ b/pkg/epinio/pages/c/_cluster/applications/index.vue
@@ -62,11 +62,8 @@ export default {
       if (services.length) {
         const serviceBounds = [];
 
-        // We iterate over all services.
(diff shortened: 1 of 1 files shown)


richard-cox (changes requested):
I've created https://github.com/rancher/dashboard/pull/7209 which contains an improved fix. 
- the app services getter should be updated rather than worked around
- the app services getter was incorrectly filtering via configurations (which were 1:M service to app)

Could you take a look, if that looks good this one can be closed

PR #6902, opened 2022-09-15
Add check for azuread endpoint type

### LINKED ISSUE #6428: Azure AD Edit Config always picks Standard endpoint in radio button

### LINKED ISSUE #6868: Editing the azure AD auth provider config before updating the endpoints result in `The 'resource' request parameter is not supported.`
**Setup**
- Rancher version: v2.6-head 79a5328

**Describe the bug**
When a rancher server is upgraded from versions v2.6.6 and below or v2.5.15 and below, we would have the old graph end points set. On a rancher server upgrade before updating the endpoints, if we try to edit the config and change the endpoints from custom to standard and input the secret we get an error upon authenticating.
 

**To Reproduce**
1. On v2.6.6 rancher server enable Azure AD auth config with standard endpoints
2. Upgrade the rancher server to v2.6-head
3. Do not update the endpoints but click on edit config
4. Notice the endpoints are switched to custom endpoints due to the [issue](https://github.com/rancher/dashboard/issues/6428#issuecomment-1245434338)  
5. Change the end points to standard endpoints and input the secret value and save

**Result**
Upon authenticating we run into the issue `The 'resource' request parameter is not supported.`

**Expected Result**
Expected the edit config to work fine without any errors.

**Additional context**
Found while validating https://github.com/rancher/rancher/issues/38753#issuecomment-1245523775

jordojordo:
### Summary
Fixes #6868 - Fixes #6428 

If the azuread auth has been configured prior to 2.6.7 and the endpoints have not been updated, the deprecated endpoints ( e.g. `https://graph.windows.net/` ) should be respected when editing the config. Also, when setting the initial endpoint option we were not checking for the old endpoints, thus setting the endpoint radio option to `custom`.

### Technical notes summary

Added a check for a `graphEndpoint` in the config to match one of the old endpoints. When `setEndpoints` is called it use the appropriate endpoint map to update the model.
Also removed the unnecessary call to `setEndpoints` when the model as this was causing the endpoint to be reset to the new set of endpoints regardless of the original value.

- Editing the config of an existing azuread auth after upgrading to `v2.6-head`
- Creating a new azuread auth

### DIFF
--- a/shell/edit/auth/azuread.vue
+++ b/shell/edit/auth/azuread.vue
@@ -76,6 +76,8 @@ export default {
   data() {
     return {
       endpoint:          'standard',
+      oldEndpoint:        false,
+
       // Storing the applicationSecret is necessary because norman doesn't support returning secrets and when we
       // override the steve authconfig with a norman config the applicationSecret is lost
       applicationSecret: this.value.applicationSecret
@@ -147,9 +149,6 @@ export default {
       handler() {
         this.model.accessMode = this.model.accessMode || 'unrestricted';
         this.model.rancherUrl = this.model.rancherUrl || this.replyUrl;
-        if (this.endpoint !== 'custom') {
-          this.setEndpoints(this.endpoint);
-        }
 
         if (this.model.applicationSecret) {
           this.$set(this, 'applicationSecret', this.model.applicationSecret);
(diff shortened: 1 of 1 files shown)


mantis-toboggan-md (approved):
lgtm

PR #6904, opened 2022-09-15
Adjusts timing, messaging, control, etc of websocket growls.

### LINKED ISSUE #6785: Rancher 2.6.8 UI throwing constant "Websocket Disconnected" errors
Internal reference: SURE-5185

Issue description:

After upgrading to Rancher 2.6.8, a user started seeing these errors continuously in the UI:

```
Websocket Disconnected 
The connection to wss://SERVER.DOMAIN.TLD/v3/subscribe closed unexpectedly 
Tue, Aug 30 2022 10:14:49 am. 
Retrying...
```

Some of their RKE downstream clusters did not upgrade. ~~There was a script that another environment had used for similar issue on AKS. This was used here with a slight change to make it work for RKE. After stopping Rancher, running the script, the clusters were upgraded and healthy.~~ (Irrelevant section struck out upon follow-up.)

But the UI continued to show that websocket error continuously every few seconds. Looks like this is because of: https://github.com/rancher/dashboard/issues/5093#issuecomment-1188467756

There doesn't seem to be any other issues with the Rancher UI and all the downstream clusters are healthy. We can't tell where the errors are coming from... Rancher? Downstream cluster agents? The web browser? The customer's networking (i.e. Load Balancer)?

Business impact:
A constant stream of errors in the UI is a real hindrance for all users and the user would like a way to silence/disable these growls since these errors and find out why it's happening.

Troubleshooting steps:
We tried refreshing the browser, logging out and logging back in, using a local user instead of LDAP, and using a direct connection without VPN. All of these still caused the errors to pop up continuously in the UI. 

Actual behavior:
Constant stream of websocket errors in the UI.

Expected behavior:
No websocket errors in the UI.

Sean-McQ:
### Summary
Fixes #6785
Some sockets close quickly, given that we growl on every socket closing, this can lead to flooding the UI with growl notifications. Growl notifications were added to the dashboard because if a websocket cannot connect, that's something a user should know about.

* We don't consider ourselves actually disconnected until the reconnect attempts have failed.
* We only attempt to reconnect 3 times before declaring the socket disconnected and telling the user to reload the page (maybe later, I'll figure out how to add a "try again" button to the growl to bump the maxTries and attempt to connect again.
* Subscribe.js doesn't have any setTimeouts for displaying growls, all the timing is handled by when we dispatch the events in socket.js, keeping the timeline in my head was driving me crazy so I had to centralize it.
* We don't even notify the user unless the 1st reconnect attempt has failed since sockets closing and reconnecting is part of their normal lifecycle.
* If there's been more than 1 attempt to reconnect and the 2nd or 3rd attempt was successful then we clear all the growls and instead tell them that we've reconnected.
* Given that the growls effectively escalate the same issue, we always remove the old growl before dispatching the new one so we don't flood the screen.

I still don't have a great way to test this. Ideally, one could send socket connections through a proxy that you could arbitrarily disconnect and reconnect to cover all test cases.

The app is now a lot more careful about sending growl notifications about sockets, we'll want to make sure that they do show up when they're supposed to still.

### DIFF
--- a/shell/plugins/steve/subscribe.js
+++ b/shell/plugins/steve/subscribe.js
@@ -25,9 +25,6 @@ export const NO_SCHEMA = 'NO_SCHEMA';
 // minimum length of time a disconnect notification is shown
 const MINIMUM_TIME_NOTIFIED = 3000;
 
-// minimum time a socket must be disconnected for before sending a growl
-const MINIMUM_TIME_DISCONNECTED = 10000;
-
 // We only create a worker for the cluster store
 export function createWorker(store, ctx) {
   const { getters } = ctx;
@@ -396,8 +393,15 @@ export const actions = {
     commit, dispatch, state, getters, rootGetters
   }, event) {
     state.debugSocket && console.info(`WebSocket Opened [${ getters.storeName }]`); // eslint-disable-line no-console
-
     const socket = event.currentTarget;
+    const tries = event?.detail?.tries; // have to pull it off of the event because the socket's tries is already reset to 0
+    const t = rootGetters['i18n/t'];
+    const perfSetting = rootGetters['management/byId'](MANAGEMENT.SETTING, SETTING.UI_PERFORMANCE);
+    let disableGrowl = false;
+
+    if ( perfSetting?.value ) {
+      disableGrowl = JSON.parse(perfSetting.value).disableWebsocketNotification || false;
+    }
 
     this.$socket = socket;
 
@@ -420,19 +424,16 @@ export const actions = {
     if ( socket.hasReconnected ) {
       await dispatch('reconnectWatches');
       // Check for disconnect notifications and clear them
-      const growlErr = rootGetters['growl/find']({ key: 'url', val: this.$socket.url });
+      const growlErr = rootGetters['growl/find']({ key: 'url', val: socket.url });
 
       if (growlErr) {
-        const now = Date.now();
-
-        // even if the socket reconnected, keep the error growl for at least a few seconds to ensure its readable
-        if (now >= growlErr.earliestClose) {
-          dispatch('growl/remove', growlErr.id, { root: true });
-        } else {
-          setTimeout(() => {
-            dispatch('growl/remove', growlErr.id, { root: true });
-          }, growlErr.earliestClose - now);
-        }
+        dispatch('growl/remove', growlErr.id, { root: true });
+      }
+      if (tries > 1 && !disableGrowl) {
+        dispatch('growl/success', {
+          title:   t('growl.reconnected.title'),
+          message: t('growl.reconnected.message', { url: this.$socket.url, tries }),
+        }, { root: true });
       }
     }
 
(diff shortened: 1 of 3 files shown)


mantis-toboggan-md (changes requested):
I tested this by rebooting my droplet. Pretty slow process but I don't know how to simulate websocket disconnects/reconnects. I made the changes I  commented in-line then tested whether or not the socket would stop trying to reconnect after 3 attempts, and tell the user that. Then I greatly increased the maxTries variable because it takes a while for my cluster to come back online, and verified that the success message appears when a socket reconnects. 

I don't know how to test the problem the customer(s?) reported, however.

Sean-McQ:
Doh! Thanks for catching all that. Changes made and pushed!

shell/plugins/steve/subscribe.js
```diff
@@ -456,41 +451,54 @@ export const actions = {
...
-            icon:          'error',
-            earliestClose: time + MINIMUM_TIME_NOTIFIED + MINIMUM_TIME_DISCONNECTED,
-            url
-          }, { root: true });
-        }, MINIMUM_TIME_DISCONNECTED);
+    const t = rootGetters['i18n/t'];
+
+    // determine if websocket notifications are disabled
+    const perfSetting = rootGetters['management/byId'](MANAGEMENT.SETTING, SETTING.UI_PERFORMANCE);
+    let disableGrowl = false;
```
nwmac, line 466: This setting seems to be ignored - we still want to disable growls via this setting, just in case there is still an issue.

nwmac (changes requested):

shell/plugins/steve/subscribe.js
Sean-McQ: it was being used on EVENT_DISCONNECT_ERROR error but not EVENT_CONNECT_ERROR which has since been corrected and pushed. Thanks for catching that!

mantis-toboggan-md (dismissed):

shell/plugins/steve/subscribe.js
```diff
@@ -456,42 +451,58 @@ export const actions = {
...
-    if (e.type === EVENT_DISCONNECT_ERROR) {
-      // determine if websocket notifications are disabled
-      const perfSetting = rootGetters['management/byId'](MANAGEMENT.SETTING, SETTING.UI_PERFORMANCE);
-      let disableGrowl = false;
+
+    // determine if websocket notifications are disabled
+    const perfSetting = rootGetters['management/byId'](MANAGEMENT.SETTING, SETTING.UI_PERFORMANCE);
+    let disableGrowl = false;
+
+    if (!disableGrowl) {
```
nwmac, line 459: We should use this guard on the success growl above as well

nwmac (changes requested):

shell/plugins/steve/subscribe.js
Sean-McQ: done

nwmac (approved):

PR #6929, opened 2022-09-19
Azuread custom endpoint radio option

### LINKED ISSUE #6428: Azure AD Edit Config always picks Standard endpoint in radio button

jordojordo:
### Summary
Fixes #6428 

This issue would arise regardless of upgrading prior to `v2.6.5` as we were checking _only_ for the `graphEndpoint`, not for the other endpoints.
Added a method to determine if the values for all of the endpoints match the values in the hard coded map. If they don't match, `custom` will be the type.

- Steps to test are in the related issue
- A test environment can be given if needed

### DIFF
--- a/shell/edit/auth/azuread.vue
+++ b/shell/edit/auth/azuread.vue
@@ -1,4 +1,5 @@
 <script>
+import isEqual from 'lodash/isEqual';
 import Loading from '@shell/components/Loading';
 import CreateEditView from '@shell/mixins/create-edit-view';
 import CruResource from '@shell/components/CruResource';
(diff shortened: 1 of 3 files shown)


shell/edit/auth/azuread.vue
```diff
@@ -190,6 +190,23 @@ export default {
...
+
+        if ( mappedValues === valuesToCheck ) {
+          out = key;
+        } else {
+          out = 'custom';
+        }
+      }
+
+      return out;
+    },
```
mantis-toboggan-md, line 208: Arrays can't be properly compared using `===` try lodash:
```
import isEqual from 'lodash/isEqual'

...

if(isEqual(mappedValues, valuesToCheck))

...

```

I think you also need to change where in this process you're setting `out = 'custom' `. As it is now, this loop is setting `out = 'custom'` for each region (standard, china, custom) that doesn't correspond to the current config. That is, if the first region checked is the one we want, out is gonna be re-set to 'custom' when the next region is iterated over. Try using 'custom' as the initial value for `out` rather than setting it within the loop.

This is, admittedly, nitpicky, but I find this loop hard to read: I think some more specific variable names would be helpful here.

mantis-toboggan-md (changes requested):

shell/edit/auth/azuread.vue
jordojordo: Great points, makes a lot of sense to set `out` as `custom` initially. Also changed the equality checking to use lodash and renamed a few of the variables.

mantis-toboggan-md: It's easier for me to read now, thank you :)

mantis-toboggan-md (approved):
LGTM

nwmac:
Force merging as drone is stuck

PR #6940, opened 2022-09-20
Add info banner for rke2 cni and network policy conflict

### LINKED ISSUE #5105: Visual indicator when selecting cilium and enabling Project Network Isolation on v2 clusters
When a user is creating a new v2 cluster and they select the cilium CNI and enable Project Network Isolation there should be a popup in the form that informs the users that the ingress controller will not be able to route requests to pods on other nodes by default.  Ideally we should link to the related documentation as well.

https://github.com/rancher/docs/issues/3868

jordojordo:
### Summary
Fixes #5105 

This adds an info banner in the rke2 cluster configuration under the `Project network isolation` checkbox when the Container Network is `cilium` and the Project network isolation checkbox is enabled. The info will read "Per default the ingress controller will not be able to route requests to pods on other nodes."

- Creating/editing an rke2 cluster

### Screenshot/Video
banner

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -1650,7 +1650,7 @@ cluster:
       label: Default Pod Security Policy
     enableNetworkPolicy:
       label: Project Network Isolation
-      warning: Per default the ingress controller will not be able to route requests to pods on other nodes.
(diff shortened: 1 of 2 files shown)


jordojordo:
@scures I fixed the translation, thanks for catching that!

PR #6943, opened 2022-09-20
Add PSP Removal banners

### LINKED ISSUE #6873: Add PSP Removal banners
PSPs are being removed in Kubernetes 1.25.

We should show a banner to the user to let them know this, if they are using PSPs

aalves08:
Fixes #6873 

- add deprecation banner for PSPs on cluster dashboard page to be displayed only if the user has any PSPs and if kubernetes version is 1.21 or higher

**Screenshot**
Screenshot 2022-09-21 at 10 58 27

**Test**
- Create a `PodSecurityPolicy`
- Make sure cluster is running kubernetes version 1.21 or higher
- Check if warning Banner appears on the cluster dashboard page

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -2446,6 +2446,7 @@ landing:
     title: Getting Started
     body: Take a look at the the quick getting started guide. For Cluster Manager users, learn more about where you can find your favorite features in the Dashboard UI.
   support: Support
+  deprecatedPsp: "In Kubernetes v1.25, Pod Security Policies will be removed following it's deprecation in v1.21. They will be removed in favor of Pod Security Admissions, for which you can follow the instructions for migration <a href='https://kubernetes.io/docs/tasks/configure-pod-container/migrate-from-psp/' target='_blank' rel='noopener nofollow noreferrer'>here</a>"
   community:
     title: Community Support
     docs: Docs
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -2446,7 +2446,7 @@ landing:
     title: Getting Started
     body: Take a look at the the quick getting started guide. For Cluster Manager users, learn more about where you can find your favorite features in the Dashboard UI.
   support: Support
-  deprecatedPsp: "In Kubernetes v1.25, Pod Security Policies will be removed following it's deprecation in v1.21. They will be removed in favor of Pod Security Admissions, for which you can follow the instructions for migration <a href='https://kubernetes.io/docs/tasks/configure-pod-container/migrate-from-psp/' target='_blank' rel='noopener nofollow noreferrer'>here</a>"
+  deprecatedPsp: Pod Security Policies are deprecated as of Kubernetes v1.21, and have been removed in Kubernetes v1.25. You have one or more PodSecurityPolicy resource(s) in this cluster.
   community:
     title: Community Support
     docs: Docs
--- a/shell/pages/c/_cluster/explorer/index.vue
+++ b/shell/pages/c/_cluster/explorer/index.vue
@@ -140,14 +140,16 @@ export default {
   },
 
   watch: {
-    // logic to be removed once kube version is 1.25 or greater
     // we need to hook up this API call to a watcher because the page logic is based on a getter
-    // as a temporary solution, this seems reasonable, so that we don't disrupt the optimal loading times of the page
+    // this seems reasonable, so that we don't disrupt the optimal loading times of the page
     currentCluster: {
       async handler(neu, old) {
         if (neu && (!old || old.id !== neu.id)) {
-          if (neu.status?.version?.major >= 1 && neu.status?.version?.minor >= 21) {
-            const psps = await this.$store.dispatch('management/request', { url: '/v1/policy.podsecuritypolicies' });
+          const major = neu.status?.version?.major ? parseInt(neu.status?.version?.major) : 0;
+          const minor = neu.status?.version?.minor ? parseInt(neu.status?.version?.minor) : 0;
+
+          if (major === 1 && minor >= 21 && minor < 25) {
+            const psps = await this.$store.dispatch('cluster/request', { url: `/k8s/clusters/${ neu.id }/v1/policy.podsecuritypolicies` });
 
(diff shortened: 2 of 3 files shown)


catherineluse:
https://github.com/rancher/dashboard/issues/6873

shell/assets/translations/en-us.yaml
```diff
@@ -2446,6 +2446,7 @@ landing:
     title: Getting Started
     body: Take a look at the the quick getting started guide. For Cluster Manager users, learn more about where you can find your favorite features in the Dashboard UI.
   support: Support
+  deprecatedPsp: "In Kubernetes v1.25, Pod Security Policies will be removed following it's deprecation in v1.21. They will be removed in favor of Pod Security Admissions, for which you can follow the instructions for migration <a href='https://kubernetes.io/docs/tasks/configure-pod-container/migrate-from-psp/' target='_blank' rel='noopener nofollow noreferrer'>here</a>"
```
nwmac, line 2449: Can we use the simpler message as per the GitHub issue?

shell/pages/home.vue
```diff
@@ -40,9 +41,14 @@ export default {
...
   mixins: [PageHeaderActions],
 
-  fetch() {
-    this.$store.dispatch('management/findAll', { type: CAPI.RANCHER_CLUSTER });
-    this.$store.dispatch('management/findAll', { type: MANAGEMENT.CLUSTER });
+  async fetch() {
+    const res = await allHash({
+      rancherClusters: this.$store.dispatch('management/findAll', { type: CAPI.RANCHER_CLUSTER }),
+      mgmtClusters:    this.$store.dispatch('management/findAll', { type: MANAGEMENT.CLUSTER }),
+      psps:            this.$store.dispatch('management/request', { url: '/v1/policy.podsecuritypolicies' }),
```
nwmac, line 48: Banner should be on the cluster dashboard page for a cluster, not the home page

shell/pages/home.vue
```diff
@@ -74,6 +84,10 @@ export default {
       return this.$store.getters['management/all'](CAPI.RANCHER_CLUSTER);
     },
 
+    hasPodSecurityPolicies() {
+      return this.psps && this.psps.length;
```
nwmac, line 88: Also want to only show the banner if the cluster is k8s version 1.21 or greater.

nwmac (dismissed):

shell/pages/home.vue
catherineluse: Right - This banner only shows if the PSPs are in the local Rancher server cluster, so this banner assumes you're an admin who has access to the local cluster. But a cluster owner for a downstream cluster should be able to see the banner too.

aalves08:
@catherineluse @nwmac thanks for all the feedback. 

I have updated the PR according to you feedback + #6943  updated requirements.

catherineluse:
I have your PR checked out locally, but the banner isn't showing for me. I have this downstream cluster with several PSPs:

Screen Shot 2022-09-22 at 11 55 23 AM

But when I go to the overview page for the cluster, the banner is not shown:

Screen Shot 2022-09-22 at 11 56 00 AM

shell/pages/c/_cluster/explorer/index.vue
```diff
@@ -138,6 +139,26 @@ export default {
     this.$store.dispatch('cluster/forgetType', MANAGEMENT.NODE);
   },
 
+  watch: {
+    // logic to be removed once kube version is 1.25 or greater
```
catherineluse, line 143: I don't think we should plan to remove the logic because there could be a situation in the future where some Rancher managed clusters are running 1.25 and some are running an older version. So I think the if-statement should also check if the minor version is less than 25.

And the banner should only show if the major version is exactly equal to "1".

shell/pages/c/_cluster/explorer/index.vue
```diff
@@ -138,6 +139,26 @@ export default {
...
+    // we need to hook up this API call to a watcher because the page logic is based on a getter
+    // as a temporary solution, this seems reasonable, so that we don't disrupt the optimal loading times of the page
+    currentCluster: {
+      async handler(neu, old) {
```
catherineluse, line 147: Is this code expected to execute if you refresh the page on cluster explorer? I'm not sure if the watcher would be triggered in that case, but the banner didn't show for me regardless of whether I refreshed or navigated from a different cluster.

catherineluse (changes requested):

shell/pages/c/_cluster/explorer/index.vue
aalves08: updated conditions to:
`neu.status?.version?.major === 1 && neu.status?.version?.minor >= 21 && neu.status?.version?.minor < 25`

shell/pages/c/_cluster/explorer/index.vue
aalves08: Yes, as the watcher has the `immediate: true ` prop + the conditions on the watch should mean that it only runs once (not triggering multiple API calls to fetch policies)

shell/pages/c/_cluster/explorer/index.vue
aalves08: Updated conditions based on your feedback 🙏 

`if (major === 1 && minor >= 21 && minor < 25)`

aalves08:
> I have your PR checked out locally, but the banner isn't showing for me. I have this downstream cluster with several PSPs:
> 
> Screen Shot 2022-09-22 at 11 55 23 AM
> 
> But when I go to the overview page for the cluster, the banner is not shown. I tried refreshing on cluster explorer, going to the PSPs list view and navigating back, going to the local cluster and navigating back:
> 
> Screen Shot 2022-09-22 at 11 56 00 AM

So, couple of problems here (sorry about it @catherineluse ):
- major and minor come as strings not int's so I wasn't parsing them correctly (don't know why it was working on my system 🤷 when I tested it )
- wasn't using kube API to fetch the policies, so the request `v1/policy.podsecuritypolicies` was giving me the PSPs for the local cluster and not for the downstream cluster we are at. I have fixed that now.

I think it's all good to go. Will ask for another review from your side. Thank you for your time and sorry about my mess 🙏

catherineluse (approved):
LGTM

PR #6971, opened 2022-09-23
Remove deprecated v1 multi-cluster apps

### LINKED ISSUE #6970: Remove v1 Multi-cluster Apps from the UI
Remove v1 Multi-cluster Apps from the UI

Note: The menu entry in the top-level menu should be renamed as 'Global Configuration'

nwmac:
Fixes #6970 

- Renames 'Multi-cluster Apps' in the top-level menu to 'Global Configuration' (since apps are removed)
- Removed Apps from Global Configuration
- Updated the installs screen that used to link to MCM Apps to instead say that they can not be managed through the UI 

Note: For simplicity, I did not rename the product definition, so this stays as `mcapps`.

### DIFF
--- a/shell/config/product/multi-cluster-apps.js
+++ b/shell/config/product/multi-cluster-apps.js
@@ -21,17 +21,6 @@ export function init(store) {
   });
 
   virtualType({
-    labelKey:       'legacy.apps',
-    name:           'mc-apps',
-    group:          'Root',
-    namespaced:     false,
-    weight:         112,
-    icon:           'folder',
-    route:          { name: 'c-cluster-mcapps-pages-page', params: { cluster: 'local', page: 'apps' } },
-    exact:          true
-  });
-
-  virtualType({
     labelKey:       'legacy.catalogs',
     name:           'mc-catalogs',
     group:          'Root',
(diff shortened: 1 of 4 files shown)


shell/config/product/multi-cluster-apps.js
```diff
@@ -65,7 +54,6 @@ export function init(store) {
   });
 
   basicType([
-    'mc-apps',
     'mc-catalogs',
```
richard-cox, line 57: Should `Catalogs` also be removed, not sure they're used anywhere outside of mc apps...  or do we need the deprecated process there as well?

shell/pages/c/_cluster/apps/charts/install.vue
```diff
@@ -1546,9 +1542,12 @@ export default {
             {{ t('catalog.install.error.legacy.enableLegacy.goto') }}
           </nuxt-link>
         </template>
+        <template v-else-if="mcapp">
```
richard-cox, line 1545: This kind of wiggles around the content from the first span

image

Can we gate the first `span` of this on `!mcapp`?

richard-cox (changes requested):

richard-cox (approved):

PR #6978, opened 2022-09-23
Plugins: Move formatter cache loading out of Sortable Table

nwmac:
Reduces size of plugin packages significantly.

Advis, if a plug-in uses sortable table, it pulls in all formatters.

This PR moved the formatted cache initialisation to a plug-in that runs once.


mantis-toboggan-md (approved):
LGTM - verified formatters load in the explorer, loaded a harvester plugin ui, verified both plugin formatters and shell formatters loaded, re-checked explorer w/ plugin loaded

PR #6985, opened 2022-09-23
Fix: Service selector preview does not consider namespace

bashofmann:
### Summary
This fixes the selector preview box on the service create and edit form to only show selected pods from the same namespace

Fixes https://github.com/rancher/dashboard/issues/6984

The matching function was missing a filter to only show pods of the namespace of the service and to update itself when the namespace changes.

Preview box when creating or editing services

### Screenshot/Video
With this deployment:

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
  namespace: default
spec:
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - image: nginx
        name: container-0
```

Before the patch, there is no namespace filter:

Bildschirmfoto 2022-09-23 um 16 26 44

After, the namespace filtering now works correctly:

Bildschirmfoto 2022-09-23 um 16 28 16
Bildschirmfoto 2022-09-23 um 16 28 23

### DIFF
--- a/shell/edit/service.vue
+++ b/shell/edit/service.vue
@@ -198,7 +198,8 @@ export default {
   },
 
   watch: {
-    'value.spec.selector': 'updateMatchingPods',
+    'value.metadata.namespace': 'updateMatchingPods',
+    'value.spec.selector':      'updateMatchingPods',
     'value.spec.sessionAffinity'(val) {
       if (val === 'ClientIP') {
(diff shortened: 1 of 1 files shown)


richard-cox (approved):

bashofmann:
@ronhorton I don't think the bot created a backport PR or issue. Is there anything, that I can do manually?

Issue #6992, opened 2022-09-24
[2.7] Disable growl on socket disconnect by default


mantis-toboggan-md:
- There's been issues with the growl shown when web sockets disconnect
- A setting to disable these was added which defaults to off
- Given new issues we should set the default disabled value to on
  - See slack threads 
  - https://github.com/rancher/dashboard/issues/6959

As part of this fix we need to make sure we're still trying to reconnect in the background if this setting is on

2.6.9 issue: https://github.com/rancher/dashboard/issues/6960

Josh-Diamond:
Ticket #6992 - Test Results
---

With Docker on a single-node instance:

Verified on rancher `v2.7-40c390d79faef993e8509d545a4ed1e18a78a4de-head`:
1. Fresh install of rancher `v2.7-head`
2. Navigate to `Global Settings > Performance > Websocket Notifications`
3. Verified - `Disabled` by default

Screenshot:

<img width="1562" alt="Screen Shot 2022-09-29 at 11 02 30 AM" src="https://user-images.githubusercontent.com/46494969/193107905-8198318a-5f01-426a-adc7-5b2e02d1699b.png">

PR #6993, opened 2022-09-24
[2.7] adjust socket reconnect logic

### LINKED ISSUE #6972: Improve performance setting fetch
In `subscrube.js` we have duplicated code (see below) and we have the default set there in 2 places as well as in the default for the setting itself.

```
    let disableGrowl = true;

    if ( perfSetting?.value ) {
      disableGrowl = JSON.parse(perfSetting.value).disableWebsocketNotification !== false;
    }
```

We already have some code in the `resource-fetch` mixin that does something similar.
We should create a single function for getting this setting and put it in `settings.js` and use it in all 3 places.

Note, the current code in `resource-fetch.js` does not support upgrade - i.e. we have added `disableWebsocketNotification` to the JSON, but for a setting created before that was added, we will not return the default in the code.

I think we should do something like this:

```
    // fetching the settings related to manual refresh from global settings
    const perfSetting = this.$store.getters['management/byId'](MANAGEMENT.SETTING, SETTING.UI_PERFORMANCE);
    let perfConfig = {};

    if (perfSetting && perfSetting.value) {
      try {
        perfConfig = JSON.parse(perfSetting.value);
      } catch (e) {
        console.warn('ui-performance setting contains invalid data'); // eslint-disable-line no-console
      }
    }

    // Start with the default and overwrite the values from the setting - ensures we have defaults for newly added options
    perfConfg = Object.assign(DEFAULT_PERF_SETTING, perfConfig);
```

mantis-toboggan-md:
…factor performance setting getting logic

### Summary
 fixes #6972, #6992
 
 related 2.6.9 issue: https://github.com/rancher/dashboard/issues/6960

This is partly addressed in a previous PR here https://github.com/rancher/dashboard/pull/6968 but per slack convo w/ Neil we should not be giving up on reconnect attempts if we're not visualizing the failure prominently. (eg when growls are disabled). I snuck in #6972 given that the request stems from code duplication and _not_ tacking it on would just be more code duplication.

### DIFF
--- a/shell/config/settings.js
+++ b/shell/config/settings.js
@@ -135,3 +135,21 @@ export const setSetting = async(store, id, val) => {
 
   return setting;
 };
+
+export const getPerformanceSetting = (rootGetters) => {
+  const perfSetting = rootGetters['management/byId'](MANAGEMENT.SETTING, SETTING.UI_PERFORMANCE);
+  let perfConfig = {};
+
+  if (perfSetting && perfSetting.value) {
+    try {
+      perfConfig = JSON.parse(perfSetting.value);
+    } catch (e) {
+      console.warn('ui-performance setting contains invalid data'); // eslint-disable-line no-console
+    }
+  }
+
+  // Start with the default and overwrite the values from the setting - ensures we have defaults for newly added options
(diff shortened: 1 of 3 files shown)


Sean-McQ (approved):
LGTM

PR #6997, opened 2022-09-26
HARVESTER: Hide read only namespaces

n313893254:
### Summary
Fixes Project Member cannot create resources in harvester-public ns

- https://github.com/harvester/harvester/issues/1986

### Technical notes summary

- Add a product prop: `hideRONamespace`


mantis-toboggan-md (changes requested):
I think this should apply to NameNsDescription for all products: a cluster explorer user with read-only access to namespaces will see the same error if they try to create a resource in that namespace: 
Screen Shot 2022-09-26 at 1 56 05 PM

In this example, the `cattle-system` namespace does not have `links.update` so the Harvester fix works here, too.

mantis-toboggan-md (approved):
LGTM

PR #7009, opened 2022-09-27
Allow to generate new yarn.lock by reviewing version

### LINKED ISSUE #6035: Allow to generate new yarn.lock by reviewing version
**Detailed Description**
Review used libraries versions in the `package.json` file, so it can be possible to generate a new `yarn.lock`.

**Context**
We currently cannot generate a new `yarn.lock` file as not pinned or verified libraries version break the app. This creates issues especially in case of merge conflicts, as the files needs to be modified by hand, which is totally contrary at the reason of the file lock itself.

cnotv:
### Summary
Fixes #6035
As defined in ticket

All the libraries versions have been pinned and a new yarn.lock has been generated.

### Technical notes summary

All the version are now pinned and updated to the latest `resolved` value defined in `yarn.lock`.  These are the one actually used in the end, which do not cause issues.

List can be printed out with the Yarn and OS related copy commands (in this case Mac)
```
yarn list | pbcopy
```

Each package in `pkg` needs to have the same versions or it would generate issues, so we'll need to define some guidelines for updating them, on top of automations

All the commands, which are mainly on CI.

All the commands, which are mainly on CI.

### Screenshot/Video
CLI.


rak-phillip (approved):
LGTM - @cnotv and I discussed in-depth offline about the need to generate new a new lockfile, as described in the issue. Pinning our dependencies in `package.json` definitely gives us more certainty moving forward.

PR #7011, opened 2022-09-27
Adds a default return to home when no route history

scures:
### Summary
Fixes # https://github.com/epinio/ui/issues/159

### Technical notes summary

The current `back-button` component works when a route history is presented (from where you landed there)
If none is found, it would not render.
With the change, we have a generic return to Home (`/`) button.

Normal scenario:
1. Click on the `dev` link at the bottom of the footer.
2. You should see the `< Back` button at the top of the new page.

image

No history scenario:
1. Refresh the page
2. You should see the `< Home` button at the top of the new page.

image


mmartin24:
tested on branch as commented here:
https://github.com/epinio/ui/issues/159#issuecomment-1260977457

PR #7016, opened 2022-09-27
remove the user preference for language toggle

aalves08:
- Removed the user preference `Enable Language Selection keyboard shortcut toggle (shift+L)` as it's only meant to be used by developers and gate it to be available on DEV builds only, as agreed with @nwmac 

**Test**
- Go to user preferences and check if option is no longer there
- make sure shortcut `SHIFT+L` still works on DEV builds

**Screenshots**
Screenshot 2022-09-27 at 15 51 39


jordojordo (approved):
lgtm

PR #7025, opened 2022-09-28
Allow wildcards in ingress hostnames

### LINKED ISSUE #6957: Invalid hostname wildcards for the ingress rule
**Setup** 
- Rancher version: v2.6.8
- Browser type & version: Chrome 105.0.5195.125 (x86_64) 

**Describe the bug**

The UI will prompt for invalid characters when using wildcards to configure the request host.

However, since k8s 1.18 hostname wildcards have been supported.
https://kubernetes.io/blog/2020/04/02/improvements-to-the-ingress-api-in-kubernetes-1.18/#support-for-hostname-wildcards

At the same time, the UI also supports ingressclass, and there should be a lot of ingress providers that comply with the API specification. The UI is consistent with the Ingress API specification, which should be a better choice.

**Result**
UI forms do not allow hostname wildcards.

**Expected Result**

UI forms allow using hostname wildcards.

**Screenshots**

截屏2022-09-22 上午9 48 03

bashofmann:
### Summary
Wildcards like `*.example.com` are allowed in ingress hostnames since Kubernetes 1.18: https://kubernetes.io/blog/2020/04/02/improvements-to-the-ingress-api-in-kubernetes-1.18/#support-for-hostname-wildcards.

Fixes #6957

This adds a new validation rule `wildcardHostname` including tests, which extends the `hostname` rule to also allow a wildcard in the first part of the hostname:

* `*.foo.example.com` is valid
* `foo.*.example.com` is not valid
* `foo.*example.com` is not valid
* `*foo.example.com` is not valid

### Technical notes summary
Allowing wildcards was added in https://github.com/kubernetes/kubernetes/pull/29204.

Creating and editing ingresses.

The hostname and tls hostname fields.

### Screenshot/Video

Working validations:

Bildschirmfoto 2022-09-28 um 07 36 58
Bildschirmfoto 2022-09-28 um 07 36 49

### DIFF
--- a/shell/edit/networking.k8s.io.ingress/index.vue
+++ b/shell/edit/networking.k8s.io.ingress/index.vue
@@ -65,7 +65,7 @@ export default {
           path: 'metadata.name', rules: ['required', 'hostname'], translationKey: 'nameNsDescription.name.label'
         },
         {
-          path: 'spec.rules.host', rules: ['hostname'], translationKey: 'ingress.rules.requestHost.label'
+          path: 'spec.rules.host', rules: ['wildcardHostname'], translationKey: 'ingress.rules.requestHost.label'
         },
         {
           path: 'spec.rules.http.paths.path', rules: ['absolutePath'], translationKey: 'ingress.rules.path.label'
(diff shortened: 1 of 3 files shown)


Sean-McQ (approved):
Looks great, thanks for the tests!

PR #7031, opened 2022-09-28
Add gke hosted provider to token creation

### LINKED ISSUE #6838: When provisioning a private GKE cluster, there is no `Registration` tab showing the registration command to connect the cluster to Rancher
**Setup**
- Rancher version: `v2.6.8`

**Describe the bug**

Similar to previously-logged issue https://github.com/rancher/dashboard/issues/6036, when you go to create a private GKE cluster within Rancher, there is no registration tab present. That means, if you do not copy the registration command presented after you provision the cluster, you need to go to the old UI and get the UI that way:

image

As we did with EKS and AKS, we need to ensure that we expose the registration tab so that it is easier for users to be able to find the registration command needed to finish configuring their private GKE clusters in Rancher.

**To Reproduce**

1. Setup Rancher, create a standard user and login as that user.
2. Navigate to Cluster Management and choose GKE.
3. Fill out the prompted information and ensure that you specify that this is a private cluster.
4. Once the cluster appears in Cluster Management, click the cluster and note that there is no `Registration` tab to grab the registration command.

**Result**
There is no registration tab containing the `kubectl apply` command to register the cluster.

**Expected Result**
There needs to be a registration tab, similar to what we have for private AKS and private EKS clusters.

jordojordo:
### Summary
Fixes #6838 

Similar to https://github.com/rancher/dashboard/issues/6036 where the cluster token was not being generated for a GKE hosted cluster, which would result in the registration tab not showing.

### Technical notes summary

Reworked how we were determining if a cluster was from a hosted k8s provider into one method (`isHostedKuberenetesProvider`). Also changed the conditional in `showEksNodeGroupWarning` to check directly for `this.value.provisioner === 'EKS'` since I merged the original `isEKS` into the new method.

- Creating a GKE cluster with a private endpoint

### Screenshot/Video

gke-private

### DIFF
--- a/shell/detail/provisioning.cattle.io.cluster.vue
+++ b/shell/detail/provisioning.cattle.io.cluster.vue
@@ -93,7 +93,7 @@ export default {
       fetchOne.snapshots = this.$store.dispatch('management/findAll', { type: SNAPSHOT });
     }
 
-    if ( this.value.isImported || this.value.isCustom || this.value.isAKS || this.value.isEKS ) {
+    if ( this.value.isImported || this.value.isCustom || this.value.isHostedKubernetesProvider ) {
       fetchOne.clusterToken = this.value.getOrCreateToken();
     }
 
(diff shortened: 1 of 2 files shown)


nwmac (approved):
LGTM

PR #7049, opened 2022-09-29
Adds validation for git URL

scures:
### Summary
Fixes # https://github.com/epinio/ui/issues/107

### Technical notes summary

1. Adds validation on input based on regex, it makes sure it matches the correct format of `{user or org}/{repository}`
2. Disables the branch input unless the URL is valid.

1. Refer to the issue https://github.com/epinio/ui/issues/107 to understand the problem.
2. Try adding an invalid URL, it should disable the branch input and throw an error
3. Try adding a valid URL, it should enable the `branch` input and allows you to continue.

### DIFF
--- a/pkg/epinio/components/application/AppSource.vue
+++ b/pkg/epinio/components/application/AppSource.vue
@@ -164,7 +164,7 @@ export default Vue.extend<Data, any, any, any>({
 
   methods: {
     urlRule() {
-      const gitRegex = /(git@|https?:\/\/)([a-zA-Z0-9\.\-_]+)(\/|:)([a-zA-Z0-9\-]+)\/([a-zA-Z0-9\-]+)/gm;
+      const gitRegex = /(https?:\/\/(?:www\.|(?!www))[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]\.[^\s]{2,}|www\.[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]\.[^\s]{2,}|https?:\/\/(?:www\.|(?!www))[a-zA-Z0-9]+\.[^\s]{2,}|www\.[a-zA-Z0-9]+\.[^\s]{2,})/gm;
 
       if (!this.gitUrl.url) {
         return;
--- a/pkg/epinio/l10n/en-us.yaml
+++ b/pkg/epinio/l10n/en-us.yaml
@@ -149,7 +149,7 @@ epinio:
             inputLabel: Branch
             tooltip: Branch to deploy
           error:
-            label: Invalid Github/Gitlab URL, please check the URL and try again
+            label: Invalid URL, please check the URL and try again.
         gitHub:
           label: GitHub
 


mmartin24:
I checked the following scenarios in branch:

### Case 1: URL with structure different from `{user or org}/{repository}` throws an error and prevents user from entering branch:
Example: `https://github.com/epinio/example-go/wrong`
image

### Case 2: User is allowed to enter if structure is correct:
Example: `https://github.com/epinio/example-go`
image

However I found 2 scenarios where an error may still be produced:

### Case 3: Changing to bad URL after a correct previous one with branch included

Steps to reproduce: 
- Enter valid URL (ie: `https://github.com/epinio/example-go`)
- Enter valid branch (ie: `main`)
- Change URL to invalid (ie: `https://github.com/epinio/example-go/wrong`

Result: user not prevented from going to next screen:
image

### Case 4: Entering a bad URL with correct structure

Steps to reproduce: 
- Enter valid URL with bad ending (ie: `- Enter valid URL (ie: `https://github.com/epinio/example-gobadending`)

Result: user not preventing from next step:
image

@scures, cases 1 and 2 are simple cases strictly related to the pr; cases 3 and 4 are edge cases that may be addressed here or in another issue.

mmartin24:
thanks @scures .
Checked Case 3 and working:
image

I hear what you say about case 4. If too complicated; let's focus for the moment other place.

richard-cox (changes requested):
The Git URL can be any URL that points to a publicly available git instance, so validation should only match that of a URL rather than something github/gitlab specific. @enrichman could you confirm?

Looking back at the issue, we'd need to do some kind of pre-fetch to determine if there was an git instance there (and not just _something_ there) and i'm not sure that's possible

enrichman:
> The Git URL can be any URL that points to a publicly available git instance, so validation should only match that of a URL rather than something github/gitlab specific. @enrichman could you confirm?
> 
> Looking back at the issue, we'd need to do some kind of pre-fetch to determine if there was an git instance there (and not just _something_ there) and i'm not sure that's possible (in a simple UX flow)

Yes, I would expect this working also with Gitlab, and maybe in the future with private Github enterprise instances.
Is it working with Gitlab?

richard-cox:
> > The Git URL can be any URL that points to a publicly available git instance, so validation should only match that of a URL rather than something github/gitlab specific. @enrichman could you confirm?
> > Looking back at the issue, we'd need to do some kind of pre-fetch to determine if there was an git instance there (and not just _something_ there) and i'm not sure that's possible (in a simple UX flow)
> 
> Yes, I would expect this working also with Gitlab, and maybe in the future with private Github enterprise instances. Is it working with Gitlab?

@enrichman Would it be expected to work with non-github/gitlab urls though, as in, anything supported by `git clone`?

enrichman:
> > > The Git URL can be any URL that points to a publicly available git instance, so validation should only match that of a URL rather than something github/gitlab specific. @enrichman could you confirm?
> > > Looking back at the issue, we'd need to do some kind of pre-fetch to determine if there was an git instance there (and not just _something_ there) and i'm not sure that's possible (in a simple UX flow)
> > 
> > 
> > Yes, I would expect this working also with Gitlab, and maybe in the future with private Github enterprise instances. Is it working with Gitlab?
> 
> @enrichman Would it be expected to work with non-github/gitlab urls though, as in, anything supported by `git clone`?

I would say yes, but I'm not sure if this is too much of work. I haven't followed much the issue, so I'm not sure if it's something feasible or not.

richard-cox:
Ok, i think for this issue then we should just validate that the URL is in the correct format. We have another issue that should address how to recover from a failed app deploy step - https://github.com/epinio/ui/issues/49

richard-cox (changes requested):
Following on from the PR conversation the Git Url validation should only validate if the URL is valid and not if it's a valid github/gitlab url. 

This doesn't quite resolve the original issue, but validating if a git instance is at the end of a url and that it supplies a valid point to checkout would be quite involved. 

There is an existing general issue that would help resolve the original issue's problem though https://github.com/epinio/ui/issues/49 will he

pkg/epinio/components/application/AppSource.vue
```diff
@@ -443,10 +465,14 @@ export default Vue.extend<Data, any, any, any>({
         <h3>{{ t('epinio.applications.steps.source.gitUrl.url.label') }}</h3>
         <LabeledInput
           v-model="gitUrl.url"
+          v-focus
```
richard-cox, line 468: At some point this needs to be added to the first input field of the other types as well

richard-cox (approved):

PR #7051, opened 2022-09-29
Build side nav with fewer watchers and more computed properties

catherineluse:
Fixes https://github.com/rancher/dashboard/issues/4849, related to https://github.com/rancher/dashboard/issues/6485

When looking into https://github.com/rancher/dashboard/issues/4849, I saw that the code in the `allSchemas` watcher wasn't running when the schemas changed. I experimented with what would happen if `getGroups` was replaced with a computed property `groups` so that it would automatically update when its dependencies changed, instead of relying on watchers.

This PR is a refactor of the file that builds the side nav. It removes 10 watchers whose only purpose was to update the side nav links, replacing them with dependencies of the new `groups` computed property.

To test this PR, I installed and reinstalled Helm charts such as CIS Benchmark and Longhorn, and confirmed that they were added and removed from the side nav as expected. I also did context changes such as changing `locale` and `clusterId` to make sure the side nav behaves the same as previously.

### DIFF
--- a/shell/layouts/default.vue
+++ b/shell/layouts/default.vue
@@ -1,5 +1,4 @@
 <script>
-import debounce from 'lodash/debounce';
 import { mapState, mapGetters } from 'vuex';
 import {
   mapPref,
@@ -24,7 +23,7 @@ import {
   COUNT, SCHEMA, MANAGEMENT, UI, CATALOG
 } from '@shell/config/types';
 import { BASIC, FAVORITE, USED } from '@shell/store/type-map';
-import { addObjects, replaceWith, clear, addObject } from '@shell/utils/array';
+import { addObjects, replaceWith, addObject } from '@shell/utils/array';
 import { NAME as EXPLORER } from '@shell/config/product/explorer';
 import { NAME as NAVLINKS } from '@shell/config/product/navlinks';
 import { HARVESTER_NAME as HARVESTER } from '@shell/config/product/harvester-manager';
@@ -39,7 +38,6 @@ import { getProductFromRoute } from '@shell/middleware/authenticated';
 const SET_LOGIN_ACTION = 'set-as-login';
 
 export default {
-
   components: {
     PromptRemove,
     PromptRestore,
@@ -61,7 +59,6 @@ export default {
   data() {
     return {
       noLocaleShortcut: process.env.dev || false,
-      groups:           [],
       gettingGroups:    false,
       wantNavSync:      false,
     };
@@ -76,6 +73,10 @@ export default {
     ...mapState(['managementReady', 'clusterReady']),
     ...mapGetters(['productId', 'clusterId', 'namespaceMode', 'isExplorer', 'currentProduct', 'isSingleProduct']),
     ...mapGetters({ locale: 'i18n/selectedLocaleLabel', availableLocales: 'i18n/availableLocales' }),
+    ...mapGetters({
+      locale:           'i18n/selectedLocaleLabel',
+      availableLocales: 'i18n/availableLocales',
+    }),
     ...mapGetters('type-map', ['activeProducts']),
 
     afterLoginRoute: mapPref(AFTER_LOGIN_ROUTE),
@@ -101,7 +102,7 @@ export default {
       if (canSetAsHome) {
         pageActions.push({
           labelKey: 'nav.header.setLoginPage',
-          action:   SET_LOGIN_ACTION
+          action:   SET_LOGIN_ACTION,
         });
       }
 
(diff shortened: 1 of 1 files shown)


shell/layouts/default.vue
```diff
@@ -194,224 +209,101 @@ export default {
...
-      // Always show cluster-level types, regardless of the namespace filter
-      const namespaceMode = 'both';
       const out = [];
-      const loadProducts = this.isExplorer ? [EXPLORER] : [];
+
       const productMap = this.activeProducts.reduce((acc, p) => {
         return { ...acc, [p.name]: p };
       }, {});
 
-      if ( this.isExplorer ) {
```
catherineluse, line 377: moved this code to a separate `loadedProducts` computed property for legibility

shell/layouts/default.vue
```diff
@@ -61,21 +59,27 @@ export default {
...
-    'authenticated'
-  ],
+  middleware: ['authenticated'],
 
   computed: {
     ...mapState(['managementReady', 'clusterReady']),
-    ...mapGetters(['productId', 'clusterId', 'namespaceMode', 'isExplorer', 'currentProduct', 'isSingleProduct']),
-    ...mapGetters({ locale: 'i18n/selectedLocaleLabel', availableLocales: 'i18n/availableLocales' }),
+    ...mapGetters([
+      'productId',
```
catherineluse, line 72: formatting changes only

shell/layouts/default.vue
```diff
@@ -84,24 +88,27 @@ export default {
...
       const product = this.$store.getters['currentProduct'];
 
-      if ( !product ) {
+      if (!product) {
         return [];
       }
 
       // Only show for Cluster Explorer or Global Apps (not configuration)
-      const canSetAsHome = product.inStore === 'cluster' || (product.inStore === 'management' && product.category !== 'configuration');
+      const canSetAsHome =
```
catherineluse, line 103: formatting changes only

shell/layouts/default.vue
```diff
@@ -112,15 +119,18 @@ export default {
...
+      if (!managementReady || !product) {
         return [];
       }
 
       return this.$store.getters[`${ product.inStore }/all`](SCHEMA);
     },
 
     allNavLinks() {
-      if ( !this.clusterId || !this.$store.getters['cluster/schemaFor'](UI.NAV_LINK) ) {
+      if (
```
catherineluse, line 130: formatting changes only

shell/layouts/default.vue
```diff
@@ -194,224 +209,101 @@ export default {
...
         this.clusterId === this.$route?.params?.cluster &&
-        this.currentProduct?.name === getProductFromRoute(this.$route);
-    },
-
-  },
-
-  watch: {
-    counts(a, b) {
-      if ( a !== b ) {
-        this.queueUpdate();
```
catherineluse, line 207: Removed all watchers that only do `queueUpdate` which just updates the nav tree, as they are now dependencies of the `groups` computed property

shell/layouts/default.vue
```diff
@@ -194,224 +209,101 @@ export default {
...
-      if ( !isEqual(a, b) ) {
-        // Store the last visited route when the cluster changes
-        this.setClusterAsLastRoute();
-      }
-    },
-
-    namespaceMode(a, b) {
-      if ( !isEqual(a, b) ) {
-        // Immediately update because you'll see it come in later
-        this.getGroups();
```
catherineluse, line 252: Deleted the watchers that only update the nav tree because it is automatically updated now.

shell/layouts/default.vue
```diff
@@ -194,224 +209,101 @@ export default {
...
-    },
-
-    $route(a, b) {
-      this.$nextTick(() => this.syncNav());
-    },
-
-  },
-
-  async created() {
-    this.queueUpdate = debounce(this.getGroups, 500);
```
catherineluse, line 300: I removed the debounce because I think it's no longer necessary because we now limit how often are schemas are updated, so that should have a similar effect as the debounce, unless I'm mistaken.

shell/layouts/default.vue
```diff
@@ -194,224 +209,101 @@ export default {
...
 
-    getLoginRoute() {
-      return {
-        name:   this.$route.name,
-        params: this.$route.params
-      };
+      return loadProducts;
     },
 
-    collapseAll() {
```
catherineluse, line 339: For now, I left the manual updates of `collapseAll` and the nav sync method, though they could be refactored in a similar way if we wanted to.

shell/layouts/default.vue
```diff
@@ -194,224 +209,101 @@ export default {
...
 
+      // Always show cluster-level types, regardless of the namespace filter
       if ( !this.$store.getters['isAllNamespaces'] ) {
         const namespacesObject = this.$store.getters['namespaces']();
 
         namespaces = Object.keys(namespacesObject);
       }
-
-      // Always show cluster-level types, regardless of the namespace filter
-      const namespaceMode = 'both';
```
catherineluse, line 370: Removed the `namespaceMode` here because there was already a computed property for it which was 'both' by default.

shell/layouts/default.vue
```diff
@@ -194,224 +209,101 @@ export default {
...
       const out = [];
-      const loadProducts = this.isExplorer ? [EXPLORER] : [];
```
catherineluse, line 372: Removed `loadProducts` here as it changed to a separate computed property `loadedProducts`.

shell/layouts/default.vue
```diff
@@ -466,29 +358,104 @@ export default {
...
+    },
+  },
+
+  async created() {
+    await this.$store.dispatch('prefs/setLastVisited', this.$route);
+  },
+
+  mounted() {
+    // Sync the navigation tree on fresh load
+    this.$nextTick(() => this.syncNav());
```
catherineluse, line 425: If we wanted this to be more declarative, the open/closed group state could also be refactored so that we wouldn't need `syncNav`.

shell/layouts/default.vue
```diff
@@ -593,22 +560,27 @@ export default {
             />
           </template>
         </div>
-        <n-link v-if="showClusterTools" tag="div" class="tools" :to="{name: 'c-cluster-explorer-tools', params: {cluster: clusterId}}">
+        <n-link
+          v-if="showClusterTools"
+          tag="div"
```
catherineluse, line 565: formatting changes only

shell/layouts/default.vue
```diff
@@ -194,224 +209,101 @@ export default {
...
-
-          if ( productId === EXPLORER || !this.isExplorer ) {
+        for (const mode of modes) {
+          const types =
+            this.$store.getters['type-map/allTypes'](productId, mode) || {};
+
+          // Because the "clusterId", "namespaces" and "namespaceMode"
+          // computed properties are dependencies of "groups", the nav tree will be
+          // recomputed if they change.
+          const more = this.$store.getters['type-map/getTree'](
```
catherineluse, line 281: This line references a bunch of computed properties that will now trigger updates.

shell/layouts/default.vue
```diff
@@ -148,9 +158,11 @@ export default {
     },
 
     showClusterTools() {
-      return this.isExplorer &&
-             this.$store.getters['cluster/canList'](CATALOG.CLUSTER_REPO) &&
-             this.$store.getters['cluster/canList'](CATALOG.APP);
+      return (
```
catherineluse, line 161: formatting change only

shell/layouts/default.vue
```diff
@@ -179,7 +191,10 @@ export default {
       const product = this.$store.getters['currentProduct'];
 
       if (product?.supportRoute) {
-        return { ...product.supportRoute, params: { ...product.supportRoute.params, cluster: this.clusterId } };
+        return {
```
catherineluse, line 194: formatting change only

shell/layouts/default.vue
```diff
@@ -194,224 +209,101 @@ export default {
...
-      if ( !isEqual(a, b) ) {
-        if (a.inStore !== b.inStore || a.inStore !== 'cluster' ) {
-          const route = {
-            name:   'c-cluster-product',
-            params: {
-              cluster: this.clusterId,
-              product: a.name,
-            }
-          };
+    loadedProducts() {
```
catherineluse, line 219: separated `loadedProducts` from what was previously `getGroups` just for legibility

shell/layouts/default.vue
```diff
@@ -194,224 +209,101 @@ export default {
...
-
-    await this.$store.dispatch('prefs/setLastVisited', this.$route);
-  },
-
-  mounted() {
-    // Sync the navigation tree on fresh load
-    this.$nextTick(() => this.syncNav());
-  },
-
-  methods: {
```
catherineluse, line 312: I didn't change the methods

shell/layouts/default.vue
```diff
@@ -194,224 +209,101 @@ export default {
...
-        return;
-      }
-      this.gettingGroups = true;
-
-      if ( !this.clusterReady ) {
-        clear(this.groups);
-        this.gettingGroups = false;
-
-        return;
+    groups() {
```
catherineluse, line 241: The whole nav tree structure is now this `groups` computed property.

shell/layouts/default.vue
```diff
@@ -194,224 +209,101 @@ export default {
...
             addObjects(out, more);
           } else {
             const root = more.find(x => x.name === 'root');
             const other = more.filter(x => x.name !== 'root');
 
             const group = {
-              name:     productId,
-              label:    this.$store.getters['i18n/withFallback'](`product.${ productId }`, null, ucFirst(productId)),
+              name:  productId,
+              // Pass in the locale as an argument, even though it is not needed,
```
catherineluse, line 299: formatting changes only

shell/layouts/default.vue
```diff
@@ -466,29 +358,104 @@ export default {
...
+        return out;
       }
 
       replaceWith(this.groups, ...sortBy(out, ['weight:desc', 'label']));
-      this.gettingGroups = false;
+
+      return out;
+    },
+  },
+
```
catherineluse, line 389: For now, I left the remaining 3 watchers which have to do with routing.

shell/layouts/default.vue
```diff
@@ -466,29 +358,104 @@ export default {
...
+    async setClusterAsLastRoute() {
+      const route = {
+        name:   this.$route.name,
+        params: {
+          ...this.$route.params,
+          cluster: this.clusterId,
+        },
+      };
+
+      await this.$store.dispatch('prefs/setLastVisited', route);
```
catherineluse, line 438: I didn't change any methods

shell/layouts/default.vue
```diff
@@ -199,175 +200,30 @@ export default {
...
+      if (!this.clusterReady || !this.allSchemas || !this.counts || !this.locale || !this.currentProduct) {
+        // Referencing these computed properties
+        // to trigger the nav tree to be recomputed if they change.
+        return [];
       }
-
       const clusterId = this.$store.getters['clusterId'];
       const currentProduct = this.$store.getters['productId'];
       const currentType = this.$route.params.resource || '';
       let namespaces = null;
```
jordojordo, line 216: Would using the computed namespaces be more efficient as it uses the `activeNamespaceCache`? If so, the conditional and getter at L221-L225 could be removed as well.

jordojordo (changes requested):
I generally advocate for replacing these methods with computed properties, but there are a couple of issues that arise as consequence (see comments). But the main issue is that this doesn't fix the problem of resources persisting in the side-nav once they're uninstalled/deleted.

shell/layouts/default.vue
catherineluse: I think the blank home page was a separate bug because after I rebased on master, the Home page worked again.

PR #7058, opened 2022-09-30
Remove uneeded check - add custom check for old endpoints

### LINKED ISSUE #6428: Azure AD Edit Config always picks Standard endpoint in radio button

jordojordo:
### Summary
Fixes #6428

### Technical notes summary

If the `graphEndpoint` matches the old version (`https://graph.windows.net/`), the conditional will now return the proper endpoint option.

### Testing
A way of testing a custom endpoint while using the standard set of URIs ( e.g. Graph Endpoint: `https://graph.windows.net/`, Token Endpoint: `https://login.microsoftonline.com/${ TENANT_ID_TOKEN }/oauth2/token` ) would be to change the Token or Auth endpoint property in `OLD_ENDPOINTS` to an arbitrary string.

### DIFF
--- a/shell/edit/auth/azuread.vue
+++ b/shell/edit/auth/azuread.vue
@@ -182,12 +182,10 @@ export default {
       const oldEndpointKey = Object.keys(OLD_ENDPOINTS).find(key => OLD_ENDPOINTS[key].graphEndpoint === endpoint);
 
       if ( oldEndpointKey ) {
-        this.endpoint = oldEndpointKey;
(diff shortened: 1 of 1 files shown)


Sean-McQ (approved):
Jordan walked me through testing it, code looks good. Approved.

PR #7061, opened 2022-09-30
Fix unique user count on bound user role deletion

### LINKED ISSUE #7046: Remove a ROLE : The Confirmation window shows incorrect user count
**Setup**
- Rancher version: v2.6-2e0026699d7f3498979f9007a17b20668cb076ef-head
- Browser type & version: Chrome (Version 105.0.5195.125 ) on MAC

**Describe the bug**

While removing a ROLE, the user is getting a confirmation window. On the window the Caution in RED color - the 'number' of bounded users to the roles is incorrect. It is displaying the SUM of all the users counts from all the ROLES which are associate to the each role. It would be good to have the actual number of users bounded as a total.
Ex: Role 1 has 2 users (U1, U2)
Role 2 has users (U1, U3)

The dialog window is showing as 4 users instead of 3 users as U1 is common in both roles

Note: In the attached screen the total users shows 76 which leads some confusion as this node has just 8 users all together.

**To Reproduce**
**Steps:**

1. Create a User
2. Create a Role
3. Create a Cluster
4. Create a Project Name Space
5. Create a Project
6. Associate Users to multiple roles
7. Roles : Select multiple Roles (Global / Custom)
8. Try to delete multiple Roles and check the text 'Caution: There are 76 users bound to the role(s) about to be deleted. Do you want to proceed?'

**Actual Behavior:**
1. On the window the Caution in RED color - the 'number' of bounded users to the roles is incorrect. It is displaying the SUM of all the user counts from all the ROLES which are associate to the each role. It would be good to have the actual number of users bounded as a total.
2. In the attached screen the total users shows 76 which leads some confusion as this node has just 8 users all together.
	

**Expected Behavior:**
1. Appropriate bounded users should display instead of summing all the users counts
2. In the attached screen the total users shows 76 which leads some confusion as this node has just 8 users all together and it should say 8 users instead of 76 users.

image
rancher/dashboard#6578

jordojordo:
### Summary
Fixes #7046

The count of unique users bound to multiple roles would stack regardless if they were unique. This changes the number of unique users to a set, which will remove the duplicate user names from the count.

### Technical notes summary

Converted the `numberUniqueUsersWithBinds` variable to a Set, adding the unique users from each role checked, then returning the size of the set as the user count.

- Removing multiple cluster member roles that are bound to multiple users
- The same should be done with project roles

### Screenshot/Video

https://user-images.githubusercontent.com/40806497/193354591-9013c073-83e8-4cb5-829b-2ad77bbe0eac.mp4

### DIFF
--- a/shell/promptRemove/mixin/roleDeletionCheck.js
+++ b/shell/promptRemove/mixin/roleDeletionCheck.js
@@ -41,7 +41,7 @@ export default {
       let resourceToCheck;
       let propToMatch;
       let numberOfRolesWithBinds = 0;
-      let numberUniqueUsersWithBinds = 0;
+      const uniqueUsersWithBinds = new Set();
 
       this.info = this.t('rbac.globalRoles.waiting', { count: rolesToRemove.length });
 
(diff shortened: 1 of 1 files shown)


nwmac (approved):
LGTM

Issue #7036, opened 2022-09-28
k3s cluster in `error` state still accessible through `explore cluster` and home page


slickwarren:
**Setup**
- Rancher version: v2.7-head (c567ef3)
- Browser type & version: chrome

**Describe the bug**

if a k3s cluster gets into an error state, the cluster is still selectable from the `explore cluster` menu and home page

**To Reproduce**

* deploy a k3s cluster with dedicated nodes-per-role
* wait for the cluster to get to an active state
* edit the cluster.yaml and add the following to the `rkeConfig` section:
```
machineSelectorConfig:
    - config:
      kubelet-arg:
        - 'streaming-connection-idle-timeout=5m'
      flannel-conf: /tmp/fake.conf
      flannel-iface: eth0
      kube-proxy-arg: feature-gates=CSIMigrationAWS=false
      resolv-conf: /tmp/resolv.conf
```
* wait for the worker node to get to an `error` state (~5 minutes)

**Result**
the cluster gets to an error state, but the cluster is still selectable from both the home page and in the `explore cluster` menu
clicking the cluster results in a 500 error

**Expected Result**

cluster should not be accessible if it is in an error state

**Screenshots**

<img width="782" alt="Screen Shot 2022-09-28 at 4 22 37 PM" src="https://user-images.githubusercontent.com/16691014/192906018-ae23adb2-11c0-4384-9744-1caab2d58dba.png">

**Additional context**

PR #7068, opened 2022-10-03
Disabling clusters in the top level menu if the cluster is in an error state

codyrancher:
### Summary
Fixes rancher/dashboard#7036

This should be isolated to the top level menu where the clusters are listed

Broken and healthy clusters
A simpler way to reproduce:
- Create a cluster with a single node
- Edit yaml of cluster
- Change `cloudCredentialSecretName: cattle-global-data:cc-4twkpo` to be a secret that doesn't exist.

This should be isolated only to the top level menu where the clusters are listed

### Screenshot/Video
image
image
image

### DIFF
--- a/shell/components/nav/TopLevelMenu.vue
+++ b/shell/components/nav/TopLevelMenu.vue
@@ -65,9 +65,10 @@ export default {
     clusters() {
       const all = this.$store.getters['management/all'](MANAGEMENT.CLUSTER);
       let kubeClusters = filterHiddenLocalCluster(filterOnlyKubernetesClusters(all), this.$store);
+      let pClusters = null;
 
       if (this.hasProvCluster) {
-        const pClusters = this.$store.getters['management/all'](CAPI.RANCHER_CLUSTER);
+        pClusters = this.$store.getters['management/all'](CAPI.RANCHER_CLUSTER);
         const available = pClusters.reduce((p, c) => {
           p[c.mgmt] = true;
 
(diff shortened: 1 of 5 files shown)


nwmac:
@codyrancher There was an issue that some dependencies got updated that broke the plugins validation action - fixes have been merged to master to address this - you'll need to rebase this PR to get the fix so that the GH Action passes

jordojordo (approved):
LGTM - Should we disable the ability to navigate to error'ed clusters from the mgmt and home page in a separate PR?

shell/components/nav/TopLevelMenu.vue
```diff
@@ -182,6 +182,11 @@ export default {
     $route() {
       this.shown = false;
     },
+    shown() {
+      // We need to reload the clusters so we can get the status from the cluster relationships.
+      // Specifically this is to determine if the cluster is in an error state.
+      this.$store.dispatch('management/findAll', { type: MANAGEMENT.CLUSTER, opt: { force: true } });
```
nwmac, line 188: Isn't this going to make an API call every time someone clicks to show the menu?

I'm not sure we want that - don't we get the updates via the watches on the resource?

nwmac (changes requested):

shell/components/nav/TopLevelMenu.vue
codyrancher: We don't get updates from the watches. It does make a request for each click.

codyrancher: It looks like provisioning cluster receives updates but management cluster doesn't. I'll see if I can use provisioning cluster instead.

codyrancher:
> LGTM - Should we disable the ability to navigate to error'ed clusters from the mgmt and home page in a separate PR?

I updated both of those places.

PR #7070, opened 2022-10-04
Fix shell package dependencies

nwmac:
This PR fixes the dependency problem that is causing the Plugins Validation GitHub action to fail.

The dependencies for the `shell` package were not pinned, like the top-level, this was the root cause of the failure, as newer aws dependencies were being picked up.

This PR adds a script that can be run to sync the dependencies between the top-level package file and the one in the `shell` folder.

This PR also contains the updated package file having run this script.


aalves08 (approved):
LGTM

PR #7074, opened 2022-10-04
[FP] Remove uneeded check - add custom check for old endpoints (#7058)

jordojordo:
### Summary
Forward port for #6428 

Original PR #7058


Sean-McQ (approved):
LGTM

PR #7077, opened 2022-10-04
Plugin updates for custom charts within pkg

jordojordo:
This adds the ability to use custom charts from within plugin.

### DIFF
--- a/shell/pages/c/_cluster/apps/charts/install.vue
+++ b/shell/pages/c/_cluster/apps/charts/install.vue
@@ -762,7 +762,6 @@ export default {
 
         if ( hasChartComponent ) {
           this.valuesComponent = this.$store.getters['type-map/importChart'](component);
-          this.valuesComponent = this.$store.getters['catalog/importComponent'](component);
           const loaded = await this.valuesComponent();
 
           this.showValuesComponent = true;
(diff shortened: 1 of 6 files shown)


shell/pages/c/_cluster/apps/charts/install.vue
```diff
@@ -758,7 +758,10 @@ export default {
 
       // Load a values component for the UI if it is named in the Helm chart.
       if ( component ) {
-        if ( this.$store.getters['catalog/haveComponent'](component) ) {
+        const hasChartComponent = this.$store.getters['type-map/hasCustomChart'](component);
+
+        if ( hasChartComponent ) {
+          this.valuesComponent = this.$store.getters['type-map/importChart'](component);
```
nwmac, line 764: Don't think we need both lines that set `this.valuesComponent` - assume its only the first one?

nwmac (changes requested):
LGTM - one thing to fix

shell/pages/c/_cluster/apps/charts/install.vue
jordojordo: Ah good catch, left that in there by mistake. Removed `'catalog/importComponent'`

nwmac (approved):

PR #7088, opened 2022-10-05
Allow /rancherversion endpoint override

nwmac:


shell/nuxt.config.js
```diff
@@ -584,25 +586,26 @@ export default function(dir, _appConfig) {
...
+      '/k8s':            proxyWsOpts(api), // Straight to a remote cluster (/k8s/clusters/<id>/)
+      '/pp':             proxyWsOpts(api), // For (epinio) standalone API
+      '/api':            proxyWsOpts(api), // Management k8s API
+      '/apis':           proxyWsOpts(api), // Management k8s API
+      '/v1':             proxyWsOpts(api), // Management Steve API
+      '/v3':             proxyWsOpts(api), // Rancher API
+      '/v3-public':      proxyOpts(api), // Rancher Unauthed API
+      '/api-ui':         proxyOpts(api), // Browser API UI
+      '/meta':           proxyMetaOpts(api), // Browser API UI
+      '/rancherversion': proxyPrimeOpts(api), // Rancher version endpoint
```
Shavindra, line 598: Don't we need to update the standalone version as well?

richard-cox (approved):
Note for all, remember to `yarn clean` after changing this env (and others). Nuxt bakes them in to cached files which won't be updated when the env var changes

PR #7105, opened 2022-10-06
Edit labels and annotations.

Shavindra:
### Summary
Fixes:
 - #7094


cnotv:
@Shavindra could you please describe what was missing and required to be done? 
I've seen that was missing the `metadata`, the `WorkloadDeployments` name changed and something else, but this one I dont get it:

```
    if ( this.type === POD ) {
      return this.spec.initContainers;
    }
```

Should we maybe add a comment about why is this logic an exception?
Sorry to bother 😬

cnotv (changes requested):
I am not able to save the creation while I created the comparison video.


https://user-images.githubusercontent.com/5009481/194555624-ac80d293-33dd-4f24-af33-68d20b591578.mov

cnotv:
Now it seems adding some extra stuff which I dont get.

Screenshot 2022-10-07 at 14 53 32

Shavindra:
> Now it seems adding some extra stuff which I dont get.
That is expected because it's a system label.

cnotv (dismissed):
If system label are not a problem LGTM.

Please mind linters.

shell/edit/workload/index.vue
```diff
@@ -110,7 +110,7 @@ export default {
               <div>
                 <div :style="{'align-items':'center'}" class="row mb-20">
                   <div class="col span-6">
-                    <LabeledInput v-model="container.name" :mode="mode" :label="t('workload.container.containerName')" />
+                    <LabeledInput v-model="allContainers[0].name" :mode="mode" :label="t('workload.container.containerName')" />
```
cnotv, line 113: Should it not be `tab.name`?
Also why always the first one? If there's a valid reason, maybe we could write a comment.

shell/edit/workload/index.vue
```diff
@@ -127,7 +127,7 @@ export default {
                 <div class="row mb-20">
                   <div class="col span-6">
                     <LabeledInput
-                      v-model.trim="container.image"
+                      v-model.trim="allContainers[0].image"
```
cnotv, line 130: Same here as above.

cnotv (changes requested):

cnotv (approved):
LGTM

PR #7117, opened 2022-10-07
Can pull Helm chart app images from cluster registry or custom registry

catherineluse:
https://github.com/rancher/dashboard/issues/6910

To test:

1. Provision two downstream RKE2 clusters, one with a private registry in the cluster config and one without.
2. In the cluster that has a registry, install a helm chart with the default settings. (I used Rancher alerting drivers) Confirm that the cluster registry from the cluster config is used in the values.yaml under `values.global.cattle.systemDefaultRegistry`. 
3. In the cluster that does not have a registry, install a helm chart with the default settings. The `values.global.cattle.systemDefaultRegistry` should be an empty string because the global default registry in the settings is also an empty string.
4. In the global settings of Rancher, change `system-default-registry` to a different string. It can be any string.
5. In the cluster that does not have a registry, install or upgrade a chart and confirm that the systemDefaultRegistry now reflects the new default registry from the global settings. Note: This is where I got stuck - the new default setting is not sent in the network request.

I have not tested the following yet, but this is what I intend to test:
6. In any cluster, install or upgrade a chart. In the first page of the helm chart install workflow, select the new checkbox that says "Use Custom Container Registry for Rancher system images"
7. Enter a custom registry
8. Confirm that the new registry is in the `values.global.cattle.systemDefaultRegistry` in the values.yaml

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -767,10 +767,10 @@ catalog:
   chart:
     registry:
       label: Container Registry
-      tooltip: By default, container images are pulled from the Cluster Container Registry in the cluster configuration, and the Global Container Registry is used as a fallback. To use images from a non-default registry, enter a custom container registry here.
+      tooltip: Container images are pulled from the Cluster Container or, failing that, the System Container Registry Setting. To change this default behaviour enter or update the registry here
       custom:
-        checkBoxLabel: Use Custom Container Registry for {vendor} System Container Images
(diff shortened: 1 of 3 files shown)


catherineluse:
Can't fully test this PR until https://github.com/rancher/dashboard/issues/7127 is resolved. We need Team 2 to confirm the YAML format for configuring the cluster scoped registry

nwmac (approved):

nwmac (approved):

PR #7118, opened 2022-10-07
Added blue color only to creation action

### LINKED ISSUE #6412: In namespace selection dropdown, only "Create a New Namespace" should be blue
I opened this issue based on Kenneth's feedback from a demo.

Currently, all list items in the namespace dropdown are blue:
Screen Shot 2022-07-19 at 6 32 41 PM

But if only the "Create a New Namespace" was blue and the others were a normal text color, the creation action would stand out more.

bisht-richa:
### Summary
Fixes #6412

In the namespace selection dropdown, only "Create a New Namespace" should be blue

### How to test
Go to : https://192.168.43.142:8005/c/local/explorer/apps.deployment
Click on the `Create` button => Namespace dropdown should have **"Create a New Namespace"** in blue color and rest of the options should be in standard text color.


shell/assets/styles/vendor/vue-select.scss
```diff
@@ -103,7 +103,7 @@
   }
 
   &.vs__dropdown-option--highlight {
-    color: var(--dropdown-hover-text);
+    color: var(--dropdown-active-text);
```
cnotv, line 106: These names have been created semantically, so if this is for hover, it should stay like this and change the color for it.

shell/components/form/LabeledSelect.vue
```diff
@@ -284,6 +285,11 @@ export default {
         <template v-else-if="option.kind === 'divider'">
           <hr />
         </template>
+        <template v-else-if="option.kind === 'link'">
```
cnotv, line 288: Can we use a different term from link, as this is just highlighted and not a link?

shell/components/form/LabeledSelect.vue
```diff
@@ -284,6 +285,11 @@ export default {
...
+          <a>
```
cnotv, line 289: Could you also please replace the `a` element with a `div` as it's done for the other options?

Screenshot 2022-10-07 at 12 17 18

cnotv (changes requested):
On top of the code specific comments, there's 3 issues which I wanted to mention.

The Suse theme is still the old one and I cannot be sure that it is working fine in that case, see screenshots:

Screenshot 2022-10-07 at 12 13 33

In any case, the dark theme seems not be updated for the dropdown, so we may take the chance to do so also for the normal option text in dark mode.

Screenshot 2022-10-07 at 12 13 55

The whole behavior seems quite odd when clicking multiple times, as the value `Create New Namespace` seems kept in somehow and allows me to create the deployment and finally get a weird unclear error if I submit.


https://user-images.githubusercontent.com/5009481/194532849-744b10dd-038a-4838-97de-7a060c761f57.mov

shell/assets/styles/vendor/vue-select.scss
bisht-richa: Fixed

cnotv:
Created new issue for the weird behavior: https://github.com/rancher/dashboard/issues/7120

cnotv (dismissed):
Screenshot 2022-10-07 at 15 35 43
Screenshot 2022-10-07 at 15 36 00

Now it seems fine also for Prime/Suse.

cnotv (approved):
LGTM

PR #7132, opened 2022-10-10
Fix for user being logged out

nwmac:
Fixes issue where if call to fetch plugins returns 401, the user is logged out.

Should address (some) issues we're seeing with e2e test failures.


Shavindra (approved):
LGTM

PR #7135, opened 2022-10-10
Removed btn-sm class to make button size nomal

### LINKED ISSUE #7042: Buttons on home page have a different style than other buttons in the UI
The buttons on the home page should have the same height, width, margin, and padding as buttons on other pages.

bisht-richa:
### Summary
Fixes #7042

The buttons on the home page should have the same height, width, margin, and padding as buttons on other pages.
- Removed `btn-sm` class to fix this issue

### DIFF
--- a/shell/pages/home.vue
+++ b/shell/pages/home.vue
@@ -296,13 +296,13 @@ export default {
                   <div class="table-heading">
                     <n-link
                       :to="importLocation"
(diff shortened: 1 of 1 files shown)


cnotv (approved):
Screenshot 2022-10-11 at 12 16 31
LGTM

Issue #7126, opened 2022-10-07
remove windows registration command from k3s custom clusters


slickwarren:
**Setup**
- Rancher version: v2.6.9-rc3
- Browser type & version: chrome (any)

**Describe the bug**

k3s does not support windows clusters. However when provisioning a custom k3s cluster through rancher, after registering the first linux node a windows registration command pops up for the k3s cluster

**To Reproduce**

* provision a k3s custom cluster through rancher
* register 1 linux node to check if the windows registration command appears

**Result**
the windows registration command appears

**Expected Result**

windows is not supported on k3s so it should not appear for custom k3s clusters

slickwarren:
tested on v2.7-head (0dcfd16):
this appears to be fixed on k3s custom clusters (windows registration command no longer appears after registering linux nodes)

PR #7139, opened 2022-10-10
Hide the windows command from k3s cluster

codyrancher:
### Summary
Fixes rancher/dashboard#7126

### Technical notes summary
It looks like we determine the provisioner is k3s in a couple different ways.

Anywhere that requires windows support to be show or used could be affected but it's unlikely.

Anywhere that requires windows support to be show or used could be affected but it's unlikely.

### Screenshot/Video
image


rak-phillip (approved):
LGTM

PR #7140, opened 2022-10-10
Re-initialize default values when new kubernetes version is selected

### LINKED ISSUE #6881: [Bug] Saving unchanged cluster leads to update status and cluster changes
**Setup**
- Rancher version:
2.6.6 and 2.6.7 backend pointing at latest UI for dev.

**Describe the bug**

Saving unchanged cluster leads to update status and cluster changes.

**To Reproduce**

1. Create a cluster.
2. Wait for cluster to go to active.
3. Go to edit the cluster.
4. Switch to yaml (I think this is actually optional).
5. Hit save.

**Result**
Fields are added and the cluster goes into updating.

**Expected Result**

The cluster should remain unchanged and should not go into updating.

**Additional context**
```
<!--Add any other context about the problem here. -->
The diff between cluster before save and after save:
<  generation: 27
---
>  generation: 28
25c25
<  resourceVersion: "114966"
---
>  resourceVersion: "116054"
28c28,29
<  agentImageOverride: ""
---
>  agentEnvVars: []
>  agentImageOverride: null
30,32c31,36
<  description: ""
<  desiredAgentImage: ""
<  desiredAuthImage: ""
---
>  clusterTemplateRevisionName: null
>  defaultClusterRoleForProjectMembers: null
>  defaultPodSecurityPolicyTemplateName: null
>  description: null
>  desiredAgentImage: null
>  desiredAuthImage: null
50c54,57
<    nodelocal: {}
---
>    linearAutoscalerParams: {}
>    nodelocal:
>     updateStrategy: {}
>    updateStrategy: {}
98c105,113
<  scheduledClusterScan: {}
---
>  scheduledClusterScan:
>   enabled: false
>   scanConfig:
>    cisScanConfig:
>     overrideBenchmarkVersion: rke-cis-1.5
>     profile: permissive
>   scheduleConfig:
>    cronSchedule: 0 0 * * *
>    retention: 24
```

mantis-toboggan-md:
### Summary
Fixes #6881 
There are two underlying problems causing extra fields to be added when editing rke2 and k3s clusters. One is that we're using rke2 default values in `machineGlobalConfig` and `agentConfig` when creating k3s clusters, then overriding with the correct k3s defaults on edit. The other problem is that the UI is trying to set default values for a handful of fields that don't appear to exist.

This PR updates the `machineGlobalConfig` and `agentConfig` logic to re-run every time a different kubernetes version is selected, so when a user selects a k3s release, their cluster spec will be populated with correct defaults. 

`hostNamePrefix` and `ignoreErrors` defaults have been removed. afaict `localClusterAuthEndpoint` doesn't need defaults explicitly set by the UI.

### DIFF
--- a/shell/edit/provisioning.cattle.io.cluster/DrainOptions.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/DrainOptions.vue
@@ -10,7 +10,6 @@ const DEFAULTS = {
   force:                           false, // Show; true = Delete standalone pods, false = fail if there are any
   gracePeriod:                     -1, // Show; Pod shut down time, negative value uses pod default
   ignoreDaemonSets:                true, // Hide; true = work, false = never work because there's always daemonSets
-  ignoreErrors:                    false, // Hide; profit?
   skipWaitForDeleteTimeoutSeconds: 0, // Hide; If the pod deletion time is older than this > 0, don't wait, for some reason
   timeout:                         120, // Show; Give up after this many seconds
 };
--- a/shell/edit/provisioning.cattle.io.cluster/rke2.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/rke2.vue
@@ -902,7 +902,9 @@ export default {
 
       // Allow time for addonNames to update... then fetch any missing addons
       this.$nextTick(() => this.initAddons());
-      this.initServerAgentArgs();
+      if (this.mode === _CREATE) {
+        this.initServerAgentArgs();
+      }
     },
 
     showCni(neu) {
--- a/shell/edit/provisioning.cattle.io.cluster/rke2.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/rke2.vue
@@ -1096,10 +1096,6 @@ export default {
           entry.config = await entry.config.save();
         }
 
-        if ( !entry.pool.hostnamePrefix ) {
-          entry.pool.hostnamePrefix = `${ prefix }-`;
-        }
-
         finalPools.push(entry.pool);
       }
 
--- a/shell/edit/provisioning.cattle.io.cluster/rke2.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/rke2.vue
@@ -202,26 +202,6 @@ export default {
       set(this.value.spec, 'kubernetesVersion', this.defaultVersion);
     }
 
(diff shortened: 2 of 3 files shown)


codyrancher (approved):
LGTM

shell/edit/provisioning.cattle.io.cluster/rke2.vue
```diff
@@ -1113,10 +1096,6 @@ export default {
           entry.config = await entry.config.save();
         }
 
-        if ( !entry.pool.hostnamePrefix ) {
```
nwmac, line 1116: Why did this go? Not needed?

shell/edit/provisioning.cattle.io.cluster/DrainOptions.vue
```diff
@@ -10,7 +10,6 @@ const DEFAULTS = {
   force:                           false, // Show; true = Delete standalone pods, false = fail if there are any
   gracePeriod:                     -1, // Show; Pod shut down time, negative value uses pod default
   ignoreDaemonSets:                true, // Hide; true = work, false = never work because there's always daemonSets
-  ignoreErrors:                    false, // Hide; profit?
```
nwmac, line 13: What does removing this do?

nwmac (commented):
Questions in comments

shell/edit/provisioning.cattle.io.cluster/DrainOptions.vue
mantis-toboggan-md: This field is also left out of the API response on creation. Removing this and `hostnamePrefix`'s ensures that when a user edits a cluster the UI doesn't add anything behind-the-scenes, because that causes the cluster to go into updating

shell/edit/provisioning.cattle.io.cluster/rke2.vue
mantis-toboggan-md: No, from what I can tell that field doesn't exist: the API response to the POST request doesn't include it nor is it listed in the relevant schema. I can see it referenced in rke1 provisioning, so I'm guessing that there was just a false assumption it'd be needed in machine pools too.

nwmac (approved):
@mantis-toboggan-md Thanks for the answers - looks good - go ahead and merge

PR #7151, opened 2022-10-11
Update favicon when suse brand is used

### LINKED ISSUE #6493: Update Home Page
Want to clean up home page and simplify the Community and Commercial support links/cards.

For the home page:

- Remove the cards on the right-hand side for commercial support
- Remove the 'getting started' card
- Remove the 'what do you want to see...' card
- Rename 'Community Support' as 'Links'
- All links in this 'Links' box should now be configurable
- Add 'Commercial Support' link to the 'Links' card - always shown at the bottom unless commercial support enabled

New settings page - 'Home Links':

- Remove 'Support Links' from 'Branding' settings page'
- The new 'Home Links' should allow user to configure links - Name and URL for each.
- Include the current links as default values
- Add a new setting that stores the links as a JSON blob
- Ensure the old two 'Support Link' settings are migrated to this new world
- These old two settings would no longer be used other than when upgrading
- Add ability to restore the default links, to support the case that the user removes or edits a link and wants to get the default back
- Getting Started should be added as a default link (since the card for it is being removed)
- Defaults:
  - Docs
  - Forums
  - Slack
  - File an Issue
  - Getting Started

nwmac:
Fixes #6493 

Changes the favicon when brand is suse.

This will be refactored when this is supported by extensions.


Shavindra (changes requested):
There is a Flash of content. It shows the old icon before the new one. Could it be related to nuxt config?

```
      link: [{
        hid:  'icon',
        rel:  'icon',
        type: 'image/x-icon',
        href: `${ resourceBase || '/' }favicon.png`
      }]
```

nwmac:
@Shavindra theres not much we can do. The default one is built in. We replace it once we have loaded the app.

I think that's okay for now.

Shavindra (approved):
LGTM

FoC of favicon that built in acceptable for now

PR #7157, opened 2022-10-11
Fix error response to return message string

### LINKED ISSUE #7138: Non-user friendly message when attempting to delete Cloud Credentials being used
**Setup**
- Rancher version: 2.7-head (Commit ID: 4092002)
- Browser type & version: Brave: v1.44.108 (Chromium: v106.0.5249.103) and Firefox v105.0.3

**Describe the bug**
When attempting to delete a Cloud Credential that is being used by an active cluster, the user is presented with a non-user friendly error message.

**To Reproduce**
1. Create AWS Cloud Credentials in Rancher
2. Create an EC2 Node Driver cluster using the newly created AWS Cloud Credentials
3. When the cluster is Active, attempt to delete the AWS Cloud Credentials

**Result**
Note the user is presented with a non-user friendly error message in the "Are you sure?" dialogue box.

**Expected Result**
A clear and concise error message is presented that explains to the user why they are not able to delete that set of Cloud Credentials.

**Screenshots**
image

jordojordo:
### Summary
Fixes #7138 

This fixes the error message when deleting cloud credentials that are in use to return the `error.message` rather than the entire error object.

### Screenshot/Video

error

### DIFF
--- a/shell/components/PromptRemove.vue
+++ b/shell/components/PromptRemove.vue
@@ -278,7 +278,7 @@ export default {
         await this.refreshSpoofedTypes(spoofedTypes);
         this.done();
       } catch (err) {
-        this.error = err;
+        this.error = err.message ? err.message : err;
         btnCB(false);
       }
     },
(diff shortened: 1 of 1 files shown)


richard-cox (approved):
I sneaked in a tweak (see commit). 

Think we generally do a bad job of handling the multiple resouce use case (only the first error will be shown). For this PR though, and given where we are in the timeline, i think the approach is fine.

codyrancher (approved):

PR #7168, opened 2022-10-12
Remaining shell items for Elemental in 2.7.0

aalves08:
- Adds the remaining shell changes, which are completely isolated from any reference to Elemental, in 2.7.0

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -48,6 +48,7 @@ generic:
   plusMore: "+ {n} more"
   readFromFile: Read from File
   readFromFolder: Read from Folder
+  reload: Reload
   register: Register
   remove: Remove
   resource: |-
@@ -189,7 +190,7 @@ product:
   neuvector: NeuVector
   harvesterManager: Virtualization Management
   rancher: Rancher
-  uiplugins: UI Plugins
+  uiplugins: Extensions
   elemental: OS Management
 
 suffix:
@@ -1650,7 +1651,7 @@ cluster:
     upcloud: UpCloud
     vmwarevsphere: VMware vSphere
     zstack: ZStack
-    machineinventoryselectortemplate: Elemental
+    machineinventoryselectortemplate: Elemental Cluster
   providerTag:
     rke2:
       harvester: '{tag}'
@@ -3675,7 +3676,7 @@ persistentVolumeClaim:
   status:
     label: Status
 
-# UI Plugins
+# Rancher Extensions
 plugins:
   labels:
     builtin: Built-in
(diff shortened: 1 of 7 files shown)


shell/detail/provisioning.cattle.io.cluster.vue
```diff
@@ -631,11 +631,11 @@ export default {
                 </div>
               </div>
               <div v-if="group.ref" class="right mr-45">
-                <template v-if="value.hasLink('update')">
+                <template v-if="value.hasLink('update') && (group.ref.canScaleDownPool() || group.ref.canScaleUpPool())">
```
nwmac, line 634: I don't think both of these canScaleX methods are there on the model - I think you need to add the missing one and have it return true by default to represent the current behvaiour.

shell/assets/translations/en-us.yaml
```diff
@@ -1566,6 +1566,9 @@ cluster:
       label: Roles
     labels:
       label: Kubernetes Node Labels
+    noAccessBanner: "You do not have access to see this machine pool's configuration."
```
nwmac, line 1569: This whole file is out of data with master - it has over-written the latest string with old ones - can you rebase again, or revert back

nwmac (changes requested):

shell/detail/provisioning.cattle.io.cluster.vue
aalves08: Correct. Thanks for pointing it out. Fixed.

shell/assets/translations/en-us.yaml
aalves08: Should be fixed. Thanks for the feedback

shell/components/form/MatchExpressions.vue
```diff
@@ -194,11 +201,12 @@ export default {
         <div v-if="isView">
           {{ row.key }}
         </div>
-        <input
+        <input v-else-if="!keysSelectOptions || !keysSelectOptions.length" v-model="row.key" :mode="mode" @input="update" />
```
richard-cox, line 204: Optimisation, could just do `!keysSelectOptions?.length`

shell/components/GrowlManager.vue
```diff
@@ -114,6 +114,11 @@ export default {
 
     @media only screen and (min-width: map-get($breakpoints, '--viewport-7')) {
       width: 350px;
+
+      // extra padding on the right so that the close X doesn't overlap the text
+      .growl-text {
+        padding-right: 30px !important;
```
richard-cox, line 120: `growl-text` does already contain padding, so that should be updated instead of of an `!important` here. however it looks like the existing padding works fine with long text. was there a specific scenario you hit that it didn't?

image

richard-cox (changes requested):

shell/detail/provisioning.cattle.io.cluster.vue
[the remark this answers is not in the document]
aalves08: updated

shell/components/form/MatchExpressions.vue
aalves08: fixed

shell/components/GrowlManager.vue
aalves08: maybe it was something that was fixed in these months already... going to remove the padding change

shell/edit/provisioning.cattle.io.cluster/MachinePool.vue
```diff
@@ -219,7 +220,8 @@ export default {
       :machine-pools="machinePools"
       @error="e=>errors = e"
     />
-    <Banner v-else color="info" label="You do not have access to see this machine pool's configuration." />
+    <Banner v-else-if="value.configMissing" color="error" label-key="cluster.machinePool.configNotFound" />
```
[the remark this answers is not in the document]
aalves08: This is something which currently only belongs to Elemental and the prop is based on a `canList` request which has always belonged there... 

https://github.com/aalves08/dashboard/blob/elemental-advanced-filtering/shell/edit/provisioning.cattle.io.cluster/rke2.vue#L1007

Can we let this one fly? Just that the fix should be moving this to the machine pool base model then extend the elemental `machineinventoryselectortemplate` model with this `configMissing` and testing all of them to make sure we don't break anything... That's quite a task for now, imo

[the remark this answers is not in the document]
richard-cox: Ok, we might want to tweak that process post 2.7.0 (we'll need to move our elemental specific code from the prov cluster model at some point)

richard-cox (approved):

nwmac (approved):

nwmac (approved):

PR #7172, opened 2022-10-12
begin update for 2.7.0

### LINKED ISSUE #4777: What's New process
Going forward on tickets like https://github.com/rancher/dashboard/issues/4678, once PMs provide us with the content for the What's New section, as part of the review process, we need a way to poke the other engineering managers to validate that the wording is correct.

The two scenarios to consider:
1) the information is accurate
2) if a feature or other documented component is pulled at the last minute, we reflect that information appropriately

gaktive:
Signed-off-by: Gary A Korhonen <gkorhonen@suse.com>

### Summary
Fixes #4777

Updating the What's New section in Dashboard for 2.7.0 (with references to 2.6.8 and 2.6.9)

The UI has a section that details, from a high level, what's new with each version of Rancher since 2.6.0.

- Make sure the copy is legible, rendered logically and accurate

### DIFF
--- a/shell/content/docs/en-us/whats-new.md
+++ b/shell/content/docs/en-us/whats-new.md
@@ -1,7 +1,29 @@
 ---
+title: What's New in 2.7
+---
+
+## New in 2.7.0
+
+### Rancher Extensions
+
+- You can now connect other projects to the Rancher Manager Dashboard. 
+
+---
 title: What's New in 2.6
 ---
 
+## New in 2.6.9
+
+### Bug fixes
+
+- 
+
+
+## New in 2.6.8
+
+### Bug fixes
+
+- 
 
 ## New in 2.6.7
 


shell/content/docs/en-us/whats-new.md
```diff
@@ -1,7 +1,27 @@
+---
+title: What's New in 2.7
+---
+
+## New in 2.7.0
+
+### Rancher Extensions
+
+- You can now connect other apps to the Rancher Manager Dashboard to leverage a single UI. This relies on feature charts that would be manageable via a catalog.
```
deniseschannon, line 9: I'm not sure how exactly to elaborate on this, but this doesn't quite feel right for the product. 

It doesn't give indication that this would allow you to bring in your own UI for your own app to be integrated into the Rancher Manager dashboard.

Also note, that Rancher Manager is still not an external facing term so we should be careful of how and where we use these words.

gaktive: I reworded it with your feedback, though it's agreed that we need someone to refine this.

shell/content/docs/en-us/whats-new.md
```diff
@@ -1,7 +1,26 @@
+---
+title: What's New in 2.7
+---
+
+## New in 2.7.0
+
+### Rancher Extensions
+
+- You can now bring your own apps' UI and integrate it with the Rancher Dashboard to leverage a single UI. This relies on feature charts that would be manageable via a catalog. 
+
```
deniseschannon, line 10: @leflambeur Should we add something in here for support for OCI for helm for fleet?

shell/content/docs/en-us/whats-new.md
```diff
@@ -1,7 +1,26 @@
...
 ---
 title: What's New in 2.6
 ---
 
+## New in 2.6.9
+
+- Alerts within the UI regarding network connectivity are now off by default with a toggle to enable them. This is to address usability uncovered as part of disconnected web sockets.
```
cbron, line 17: The second sentence is a bit confusing to me, I would say more like: 
```suggestion
- Alerts within the UI regarding network connectivity are now off by default with a toggle to enable them. This will address usability concerns when web sockets are disconnected.
```

cbron (commented):
lgtm

shell/content/docs/en-us/whats-new.md
gaktive: @deniseschannon I've grabbed top-level blurbs from a release notes draft, so the 2.7.0 material has more meat. At this point, I've not received much else from other teams but based on what's going into this version, this may be all.

nwmac:
@gaktive I've updated the text for extensions - please take a look. Thanks.

nwmac (approved):

PR #7173, opened 2022-10-12
ui:Job List, fix Duration field sorting

### LINKED ISSUE #6899: Job list Duration field is sorted incorrectly
**Setup**
- Rancher version: v2.6.8
- Browser type & version: chrome 105.0.5195.125

**Describe the bug**

Job list Duration field is sorted incorrectly

**To Reproduce**

1. Navigate to the Job page
2. Click Duration to sort

**Result**
Not sorted by time

**Screenshots**

image

image

**Others**

Same problem with Last Schedule field in CronJobs

torchiaf:
- fix sorting
- code refactoring

### Summary
Fixes #6899


shell/config/product/explorer.js
```diff
@@ -192,7 +192,30 @@ export function init(store) {
...
   headers(WORKLOAD_TYPES.REPLICA_SET, [STATE, NAME_COL, NAMESPACE_COL, WORKLOAD_IMAGES, WORKLOAD_ENDPOINTS, 'Ready', 'Current', 'Desired', POD_RESTARTS, AGE, WORKLOAD_HEALTH_SCALE]);
   headers(WORKLOAD_TYPES.STATEFUL_SET, [STATE, NAME_COL, NAMESPACE_COL, WORKLOAD_IMAGES, WORKLOAD_ENDPOINTS, 'Ready', POD_RESTARTS, AGE, WORKLOAD_HEALTH_SCALE]);
-  headers(WORKLOAD_TYPES.JOB, [STATE, NAME_COL, NAMESPACE_COL, WORKLOAD_IMAGES, WORKLOAD_ENDPOINTS, 'Completions', 'Duration', POD_RESTARTS, AGE, WORKLOAD_HEALTH_SCALE]);
+  headers(WORKLOAD_TYPES.JOB, [
+    STATE,
+    NAME_COL,
+    NAMESPACE_COL,
+    WORKLOAD_IMAGES,
+    WORKLOAD_ENDPOINTS,
+    'Completions',
```
aalves08, line 201: this should be defined on `table-headers` and imported to `explorer.js` to be used as a constant, just to keep the same pattern of usage/definition

shell/config/product/explorer.js
```diff
@@ -192,7 +192,30 @@ export function init(store) {
...
+    {
+      name:      'duration',
+      labelKey:  'tableHeaders.duration',
+      value:     (row) => {
```
aalves08, line 205: It think we should here `getValue` instead of `value` prop. `getValue` generally represents a method while `value` represents a row property

aalves08 (commented):
Left a couple of comments, small stuff imo, but fits better into our current code structure/definition. Otherwise, LGTM. Interesting the creation of this new getter and it's usage. Good work 🙏

shell/config/product/explorer.js
```diff
@@ -192,7 +192,30 @@ export function init(store) {
...
+        const getValue = store.getters['type-map/rowValueGetter'](
+          store.getters['cluster/schemaFor'](WORKLOAD_TYPES.JOB),
```
aalves08, line 207: I am not sure about the full implications, but a user might not have access to a certain schema. We usually use the `canList` getter https://github.com/rancher/dashboard/blob/master/shell/plugins/dashboard-store/getters.js#L199 to check first...
Ex: not the same store, but works the same... https://github.com/rancher/dashboard/blob/master/shell/list/provisioning.cattle.io.cluster.vue#L42

Maybe it's wise to add it here and prepare for a fallback.

aalves08 (commented):
added one final comment

shell/config/product/explorer.js
aalves08: Ignore this comment because of the getter stated here... Moving this to table-headers would imply that we would need state access there and that's not the goal. Thanks 🙏

shell/config/product/explorer.js
aalves08: Let's leave this for now as a user without permissions to access JOBS won't even be allowed to click on the list view link to access it, so therefore this possible scenario might not even manifest itself.

shell/config/product/explorer.js
torchiaf: @aalves08  Just pushed, thanks!

aalves08 (dismissed):
LGTM. Careful not to merge, since we haven't branched for 2.7.1 🙏

torchiaf:
> LGTM. Careful not to merge, since we haven't branched for 2.7.1 pray

Hi @aalves08 . I am sorry but I pushed a new change because I noticed the sorting wasn't actually working making some tests, so please, could you take another quick look? Thanks.

shell/models/workload.js
```diff
@@ -618,4 +619,26 @@ export default class Workload extends WorkloadService {
 
     return matching(allInNamespace, selector);
   }
+
+  get duration() {
```
aalves08, line 623: I believe this should be applied to the model `batch.job` rather than `workload` because `workload` is the parent class to all "workload" types (jobs, deployments, daemonsets, cronjobs,etc) and it looks to me this is only applied to `batch.job`

aalves08 (commented):
Let 1 comment, but overall the feature stopped working for me. You should check what is going on. Thanks 🙏 

Screenshot 2022-10-25 at 09 26 36
Screenshot 2022-10-25 at 09 26 40

shell/models/workload.js
torchiaf: Agree with you, moved to batch.job, thanks!

torchiaf:
> Left 1 comment, but overall the feature stopped working for me. You should check what is going on. Thanks pray
> 
Fixed!
There was a bug in case of Pods inside the Job never start.

In that case, `completionTime` is `undefined`, so this piece of code return always `seconds` = `undefined` and the jobs are sorted by `jobName` (default behavior).

```
  let seconds;
  
  if (value && startTime && completionTime) {
    seconds = getSecondsDiff(startTime, completionTime);
  }
```

It should works after latest fix:

```
  let seconds;

  if (value && startTime) {
    seconds = getSecondsDiff(startTime, completionTime || new Date());
  }
```

duration_sort_1

duration_sort_2

aalves08 (approved):
LGTM

catherineluse:
@torchiaf I believe this can go into master now

aalves08 (approved):
LGTM

PR #7187, opened 2022-10-13
move harvester service add on translations back into shell

mantis-toboggan-md:
This fixes a problem with the harvester-specific section of loadbalancers: https://github.com/harvester/harvester/issues/2929

There is a harvester component still in `shell` that adds some UI when creating loadbalancers in an rke2 cluster provisioned with harvester. There's also a related bit in workload creation port config (which can create services) During the pluginization process, we accidentally moved the translations for this component into the harvester pkg; they need to stay in `shell` where the component is used.

The easiest way to test this is to alter the service edit component's `showHarvesterAddOnConfig` computed prop and WorkloadPort's `showIpam` computed prop to always return `true`.

Screen Shot 2022-10-13 at 10 34 46 AM
Screen Shot 2022-10-13 at 10 35 13 AM

forwardport: https://github.com/rancher/dashboard/pull/7189

### DIFF
--- a/pkg/harvester/l10n/en-us.yaml
+++ b/pkg/harvester/l10n/en-us.yaml
@@ -748,25 +748,6 @@ harvester:
   projectNamespace:
     label: Projects/Namespaces
 
-  service:
-    title: Add-on Config
-    ipam:
-      label: IPAM
-    healthCheckPort:
-      label: Health Check Port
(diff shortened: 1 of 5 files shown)


rak-phillip (approved):
LGTM - this change is pretty straightforward in that it looks to revert a previous change. I was able to verify that text renders properly for the affected components.

PR #7189, opened 2022-10-13
[forwardport] move harvester service add on translations back into shell

mantis-toboggan-md:
2.6.9 issue:  https://github.com/rancher/dashboard/pull/7187 - this is a fix for UI that only appears for harvester-provisioned clusters


rak-phillip (approved):
LGTM

PR #7218, opened 2022-10-18
Fix createYaml data deletion

### LINKED ISSUE #6881: [Bug] Saving unchanged cluster leads to update status and cluster changes

mantis-toboggan-md:
### Summary
Fixes #6881

The remaining issue was specifically caused by using the 'view as yaml' option in the edit form then saving, then editing again without using that option before saving, and vice versa. In other words, clicking 'view as yaml' from the form changed the cluster spec. 

The problem was in `createYaml` and therefore not specific to cluster provisioning. I previously mentioned that I thought the problem was the use of the `cleanUp` function in `createYaml` but it was more subtle. `createYaml` uses schemas' `resourceFields` to generate commented out fields: if the schema for a given type had no properties defined in `resourceFields`, `createYaml` would obliterate any existing data for that type and set the field to `null`.  An example of such a schema is `provisioning.cattle.io.v1.cluster.spec.rkeConfig.machineGlobalConfig`

I also cleaned up some code around comment indentation to fix stuff like this:
Screen Shot 2022-10-18 at 12 20 41 PM

I also realized the change I made to the `ACE` component in my previous PR for #6881 caused a console error so I reverted that. It was the wrong fix: the changes made to `createYaml` are sufficient. 

You should provision a new rke2 cluster to test this. 
Go to the 'edit config' view and save without clicking 'view as yaml', then go back to 'edit config', _do_ click view as yaml, save again. Then go back to 'edit config' and save without clicking 'view as yaml'. It may be worthwhile to also test alternating between saving from 'edit config' and 'edit yaml'.

### DIFF
--- a/shell/edit/provisioning.cattle.io.cluster/ACE.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/ACE.vue
@@ -4,7 +4,6 @@ import { LabeledInput } from '@components/Form/LabeledInput';
 import FileSelector, { createOnSelected } from '@shell/components/form/FileSelector';
 import { set } from '@shell/utils/object';
 import isEmpty from 'lodash/isEmpty';
-import { _CREATE } from '@shell/config/query-params';
 
 export default {
   components: {
--- a/shell/edit/provisioning.cattle.io.cluster/ACE.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/ACE.vue
@@ -24,7 +24,7 @@ export default {
   },
 
   data() {
-    if ( isEmpty(this.value?.spec?.localClusterAuthEndpoint) && this.mode === _CREATE ) {
+    if ( isEmpty(this.value?.spec?.localClusterAuthEndpoint) ) {
       set(this.value, 'spec.localClusterAuthEndpoint', {
         enabled: false,
         caCerts: '',
--- a/shell/utils/create-yaml.js
+++ b/shell/utils/create-yaml.js
@@ -187,11 +187,10 @@ export function createYaml(schemas, type, data, processAlwaysAdd = true, depth =
       out = 'type:';
     }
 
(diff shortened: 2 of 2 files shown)


catherineluse (approved):
Tested as described - works great.

PR #7222, opened 2022-10-19
Update Chinese translation

vickyhella:


vickyhella:
@niusmallnan @kingsd041 would you guys please take a look at the translation update? Thanks!

PR #7229, opened 2022-10-19
Extensions: Fix minor issues

### LINKED ISSUE #7208: [feature testing bug] UI Extensions page- when an extension is 'upgraded' or 'rolled back' to the version that is currently installed the 'installing' process in the UI hangs indefinitely
**Setup**
Rancher HA 
rancher image - v2.7head- 0dcfd16

**Describe the bug**
as an admin, when i navigate to hamburger menu > configuration > extensions page and install a chart with multiple versions, when I select that chart and decide to 'upgrade' or 'roll it back' i'm offered the ability to select the extensions' current version number. if i select that, the UI extensions page hangs in an 'installing' state indefinitely. 

note: a minor UI bug, the UI extension does not crash or become unavailable to users, if the user navigates away from that page and returns everything appears as normal.

**To Reproduce**
install UI operator
install a repro with an example plugin of multiple versions - https://github.com/rancher/ui-plugin-examples
select a extension with multiple versions - homepage for instance.
install the extension with latest version
 you will find the extension in the 'installed' tab
click 'rollback'
select the latest version you just installed.
click update

**Result**

**Expected Result**
the 'installing' grayed out button displays on the extension and hangs there indefinitely. 

**Screenshots**

image

**Additional context**

nwmac:
Fixes #7208 

PR:

- Prevents rollback or upgrade to the version already installed
- Fixes to 'nor'
- Adds more permissions checks for 'Extensions' to show up which should mean users that can get to do actually have permissions to use it

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -3712,7 +3712,7 @@ plugins:
     versions: Versions
     versionError: Could not load version information
   empty:
-    all: Extensions are neither installed or available
+    all: Extensions are neither installed nor available
     available: No Extensions available
     installed: No Extensions installed
     updates: No updates available for installed Extensions
--- a/shell/pages/c/_cluster/uiplugins/InstallDialog.vue
+++ b/shell/pages/c/_cluster/uiplugins/InstallDialog.vue
@@ -24,6 +24,7 @@ export default {
 
   data() {
     return {
+      currentVersion:         '',
       defaultRegistrySetting: null,
       plugin:                 undefined,
       busy:                   false,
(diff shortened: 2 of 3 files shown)


richard-cox (changes requested):
Found a issue probably not related, but just in case... the `version` shown in the card when the chart is installing is the latest version rather than the one being installed. Reproduced with both clock and homepage extensions. 

On a side note homepage took long time to install, so went to check (in another tab) the helm operation which showed as `Active` (separate bug there, we should show a better state text for successfully completed operations), returned to the original tab and it did update to the correct state (installed, version shows `0.2.0` --> `0.3.0`). This was on a system where i'd previously installed v1 and 2.

richard-cox (approved):
Changes will come in second incoming PR

PR #7234, opened 2022-10-19
Fix windows builds by using posix path for all dirs in nuxt.config

rak-phillip:
### Summary
This attempts to produce consistent results when working with `dir` paths on Windows, Linux, and macOS by using `path.posix` for all values specified for `dir`. 

The Nuxt documentation doesn't specify much regarding formatting for `dir` values but I was able to identify an issue with imports repeating (e.g. `require('..\\shell\\store\\shell\\store\\index.js')`) in the generated Nuxt output when `dir` has paths specified in the Windows-style. Always deferring to the posix-style resolves this issue allows for building Dashboard on Windows.

### DIFF
--- a/shell/nuxt.config.js
+++ b/shell/nuxt.config.js
@@ -326,12 +326,12 @@ export default function(dir, _appConfig) {
     ],
 
     dir: {
-      assets:     path.join(SHELL, 'assets'),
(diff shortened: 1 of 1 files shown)


nwmac (approved):
LGTM

catherineluse:
@rak-phillip I believe this can go into master now

Issue #7232, opened 2022-10-19
Missing word on Helm chart install registry tooltip


catherineluse:
This should say "Cluster Container Registry," not "Cluster Container."
<img width="875" alt="Screen Shot 2022-10-19 at 11 51 50 AM" src="https://user-images.githubusercontent.com/20599230/196779193-4c85c612-6d91-403d-bbf2-6f935d48cce1.png">

nickwsuse:
`Verified on v2.7.0-rc8`

The helper text is now correct

![image](https://user-images.githubusercontent.com/107147272/199604436-8a62d9d9-fd81-4605-9574-adce2297b9b0.png)

PR #7237, opened 2022-10-19
Can use RKE1 cluster scoped registry for Helm chart installs

### LINKED ISSUE #7233: [BUG] Downstream RKE1 clusters' fail to install charts
**Rancher Server Setup**
- Rancher version: `v2.7-9f1e0435bb45e104ba897c5c329e1f23735843d9-head`, `v2.7.0-rc4`
- Installation option (Docker install/Helm Chart): `seen with both`
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): `RKE1`
- Proxy/Cert Details: `byo-valid` with helm; insecure w/ Docker

- Kubernetes version: `v1.24.6-rancher1-1`
Downstream RKE1 AWS Node driver

**User Information**
`Admin`

**Describe the bug**
Downstream RKE1 clusters are unable to install charts

**To Reproduce**
1. Fresh install of rancher `v2.7-head` or `v2.7.0-rc4`
2. Provision a downstream RKE1 AWS Node driver cluster w/ k8s `v1.24.6-rancher1-1`
3. Once `Active`, Navigate to `Cluster Explorer > Apps > Charts > [select a chart] > Install`
4. Reproduced

**Result**
Server responded with status code 404; 

**Expected Result**
Chart is installed

**Screenshots**

`Monitoring`
Screen Shot 2022-10-18 at 5 32 40 PM

`Monitoring > Install`
Screen Shot 2022-10-18 at 5 32 59 PM

`CIS Benchmark`
Screen Shot 2022-10-18 at 5 33 13 PM

`CIS Benchmark > Install`
Screen Shot 2022-10-18 at 5 33 25 PM

`Dev Tools: Networking`
Screen Shot 2022-10-18 at 6 04 10 PM

`Helm Install`
Screen Shot 2022-10-18 at 6 05 18 PM

**Additional context**
This was tested via terminal as well, which resulted in the following error message:
```
Error: INSTALLATION FAILED: create: failed to create: the server responded with the status code 413 but did not return more information (post secrets)
```

catherineluse:
Fixes #7233, https://github.com/rancher/dashboard/issues/7232

# QA Template

## Root cause
Last week, we changed the Helm chart install page so that it would check for the existence of a cluster scoped private registry, then add that value into the values.yaml. https://github.com/rancher/dashboard/issues/6910 However, this only worked for K3s/RKE2.

Some additional code still needed to be written for RKE1 because the cluster registry needed to be taken from the cluster management resource for RKE1, as opposed to the provisioning cluster resource for RKE2.
 
- Fixed a bug in the `provClusterId` getter on the management cluster model in which it returned undefined for RKE1 clusters. (Because RKE1 clusters are not created with CAPI resources, they don't have the same relationship to CAPI resources as RKE2/K3s.)
- The Helm chart install page now gets the cluster registry from the management cluster resource if it's an RKE1 cluster.
 
## Areas or cases that should be tested

1. Should create two RKE1 clusters, one that has a cluster-level private registry when the cluster is first created, and one that does not. 
2. Then make sure that when a Rancher Helm chart app is installed, such as alerting drivers or CIS benchmark,
  - For the RKE1 cluster with the cluster scoped registry, images are pulled from that cluster registry
  - For the RKE1 cluster without the cluster scoped registry, see that the images are pulled from the global registry
  - If you select the checkbox for a custom registry on the Helm chart install page and enter a value there, the images should be pulled from that registry
 
## What areas could experience regressions?
I don't expect behavior to change for RKE2/K3s, but if it did change, it would be in the same areas that were affected when the feature was added for RKE2/K3s https://github.com/rancher/dashboard/issues/6910#issuecomment-1266143075

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -768,7 +768,7 @@ catalog:
   chart:
     registry:
       label: Container Registry
-      tooltip: Container images are pulled from the Cluster Container or, failing that, the System Container Registry Setting. To change this default behavior enter or update the registry here
+      tooltip: Container images are pulled from the Cluster Container Registry or, failing that, the System Container Registry Setting. To change this default behavior enter or update the registry here
       custom:
         checkBoxLabel: Container Registry for Rancher System Container Images
         inputLabel: Container Registry
--- a/shell/models/management.cattle.io.cluster.js
+++ b/shell/models/management.cattle.io.cluster.js
@@ -432,7 +432,12 @@ export default class MgmtCluster extends HybridModel {
   }
 
   get provClusterId() {
-    const verb = this.isLocal ? 'to' : 'from';
+    const isRKE1 = !!this.spec?.rancherKubernetesEngineConfig;
+    // Note: RKE1 provisioning cluster IDs are in a different format. For example,
+    // RKE2 cluster IDs include the name - fleet-default/cluster-name - whereas an RKE1
+    // cluster has the less human readable management cluster ID in it: fleet-default/c-khk48
+
+    const verb = this.isLocal || isRKE1 ? 'to' : 'from';
     const from = `${ verb }Type`;
     const id = `${ verb }Id`;
 
--- a/shell/pages/c/_cluster/apps/charts/install.vue
+++ b/shell/pages/c/_cluster/apps/charts/install.vue
@@ -799,6 +799,19 @@ export default {
             return clusterRegistry;
           }
         }
+        if (provCluster.isRke1) {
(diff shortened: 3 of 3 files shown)


richard-cox (approved):
I made a tweak to cover rke1 clusters without a custom registry. With that in the local, rk1 (with & without registry) and rke2 (with and without registry) cluster types all load the install stepper and display the correct registry value

I also snuck in a fix for https://github.com/rancher/dashboard/issues/7232

richard-cox (approved):

PR #7245, opened 2022-10-20
Fix UI not refreshing when cluster members are added or removed.

### LINKED ISSUE #4904: [BUG] UI not refreshing when cluster members are added or removed
Rancher v2.6.3

**To Reproduce**
1. Navigate to Cluster -> Explore -> RBAC -> Cluster Members
2. Remove a member (Context Menu -> Delete)
3. Notice that the resource remains stuck in `Removing` state
4. While keeping this browser session open, login with another user account in a different browser
5. Add a member to the cluster in the second browser session
6. Notice that the new member is not dynamically displayed in the first browser session (browser refresh required)

**Expected Result**
UI should be refreshed when resources are added or removed.

Screenshot 2022-01-12 at 10 41 08

Shavindra:
### Summary
Fixes #4904 - UI not refreshing when cluster members are added or removed 

1. On 2 separate browser windows navigate to Explore Cluster > Cluster > Cluster Members
2. On one of the windows Select `Add` from top right
3. Add/Delete an existing user. 
4. It should update the cluster members on the second screen

### Screenshot/Video

https://user-images.githubusercontent.com/1387263/196994138-f2413408-a4de-4e23-b4ea-9fc013af1503.mov


shell/components/ExplorerMembers.vue
```diff
@@ -70,13 +66,15 @@ export default {
...
     };
   },
 
   computed: {
+    clusterRoleTemplateBindings() {
+      return this.$store.getters[`rancher/all`](NORMAN.CLUSTER_ROLE_TEMPLATE_BINDING).map(b => b.clusterroletemplatebinding) ;
+    },
     ...mapGetters(['currentCluster']),
     filteredClusterRoleTemplateBindings() {
       return this.clusterRoleTemplateBindings.filter(
```
mantis-toboggan-md, line 80: When I add a user the dash breaks in both windows with the error message: `Cannot read properties of undefined (reading 'clusterName')`

Changing to `b?.clusterName` appears to be enough to resolve it then everything works well (deleting members is fine too). If I had to guess, I think there is probably a split second after saving where the dashboard has a norman cluster role template binding, but it hasn't gotten the binding from steve yet.

mantis-toboggan-md, line 80: When I add a user the dash breaks in both windows with the error message: `Cannot read properties of undefined (reading 'clusterName')`

Changing to `b?.clusterName` appears to be enough to resolve it then everything works well (deleting members is fine too). If I had to guess, I think there is probably a split second after saving where the dashboard has a norman cluster role template binding, but it hasn't gotten the binding from steve yet.

mantis-toboggan-md (changes requested):

shell/components/ExplorerMembers.vue
Shavindra: Yes... I thought I pushed that change. It should be OK now

mantis-toboggan-md (approved):
LGTM

PR #7267, opened 2022-10-24
component: CruResource

### LINKED ISSUE #7112: Create Cloud Credentials: Pressing enter navigates to Select Existing instead of submitting form
**Setup**
- Rancher version: 78c7ac0 (latest build from master at time of writing)
- Browser type & version: Firefox 106.0b8

**Describe the bug**

Pressing enter on the Create Cloud Credentials page navigates to Select Existing instead of submitting the form.

**To Reproduce**

1. Navigate to Clusters => Create
2. Toggle RKE2/K3s so that it's on
3. Select Digital Ocean
4. Enter a Credential Name and an Access Token
5. Press Enter to submit the form

**Result**

The page for selecting existing credentials is shown.

**Expected Result**

I expect the cloud credential to be created and to navigate to the cluster creation form. 

**Additional context**

The Continue button is the primary action on this form and I expect this to be action that will be taken when pressing enter. Select Existing is secondary and I would expect that I would need to select this option in order to use it.

torchiaf:
- disable Enter key to submit form

### Summary
Fixes #7112

### Technical notes summary

IMO Submit action through keyboard should already have been disabled by design:

https://github.com/rancher/dashboard/blob/a07cf762249c167ff326bd477f74ec96fca58e8f/shell/components/CruResource.vue#L364

CruResource.vuew component.

### DIFF
--- a/shell/components/CruResource.vue
+++ b/shell/components/CruResource.vue
@@ -362,6 +362,7 @@ export default {
       :is="(isView? 'div' : 'form')"
       class="create-resource-container cru__form"
       @submit.prevent
+      @keydown.enter.prevent
     >
       <div
         v-if="hasErrors"


aalves08:
Hi @torchiaf . Your solution seems to not address the issue as it's described in https://github.com/rancher/dashboard/issues/7112. We can chat a bit about this if you want 🙏

aalves08:
@nwmac or @richard-cox I agree to the solution @torchiaf as come up with despite being different than what was the proposed solution on https://github.com/rancher/dashboard/issues/7112.

Imo, I think that a "submit on enter" is dangerous to have on all our CRU forms where and `enter` might just be a line break or a selection of an input + the fact that I tried other CRU forms and could never submit them with `enter`.

Therefore I think it doesn't make sense to support it in any CRU screens as a submit action, but I just wanted to confirm with you guys before approving this. Thanks in advance! 🙏

aalves08 (approved):
LGTM

PR #7282, opened 2022-10-25
Add retry option for E2E tests

### LINKED ISSUE #7281: Add retry option for E2E tests
# Description

Allow tests to be executed more than once, due presence of flaky tests and other issues which can be tracked by the dashboard.

## Context

Multiple times tests needs to be re-run manually. This would allow us to re-run a test to be sure it's not a flaky scenario.

cnotv:
### Summary
Fixes #7281

Added [`retry`](https://docs.cypress.io/guides/guides/test-retries) option in the Cypress config.

### Technical notes summary
The retry will generate multiple attempts to a total of 3, if any error occurs, which can be inspected separately and will have separated screenshots.

Retrying 3 times with a long waiting time in case of error may eventually have a very length test, in case everything fails.

### DIFF
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -26,7 +26,8 @@ const baseUrl = (process.env.TEST_BASE_URL || 'https://localhost:8005').replace(
 export default defineConfig({
   projectId:              process.env.TEST_PROJECT_ID,
   defaultCommandTimeout:  60000,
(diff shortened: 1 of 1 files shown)


nwmac:
Does not need QA validation

nwmac (approved):

PR #7287, opened 2022-10-25
storage class provisioner dropdown tweaks

### LINKED ISSUE #6091: Usage of Storage Provisioner for k3s confusing
Internal reference: SURE-4008
Reported in 2.6.3
K3s version: 1.21

(Wording gets tricky as there's the initial report and then subsequent investigation that points to UI.)

Issue description:
There is some confusion with respect to the Azure File Share Usage in Rancher 2.6.0 for storage class.

There are two provisioners and the user wanted to know the difference in usage and use case.

- provisioner: kubernetes.io/azure-file
- provisioner: file.csi.azure.com

When using K3S cluster, there is option to select the provisioner.

Business impact:
Without knowing the impact, unable to proceed the further installation. The user is stuck at K3S provisioning on Azure without knowing the impact of using different storage provisioner for K3S. If you could validate the approach and provide the feedback can be the solution here and we don't have much visibility due to the lack of documentation for same.

Two approaches considered:

**Approach 1:**

The Storage class provisioner: kubernetes.io/azure-file which is also in the Rancher UI in any imported cluster.

With the above approach and K3S as the imported cluster, failed to use azure file share. ( as k3s doesn’t have storage driver in it).

And we are not sure which particular driver is needed for this provisioner : kubernetes.io/azure-file

**Approach 2:**

In this approach , installed the Azure File Share CSI driver (as helm chart) in K3S cluster and azure secret from  storage account credentials.

The Storage class in this case has provisioner as: provisioner: file.csi.azure.com and we were able to get the file share created dynamically in the Azure Storage account with a file share.

Installing CSI driver Helm chart: 

https://github.com/kubernetes-sigs/azurefile-csi-driver/tree/master/charts#install-a-specific-version

Creating azure-secret and storage class and then deploying this custom application which successfully created file share:

https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/master/deploy/example/e2e_usage.md#option2-bring-your-own-storage-account-only-for-smb-protocol

----

Backend already looked into this and it was determined it tied into the UI; they looked into this issue to see where the UI gets the information for the CSI driver and to see how Rancher might be able to display this properly. 

From their understanding, this is a UI issue, and needs to be updated here: https://github.com/rancher/dashboard/blob/master/shell/models/storage.k8s.io.storageclass.js as well as a few other places.

Repro steps:

There is no relevant documentation which is giving the hint to approach this scenario. 

Workaround: none

Actual behavior:
No options to select storage provisioner from Rancher UI for K3S imported cluster.

Expected behavior:
The storage provisioner should be selected from Rancher UI

Files, logs, traces:
No Logs

Additional notes:
Kubernetes project in using CSI driver for Kubernetes instead of traditional in tree Volume plugin

mantis-toboggan-md:
### Summary
Fixes #6091

The storage class `provisioner` dropdown contains a hardcoded list of in-tree volume plugins. Some of these plugins have been deprecated in favor of external CSI drivers (which is now the preferred method of writing provisioners). Rather than being built in with kubernetes these drivers are maintained separately and installed via Helm or kubectl. 

This PR adds deprecation warnings for the relevant hard-coded provisioner options, and includes any installed CSI drivers in the dropdown (which also accepts user-defined provisioners)

Screen Shot 2022-10-25 at 3 48 44 PM
Screen Shot 2022-10-25 at 3 48 57 PM


catherineluse:
The changes look reasonable to me, but I couldn't test locally due to a strange bug preventing me from logging into the UI in v2.7-head.

catherineluse (approved):

mantis-toboggan-md:
I expanded the scope of this PR because there was so much overlap with another issue (edited the OP with info). Sorry about the premature review request on #6091 @catherineluse

catherineluse:
The code changes all look good to me, but now that I'm thinking about the labels more, I think they could be a little clearer because it's not always obvious whether the dropdown list item represents an in-tree plugin or a third-party CSI driver. 

For example, take the label `Azure Disk (Deprecated)`. I know that's an in-tree plugin because the Azure Disk CSI driver would only show up if I installed it on my cluster, but I'm not sure everyone would immediately infer that. Basically I'm thinking it could be misinterpreted as meaning that the Azure CSI driver is deprecated when it isn't.

I know this is verbose, but I wonder if it would be worthwhile to always explicitly state which one it is. Like this:

- Amazon EBS Disk in-tree plugin
- Azure Disk in-tree plugin (Deprecated)
- Longhorn CSI driver
- ...etc

And for the ones without labels, I think you can just add the CSI driver for anything that doesn't start with `kubernetes.io`.  So the label for `file.csi.azure.com` would be `file.csi.azure.com CSI driver`.

mantis-toboggan-md:
It is getting to be a lot of text but I think adding 'in-tree plugin' is fine on this view since the dropdown is huge anyway:
Screen Shot 2022-11-22 at 8 45 23 AM

shell/assets/translations/en-us.yaml
```diff
@@ -3449,6 +3452,21 @@ persistentVolume:
     controllerPublishSecretNamespace:
       label: Controller Publish Secret Namespace
       placeholder: e.g. default
+    drivers:
+      disk-csi-azure-com: Azure Disk (CSI)
+      file-csi-azure-com: Azure File (CSI)
+      driver-longhorn-io: Longhorn
```
catherineluse, line 3458: I think there should be just a couple more changes to the labels:

- the `driver.longhorn.io` should say that it's a CSI driver (the screenshot showed it's an in-tree plugin but it's not)
- the Harvester one should say it's a CSI as well
- Amazon EBS Disk in-tree plugin should say it's deprecated

catherineluse (changes requested):

shell/assets/translations/en-us.yaml
mantis-toboggan-md: Done
Screen Shot 2022-12-06 at 8 43 40 AM

catherineluse (approved):
LGTM

PR #7319, opened 2022-11-01
update azuread migration warning text

### LINKED ISSUE #7317: Azure AD - update the modal text for endpoint update
Needed for v2.6.11 and v2.7.0.

We need to change the text that warns users about the need to update their Azure apps to include a set of permissions. We need to change the permission required. It should be `Directory.Read.All` now instead of the current `Group.Read.All` and `User.Read.All`.

mantis-toboggan-md:
Fixes #7317


shell/assets/translations/en-us.yaml
```diff
@@ -487,7 +487,7 @@ authConfig:
         linkText: 'Update it here.'
       modal:
         title: Are you sure? This update is irreversible.
-        body: '<p><b>You may need to make some additional changes</b>. Please ensure the Azure AD app has the following application permissions for Microsoft Graph: User.Read.All, Group.Read.All<br> If any endpoints were customized while configuring Azure AD authentication in Rancher, they will not be automatically updated. </p>'
+        body: '<p><b>You may need to make some additional changes</b>. Please ensure the Azure AD app has the Directory.Read.All application permission added to Microsoft Graph.<br> If any endpoints were customized while configuring Azure AD authentication in Rancher, they will not be automatically updated. </p>'
```
maxsokolovsky, line 490: Could we display the word **Application** capitalized and in bold to draw attention to the need to choose the permission of type Application? This is an Azure concept, and many users don't notice that part, but it's important for the backend.

mantis-toboggan-md: Sure thing, I've updated the warning to:
<p><b>You may need to make some additional changes</b>. Please ensure the Azure AD app has the Directory.Read.All <b>Application</b> permission added to Microsoft Graph.<br> If any endpoints were customized while configuring Azure AD authentication in Rancher, they will not be automatically updated. </p>

maxsokolovsky (approved):

PR #7323, opened 2022-11-03
Removed required from ApiGroup feild

### LINKED ISSUE #7084: Issue with input in Role apiGroups
Internal reference: SURE-5158
Reported in 2.6.5 & 2.6.7

Issue description:
Minor issue -- role creation through UI with double quotes `""` in input section of API Groups textbox resulted in a bad role with invalid API groups `'""'`.

Troubleshooting steps:
After creation of role, edit the the specific role and check the apiGroups section

```
rules:

apiGroups:
  - '""'
  resources:
  - configmaps
  verbs:
  - create
```

Repro steps:

Try to create a role through Rancher >> Cluster Explorer >> RBAC >> Roles >> Create
In create window, enter mandatory fields and put `""` in API Groups textbox for core API groups and click create

Workaround:
Use blank value in API Groups textbox or edit the role and remove the singlequotes `''`

Actual behavior:

- Mandatory Asterix `*` sign on API Groups (Asterix is not there in cluster role creation)
- Double quotes are accepted as literal value and enclosed with a single quote `'""'`

Expected behavior:

- Shouldn't have the mandatory Asterix `*` sign on API Groups, OR
- Accept double quotes `""` as an input value

bisht-richa:
### Summary
Fixes #7084

Removed mandatory Asterix * sign on API Groups

### To test 
Try to create a role through Rancher >> Cluster Explorer >> RBAC >> Roles >> Create
The `API Groups` field should not have a mandatory Asterix * sign

### DIFF
--- a/shell/components/CruResource.vue
+++ b/shell/components/CruResource.vue
@@ -357,7 +357,6 @@ export default {
 
 <template>
   <section class="cru">
-    test
     <slot name="noticeBanner" />
     <form
       :is="(isView? 'div' : 'form')"
--- a/shell/components/CruResource.vue
+++ b/shell/components/CruResource.vue
@@ -357,6 +357,7 @@ export default {
 
 <template>
   <section class="cru">
+    test
     <slot name="noticeBanner" />
     <form
       :is="(isView? 'div' : 'form')"
--- a/shell/components/auth/RoleDetailEdit.vue
+++ b/shell/components/auth/RoleDetailEdit.vue
@@ -447,6 +447,21 @@ export default {
       this.done();
     },
     async actuallySave(url) {
+      // Go through all of the grules and replace double quote apiGroups
+      // k8S documentation shows using empty rules as "" - we change this to empty string when used
+      this.value.rules?.forEach((rule) => {
+        if (rule.apiGroups) {
+          rule.apiGroups = rule.apiGroups.map((group) => {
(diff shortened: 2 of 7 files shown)


shell/components/CruResource.vue
```diff
@@ -357,6 +357,7 @@ export default {
 
 <template>
   <section class="cru">
+    test
```
nwmac, line 360: @bisht-richa I don't think this should be here. You should self-review PRs once created just to check for this kind of thing.

nwmac (changes requested):

nwmac:
@bisht-richa I've updated this PR, removed the test text and added code to replace the double quotes in the API group field

shell/components/auth/RoleDetailEdit.vue
```diff
@@ -447,6 +447,21 @@ export default {
...
+            // If the group is two double quotes ("") replace if with empty string
+            if (group.trim() === '\"\"') {
+              group = '';
+            }
+
+            return group;
+          });
+        }
+      });
+
```
Shavindra, line 464: I think this need to be handled where the rules are actually set. Not before it just save? I think this might be in the model.

Shavindra (changes requested):

Shavindra (approved):
LGTM

PR #7326, opened 2022-11-03
StringList, new component

### LINKED ISSUE #7325: pkg/rancher-components: new component StringList
We need a new component to handle a simple list of strings, in Rancher Desktop preferences dialog; it will be shared across Rancher products.

List of feature:

- add a new item, by click Add button or db-click on empty space.
- select an item, by mouse-click or up/down arrows.
- remove an item, by select an item and click Remove button.
- inline edit an item, by db-click on existing items.
- scrollbar is visible when the box is full.
- up/down arrows makes scrollbar moves when it is visible.
- Add button is disabled when create/edit mode is active.
- Remove button is disabled when no item is selected or create mode is not active.

List of options:

- `caseSensitive` determines if items with same text will be treated differently, depending on the letters case.
    It is used to compare items during create/edit actions and detect duplicates.
- `readonly` determines if the list of items is editable or view-only.
- `placeholder` determines what to show in edit/create input fields.
- `actionsPosition` determines the action buttons position (left, right).
- `errorMessages` defines error messages.

### Screenshot/Video
string-list-light-1

string-list-dark-1

torchiaf:
- add StringList.vue
- add unit tests

### Summary
Fixes #7325

### Technical notes summary

This is not a breaking change, so there are no impacts on other functionalities in the Dashboard UI.

List of feature:

- add a new item, by click Add button or db-click on empty space.
- select an item, by mouse-click or up/down arrows.
- remove an item, by select an item and click Remove button.
- inline edit an item, by db-click on existing items.
- scrollbar is visible when the box is full.
- up/down arrows makes scrollbar moves when it is visible.
- Add button is disabled when create/edit mode is active.
- Remove button is disabled when no item is selected or create mode is not active.

List of options:

- `caseSensitive` determines if items with same text will be treated differently, depending on the letters case.
    It is used to compare items during create/edit actions and detect duplicates.
- `readonly` determines if the list of items is editable or view-only.
- `placeholder` determines what to show in edit/create input fields.
- `actionsPosition` determines the action buttons position (left, right).
- `errorMessages` defines error messages.

Here is a branch which can be used to test the new component:
https://github.com/torchiaf/dashboard/tree/test/string-list
Run it locally and go to `cluster`.

No needs of regression tests

### Screenshot/Video
string-list-light-1

string-list-dark-1

### DIFF
--- a/pkg/rancher-components/src/components/StringList/StringList.test.ts
+++ b/pkg/rancher-components/src/components/StringList/StringList.test.ts
@@ -1,0 +1,72 @@
+import { mount, shallowMount } from '@vue/test-utils';
+import { StringList } from './index';
+
+describe('StringList.vue', () => {
+
+  it('is empty', () => {
+    const wrapper = mount(StringList, { 
+      propsData: { items: [] },
+    });
+    const box = wrapper.find('.string-list-box').element as HTMLElement;
+
+    expect(box.children.length).toBe(0);
+  });
+
+  it('is showing one element', () => {
+    const wrapper = mount(StringList, { 
+      propsData: { items: ['test'] },
+    });
+    const box = wrapper.find('.string-list-box').element as HTMLElement;
+    const actionButtons = wrapper.find('.action-buttons').element as HTMLElement;
+
+    expect(box.children.length).toBe(1);
+    expect(actionButtons).not.toBe(undefined);
+  });
+
+  it('action-buttons are hidden when is view-only mode', () => {
+    const wrapper = mount(StringList, { 
+      propsData: {
+        items: ['test'],
+        readonly: true,
+      },
+    });
+    const actionButtons = wrapper.find('.action-buttons').element as HTMLElement;
+
+    expect(actionButtons).toBe(undefined);
+  });
+
+  it('show new item when "items" property change', async () => {
+    const items = [ 'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'k'];
+
+    const wrapper = mount(StringList, { 
+      propsData: {
+        items,
+      },
+    });
+    const elements = wrapper.findAll('.item');
+    expect(elements.length).toBe(10);
+
+    await wrapper.setProps({ items: [ ...items, 'new' ] });
+
+    const newElements = wrapper.findAll('.item');
+    expect(newElements.length).toBe(11);
+  });
+
+  it('remove item when "items" property change', async () => {
+    const items = [ 'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'k'];
+
+    const wrapper = mount(StringList, { 
+      propsData: {
+        items,
+      },
+    });
+    const elements = wrapper.findAll('.item');
+    expect(elements.length).toBe(10);
+
+    await wrapper.setProps({ items: [ ...items.filter(f => f !== 'a') ] });
+
+    const newElements = wrapper.findAll('.item');
+    expect(newElements.length).toBe(9);
+  });
+
+});
--- a/pkg/rancher-components/src/components/StringList/StringList.test.ts
+++ b/pkg/rancher-components/src/components/StringList/StringList.test.ts
@@ -1,4 +1,4 @@
-import { mount, shallowMount } from '@vue/test-utils';
+import { mount } from '@vue/test-utils';
 import { StringList } from './index';
 
 describe('StringList.vue', () => {
@@ -7,7 +7,7 @@ describe('StringList.vue', () => {
     const wrapper = mount(StringList, { 
       propsData: { items: [] },
     });
-    const box = wrapper.find('.string-list-box').element as HTMLElement;
+    const box = wrapper.find('[data-testid="div-string-list-box"]').element as HTMLElement;
 
     expect(box.children.length).toBe(0);
   });
@@ -17,9 +17,16 @@ describe('StringList.vue', () => {
       propsData: { items: ['test'] },
     });
     const box = wrapper.find('.string-list-box').element as HTMLElement;
-    const actionButtons = wrapper.find('.action-buttons').element as HTMLElement;
 
     expect(box.children.length).toBe(1);
+  });
+
+  it('action-buttons are visible', () => {
+    const wrapper = mount(StringList, { 
+      propsData: { items: ['test'] },
+    });
+    const actionButtons = wrapper.find('[data-testid="div-action-buttons"]').element as HTMLElement;
+
     expect(actionButtons).not.toBe(undefined);
   });
 
@@ -30,7 +37,7 @@ describe('StringList.vue', () => {
         readonly: true,
       },
     });
-    const actionButtons = wrapper.find('.action-buttons').element as HTMLElement;
+    const actionButtons = wrapper.find('[data-testid="div-action-buttons"]').element as HTMLElement;
 
     expect(actionButtons).toBe(undefined);
   });
@@ -43,12 +50,13 @@ describe('StringList.vue', () => {
         items,
       },
     });
-    const elements = wrapper.findAll('.item');
+    const elements = wrapper.findAll('[data-testid^="div-item"]');
+
     expect(elements.length).toBe(10);
 
     await wrapper.setProps({ items: [ ...items, 'new' ] });
 
-    const newElements = wrapper.findAll('.item');
+    const newElements = wrapper.findAll('[data-testid^="div-item"]');
     expect(newElements.length).toBe(11);
   });
 
@@ -60,12 +68,12 @@ describe('StringList.vue', () => {
         items,
       },
     });
-    const elements = wrapper.findAll('.item');
+    const elements = wrapper.findAll('[data-testid^="div-item"]');
     expect(elements.length).toBe(10);
 
     await wrapper.setProps({ items: [ ...items.filter(f => f !== 'a') ] });
 
-    const newElements = wrapper.findAll('.item');
+    const newElements = wrapper.findAll('[data-testid^="div-item"]');
     expect(newElements.length).toBe(9);
   });
 
--- a/pkg/rancher-components/src/components/StringList/StringList.vue
+++ b/pkg/rancher-components/src/components/StringList/StringList.vue
@@ -1,0 +1,573 @@
+<script lang="ts">
+import Vue from 'vue';
+
+type Error = 'duplicate' | 'empty';
+
+type Arrow = 'up' | 'down';
+
+const DIRECTION: Record<Arrow, number> = {
+  up:   -1,
+  down: 1,
+};
+
+const INPUT = {
+  create: 'item-create',
+  edit:   'item-edit',
+};
+
+const BOX = 'box';
+
+const CLASS = {
+  item:  'item',
+  error: 'error',
+};
+
+function findIndex(items: string[], item: string, trim = true) {
+  return items.indexOf(trim ? item?.trim() : item);
+}
+
+function findDuplicates(items: string[], caseSensitive = true) {
+  return items.filter((item, idx) => {
+    const index = findIndex(
+      items.map(i => (caseSensitive ? i : i.toLowerCase()).trim()),
+      (caseSensitive ? item : item.toLowerCase()),
+    );
+
+    return idx !== index;
+  }).length > 0;
+}
+
+/**
+ * Manage a list of strings
+ */
+export default Vue.extend({
+  name:  'string-list',
+  props: {
+    /**
+     * The items source
+     */
+    items: {
+      type: Array,
+      default() {
+        return [];
+      },
+    },
+    /**
+     * The caseSensitive option when create or edit items
+     */
+    caseSensitive: {
+      type:    Boolean,
+      default: false,
+    },
+    /**
+     * Determines if the list of items is editable or view-only
+     */
+    readonly: {
+      type:    Boolean,
+      default: false,
+    },
+    /**
+     * The placeholder to show in input field when create or edit items
+     */
+    placeholder: {
+      type:    String,
+      default: null,
+    },
+    /**
+     * Determines the action buttons position at the bottom of the main box
+     */
+    actionsPosition: {
+      type:    String,
+      default: 'right',
+    },
+    /**
+     * Custom Error messages
+     */
+    messages: {
+      type: Object,
+      default() {
+        return { duplicate: 'Error, item is duplicate!' };
+      },
+    },
+  },
+  data() {
+    return {
+      value:        null as string | null,
+      selected:     null as string | null,
+      isEditItem:   null as string | null,
+      isCreateItem: false,
+      errors:       { duplicate: false } as Record<Error, boolean>,
+    };
+  },
+
+  computed: {
+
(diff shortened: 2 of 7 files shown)


pkg/rancher-components/src/components/StringList/StringList.vue
```diff
@@ -0,0 +1,573 @@
...
+    setFocus(refId: string) {
+      setTimeout(() => {
+        this.getRef(refId)?.focus();
+      }, 0);
+    },
+
+    /**
+     * Move scrollbar when the selected item is over the top or bottom side of the box
+     */
+    moveScrollbar(arrow: Arrow, value = 25) {
```
cnotv, line 217: Why is value default 25? What is aimed to?

pkg/rancher-components/src/components/StringList/StringList.vue
```diff
@@ -0,0 +1,573 @@
...
+</template>
+
+<style lang="scss" scoped>
+
+.string-list {
+  display: flex;
+  flex-direction: column;
+  width: auto;
+
+  input {
```
cnotv, line 470: Please avoid use of HTML tags for styling.

pkg/rancher-components/src/components/StringList/StringList.test.ts
```diff
@@ -0,0 +1,72 @@
+import { mount, shallowMount } from '@vue/test-utils';
+import { StringList } from './index';
+
+describe('StringList.vue', () => {
```
cnotv, line 4: I would expect some tests related to the UI behaviors, rather than props binding only.
Kudos in any case for adding some tests in a first place, really! 👏

cnotv (changes requested):
Great job there, also thanks for adding unit tests! 🥇 

Please remember to create an issue for adding the component to the style guide (Storybook) so we don't lose track of it.

Tiny remarks about the styling, when selected we invert the colors just as for the CTA buttons.
Screenshot 2022-11-03 at 13 23 19

Please add some information about how to add further validation for that input, as it may be required. You may coordinate with your team or the scope of the component for current use and eventually postpone this or write an issue for it or make it clear of the need to be extended afterwards. You may coordinate with Sean for that.

Remove button (-) should be set disabled if nothing is selected.

Remember to use i18n for all the strings, e.g. `Error, item is duplicate!`

Finally I would recommend the use of existing `LabeledInput` as input components, to keep consistency along the library for both visual and functionalities.

PS: The component can be seen in the cluster dashboard node section, not node resource.

pkg/rancher-components/src/components/StringList/StringList.vue
```diff
@@ -0,0 +1,573 @@
...
+      }
+      if (this.selected) {
+        const index = this.deleteItem(this.selected);
+
+        if (index !== -1) {
+          /**
+           * Select the next item in the list when an item is deleted.
+           * If the deleted item was the last item, select the previous item (this.items[index - 1]). 
+           */
+          setTimeout(() => {
```
cnotv, line 198: Oh I missed this timeout. May ask you why? Usually is a sign of issues with some reactivity. Plus you should always unset as garbage collection logic.

pkg/rancher-components/src/components/StringList/StringList.vue
```diff
@@ -0,0 +1,573 @@
...
+     * The items source
+     */
+    items: {
+      type: Array,
+      default() {
+        return [];
+      },
+    },
+    /**
+     * The caseSensitive option when create or edit items
```
rak-phillip, line 56: It might be a good idea to mention what `caseSensitive` prop controls. It looks like it's used for detecting duplicate items.

rak-phillip (changes requested):
I think this has come together pretty well. I tried to leave some comments on some additional points that you could take into account.

pkg/rancher-components/src/components/StringList/StringList.vue
torchiaf: When an item is removed from the list (`this.items`), the new list is emitted by `change` event, then the list is refreshed by the parent of `StringList`and this could take some times. The selection of the next item in the list cannot happen until the new list is pushed back to the component.

pkg/rancher-components/src/components/StringList/StringList.vue
torchiaf: 25 is the height of each div in the list box, in px:

```
    .static {
      height: 25px;
      padding: 3px;
    }
```

pkg/rancher-components/src/components/StringList/StringList.vue
```diff
@@ -0,0 +1,573 @@
...
+};
+
+const BOX = 'box';
+
+const CLASS = {
+  item:  'item',
+  error: 'error',
+};
+
+function findIndex(items: string[], item: string, trim = true) {
```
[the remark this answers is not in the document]
torchiaf: I have renamed those functions in `findStringDuplicates`, `findStringIndex` and moved in `array.ts`.

pkg/rancher-components/src/components/StringList/StringList.vue
```diff
@@ -0,0 +1,573 @@
...
+
+function findDuplicates(items: string[], caseSensitive = true) {
+  return items.filter((item, idx) => {
+    const index = findIndex(
+      items.map(i => (caseSensitive ? i : i.toLowerCase()).trim()),
+      (caseSensitive ? item : item.toLowerCase()),
+    );
+
+    return idx !== index;
+  }).length > 0;
```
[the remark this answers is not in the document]
torchiaf: I have fixed the nested loop and changed it to return immediately after finding a duplicate :+1:

torchiaf:
Thanks for the review guys:

@cnotv I improved the code following your suggestions, just a quick recap about the points which are still open:

- I will open a separated issue to add the `StringList` to the Storybook.
- I would like to open a separated issue to add more unit tests. In this way we can go ahead in parallel with that tasks which are requiring the StringList component on Rancher Desktop, once it is merged and a new version of the `pkg/rancher-component` lib is created, and then implement additional tests.
- Regarding the use of `LabeledInput` , are we sure we need it in this context? My thoughts about it is that probably we don't really need all the features implemented in it, maybe `LabeledInput` better fits in case of implementing forms, but here we need only a simple input field. Furthermore, it would be tricky to adapt the component in some sections, like here  for example:

https://github.com/rancher/dashboard/blob/1da4e77545d125fc33a61e9c355ee9914bbb786a/pkg/rancher-components/src/components/StringList/StringList.vue#L407-L409

where the v-model (if I am not wrong) cannot be used.

What do you think about?

@rak-phillip I made all the changes you suggested.

pkg/rancher-components/src/components/StringList/StringList.vue
cnotv: Should we try to get the ref height instead so we do not have to hard code it??

shell/utils/array.ts
```diff
@@ -155,6 +155,27 @@ export function findBy<T, K, V>(
   return findOrFilterBy('find', ary, keyOrObj, val);
 }
 
+export function findStringIndex(items: string[], item: string, trim = true): number {
+  return items.indexOf(trim ? item?.trim() : item);
+}
+
+export function findStringDuplicates(items: string[], caseSensitive = true): boolean {
```
cnotv, line 162: I am very sorry to be this annoying, but since it's returning boolean and not a string, I would rather use `hasDuplicatedStrings` as name.

cnotv (changes requested):

pkg/rancher-components/src/components/StringList/StringList.vue
```diff
@@ -0,0 +1,572 @@
...
+        <button
+          class="btn btn-sm role-tertiary add-button"
+          :disabled="!selected && !isCreateItem && !isEditItem"
+          @mousedown.prevent="onClickMinusButton"
+        >
+          <span class="icon icon-minus icon-sm" />
+        </button>
+        <button
+          class="btn btn-sm role-tertiary remove-button"
+          :disabled="isCreateItem"
```
cnotv, line 446: Remove button (-) should be set disabled if nothing is selected.

pkg/rancher-components/src/components/StringList/StringList.vue
```diff
@@ -0,0 +1,572 @@
...
+      height: 25px;
+      padding: 3px;
+    }
+
+    .item {
+      outline: none;
+
+      &.selected {
+        background: var(--primary);
+        color: var(--primary-hover-text);
```
cnotv, line 498: Tiny remarks about the styling, when selected we invert the colors just as for the CTA buttons.

Screenshot 2022-11-03 at 13 23 19

pkg/rancher-components/src/components/StringList/StringList.vue
```diff
@@ -0,0 +1,572 @@
...
+          @input.prevent="onChange($event.target.value)"
+          @blur.prevent="toggleEditMode(false)"
+          @keydown.enter.prevent="updateItem(item)"
+        />
+      </div>
+      <div
+        v-if="isCreateItem"
+        class="create-field static"
+      >
+        <input
```
cnotv, line 418: Finally I would recommend the use of existing `LabeledInput` as input components, to keep consistency along the library for both visual and functionalities.

pkg/rancher-components/src/components/StringList/StringList.vue
```diff
@@ -0,0 +1,572 @@
...
+        @keydown.up.prevent="onSelectNext('up')"
+        @blur="onSelectLeave(item)"
+      >
+        <span
+          v-if="!isEditItem || isEditItem !== item"
+          class="label static"
+        >
+          {{ item }}
+        </span>
+        <input
```
cnotv, line 404: I just realized that we could even edit the string 😬 
Has it been requested like this? Maybe we should give some hints? I would not make it a blocker issue though, just giving a feedback.

cnotv (changes requested):

pkg/rancher-components/src/components/StringList/StringList.vue
```diff
@@ -0,0 +1,572 @@
...
+          readonly
+        }"
+        :data-testid="`div-item-${item}`"
+        class="item static"
+        tabindex="0"
+        @mousedown="onSelect(item)"
+        @dblclick.stop="toggleEditMode(true, item)"
+        @keydown.down.prevent="onSelectNext('down')"
+        @keydown.up.prevent="onSelectNext('up')"
+        @blur="onSelectLeave(item)"
```
cnotv, line 396: I just realized that the input closes in different ways between creation and modification. It may be worth to keep it consistent and allow to close it in the same way.

pkg/rancher-components/src/components/StringList/StringList.vue
cnotv: `$nextTick` is definitely better and a valid option.
I would however recommend you 2 different solutions for the 2 cases.

For input focus in `setFocus()`, you should rely on autofocus while toggling it with the condition. In the case of the `LabelInput`, it has a prop for that.

For the string selection happening in `onClickMinusButton()`, you could find and assign the new string into `selected` before waiting the DOM gets updated by deletion. Since the 2 operations are not bound with the Node but rather from an UX perspective, we have no constrains. 
You should be able to compute the new index based on the current one. If you are using strings over index, as you may want to have the `selected` value, you would have to map it back and forth to an index.

cnotv:
> * Regarding the use of `LabeledInput` , are we sure we need it in this context? My thoughts about it is that probably we don't really need all the features implemented in it, maybe `LabeledInput` better fits in case of implementing forms, but here we need only a simple input field. Furthermore, it would be tricky to adapt the component in some sections, like here  for example:
> 
> https://github.com/rancher/dashboard/blob/1da4e77545d125fc33a61e9c355ee9914bbb786a/pkg/rancher-components/src/components/StringList/StringList.vue#L407-L409
> 
> where the v-model (if I am not wrong) cannot be used.
> 
> What do you think about?
> 
> @rak-phillip I made all the changes you suggested.

I would ask @kwwii or who gave you the requirement for the design at this point, because personally I would never rewrite an input twice.

pkg/rancher-components/src/components/StringList/StringList.vue
torchiaf: I have onfused remove with add button :face_with_spiral_eyes:

pkg/rancher-components/src/components/StringList/StringList.vue
torchiaf: I'll try to introduce the `LabeledInput`

kwwii:
I think that in general it would be better to use LabeledInput unless there is a strong technical or design reason against it. For example, if we have to up over-riding lots of things to make it somehow fit, we should NOT use LabeledInput.

torchiaf:
@cnotv thanks, I fixed the last issues:

- scroll value is not hardcoded anymore.
- `findStringDuplicates` changed to `hasDuplicatedStrings`.
- `setTimeout` changed into `$nextTick`, and eliminated that on item selection.
- `errorMessages` are demanded to the parent to be defined.

The open point is the use of `LabeledInput` . I'll give it a try to introduce it in the component and evaluate if it is convenient to use it or leave the simple `input`.

cnotv (approved):
LGTM

pkg/rancher-components/src/components/StringList/StringList.vue
```diff
@@ -0,0 +1,573 @@
...
+
+        this.isEditItem = null;
+      }
+    },
+
+    getElemByRef(id: string) {
+      const ref = this.$refs[id];
+
+      return (Array.isArray(ref) ? ref[0] : ref) as any;
+    },
```
rak-phillip, line 290: I'm noticing two different usages of `getElemByRef`:

1. Line 238: `const input = this.getElemByRef(refId)?.$el`
2. Lines 209-210:
   ``` 
      const box = this.getElemByRef(BOX);
      const item = this.getElemByRef(this.selected || '');
   ```

I think we can improve the type of `getElemByRef` can be improved by introducing a new function (or changing the return type of the existing function by always returning `this.$refs[id].$el`). We could probably attempt something like this to make sure that we are working with a stronger type than `any` 

```suggestion
    getElemByRef(id: string): Vue {
      const ref = this.$refs[id];

      return Array.isArray(ref) ? (ref[0] as Vue) : ref as Vue;
    },

    getHtmlDivElement(id: string) {
      return this.getElemByRef(id).$el as HTMLDivElement;
    },
```

rak-phillip (changes requested):
I left one suggestion for how we might be able to avoid the usage of `any` for our type. As it stands, it looks like `getElemByRef` can return either a Vue component for an HTML element. I think we should be more explicit in which scenario we expect so we can reduce the chance for introducing an error when making changes to this component in the future.

rak-phillip (approved):
@torchiaf and I met out of band. There are some issues that we will need to resolve with `LabeledInput` so we can improve the behavior of focus. 

The changes here look good to me.

pkg/rancher-components/src/components/StringList/StringList.vue
cnotv: I missed that, good point

PR #7337, opened 2022-11-04
when using pv to make pvc set storageClassName if present on pv

### LINKED ISSUE #7332: error "ClassName does not match" when creating PVC for existing PV using the Rancher 2.1.8 UI (and 2.6.9)
bug

- create a persistent storage class
- create a PV using the storage class
- create a PVC from a namespace using the new UI option  "Use an existing persistent volume"
- select the previously created PV and create the PVC

**Result:**
The PVC is stuck in status pending.
The problem is rancher when creating the PVC don't set the same storage class name for the PVC as the PV. The storage class name field in the PVC spec is an empty string instead of the correct storage class name. As a result it is impossible to create a working PVC using the UI, also it is impossible to modify the storage class name of an existing PVC the only way is to delete the PVC and create it manually using kubectl.

mantis-toboggan-md:
### Summary
Fixes #7332 

When a PV is assigned to a storage class (has `spec.storageClassName` set) PVC's created from that PV also need `spec.storageClassName` set.

A note about the repro step [here](https://github.com/rancher/dashboard/issues/7332#issuecomment-1302586528) : when creating your PV you need to go to the customize tab and choose 'longhorn' in the 'assign to storage class' dropdown, not just select 'longhorn' as the volume plugin (I made this mistake initially)

### DIFF
--- a/shell/edit/persistentvolumeclaim.vue
+++ b/shell/edit/persistentvolumeclaim.vue
@@ -145,12 +145,16 @@ export default {
       set(value) {
         const persistentVolume = this.persistentVolumes.find(pv => pv.metadata.name === value);
 
+        this.$set(this.value.spec, 'storageClassName', '');
+
         if (persistentVolume) {
           this.$set(this.value.spec.resources.requests, 'storage', persistentVolume.spec.capacity?.storage);
+          if (persistentVolume.spec?.storageClassName) {
+            this.$set(this.value.spec, 'storageClassName', persistentVolume.spec?.storageClassName );
+          }
         }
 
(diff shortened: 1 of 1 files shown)


catherineluse (approved):
LGTM

1. I installed Longhorn
2. Created a Longhorn StorageClass called `longhorn-class`
3. Went to create a PersistentVolume. Selected the Longhorn volume plugin. On the Customize tab, I selected Assign To Storage Class > `longhorn-class`
4. Went to create a PersistentVolumeClaim. Selected "Use an existing PersistentVolume." Selected the new PV.
5. Finished creating the PVC and confirmed that it was successfully bound.

Issue #7369, opened 2019-04-12
error "ClassName does not match" when creating PVC for existing PV using the Rancher 2.1.8 UI (and 2.6.9)


3pns:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
- create a persistent storage class
- create a PV using the storage class
- create a PVC from a namespace using the new UI option  "Use an existing persistent volume"
- select the previously created PV and create the PVC

**Result:**
The PVC is stuck in status pending.
The problem is rancher when creating the PVC don't set the same storage class name for the PVC as the PV. The storage class name field in the PVC spec is an empty string instead of the correct storage class name. As a result it is impossible to create a working PVC using the UI, also it is impossible to modify the storage class name of an existing PVC the only way is to delete the PVC and create it manually using kubectl.

mantis-toboggan-md:
You're completely right @MbolotSuse - I missed this when I was testing with the local cluster, but we're checking the wrong store for storage class and pv schemas. I've transferred this issue back to dashboard, sorry to rope you in on this, that's my bad!

gaktive:
Internal reference: SURE-4826

For QA, please test to make sure that the appropriate user that does have permissions to list SC's, just not PVs, doesn't show "You don't have permission to list Storage Classes, enter a name manually" in 2.7.0. This popped up in 2.6.9 again, as noted, so hopefully this would show a list of SCs the user can select from, rather than having to enter it manually.

slickwarren:
tested on rancher v2.7-head (4f6a03e):
* standard user, cluster owner can create the following:
  * manually create PV -- pass
  * manually create PVC from existing PV -- pass (original use case)
  * manually create PVC with new PV -- pass
  * create a PV/PVC from a new workload -- pass
  * check that user can list SC and PV -- pass

PR #7370, opened 2022-11-08
[2.7.0] fix pvc creation in downstream clusters for standard users

mantis-toboggan-md:
### Summary
Fixes [#7369 ](https://github.com/rancher/dashboard/issues/7369#issuecomment-1307658187)

edit: this also addresses a lingering issue in https://github.com/rancher/dashboard/issues/6444

We're checking the availability of storage classes and pvs using the management store: this seems to work fine for the local cluster, but standard user cluster owners encounter this bug on downstream clusters.

forwardport: https://github.com/rancher/dashboard/pull/7371


catherineluse (approved):
LGTM. Confirmed that as a standard user, I could create a PersistentVolumeClaim from existing PersistentVolumes and StorageClasses.

PR #7373, opened 2022-11-08
Add pr to docs workflow trigger

### LINKED ISSUE #7314: Create GH action for PRs that checks the devkit doc build
Docs are only built on merge to master, so if you break it, you don't know until that fails.

Add a PR check to check that a PR has not broken the docs.

jordojordo:
### Summary
Fixes #7314 

This adds a step in the Docusaurus workflow to run anytime a PR is opened.

### DIFF
--- a/.github/workflows/docusaurus.yaml
+++ b/.github/workflows/docusaurus.yaml
@@ -4,6 +4,9 @@ on:
   push:
     branches:
       - master
+  pull_request:
+    types:
+      - opened
 
 jobs:
   deploy:
--- a/.github/workflows/docusaurus.yaml
+++ b/.github/workflows/docusaurus.yaml
@@ -4,9 +4,11 @@ on:
   push:
     branches:
       - master
+      - 'release-*'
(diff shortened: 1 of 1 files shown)


.github/workflows/docusaurus.yaml
```diff
@@ -4,6 +4,9 @@ on:
   push:
     branches:
       - master
+  pull_request:
+    types:
+      - opened
```
nwmac, line 9: Think we also need `edited` and `reopened` ?

or remove `types` and set `branches` as per the `test` workflow, e.g.
```
  pull_request:
    branches:
      - master
      - 'release-*'
```

nwmac (changes requested):

.github/workflows/docusaurus.yaml
jordojordo: I added the same cases for `push` and `pull_request` as they are in the `test` workflow.

nwmac (approved):

PR #7378, opened 2022-11-08
account for no currentCluster on chart install

### LINKED ISSUE #7372: unable to create cluster from rke2 template as standard user
**Setup**
- Rancher version: v2.7.0-rc9

**Describe the bug**

as a standard user, when creating a cluster from an rke2 template the user is presented with a blank page. 

**To Reproduce**

* (as admin) add an rke2 cluster template repository (i.e. the [example we have in gh](https://github.com/rancher/cluster-template-examples.git))
* (as standard user) go to cluster management -> create -> select a rke2 template 

**Result**
user is presented with a blank page 

**Expected Result**

user should be able to configure the template and create from it as a standard user

**Screenshots**

Screen Shot 2022-11-08 at 11 23 54 AM

**Additional context**

an admin user can correctly configure the template and does not hit this issue
this is a regression, as it is not an issue on 2.6.9

mantis-toboggan-md:
### Summary
Fixes #7372 
The blank page reported in the issue appears to be caused by this page's fetch throwing an error. I've updated references to `currentCluster` to account for the possibility that it is undefined. This seems to be enough to resolve the problem: I was able to deploy an rke2 cluster using the template repo linked in the issue.

### DIFF
--- a/shell/pages/c/_cluster/apps/charts/install.vue
+++ b/shell/pages/c/_cluster/apps/charts/install.vue
@@ -434,7 +434,7 @@ export default {
       const cluster = this.currentCluster;
       const projects = this.$store.getters['management/all'](MANAGEMENT.PROJECT);
 
-      const out = projects.filter(x => x.spec.clusterName === cluster.id).map((project) => {
+      const out = projects.filter(x => x.spec.clusterName === cluster?.id).map((project) => {
         return {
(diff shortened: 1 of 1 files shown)


catherineluse (approved):
LGTM - successfully provisioned RKE2 cluster from a cluster template as a standard user.

Issue #6022, opened 2022-05-24
Rollback revision list shows relative times only


hameno:
**Setup**
- Rancher version: 2.6.5
- Browser type & version: Chrome 101.0.4951.67

**Describe the bug**
The revision list only shows relative times, there is no way to see the actual date.
See
![image](https://user-images.githubusercontent.com/365084/169959482-2710bc2f-f5cb-4bbd-bc2a-76fa07bf77bf.png)
This is not very helpful in finding the correct time

rak-phillip:
@gaktive @kwwii I asked that we get an issue together for this so we can take a closer look. I think we have an opportunity to improve the UX a little bit here, what do you think?

kwwii:
@rak-phillip Agreed, perhaps something like this would be better:

Rev. 179 - 24.05.2022 9:45am
Rev. 180 - 24.05.2022 7:00am
Rev. 181 - 17.05.2022 9:45am
Rev. 182 - 15.05.2022 7:00am
Rev. 183 - 17.04.2022 9:45am
Rev. 184 - 15.04.2022 7:00am

rak-phillip:
I think it looks good. My only feedback is that we should make sure we're consistent with how we format date/time in other areas of dashboard. I took a peek to find other examples of formatting dates and we definitely prefer the "created `n` days ago" approach. Perhaps, this isn't too much a of a concern for us..

I'm not too sure on the original motivation for formatting dates this way, but I find it to be a little informal for something like this.

catherineluse:
We should probably use the user's preferred date format that is configured in the global settings.

catherineluse:
@gaktive  @kwwii Can we prioritize this one? There are a lot of people asking for a fix on the backend ticket. https://github.com/rancher/rancher/issues/35138

markusewalker:
**QA TEST PLAN**

**Scenarios**
|Scenario|Test Case|
|---|---|
|1|As a standard user, provision a downstream RKE2 cluster. Create a test deployment, edit the image pull policy, validate that the rollback shows the timestamp in the format of user preferences|
|2|Repeat scenario 1 with a new deployment. Afterwards, modify the user preferences time format and validate that the rollback shows the updated format|

markusewalker:
Validated this against `v2.7-head` https://github.com/rancher/rancher/commit/edf59f620ceec6b3f22507b4d1ac780d392568df.

**ENVIRONMENT DETAILS**
- Rancher install: Helm
[...]

PR #7383, opened 2022-11-09
Rollback options are timestamped in format from user preferences

catherineluse:
Fixes https://github.com/rancher/dashboard/issues/6022

This PR adds the exact time of the revision creation to the dropdown menu for rolling back a workload.

To test this PR,

1. I created a Deployment
2. Went to Edit Config on the Deployment, changed the image pull policy, saved, and changed it back to create new revisions of the workload
3. From the Deployment list view, I went to the new Deployment and clicked **⋮ > Rollback**
4. Confirmed that the timestamp of each workload was in the format in the user preferences

Screenshot 2022-11-08 at 6 47 22 PM

6. Went to the user preferences
7. Changed the time format
8. Went back to rollback the workload again
9. Confirmed that the timestamps were updated to the new format in the user preferences 
Screenshot 2022-11-08 at 6 48 04 PM


aalves08 (approved):
Test done and working. Code looks good. Approved 🙏

Issue #7401, opened 2022-11-10
[unit-test] ApplicationCard component


scures:
Issues for tracking the unit-test of `ApplicationCard` component.

PR #7402, opened 2022-11-10
Application card test

scures:
### Summary
Fixes rancher/dashboard#7401

Introduces some unit-test for the component.

### DIFF
--- a/shell/components/cards/ApplicationCard.vue
+++ b/shell/components/cards/ApplicationCard.vue
@@ -4,7 +4,10 @@ export default {};
 
 <template>
   <div class="application-card mt-20">
-    <div class="application-card-icon" data-testid="cardIcon-section">
+    <div
+      class="application-card-icon"
+      data-testid="cardIcon-section"
+    >
       <slot name="cardIcon" />
     </div>
     <div class="application-card-details">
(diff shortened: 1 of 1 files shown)


shell/components/__tests__/ApplicationCard.test.ts
```diff
@@ -0,0 +1,51 @@
...
+  it('should have a populated card-icon slot', () => {
+    expect(wrapper.find('[data-testid="icon-slot-section"]').exists()).toBe(true);
+    expect(wrapper.find('[data-testid="icon-slot-section"]').find('i').exists()).toBe(true);
+    expect(wrapper.find('[data-testid="icon-slot-section"]').find('i').classes()).toContain('icon-epinio');
+  });
+
+  it('should have a populated top-left slot', () => {
+    const ele = wrapper.find('[data-testid="top-left-section"]');
+
+    expect(ele.find('ul').exists()).toBe(true);
```
cnotv, line 23: As the content has no constrain, the test is only testing that the markup you are writing is correct, which does not bring any use. It could be even just text or a simpler HTML element.

The idea is to check that the content is present and is in the correct section, regardless of what is passed. For this reason you could parameterize as defined above.

cnotv (approved):
LGTM

PR #7408, opened 2022-11-10
StringList - Commit changes when losing focus

### LINKED ISSUE #7392: StringList, avoid item creation in readonly mode
**Setup**
- Rancher version: any
- Browser type & version: any

**Describe the bug**

It is possible to double click on the box and create an item when `readonly` mode is active

**To Reproduce**

- set `readonly` property.
- double click on the empty space of the list box.

**Result**
The input field is open and a new item can be added to the list.

**Expected Result**

It is not possible to create items.

torchiaf:
### Summary
Fixes [#3369](https://github.com/rancher-sandbox/rancher-desktop/issues/3369)

Fixes #7392
Fixes https://github.com/rancher-sandbox/rancher-desktop/issues/3373 (duplicate)

### Technical notes summary

The list of items editing has changed (both create and edit mode):

- Items are saved when losing focus or when pressing Enter key.
- when there is a duplicate and focus is lost, input field is closed and the item is not saved.
- when there is a duplicate and Enter key is pressed, nothing happens.
- a new event `errors` is emitted when there is an error. We have only a `duplicate` error for the moment.
- duplicate error messages are displayed during user's typing.

StringList

### DIFF
--- a/pkg/rancher-components/src/components/StringList/StringList.vue
+++ b/pkg/rancher-components/src/components/StringList/StringList.vue
@@ -129,7 +129,7 @@ export default Vue.extend({
     },
 
     onSelect(item: string) {
-      if (this.isCreateItem || this.editedItem === item) {
+      if (this.readonly || this.isCreateItem || this.editedItem === item) {
         return;
       }
       this.selected = item;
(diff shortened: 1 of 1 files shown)


rak-phillip (approved):
LGTM - tested in Rancher Desktop and this change works as advertised. I initially considered if we want the watcher for `errors` but I think it does the job of preventing the emit until there's an actual change quite succinctly.

PR #7426, opened 2022-11-12
Add ArrayList to Storybook and add unit tests for it

catherineluse:
- Adds five unit tests
- Adds ArrayList to Storybook:
Screenshot 2022-11-16 at 2 10 26 AM


cnotv:
The import is wrong as `export { default as ArrayList } from './ArrayList.vue';` is pointing to something which does not exists 😅 

Some extra notes:
- `ArrayListRow` is not exported
- Types should have separated files for each component, except global imports
- You got a 4 spaces indentation 🤔 
- There's no need to prefix the data with `div-` or is this some kind of convention?

stories/components/form/ArrayList.stories.mdx
```diff
@@ -0,0 +1,68 @@
...
+import ArrayList from '@/pkg/rancher-components/src/components/Form/ArrayList/ArrayList';
+import { _EDIT, _VIEW } from "@shell/config/query-params";
+
+export const Template = (args, { argTypes }) => ({
+  components: { ArrayList },
+  props:      Object.keys(argTypes),
+  template:   '<ArrayList v-bind="$props" />',
+})
+
+<Meta
```
cnotv, line 11: We should show controls for each variation defined as default, plus your value is not set properly.

Replace this as follow:
```jsx
<Meta 
  title="Components/Form/ArrayList"
  decorators={[]}
  argTypes={{ 
    value: [],
    mode: {
      options: [_EDIT, _VIEW],
      control: { type: 'radio' },
    },
    valueLabel: 'Value',
    valuePlaceholder: 'e.g. bar',
    valueMultiline: false,
    addLabel: 'Add value',
    addAllowed: true,
    removeLabel: 'Remove value',
    removeAllowed: true,
    loading: false,
    disabled: false,
    defaultAddValue: '',
    initialEmptyRow: false,
    title: '',
    showHeader: false
  }}
/>
```

stories/components/form/ArrayList.stories.mdx
```diff
@@ -0,0 +1,68 @@
...
+/>
+
+# ArrayList
+The ArrayList is used to input a list of strings. The list is of variable length.
+
+## Default Empty ArrayList (Button only)
+<Canvas>
+  <Story
+    name="Empty"
+    args={{
```
cnotv, line 23: We can rely on default settings and replace this with `args` without brackets, as it still needs to define attribute.
`value` and `mode` are not needed.

stories/components/form/ArrayList.stories.mdx
```diff
@@ -0,0 +1,68 @@
...
+  </Story>
+</Canvas>
+
+
+## Showing an Initial Empty Row
+<Canvas>
+  <Story
+    name="Showing an initial empty row"
+    args={{
+       value: [],
```
cnotv, line 37: Also here `value` and `mode` are not needed.

pkg/rancher-components/src/components/Form/ArrayList/ArrayList.vue
```diff
@@ -114,115 +118,117 @@ export default {
...
+    },
   },
 
   created() {
-    this.queueUpdate = debounce(this.update, 50);
+    debounce(this.update, 50);
   },
 
   methods: {
+    queueUpdate(){
```
cnotv, line 169: It has been pointed out that this would create issue in case of multiple components it would cause issues.

cnotv (changes requested):
Very good job and thanks for covering every mentioned definition of done! 👍 👍 👍 

Please check mentioned points, especially for the i18n.

We should eventually create also an issue to address styling inconsistencies.

pkg/rancher-components/src/components/Form/ArrayList/ArrayList.vue
```diff
@@ -1,111 +1,115 @@
...
-      type:    Boolean,
+      type: Boolean,
       default: false,
     },
 
     addLabel: {
       type: String,
       default() {
-        return this.$store.getters['i18n/t']('generic.add');
+        return this.$store.getters["i18n/t"]("generic.add");
```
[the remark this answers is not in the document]
rak-phillip: This is the first component that provides internationalization of default values for labels. I agree with the recommended approach. In addition to this, we will also need to bundle any supported locales along with the rest of the library. 

One more thing, the reliance on the store here is potentially problematic as we have no guarantee that the consumer of this library will have the `i18n/t` getter. We will need to figure out the appropriate refactor to remove this dependency before moving forward.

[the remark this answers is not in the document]
catherineluse: Phil sent me a wrapper component that has the dependency on the store. 

@rak-phillip Does this now need conditional logic so that if you are Rancher, you get the wrapped component, but if you are not Rancher, you get the unwrapped component?

```
export { default as ArrayList } from './ArrayListWrapper.vue';
```

stories/components/form/ArrayList.stories.mdx
catherineluse: I don't like the idea of adding all these options because it doesn't demonstrate how the component is supposed to be used. A normal use case doesn't require most of these options.

stories/components/form/ArrayList.stories.mdx
catherineluse: I have removed the args prop

stories/components/form/ArrayList.stories.mdx
catherineluse: removed value and mode

pkg/rancher-components/src/components/Form/ArrayList/ArrayList.vue
```diff
@@ -114,115 +118,117 @@ export default {
...
+    },
     add() {
       this.rows.push({ value: clone(this.defaultAddValue) });
       if (this.defaultAddValue) {
         this.queueUpdate();
       }
 
       this.$nextTick(() => {
-        const inputs = this.$refs.value;
+        const inputs: any = this.$refs.value;
```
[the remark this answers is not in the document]
catherineluse: changed to `Array<string>`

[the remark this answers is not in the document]
catherineluse: It can't be `Array<string>` because the value of refs could be undefined, an object or an array.

pkg/rancher-components/src/components/Form/ArrayList/ArrayList.vue
```diff
@@ -1,111 +1,115 @@
...
-      type:    Boolean,
+      type: Boolean,
       default: false,
     },
 
     title: {
-      type:    String,
-      default: ''
+      type: String,
+      default: "",
```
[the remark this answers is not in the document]
catherineluse: changed to single quotes (but didn't figure out IDE configuration issue)

pkg/rancher-components/src/components/Form/ArrayList/ArrayList.vue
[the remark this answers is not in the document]
catherineluse: I'm having a problem with getting the events from ArrayList to get passed to the parent component of the wrapper, or with the value of the parent component getting updated properly.

For example, in the ingress creation form, there's an ArrayList to the right on this Certificates tab:
Screenshot 2022-11-17 at 11 13 11 PM

When it works properly, it's supposed to update the YAML like this:

```yaml
  tls:
    - hosts:
        - ioeioei
        - iaieoi245
```
But when the component is in the wrapper, it doesn't update the YAML. I can't see why yet.

pkg/rancher-components/src/components/Form/ArrayList/ArrayList.vue
catherineluse: Refactored to remove queueUpdate

package.json
```diff
@@ -190,7 +190,14 @@
     "webpack-virtual-modules": "0.4.3",
     "worker-loader": "3.0.8",
     "yaml-lint": "1.7.0",
-    "yarn": "1.22.18"
+    "yarn": "1.22.18",
+    "@storybook/addon-actions": "^6.3.8",
```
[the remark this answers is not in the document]
catherineluse: removed storybook packages

stories/components/form/ArrayList.stories.mdx
cnotv: The scope of stoybook is to interact with the component and test it there, as you do not have another places for components to do so, so any possible option should be there.
The API list is for documentation.
If you think that listing all the props may be unclear, we could add a default component for this scope.
What do you think?

pkg/rancher-components/src/components/Form/ArrayList/ArrayList.vue
```diff
@@ -0,0 +1,347 @@
...
+      <slot name="empty" />
+    </div>
+    <div v-if="showAdd && !isView" class="footer">
+      <slot v-if="showAdd" name="add" :add="add">
+        <button
+          data-testid="array-list-button"
+          type="button"
+          class="btn role-tertiary add"
+          :disabled="loading"
+          @click="add"
```
cnotv, line 308: @aalves08 pointed out that emitters do not automatically trigger functions in some cases.
Expected: `@click="add()`.

stories/components/form/ArrayList.stories.mdx
catherineluse: I thought tests are done in the component library and the storybook is for documentation only. I believe that documentation should be short and succinct, not overwhelming or exhaustive.

cnotv: I did not mention unit test at all. I said in Storybook you have to see how the component works in its integrity like in any styleguide of any UI library existing. That's literally the only way to know how to use them.
https://vuetifyjs.com/en/components/

cnotv: Beside the fact, visual tests in Storybook (as it comes out of the box) should also be something in an optimal scenario.
Some people do also screenshot testing through Jest, although it requires more efforts.

stories/components/form/ArrayList.stories.mdx
```diff
@@ -0,0 +1,71 @@
+import { Meta, Canvas, Story, ArgsTable, Source } from '@storybook/addon-docs';
+import ArrayList from '@/pkg/rancher-components/src/components/Form/ArrayList/ArrayList';
+import { _EDIT, _VIEW } from "@shell/config/query-params";
+
+export const Template = (args, { argTypes }) => ({
+  components: { ArrayList },
+  props:      Object.keys(argTypes),
+  template:   '<ArrayList v-bind="$props" />',
```
cnotv, line 8: We should also bind the props for events as follow, so can be observed the emission of the value.
[From the docs](https://storybook.js.org/docs/vue/writing-stories/args#story-args).

```
  template:   '<ArrayList v-bind="$props" v-on="$props"/>',
```

It will require to list them  on  the `Meta` in something like:

```
  argTypes={{ 
    add: {
      action: 'add',
    },
  }}
```

catherineluse:
I have removed ArrayList from the component library because there is not yet a plan for how to handle the translations for these components.

stories/components/form/ArrayList.stories.mdx
catherineluse: I don't see what's the point of listing emitted events here. What's the use case? Who would use Storybook to find out events? For me, I would always just look at the actual Vue component 100 percent of the time.

cnotv: Seems like there's a way to automatically add that, so we can skip this point.
We can discuss about the style guide use in another time and agree on some guideline as a team, since I don't think is the place and is not something I can choose on my own.
Sorry if it bothered you.

cnotv (approved):
LGTM

PR #7436, opened 2022-11-14
get docs base url version from getVersionInfo

### LINKED ISSUE #7308: Documentation Base URL should be calculated off of Rancher version
In `private-label.js` we hard-code the docs base URL.

We should not hard-code the version number, but instead use the one from the backend `/rancherversion` endpoint that we read in.

Note, for a dev build, this will be a commit ID rather than a version, so in that case we should use a fallback default.

mantis-toboggan-md:
Fixes #7308 - rather than using the /rancherVersion endpoint suggested in the issue I've used the existing getVersionInfo getter, so this update will work for older cluster versions.

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -3089,7 +3089,7 @@ monitoring:
       stepTitle: Uninstall V1
       stepSubtext: Uninstall Previous Monitoring
       warning1: V1 Monitoring is currently deployed. This needs to be uninstalled before V2 monitoring can be installed.
-      warning2: <a target="blank" href="{docsBase}/how-to-guides/advanced-user-guides/monitoring-alerting-guides/migrate-to-rancher-v2.5+-monitoring#migrating-from-monitoring-v1-to-monitoring-v2" target='_blank' rel='noopener nofollow'>Learn more</a> about the migration steps to V2 Monitoring.
+      warning2: <a target="blank" href="https://docs.ranchermanager.rancher.io/v2.6/how-to-guides/advanced-user-guides/monitoring-alerting-guides/migrate-to-rancher-v2.5+-monitoring#migrating-from-monitoring-v1-to-monitoring-v2" target='_blank' rel='noopener nofollow'>Learn more</a> about the migration steps to V2 Monitoring.
       promptDescription: <div class="mt-20 mb-20">You are attempting to uninstall V1 Monitoring. Please ensure you have read the migration steps.</div>
       success1: V1 monitoring successfully uninstalled.
       success2: Press Next to continue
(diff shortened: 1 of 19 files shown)


catherineluse (changes requested):
I tested some of the links on the Rancher UI and it works fine. However, the migration docs were removed from the v2.7 documentation, so I believe that the 2.6 version should be prepended to the migration guide links. (By default, without a prepended version, the links redirect to the main docs page.)

mantis-toboggan-md:
Coming back to this PR with fresh eyes, I don't think the feature is viable with the way docs are structured. If this were merged, 2.7.x instances would just use `https://docs.ranchermanager.rancher.io` as the doc base which is fine for now, but that url will point to 2.8 docs when that is released, and 2.7 docs would be listed under `https://docs.ranchermanager.rancher.io/v2.7` - a url that doesn't currently work. So I think the doc site would have to be changed to list the latest docs  at `https://docs.ranchermanager.rancher.io/<latest version>`... @catherineluse since you were involved in the recent documentation re-structuring do you know if this url change is viable/reasonably easy to do?

Even if that change we made, the fact that we can't guarantee a particular route will work between docs versions makes this feature pretty brittle: we'd need to verify all doc links work on each minor release, which I suppose should be a set of e2e tests.

catherineluse:
I asked the docs team about how to create future-proof links to the latest docs version. They said that although it would be possible to create a generic redirect for all links to v2.7 docs, they would prefer to just create specific redirects for the docs that we need future proof links for.

In my opinion we should always hardcode the version in the URL, even when pointing to the latest docs, and open an issue in the rancher/docs repo to make those specific links redirect to the appropriate docs pages.

I checked to see how many times we link to the v2.7 docs in our codebase and by my count it's only five times (this is not counting the links to the v1 app migration guides which should point to v2.6):

- `/faq/telemetry/`
- `/pages-for-subheaders/monitoring-v2-configuration`
- `/how-to-guides/new-user-guides/launch-kubernetes-with-rancher/rke1-vs-rke2-differences`
- `/how-to-guides/advanced-user-guides/monitoring-v2-configuration-guides/advanced-configuration/alertmanager`
- `/how-to-guides/new-user-guides/authentication-permissions-and-global-configuration/authentication-config/configure-google-oauth#3-creating-service-account-credentials`

I've opened an issue about it on the docs side to track the work of adding redirects on their end. https://github.com/rancher/rancher-docs/issues/313

Note: I'm not married to this approach - if there's a better way to do it we can discuss.

mantis-toboggan-md:
What is the argument against adding a generic redirect? It seems more future-proof. My only big concern with the specific-redirect approach is that any time we want to add a new doc link, we'd have to go ask the docs team to create a redirect for it and I think there's a real good chance we'd forget to do that. I mean, I think if I were adding a link out to docs for the first time, I'd probably look at how other doc links are structured in the localization file, see that they use `docsBase`, and just use that myself without looking for more information in our developer guide.

catherineluse:
> What is the argument against adding a generic redirect?

Billy said it was because the way the generic redirect works in the Rancher Desktop documentation is because it involved creating a separate directory of duplicated files, which created overhead. @btat can you please advise?

mantis-toboggan-md:
I think an alternative approach we can take that wouldn't require the docs team to duplicate the entirety of the rancher docs would be to query the github api to find the latest version of rancher and work out if it's the same major/minor version of the current rancher instance, and avoid hardcoding the url in that case. I'll make a PR with this functionality soon(tm)

mantis-toboggan-md:
I've updated the doc url versioning functionality to (attempt to) query github for the latest rancher version. If that doesn't work for whatever reason, the docs url will default to using the latest version of docs. I've also updated links to 2.6-only docs to hardcode the doc url and avoid appending the wrong version.

It's worth noting that -head builds link to the latest doc version, so this PR should be tested with non-head 2.6 and 2.7 rancher instances.

gaktive:
As a heads up, the base URL has changed so it's no longer `.io` but there's a new sub-domain. Let me find that...

catherineluse (approved):
LGTM

PR #7456, opened 2022-11-15
unit tests: Collapse, SimpleBox

### LINKED ISSUE #7455: [unit-test] Collapse.vue, SimpleBox.vue
Create unit test for `Collapse`,  `SimpleBox` components.

torchiaf:

### Summary
Fixes #7455

### DIFF
--- a/shell/components/__tests__/Collapse.spec.ts
+++ b/shell/components/__tests__/Collapse.spec.ts
@@ -1,0 +1,44 @@
+import { mount } from '@vue/test-utils';
+import Collapse from '@shell/components/Collapse.vue';
+
+describe('component: Collapse.vue', () => {
+  describe('closed', () => {
+    const wrapper = mount(Collapse, { propsData: { open: false } });
+
+    it('is closed', () => {
+      const content = wrapper.find(`[data-testid="collapse-content"]`);
+
+      expect(content.element).toBeUndefined();
+    });
+
+    it('icon is chevron right', () => {
+      const icon = wrapper.find(`[data-testid="collapse-icon-right"]`);
+
+      expect(icon.element).toBeDefined();
+    });
+  });
+
+  describe('collapsed', () => {
+    const wrapper = mount(Collapse, { propsData: { open: true } });
+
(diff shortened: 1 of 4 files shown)


shell/components/__tests__/Collapse.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+
+      expect(icon.element).toBeDefined();
+    });
+
+    it('emit', async() => {
+      const collapse = wrapper.find(`[data-testid="collapse-div"]`);
+
+      await collapse.trigger('click');
+
+      expect(wrapper.emitted('update:open')).toBeTruthy();
```
aalves08, line 41: a bit nitpicky but I would also check if the value sent on emit is correct:

`expect(wrapper.emitted('update:open')![0][0]).toBe('whatever-value-to-expect')`

shell/components/__tests__/SimpleBox.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+    expect(closeButton.element).toBeDefined();
+  });
+
+  it('close emit', async() => {
+    await wrapper.setProps({ canClose: true });
+    const closeButton = wrapper.find(`[data-testid="simple-box-close"]`);
+
+    await closeButton.trigger('click');
+
+    expect(wrapper.emitted('close')).toBeTruthy();
```
aalves08, line 26: same as previous comment

aalves08 (approved):
Left a couple of comments, but good for approval.

I prefer to go a bit further on my unit tests because I like to make sure stuff appears as it should. Ex:

`expect(title.element).toBeDefined();`
I would do:
`expect(title.element).toBe('whatever-value-should-be-here');`

but since we don't have guidelines for unit tests and you've covered the most important parts I think we are good

shell/components/__tests__/SimpleBox.spec.ts
torchiaf: Thanks for the review Alex, got you points, it's a good idea to be more specific in verifying expected results.

PR #7420, opened 2022-11-11
Unit testing components

aalves08:
Fixes #7393 

Adds unit tests for the following components:
- AsyncButton
- BackLink
- ButtonGroup
- ClusterBadge
- CollapsibleCard
- ConsumptionGauge
- FixedBanner
- PercentageBar
- GrowlManager

PR #7460, opened 2022-11-15
Unit test some table formatters

### LINKED ISSUE #7434: Add unit tests for generic table formatters
Add unit tests for eg:
LiveDate,
Age,
BadgeState,
CopyToClipboard,

mantis-toboggan-md:
### Summary
Fixes #7434 
 this pr adds unit tests for LiveDate, Date, LinkDetail, and Si formatters

### DIFF
--- a/shell/components/formatter/LiveDate.vue
+++ b/shell/components/formatter/LiveDate.vue
@@ -7,7 +7,7 @@ import { diffFrom } from '@shell/utils/time';
 export default {
   props: {
     value: {
-      type:    String,
+      type:    [String, Number],
       default: ''
     },
 
@@ -109,10 +109,10 @@ export default {
 
       let label = diff.label;
 
-      if ( diff === 0 ) {
+      if ( diff.diff === 0 ) {
         label = 'Just now';
       } else {
-        label += ` ${ prefix } ${ this.t(diff.unitsKey, { count: diff.label }) }`;
+        label += ` ${ prefix }${ this.t(diff.unitsKey, { count: diff.label }) }`;
         label = label.trim();
       }
 
--- a/shell/components/formatter/__tests__/LiveDate.test.ts
+++ b/shell/components/formatter/__tests__/LiveDate.test.ts
@@ -1,6 +1,8 @@
 import { mount } from '@vue/test-utils';
 import LiveDate from '@shell/components/formatter/LiveDate.vue';
 import dayjs from 'dayjs';
+import { ExtendedVue, Vue } from 'vue/types/vue';
+import { DefaultProps } from 'vue/types/options';
 
 const defaultMock = {
   $store: {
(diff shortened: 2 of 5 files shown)


shell/components/formatter/__tests__/LiveDate.test.ts
```diff
@@ -0,0 +1,82 @@
...
+    const wrapper = await mount(LiveDate, {
+      mocks:     defaultMock,
+      propsData: { value: dayjs().toString() }
+    });
+
+    let element = wrapper.find('span');
+
+    expect(element.text()).toContain('Just now');
+    await new Promise(resolve => setTimeout(resolve, 1000));
+    await wrapper.vm.liveUpdate(Date.now());
```
aalves08, line 54: This is generating a linting error during build on e2e tests "Prepare Build" step. In order to fix this (it ain't pretty...) One idea is what I used (with Giuseppe's help) for my unit tests PR:
https://github.com/rancher/dashboard/pull/7420/files#diff-3768d1217b02196ee43bf62bee0f4965eeb8265fdea94db6a5cee1949df6d7ba

import these:
```
import { ExtendedVue, Vue } from 'vue/types/vue';
import { DefaultProps } from 'vue/types/options';
```

and instead of:
```
const wrapper = await mount(LiveDate, {
```

you do:
```
const wrapper = await mount(LiveDate as unknown as ExtendedVue<Vue, {}, {}, {}, DefaultProps>, {
```

TS fun here... 😛  That should fix it

shell/components/formatter/LiveDate.vue
```diff
@@ -109,10 +109,10 @@ export default {
 
       let label = diff.label;
 
-      if ( diff === 0 ) {
+      if ( diff.diff === 0 ) {
```
aalves08, line 112: this is a duesy... did something change or did we get this horribly wrong before this change?

aalves08 (commented):
Left a couple of comments... To "catch" those e2e linting errors, you can run `yarn build` locally and should be able to replicate it

shell/components/formatter/LiveDate.vue
mantis-toboggan-md: We got it horribly wrong before this change! If you look at that `diffFrom` function you can see it's returning an object with a `diff` key. This bug causes the table formatter to show '0 secs' instead of 'Just now' - I think we just never noticed because that's shown so briefly.
https://github.com/rancher/dashboard/blob/9268de26403418c17f75d792132addcbcad0ee74/shell/utils/time.js#L33-L40

shell/components/formatter/__tests__/LiveDate.test.ts
mantis-toboggan-md: Oof, thank you! Who knows how long that would have taken me to work out myself D:

aalves08 (approved):
LGTM

PR #7462, opened 2022-11-16
Add GitHub workflow for adding triage labels

catherineluse:
Addresses https://github.com/rancher/dashboard/issues/7579

This PR creates a GitHub workflow to add a new default label, "[zube]: To Triage" to all new issues. There are exceptions for issues which are initially created with Zube labels. Based on my interpretation of the docs on expressions (https://docs.github.com/en/actions/learn-github-actions/expressions#contains) and Stack Overflow (https://stackoverflow.com/questions/65653291/conditional-to-check-if-a-string-is-included-in-a-list), it appears that you can't do a regex search for all labels that start with 'zube' if you want to conditionally run an action. So I have it set up to specifically check for the only five Zube labels that I think anyone would ever plausibly open an issue with - review, backlog, backend blocked, working, and next up.

I tested the GitHub action in a separate repository, which is here: https://github.com/catherineluse/test-github-action/issues

I have also switched from an external action to an official GitHub action which is documented here:  https://github.com/github/issue-labeler


nwmac:
@catherineluse The best way to test is on a test GitHub Repository - create a repo add the workflow and then create issues and check it does the right thing.

nwmac:
You might want to check that the issue doesn't have any Zube labels before adding Triage - e.g. someone creates an issue with the zube backend blocked label.

I see you're using a custom external action - these need to be added to an allow list on the repository - we do need check that external actions are not malicious.

catherineluse:
@nwmac Does it need to be added to the allow list if it's an official GitHub action? like this one https://github.com/github/issue-labeler

rak-phillip:
We might not need to add the action to the allow list depending on our repo security. Actions created by github and trusted creators would be permitted if we have `Allow rancher, and select non-rancher, actions and reusable workflows` enabled:

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#allowing-select-actions-and-reusable-workflows-to-run

PR #7465, opened 2022-11-16
Add unit tests for Namespace filtering in resources

### LINKED ISSUE #7391: Add unit tests for Namespace filtering in resources
# Description 

Ensure namespace filtering from the top bar to apply changes.

## Context

Part of the group: https://github.com/rancher/dashboard/issues/6613

* When a user selects a namespace in the top nav of cluster explorer, then goes to the list view of a namespaced resource such as Deployments, confirm that only resources in the selected namespaces are listed
* After refreshing the page, the namespace filter is still persisted

## Required tests

- Assure the selection of the namespace from the top bar to be applied in the store or resource (tbd source file)
- Given a faux resource list from the endpoint, render based on filtered values from the store (example case on Development)
- Ensure ~localstorage~ user preference resource saving and loading capabilities

cnotv:
### Summary
Fixes #7391
These tests are aimed to ensure the functionalities of the namespace filter an part of the Workload fetching data capabilities.

- Prevented issues in `NamespaceFilter` component in case of missing props or data, e.g. `currentProduct`
- Added `data-testid` attributes for selecting component.

### Technical notes summary
- Added  [jest.restoreAllMocks()](https://jestjs.io/docs/jest-object#jestrestoreallmocks) globally for jest setup
- Added information related on global configuration for the unit tests in the documentation

### DIFF
--- a/shell/components/__tests__/NamespaceFilter.test.ts
+++ b/shell/components/__tests__/NamespaceFilter.test.ts
@@ -1,0 +1,213 @@
+import { mount } from '@vue/test-utils';
+import NamespaceFilter from '@shell/components/nav/NamespaceFilter.vue';
+
+describe('component: NamespaceFilter', () => {
+  describe('given namespace select input', () => {
+    it('should be visible', () => {
+      const wrapper = mount(NamespaceFilter, {
+        computed: {
+          filtered: () => [],
+          options:  () => [],
+          value:    () => [],
+        },
+        directives: { shortkey: () => jest.fn() }
+      });
+      const filter = wrapper.find(`[data-testid="namespaces-filter"]`);
+
+      expect(filter).toBeDefined();
+    });
+
+    it('should display no namespace selection', () => {
+      const text = 'none';
+      const wrapper = mount(NamespaceFilter, {
+        computed: {
+          filtered: () => [],
+          options:  () => [],
+          value:    () => [],
+        },
+        mocks:      { $store: { getters: { 'i18n/t': () => text } } },
+        directives: { shortkey: () => jest.fn() }
+      });
+      const element = wrapper.find(`[data-testid="namespaces-values-none"]`).element.textContent;
+
+      expect(element).toContain(text);
+    });
+
+    it('should display the default namespace', () => {
+      const text = 'special namespace';
+      const wrapper = mount(NamespaceFilter, {
+        computed: {
+          filtered: () => [],
+          options:  () => [],
+          value:    () => ([{
+            label: text,
+            kind:  'special',
+          }]),
+        },
+        directives: { shortkey: () => jest.fn() }
+      });
+
+      const element = wrapper.find(`[data-testid="namespaces-values-label"]`).element.textContent;
+
+      expect(element).toContain(text);
+    });
+
+    it('should display the selected namespace', () => {
+      const text = 'current namespace';
+      const wrapper = mount(NamespaceFilter, {
+        computed: {
+          filtered: () => [],
+          options:  () => [],
+          value:    () => [{ label: text }],
+        },
+        directives: { shortkey: () => jest.fn() }
+      });
+
+      const element = wrapper.find(`[data-testid="namespaces-value-0"]`).element.textContent;
+
+      expect(element).toContain(text);
+    });
+
+    it('should display the selected namespace from user preferences if options are available', () => {
+      const text = 'my preference';
+      const key = 'local';
+      const preferences = {
+        [key]: [
+          `ns://${ text }`
+        ]
+      };
+
+      jest.spyOn(NamespaceFilter.computed.value, 'set');
+      const wrapper = mount(NamespaceFilter, {
+        computed: {
+          filtered:       () => [],
+          options:  () => [{
+            id:    `ns://${ text }`,
+            kind:  'namespace',
+            label: text
+          }],
+          currentProduct: () => undefined,
+          key:            () => key,
+        },
+        mocks: {
+          $store: {
+            getters: {
+              'i18n/t':    () => text,
+              'prefs/get': () => preferences
+            },
+          }
+        },
+        directives: { shortkey: () => jest.fn() }
+      });
+
+      const element = wrapper.find(`[data-testid="namespaces-value-0"]`).element.textContent;
+
+      expect(element).toContain(text);
+    });
+  });
+
+  describe('given namespace menu options', () => {
+    it('should be opened and displayed on click', async() => {
+      const wrapper = mount(NamespaceFilter, {
+        computed: {
+          filtered: () => [],
+          options:  () => [],
+          value:    () => [],
+        },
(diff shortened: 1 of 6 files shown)


shell/components/__tests__/NamespaceFilter.test.ts
```diff
@@ -0,0 +1,213 @@
...
+    it('should display the selected namespace from user preferences if options are available', () => {
+      const text = 'my preference';
+      const key = 'local';
+      const preferences = {
+        [key]: [
+          `ns://${ text }`
+        ]
+      };
+
+      jest.spyOn(NamespaceFilter.computed.value, 'set');
```
rak-phillip, line 80: I'm not understanding the purpose of the spy in this situation. I would generally expect some sort of assertion associated with the spy. Perhaps its usage is different here?

In a previous project, we always made sure to clean up our spys with `.mockRestore()`. Perhaps this isn't required but maybe something we can evaluate. https://jestjs.io/docs/mock-function-api#mockfnmockrestore

rak-phillip, line 80: I get the following warning for `NamespaceFilter.computed.value`

```
Property 'computed' does not exist on type 'VueConstructor<Vue>'.ts(2339)
```

This might be an issue with my environment but warrants calling out.

shell/components/__tests__/NamespaceFilter.test.ts
```diff
@@ -0,0 +1,213 @@
...
+        directives: { shortkey: () => jest.fn() }
+      });
+
+      await wrapper.find(`[data-testid="namespaces-dropdown"]`).trigger('click');
+      await wrapper.find(`[data-testid="namespaces-option-0"]`).trigger('click');
+
+      expect(action).toHaveBeenCalledWith(actionName, value);
+    });
+
+    it.todo('should generate the options based on the Rancher resources');
```
rak-phillip, line 211: Is there a plan for this dangling todo?

shell/components/nav/NamespaceFilter.vue
```diff
@@ -87,7 +87,8 @@ export default {
       const t = this.$store.getters['i18n/t'];
       let out = [];
 
-      if (this.currentProduct.customNamespaceFilter) {
+      // TODO: Add return info
```
rak-phillip, line 90: I think it's a good practice to link a todo in comments to a github issue. This way other developers can gather context as to what needs to be done and we can identify if this has been finished.

shell/components/nav/NamespaceFilter.vue
```diff
@@ -561,53 +571,68 @@ export default {
...
+
+    <!-- Select Dropdown control -->
     <div
       ref="dropdown"
       class="ns-dropdown"
+      data-testid="namespaces-dropdown"
       :class="{ 'ns-open': isOpen }"
       @click="toggle()"
     >
+      <!-- No filters found or available -->
```
rak-phillip, line 583: These are all great comments but they are making me think that these are describing individual components or slots that we don't currently have 🤔

Just something to ponder about. Maybe there's a sign that this might be approaching time for a refactor.

shell/list/__tests__/workload.test.ts
```diff
@@ -0,0 +1,53 @@
...
+          pending: false, error: true, timestamp: Date.now()
+        },
+        $route: { params: { resource } },
+      }
+    });
+
+    expect(wrapper.vm.schema).toStrictEqual(schema);
+  });
+
+  it.todo('should display the list of workloads');
```
rak-phillip, line 48: Do we have a plan for these unimplemented tests? Perhaps these would benefit from being linked to an issue as well so we can follow up at a later date if needed.

rak-phillip (commented):
I think the tests look great. I just have a few comments on spy usage, todos, and tsc warnings 🙂

shell/components/__tests__/NamespaceFilter.test.ts
cnotv: I'll set a global configuration to always restore mocks after each test, as [defined here](https://jestjs.io/docs/configuration#setupfilesafterenv-array).

shell/components/__tests__/NamespaceFilter.test.ts
cnotv: As defined in the ticket, it must be clarified by the author of the code what is defined here.
We may create a ticket to add documentation and bring information, so can be used in the component. As it applies to pretty much any component which includes models or controls, I would rather avoid it for the next iteration or internal discussions.

shell/components/nav/NamespaceFilter.vue
cnotv: I would personally rewrite entirely the whole logic of most of the components, but that would be out of scope and too expensive if there's not a real reason behind.
We should considering a refactoring on the next design change.

The comments however are meant to identify cases. The abstraction of the components may be related more on chips, options and such, rather than cases of the provided data.

shell/list/__tests__/workload.test.ts
cnotv: Adding issue number for the existing epic.
https://github.com/rancher/dashboard/issues/7473

shell/components/nav/NamespaceFilter.vue
cnotv: I totally agree with you, still this reason is not sufficient to need a ticket as there's not enough requirements to write down, even if we just want to refactor it (comment above). This should rather be addressed in a next issue, where the functionality is required and needs to be defined to proceed.

I keep it unresolved to know if you agree or if you have anything in mind.

shell/components/__tests__/NamespaceFilter.test.ts
cnotv: I opted for [jest.restoreAllMocks()](https://jestjs.io/docs/jest-object#jestrestoreallmocks).

rak-phillip: Thanks for verifying. I will chalk it up to my editor.

shell/components/nav/NamespaceFilter.vue
rak-phillip: I think we're fine here. It sounds like we have some potential future refactoring work but we would need to gather requirements for it. These todos would get identified during that process.

rak-phillip (approved):
LGTM

catherineluse:
@cnotv It looks like cluster explorer is no longer rendering for downstream clusters. The error says something about the namespace filter. could it have been caused by a change in this PR?
Screenshot 2022-11-17 at 5 32 28 PM

cnotv:
@catherineluse I have created a new issue here: https://github.com/rancher/dashboard/issues/7495
Could you please define how to reproduce it?

cnotv:
@catherineluse Taking a look to the code, I see nothing which may cause issues, rather the other way around, to prevent access function when undefined. In the same way I don't see any other commit either since a while.

PR #7475, opened 2022-11-17
Update key-spacing eslint rule

rak-phillip:
### Summary

This applies `mode: strict` to the `key-spacing` eslint rule to enforce exactly one space after colons in objects. 

We noticed that certain objects wouldn't auto-format to align values when there were too many spaces in Rancher Desktop rancher-sandbox/rancher-desktop#3400. Updating this eslint rule allows objects values to align nicely, as we most likely intended with the original rule.

This is a good example of what this rule fixes

https://github.com/rancher/dashboard/blob/d05676212068aecd08b8e0c64568b3a4bbeeb5ed/pkg/epinio/models/applications.js#L33-L38

Notice how the spacing is inconsistent between the values but this passes the linter. This happens because `mode: minimum` only enforces one or more spaces.

This is strictly code formatting.

This is strictly code formatting. We might want to make sure that our linting rules still behave as expected.


cnotv (approved):
I can see in manu cases it reduced the spacing, but all in all it seems way more neat, so good job there 👏 
Please give it a warning in the chat before the merge, so people do not panic 😅 

LGTM

PR #7481, opened 2022-11-17
Replace initial Verdaccio guide with published npm packages

jordojordo:
### Summary
This replaces the initial use of Verdaccio when creating plugins/pkgs as we can leverage the published npm packages. For the "Use Case: Publish a package to NPM" section I have added the info on using Verdaccio to publish locally.


rak-phillip (approved):
LGTM - I think these changes make sense. I originally had reservations about removing instructions on how to build and publish the creator packages but I think using the packages published to npm should suffice.

Issue #7495, opened 2022-11-18
Resource lists are not loaded


cnotv:
**Setup**
- Rancher version: 2.7.0
- Browser type & version: latests

**Describe the bug**
It looks like cluster explorer is no longer rendering for downstream clusters. The error says something about the namespace filter. 

**To Reproduce**
1. Check out the master branch
2. Go to cluster explorer for a downstream cluster

**Result**
Rancher crashes

**Expected Result**
Rancher does not crash

**Screenshots**
<img width="1521" alt="Screenshot 2022-11-18 at 8 05 04 AM" src="https://user-images.githubusercontent.com/20599230/202736031-7506cb6a-c40c-4ba3-bc2c-fcbe3bab13b4.png">

**Additional context**
Issue seems to have been created while modifying some conditions in https://github.com/rancher/dashboard/pull/7465

PR #7498, opened 2022-11-18
Cluster explorer no longer crashes

catherineluse:
Fixes https://github.com/rancher/dashboard/issues/7495

If a preferences object exists, the code checks that object for the namespace filtering preference for the active cluster. But sometimes, the user does not have any preexisting preferences for the cluster, in which case the UI crashes. This fix makes it so the default is used if the user has no preference set for namespace filtering for the active cluster.


cnotv (approved):
LGTM

PR #7529, opened 2022-11-23
StringList, type:item event

### LINKED ISSUE #7528: [@rancher/components] StringList, add event type:item
The `type:item` is emitted while the user type an item name or when the input field lose focus (`null` is emitted).

torchiaf:

### Summary
Fixes #7528


aalves08 (approved):
LGTM

PR #7539, opened 2022-11-23
Change the default value for 'Allow Privilege Escalation' in container security context

### LINKED ISSUE #7165: Creating a Deployment defaults to "Allow Privilege Escalation"
Internal reference: SURE-3663

Request description:
When users create Deployments in the Rancher Cluster Explorer UI, the setting under "Security Context" -> "Privilege Escalation" defaults to _"Yes, container can gain more privileges than its parent process"_

Users will fail to create deployments by default (at least when e.g. setting the _"Run as User ID"_ field). The security context parameter is not added to the deployment unless one parameter is set, however when some parameter is set the Privilege Escalation option is set to yes by default.

This could be considered a more secure means of running Kubernetes, which would help with RKE2 strategy to align with other usage within the product.

Actual behavior:
Privilege Escalation in the security context tab is set to `yes` by default

Expected behavior:

Privilege Escalation in the security context tab is set to `no` by default

Additional notes:
Instead of choosing between one default or another, it would be helpful to have a way to select the default config options in the UI for each user.

mantis-toboggan-md:
### Summary
Fixes #7165 

The ask was to change the default for `allowPrivilegeEscalation` but the ticket also mentions that these securityContext defaults are only set when a value on the tab is changed. We can avoid that by setting them when the container is initialized. I also noticed we were trying to set `runAsRoot` when the correct field is `runAsNonRoot` https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#securitycontext-v1-core

### DIFF
--- a/shell/components/form/Security.vue
+++ b/shell/components/form/Security.vue
@@ -63,10 +63,10 @@ export default {
 
     const {
       capabilities = {},
-      runAsRoot = true,
+      runAsNonRoot = false,
       readOnlyRootFilesystem = false,
       privileged = false,
-      allowPrivilegeEscalation = true,
+      allowPrivilegeEscalation = false,
       runAsUser
     } = this.value;
     const {
@@ -78,7 +78,7 @@ export default {
       privileged,
       allowPrivilegeEscalation,
       allCapabilities,
-      runAsRoot,
+      runAsNonRoot,
       readOnlyRootFilesystem,
       add,
       drop,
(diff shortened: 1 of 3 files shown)


Shavindra (approved):
LGTM

mantis-toboggan-md:
@Shavindra I looked into this more and I think the UI is behaving correctly, and unfortunately this isn't something we can validate on our end as we can't tell whether or not an image is configured to run as root.

PR #7562, opened 2022-11-28
Fix role form input event handling

### LINKED ISSUE #7549: Some input fields are broken on role creation screens
The text input fields on the Role creation screen are broken - if you start typing, you just get [object inputEven] - see below 

Looks like the input fields were changed from LabelInput to plain input fields - LabelInout emits the value of the entire  input via the @input event - the input element just outputs each keypress - this is why it is broken. 

This PR is the cause: https://github.com/rancher/dashboard/pull/7258/files

image

mantis-toboggan-md:
Fixes #7549 

This PR also fixes a bug with the 'resources' select where hitting 'backspace' while the select is focused causes a console error and fails to clear the selected value.

### DIFF
--- a/shell/components/auth/RoleDetailEdit.vue
+++ b/shell/components/auth/RoleDetailEdit.vue
@@ -380,28 +380,27 @@ export default {
       // - resourceNames
       // - resources
       // - verbs
-      const value = event.label ? event.label : event;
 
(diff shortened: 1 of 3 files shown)


catherineluse (approved):
LGTM

PR #7570, opened 2022-11-29
Fixed close icon color on hover in Banner component

cogor:
### Summary
Close icon on hover is invisible because the color is equal to banner background color
Fixes #
Fixed close icon color on hover in Banner component for primary variant

### Screenshot/Video
Before: 
before

After:
after

### DIFF
--- a/pkg/rancher-components/src/components/Banner/Banner.vue
+++ b/pkg/rancher-components/src/components/Banner/Banner.vue
@@ -224,6 +224,7 @@ $icon-size: 24px;
       text-align: center;
 
       .closer-icon {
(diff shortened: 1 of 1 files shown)


bisht-richa:
@cogor You have merge conflicts

cnotv:
I think the color picked should match the text, what do you think @kwwii ?

kwwii:
Agreed, we want this to play nice with theme switchers and such. It would be best to have it be the same color as the text. I think that the specific colours of these alerts and banners will be changed in the near-ish future so it is important that we keep it the same now.

PR #7578, opened 2022-11-29
CI, add steps for rancher-components

### LINKED ISSUE #7344: [@rancher/components] Assert that package builds in CI
It looks like builds for `@rancher/components` regressed with https://github.com/rancher/dashboard/commit/d709dcc9db06b2a846fa254e81b1774583480757#diff-1a3646597f62985362f87700c4a0752e51a8868f3849e97ba504965a44f53113 and this wasn't identified until we attempted to build and release a new component. 

It makes sense to build `@rancher/components` in CI to assert that we don't revert basic behavior in the future. 

We should be able to modify the ci script to include a line for `yarn build:lib`

https://github.com/rancher/dashboard/blob/2c22aadef43f60a30bbbbefa9962bc2c29dd7be5/scripts/ci

See also rancher/dashboard#7327

## edit
Also add a CI step to identify `lint` errors.

torchiaf:

### Summary
Fixes #7344

### DIFF
--- a/.github/workflows/rancher-components.yaml
+++ b/.github/workflows/rancher-components.yaml
@@ -1,0 +1,40 @@
+name: rancher-components
+on:
+  pull_request:
+    paths:
+      - 'pkg/rancher-components/**/*'
+    branches:
+      - master
+
+jobs:
+  validate-build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 1
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '14.x'
+      
+      - name: Install packages
+        run: yarn install
+
+      - name: Verify build
+        run: yarn build:lib
+
+  validate-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 1
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '14.x'
+      
+      - name: Install packages
+        run: yarn install
+      
+      - name: Run linters
+        run: yarn lint:lib


.github/workflows/rancher-components.yaml
```diff
@@ -0,0 +1,25 @@
...
+          node-version: '14.x'
+      
+      - name: Install packages
+        run: yarn install
+
+      - name: Verify build
+        run: yarn build:lib
+      
+      - name: Run linters
+        run: cd pkg/rancher-components && yarn lint
```
cnotv, line 25: Please add command specific for linting in the main repository, so we are able to run autofix with the flag `--fix`

.github/workflows/rancher-components.yaml
```diff
@@ -0,0 +1,25 @@
+name: rancher-components
+on:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  validate:
```
cnotv, line 8: Please split this job in `build` and `lint` as we want to have separated checks and know what is failing in a glance

.github/workflows/rancher-components.yaml
```diff
@@ -0,0 +1,25 @@
+name: rancher-components
+on:
+  pull_request:
```
cnotv, line 3: Please restrict this check on file changes [specifying the path](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore), as it should not require files outside of the library.

E.g.:

```yml
on:
  pull_request:
    paths:
      - 'pkg/rancher-components/**/*'
    branches:
      - master
```

.github/workflows/rancher-components.yaml
```diff
@@ -0,0 +1,25 @@
...
+    branches:
+      - master
```
cnotv, line 5: Please create an action by temporary removing the `master` flag, so we are sure to do not break `master` afterwards.
You can then link the [action](https://github.com/torchiaf/dashboard/actions/workflows/test.yaml) in the PR.

cnotv (changes requested):
Please create an issue for the warnings, because these will appear every time.

.github/workflows/rancher-components.yaml
cnotv: Sorry I did not realize that is a PR and not a push, so it's already here 🙃 

For the records here's the checks:
https://github.com/rancher/dashboard/actions/workflows/rancher-components.yaml

Noticed also the check below 🙃

torchiaf:
> Please create an issue for the warnings, because these will appear every time.

Thanks for the review! Just create an issue to remove lint warnings: #7596

.github/workflows/rancher-components.yaml
torchiaf: Done & tested.

cnotv (approved):
LGTM

PR #7601, opened 2022-12-01
Fix storybook rancher logo

nwmac:
Logo is broken when published:

image

This PR fixes this by using a relative link for the image rather than an absolute one.


nwmac:
@bisht-richa Could you review this for me, please?

bisht-richa (approved):
LGTM

PR #7604, opened 2022-12-01
Tabs component storybook

### LINKED ISSUE #7141: Add tabs component to Storybook
# Description

Add tabs related components to Storybook, in highlight of given situation https://github.com/rancher/dashboard/issues/7124

## Context

Required cases:
- default
- sideTabs
- nested horizontal/vertical, default -> sideTabs (as in Deployment, with flat prop)
- nested vertical/vertical, sideTabs -> sideTabs (as in NetworkPolicy)
- nested horizontal/vertical/horizontal, default -> sideTabs -> default (none yet afaik)
- ~adding/removing default~ slot `tab-row-extras` (with any content of choice)
- adding/removing sideTabs (not available for default)

## Screenshot

Screenshot 2022-11-30 at 15 30 18

Screenshot 2022-11-30 at 12 29 51

Screenshot 2022-11-30 at 12 30 11

Screenshot 2022-11-30 at 12 30 39

bisht-richa:
### Summary
Fixes #7141

Added tabs component to Storybook


package.json
```diff
@@ -190,7 +190,14 @@
     "webpack-virtual-modules": "0.4.3",
     "worker-loader": "3.0.8",
     "yaml-lint": "1.7.0",
-    "yarn": "1.22.18"
+    "yarn": "1.22.18",
```
cnotv, line 193: Please remove Storybook from the main project.

stories/components/Alert.stories.mdx
```diff
@@ -0,0 +1,50 @@
+
```
cnotv, line 1: Do you know when is the Alert used? I dont get exactly why is called like that and where/when we use it.
For instance the `BadgeState` does also look pretty much the same, it seems rly confusing.
If you also dont know, let's just mark it as resolved and we can review another time.

stories/components/Tab.stories.mdx
```diff
@@ -0,0 +1,315 @@
...
+      ],
+      useHash: false,
+      sideTabs: true,
+      showTabsAddRemove: true,
+    }}>
+    {AddRemove.bind({})}
+  </Story>
+</Canvas>
+
+### Add/Remove Default Tab
```
cnotv, line 291: Could you please change the markup and remove the logic? This is not the built in add/remove feature, so we cannot list as something ready to use. It has, sorry to say it, no purpose for the style guide.

Let's add anything else in the tab. Keeping the button may also be ok, although it may is totally misleading.

In the future we may add also specification in what and how add it, if not e refactoring.

stories/components/Tab.stories.mdx
```diff
@@ -0,0 +1,315 @@
...
+    args={{
+      useHash: false,
+      sideTabs: true,
+    }}>
+    {Template.bind({})}
+  </Story>
+</Canvas>
+
+
+#### Nested
```
cnotv, line 236: We had an issue to identify exactly cases where we had 3 nested tabs, as defined in the issue: horizontal, vertical, vertical.
Could you please add it?

stories/components/ButtonDropdown.stories.mdx
```diff
@@ -0,0 +1,63 @@
...
+  components: { ButtonDropdown },
+  props:      Object.keys(argTypes),
+  template:   '<ButtonDropdown v-bind="$props" />',
+})
+
+# Button Dropdown
+
+Dropdowns are toggleable lists of links or actions which are displayed by clicking on them.
+
+### Description
```
cnotv, line 20: Could you please add also about the `hash`:
- brief information about the functionality
- requirements: `vue-router` or `Nuxt`

cnotv (changes requested):
Good job in adding other related stories! 💯 
Please mind to mention it inside the PR as defined in the template.
Same for other out of the scope changes which reviewers must be aware about.

stories/components/Tab.stories.mdx
cnotv: As we do not have currently this case, we opted to exclude it.

shell/components/Tabbed/index.vue
```diff
@@ -161,12 +160,19 @@ export default {
         return;
       }
 
-      if (this.useHash && routeHash !== hashName) {
-        const kurrentRoute = { ...currentRoute };
+      if ( this.useHash ) {
```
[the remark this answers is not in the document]
cnotv: We should not add reference to Storybook, also the statement is not correct. I'll suggest you something like this"

```js
/**
 * Exclude logic with URL anchor (hash) for projects without routing logic (vue-router)
 */
```

cnotv (approved):
LGTM

cnotv (approved):
Issue with the hash seems fixed.

LGTM

Issue #7608, opened 2022-12-02
Growlmanager Storybook breaks


scures:
After merging rancher/dashboard#7451, seems that the store broke.

```
Object.assign(...).registerModule is not a function
```

Maybe a good solution would be to continue with the unified store until we remove the store dependency for the component.

catherineluse:
If it's merged, I'd say you should remove the to triage label and add the review label

PR #7609, opened 2022-12-02
fix: unifies the store so it won't break under some specific env

scures:
### Summary
Fixes https://github.com/rancher/dashboard/issues/7608

There was an issue where in some environments the store would not work as expected.

Ex: It would work locally but not deployed.
image

### Technical notes summary

Merges the `growl` store into the global store, this should get rid of the error.


cnotv (approved):
LGTM

Issue #7464, opened 2022-11-16
Add troubleshooting sections and further guidelines to documentation sections


cnotv:
# Description

As outcome of several faced complexities, add a troubleshooting section within [unit tests](https://rancher.github.io/dashboard/testing/unit-test) and [Storybook](https://rancher.github.io/dashboard/storybook) section in the DevKit and extend the existing guidelines.

### Unit tests

- Selecting elements and use of `data-testid` over classes as separation of concerns
- Namespaced states mocks 
- Module state mocks
- Vue mapping functions combined to modules, e.g. `mapGetters('type-map', ['activeProducts']),`
- Vm linting issues 
- Elements outside of the wrapper, e.g. appended to the document or outside the component ([`vue-select` example handled case](https://github.com/sagalbot/vue-select/blob/master/tests/unit/Dropdown.spec.js))
- Object get/set 
- Directives 
- Mapped computed properties to the state or getters
- Slots (beside documentation)
  - Simple stub
  - Multiple cases
  - Destructured slots options (case Vue2 and Vue3) with multiple attributes
- Dispatched actions with mocked state 
- Debounce  and computed/method cases with Vue
- Case sensitive import in unit tests 
- Debugging setup 
- Debug views without components with stubs or `shalloMount()`
- View components with select/checkbox events
- Mock imports
- Complex views, e.g. cluster config editor
- Nuxt `$fetch`
- Async functions
  - Errors: it crashes Jest
  - How to await for Vue methods
- Class models
- Document functions, e.g. `document.queryAll()`
- Vue `$el()`
- Jest unexpected token (ESM mismatch, e.g. use of conditional chaining in Vue templates)
- Read only properties

### Storybook

- Migrating i18n text in `@rancher/components`
- Require jsdoc comments
- Define default controls using Vue2 MDX, by default are all set automatically
- Action binding to the existing examples, currently never triggered in any story
- Example with Vuex integration, by default we have `i18n` mandatory
- Example with parent component holding data 
- Global and component styling (SCSS)
- `<nuxt-link>` integration (e.g. `DetailTop`)
- Multiple slots

### Setup

- Build issues in the monorepo (episode with `rancher/icons`)

### Typescript

- Extending JS classes
- Extending JS mixins 

### Generic

- Some global styling are applied also to other packages, e.g. the case with the headers and Harvester

aalves08:
@cnotv A couple of items more for the list

[...]

Issue #7119, opened 2022-10-07
Add support for Pod Security Admission


nwmac:
Documentation: https://kubernetes.io/docs/concepts/security/pod-security-admission/
Epic: https://github.com/rancher/rancher/issues/38701

# Issues

Grouped features
- https://github.com/rancher/dashboard/issues/7617
- https://github.com/rancher/dashboard/issues/7677
- https://github.com/rancher/dashboard/issues/7628
- https://github.com/rancher/dashboard/issues/7575
- https://github.com/rancher/dashboard/issues/7875
- https://github.com/rancher/dashboard/issues/7999

Smaller UI iterations
- https://github.com/rancher/dashboard/issues/7597
- https://github.com/rancher/dashboard/issues/7583
- https://github.com/rancher/dashboard/issues/7584
- ~https://github.com/rancher/dashboard/issues/7684~ Done in PR

# Implementations

## 1. Namespace creation/edit UI

Create new tab 'Pod Security Admission' to be shown before 'Labels & annotations' that allows user to configure the PSA for the namespace.

This would have 3 checkboxes to enable enforce, audit, warn - for each when checked, dropdown to chose policy `privileged`, `baseline`, or `restricted` and an input box to enter a version (defaults to text 'latest').

The UI will take this form and add/update the labels required for these settings.

On the labels and annotations tab, if there are PSA labels, we will by default hide these but add a button/link to allow the user to see them in the labels list - see the mock for labels further down. Note the mock shows the system labels being readonly, but I think we should allow them to be edited. 

UX design for tab contents:

![image](https://user-images.githubusercontent.com/1955897/199688475-abd0c19a-52e1-4bb7-81d6-b49f0a55f387.png)

Note: The defaults for the select box should be 'Baseline' - the mock up above shows the three possibilities.

## 2. Project creation/edit UI

As above - user can use the same UI but for Project create/edit. The backend will take care of sync'ing the labels from the project to the namespaces in that project

## 3. Project/Namespace list

When a namespace has PSA labels, we want to show an icon (lock) next to the namespace name - on hover shows the PSA config for that namespace - e.g. "Enforce Privileged (version), Audit Baseline (latest)" (as a vertical list of the up to 3 possibilities)

## 4. Namespace Detail View

[...]

PR #7619, opened 2022-12-05
Add support for Pod Security Admission in the Namespace resource

### LINKED ISSUE #7617: Add support for Pod Security Admission in the Namespace resource
# Description

As outcome of https://github.com/rancher/dashboard/issues/7119, add mentioned features:
1. Namespace creation/edit UI
2. ~Project creation/edit UI (optional)~ Not for this initial implementation
3. Project/Namespace list
4. Namespace Detail View

## Context

The ability to define label is already allowed in the current state, without the need to have anything from the server side defined. In this way we could already proceed in integrating this functionality.

cnotv:
### Summary
Fixes #7617
This PR introduces Pod Security Admission (PSA) for Namespaces and affect changes in several views.

- In Namespace and Project editors, added PSA tab with smaller PSA form
- In Namespace and Project list, added icon with tooltip for PSA resources, if any
- In Namespace details, added label for PSA

### Technical notes summary

Tests involving checkbox and select input components have been avoided due complication in identify the implementation of it. Case added to the [troubleshooting list](https://github.com/rancher/dashboard/issues/7464).

See issue for the details of each:
- Namespace CRUD

### Screenshot/Video

https://user-images.githubusercontent.com/5009481/205892111-8893ac0a-8ee1-41be-bd51-0196f9648b70.mov

Labels after PR update:

Screenshot 2022-12-16 at 14 42 42


shell/models/pod.security.admission.js
```diff
@@ -0,0 +1,2 @@
+// export default class PodSecurityPolicy {
```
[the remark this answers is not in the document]
cnotv: Not for this implementation. Removed.

shell/components/__tests__/PodSecurityAdmission.test.ts
```diff
@@ -0,0 +1,195 @@
...
+        mode: 'create',
+        labels
+      }
+    });
+
+    expect(wrapper.vm.controls).toStrictEqual(controls);
+  });
+
+  it.each([
+    // ['true', 'active'],
```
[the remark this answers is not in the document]
cnotv: We needed as we find why it's not possible to use other inputs. Suggestions are appreciated :P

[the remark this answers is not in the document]
cnotv: I add a `TODO` connected to the troubleshooting.

shell/components/__tests__/PodSecurityAdmission.test.ts
```diff
@@ -0,0 +1,195 @@
...
+      }
+    });
+
+    const input = wrapper.find(`[data-testid="pod-security-admission--0-${ inputId }"]`).find('input').element as HTMLInputElement;
+
+    expect(input.value).toStrictEqual(value);
+  });
+
+  it.each([
+    // ['true', 'active'],
```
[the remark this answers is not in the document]
cnotv: Added TODO.

richard-cox (changes requested):
There's some places we could use the generic framework rather than specific PSA code, I need to come back a mark these out and include info how it can be done

shell/components/LabelsPSA.vue
```diff
@@ -0,0 +1,72 @@
...
+  },
+  data() {
+    return { toggler: true };
+  },
+  computed: {
+    /**
+     * Generate list of present keys which can be filtered based on existing label keys and system keys
+     */
+    protectedKeys() {
+      return getPSALabels(this.value);
```
richard-cox, line 29: This one can be moved to the model for the value type as well

shell/components/LabelsPSA.vue
```diff
@@ -0,0 +1,72 @@
+<script>
```
richard-cox, line 1: Optional - I haven't seen this running yet, but low bar solution would be to just pass the `mode` through to the component's of the PodSecurityAdmission component. If the mode is something not editable the controls will then be disabled.

shell/components/PodSecurityAdmission.vue
```diff
@@ -0,0 +1,225 @@
...
+     */
+    mode: {
+      type:     String,
+      required: true
+    },
+
+    /**
+     * Conditionally show users restrictions
+     */
+    hasExceptions: {
```
richard-cox, line 54: Couldn't see where this is wired in, missing a TODO or can it be removed?

shell/components/ResourceDetail/index.vue
```diff
@@ -374,6 +378,8 @@ export default {
       <DetailTop
         v-if="isView && isDetail"
         :value="liveModel"
+        :tooltips="tooltipsDescription"
```
richard-cox, line 381: `tooltipsDescription` should be a fn in the root `resource-class` that returned undefined (or not and we check for undefined) and have an override in the model for this resource type

```
liveModel?.detailTopToolTip

get detailTopToolTip() { ... }
```

shell/edit/namespace.vue
```diff
@@ -224,14 +230,24 @@ export default {
       <Tab
         v-if="!isView"
         name="labels-and-annotations"
-        label-key="generic.labelsAndAnnotations"
+        label-key="generic.labels"
+        :show-header="false"
         :weight="-1"
       >
-        <Labels
```
richard-cox, line 230: Users should still be able to edit labels and annotations for a namespace

richard-cox (changes requested):
I didn't give much of the functionality a test but did notice there's an error when looking at the detail page of a namespace that had the new lock icon.

Once review comments are addressed will double check how it all runs

shell/components/DetailTop.vue
```diff
@@ -97,6 +99,10 @@ export default {
       return this.value?.filteredSystemLabels;
     },
 
+    filteredLabels() {
```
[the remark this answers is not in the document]
cnotv: The component is full of logic on its own on manipulating the `value`, do you want to refactor it now?

shell/components/ExplorerProjectsNamespaces.vue
```diff
@@ -235,6 +237,21 @@ export default {
     }
   },
   methods: {
+    /**
+     * Check if resource contains PSA labels
+     * @param {*} row HciNamespace
+     */
+    hasPSA(row) {
+      return hasPSALabels(row);
```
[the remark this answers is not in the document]
cnotv: You may want to use the function outside of the model, so you don't have to import everything.
For instance, you are in this case into the namespace, not in the PSA. How do you deal with that?

shell/edit/namespace.vue
cnotv: Do we have 2 types of labels then? @kwwii could you please clarify this point?

shell/components/ExplorerProjectsNamespaces.vue
```diff
@@ -382,32 +399,45 @@ export default {
...
+          class="text-muted"
         >&ndash;</span>
       </template>
-      <template
-        v-for="project in projectsWithoutNamespaces"
-        v-slot:[slotName(project)]
+      <template #cell:name="{row}">
+        <!-- /c/local/explorer/namespace/local -->
+        <n-link
+          :to="`/c/local/explorer/namespace/${row.id}`"
```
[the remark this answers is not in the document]
cnotv: Where should `row.detailLocation` be written specifically?

shell/components/LabelsPSA.vue
cnotv: How can I access to the PSA here? This resource (`value`) is a Namespace in this case, so it has not the method. Or are you asking me to write it twice for the Namespace and the PSA?

shell/components/LabelsPSA.vue
cnotv: Only some keys are disabled, not all.
All the keys are put together as all the labels are together as well.
Please refer to the design.

shell/components/ResourceDetail/index.vue
cnotv: This is for the Namespace.
The same function is used also later in the PSA.
We need 2 functions in 2 different models, where we are not needing the class.
Again we need just the utility function here, so there's no gain is moving it to the model.

cnotv:
@richard-cox as help to understand the needs for the utilities, please rely on the new PR for the resource itself.

https://github.com/rancher/dashboard/pull/7678

shell/edit/namespace.vue
kwwii: Yes, essentially this brings in a new type of label - a "system" label which can be shown or hidden by the user. A "system" label is a label that is used by Rancher itself to perform specific actions, in comparison to "user" labels which can be set by the user for use outside of the Rancher specific scope.

cnotv: What is the required action then? Should we keep `Label and Admission` or not?

kwwii: I assume you mean Annotations? If so, yes, there still needs to be a possibility for a user to enter a label or annotation that is not a PSA (or other possible system labels in the future).

cnotv: Ok but in these `Labels and Annotation` the user will see exactly the same `Labels` and on top of it the `Annotation`, so what's the sense of our custom design?

richard-cox: The `Labels and Annotations` tab should still be shown. It's `Labels` component needs updating to have a button (situational given that it's only needed for PSA world?) to reveal the system labels (PSAs)

https://github.com/rancher/dashboard/issues/7119#issuecomment-1292511032

shell/components/DetailTop.vue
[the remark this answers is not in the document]
richard-cox: The component generically handles input from `value` and converts it into a specific format that it needs to display the information. Moving type specific code into the model will maintain that generic pattern

shell/components/ExplorerProjectsNamespaces.vue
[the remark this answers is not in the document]
richard-cox: `shell/models/namespace.js` 

```
  get hasPSALabels() {
    return hasPSALabels(row); 
  }
```

Though i think we can remove the need to use it outside of the models

shell/edit/namespace.vue
richard-cox: Ok, i've seen it running again. The only thing visually missing here is the `Annotations`. The current `Labels` component which currently covers both labels and annotations should probably be beefed up to optionally include the `show system label` toggle.

shell/components/LabelsPSA.vue
richard-cox: The getter method can be added to the namespace model. This is the only place where `getPSALabels` is used

shell/components/LabelsPSA.vue
richard-cox: I've seen the design now, most of the conversation is more relevant there (given that we can hopefully use the existing `Labels` component

shell/components/ResourceDetail/index.vue
richard-cox: The same explanation given for keeping `DetailTop` generic apply here. There should be no type specific code in generic components. 

I can't see where these generic functions that rely on labels are used outside of namespaces, either here or inhttps://github.com/rancher/dashboard/pull/7678/files. Even if they are used in both they can still either use the same generic function or be in their on `BasePSA` class that the other classes inherit.

shell/components/LabelsPSA.vue
cnotv: This has been also discussed and we cannot use the existing `Labels` component, as this configuration is exceptional to Namespace only. Your option was already evaluated as first approach, but the component has nothing relevant to be reused and is rather complicating the view. I opted for keeping the tabs separated to avoid issues.

shell/components/DetailTop.vue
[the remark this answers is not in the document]
cnotv: The `value` computation from the model is already inside the component, which is not abstracted, I am not so sure to follow.
To avoid labels of certain types is part of the `value` computation like the others. Why is this an exception?

shell/components/ExplorerProjectsNamespaces.vue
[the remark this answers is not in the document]
cnotv: As we have a parameter, I'm force to turn into a simple method. 

Screenshot 2022-12-12 at 15 45 04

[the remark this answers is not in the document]
cnotv: As we have a parameter, I'm force to turn into a simple method. 

Screenshot 2022-12-12 at 15 45 04

shell/components/DetailTop.vue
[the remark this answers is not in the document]
richard-cox: `DetailTop` is a generic component, it has no reference to type specific code (PSALabelsNamespaceVersion) only generic methods exposed by models/classes. 

The model for namespaces can be updated to supply the type specific information that is accessed by the generic component in a generic way. Depending on how the Labels component works this can be the `get labels` getter in namespaces. If you still need the original labels list for other components, it looks like you can add a `get filteredSystemLabels` to namespaces that will contain the PSA label filtering

shell/components/ExplorerProjectsNamespaces.vue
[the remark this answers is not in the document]
richard-cox: Sorry, typo above. The `row` is the instance available on the model. so it would be...

```
  get hasPSALabels() {
    return hasPSALabels(this); 
  }
```

shell/edit/namespace.vue
cnotv: It's not the same, we are not using that logic somewhere else.

cnotv:
@richard-cox It seems convenient to discuss about restructuring the whole architecture in [an issue](https://github.com/rancher/dashboard/issues/7684) where we can tackle down all the other generated problems caused by moving tests and TS inside the models. We will need a full perspective of all the use cases, so we do not omit or overcomplicate anything.

As this causes no harm, I'll opt to focus on the functionalities, e.g. solve the issue with the lack of Annotations from your new requirements.

It would be also constructive to [write a documentation with the enforced guidelines](https://github.com/rancher/dashboard/issues/7682) of all these points, so we do not have them dispersed. [The point from you mentioned do not seems present in the guidelines](https://rancher.github.io/dashboard/code-base-works/api-resources-and-schemas#models).

I have also created an issue where we could separately [make some analysis to eventual issues in adopting TS and unit tests](https://github.com/rancher/dashboard/issues/7683), so it may be simpler to point it out.

shell/components/ResourceDetail/index.vue
richard-cox: The three functions getPSALabels, hasPSALabels, getPSATooltipsDescription only apply to namespaces. PodSecurityAdmissionConfigurationTemplate work with a different object.

The description in question here only applies to namespaces.

Keeping resource type code out of generic components is what improves (or does not make worse) the code. There are no other cases where we use resource type code in generic components.

This does not prevent the use of TS or creating simple tests (models are easily created).

cnotv:
@richard-cox ok.

shell/edit/namespace.vue
cnotv: After talking with @kwwii there are the changes to be applied:
- We can merge the 2 labels components
- Correct the Label title size and space to match the tab
- Border to be removed
- Toggler can be hidden if no system labels
- Description is kept

This new style is applied to all the `Labels and Annotations` tabs, which means that not having system labels, it will not have toggler and no disabled inputs.

cnotv: Updated and added screenshot.

cnotv:
@richard-cox If you want to enforce the use of the models, please help me out with the migration to Typescript in https://github.com/rancher/dashboard/issues/7683 so I don't have to delete my work, as there were no issues as standalone implementation.
I am not able to override the key `metadata` inside the class Namespace, as it would not match with the inherited values. Any plan/suggestion?
You have also the PR assigned to you, so you can check what I did and the issue in re-assigning `metadata` breaking somewhere for some reasons: https://github.com/rancher/dashboard/pull/7726

cnotv:
@richard-cox As alternative we may refer to the helpers from the classes, then when we complete [the model conversion to TS](https://github.com/rancher/dashboard/issues/7683) we can remove them.

shell/components/DetailTop.vue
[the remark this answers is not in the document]
cnotv: I'm trying to understand how to move the config into the model, but:
- We have already a config for the edit mode on top
- Labels is not a model, just a list from the value
- The value is not always namespace, so if I move the configuration it's not always present

What should I do in this case?

shell/components/LabelsPSA.vue
cnotv: Moved within model.

shell/components/ResourceDetail/index.vue
cnotv: Moved within model.

richard-cox: Resolved by https://github.com/cnotv/dashboard/pull/7

richard-cox (changes requested):
I've resolved the remaining comments, fixed one bug and improved some visual elements in https://github.com/cnotv/dashboard/pull/7. Once that's merged to this branch's PR this can go in

shell/components/LabelsPSA.vue
richard-cox: Resolved by https://github.com/cnotv/dashboard/pull/7

cnotv:
@richard-cox PR merged, still needs approval.

cnotv:
I am getting this error after merging your PR @richard-cox 
Screenshot 2023-01-06 at 11 16 40

It seems that the `value` model does not have the method `getPSALabels()` on page loaded.
This happens for all the other labels which do not use the Namespace model.

cnotv:
@richard-cox seen the bug fixed with the PSA labels. Forgot to pull the merged PR.

richard-cox (dismissed):
I've pushed one final commit (see commit message for more details, but roughly... it fixes something related but not caused by this pr, removes usages of lodash and aligns sort order of PSAs in ns list and ns edit). 

If those look good to you then feel free to merge

cnotv:
@richard-cox you still have to approve.

richard-cox (approved):

PR #7641, opened 2022-12-07
Update Chinese translation

vickyhella:


vickyhella:
The word `Cores` is changed to `CPU` as shown in the picture below, which is modified to fix the issue mentioned in #6331 for the English page.
image

PR #7642, opened 2022-12-07
[rancher-components] fix lint warnings

### LINKED ISSUE #7596: [pkg/rancher-components] resolve linter warnings
# Description

`yarn lint:lib` returns some linter warnings during CI runs.

As mention to: https://github.com/rancher/dashboard/pull/7578#pullrequestreview-1198958523

torchiaf:

### Summary
Fixes #7596

### DIFF
--- a/pkg/rancher-components/src/components/Form/Checkbox/Checkbox.vue
+++ b/pkg/rancher-components/src/components/Form/Checkbox/Checkbox.vue
@@ -2,6 +2,7 @@
 import Vue, { PropType } from 'vue';
 import { _EDIT, _VIEW } from '@shell/config/query-params';
 import { addObject, removeObject } from '@shell/utils/array';
+import cloneDeep from 'lodash/cloneDeep';
 
 export default Vue.extend({
   props: {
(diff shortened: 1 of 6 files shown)


pkg/rancher-components/src/components/Form/Checkbox/Checkbox.vue
```diff
@@ -114,6 +114,9 @@ export default Vue.extend({
   },
 
   computed: {
+    modelValue() {
```
rak-phillip, line 117: A data prop would be more appropriate for `modelValue` if we intend to mutate it via `v-model`.

We should consider computed props to be getters. We can provide a setter if we need it, but I don't see the benefit in this scenario

https://v2.vuejs.org/v2/guide/computed.html#Computed-Setter

rak-phillip: Explicitly specifying the getter still doesn't fix the issue at hand. We are still mutating `modelValue` without a setter as well as mutating the `value` prop. The computed prop implementation works, but just because we are mutating `value` in a roundabout way and the vue template compiler isn't picking up on it. 

I don't think we need to be tracking state locally at all. I hope you don't mind, but I took a closer look and ended up amending this PR so that we could address the root issue.

PR #7644, opened 2022-12-07
import machine drivers for machine pools from extensions

aalves08:
- import machine drivers for machine pools from extensions


aalves08:
closed as work done by Nancy is so much better 🙏

PR #7651, opened 2022-12-08
Do not always instrument code

### LINKED ISSUE #7649: Code is always being instrumented
Code is always being instrumented for code coverage, regardless of whether we are building for test.

nwmac:
Fixes #7649


package.json
```diff
@@ -40,9 +40,9 @@
     "cy:open": "cypress open",
     "cy:run": "cypress run --browser chrome",
     "cy:run:sorry": "./scripts/e2e",
-    "e2e:pre-dev": "yarn docker:local:stop && yarn docker:local:start && NODE_ENV=dev yarn build",
+    "e2e:pre-dev": "yarn docker:local:stop && yarn docker:local:start && NODE_ENV=dev INSTRUMENT=true yarn build",
```
cnotv, line 43: As mentioned, let's prefix the env var so we know the area and avoid to overlap with others. For E2E we use a generic `TEST_`, so it would be something like `TEST_INSTRUMENT`.

cnotv: Sorry to bother you with this but, we should definitely extend this information to the existing documentation in the [E2E coverage sections](https://rancher.github.io/dashboard/testing/e2e-test#coverage)

cnotv (changes requested):
Sorry but there's these 2 points for completeness.

nwmac:
@cnotv Updated

cnotv (approved):
LGTM
LGTM

PR #7663, opened 2022-12-09
defaults to stable release if the build is a custom/internal build

scures:
### Summary
Fixes # https://github.com/epinio/ui/issues/145

There was an issue with the tests failing because they couldn't download the files from git given that was an internal build of the Epinio BE.

This PR introduces a check for a matching semantic release versioning and defaults to `v1.5.1` when it detects a custom build.
### Screenshot/Video

Small test:
image

### DIFF
--- a/pkg/epinio/index.ts
+++ b/pkg/epinio/index.ts
@@ -5,17 +5,22 @@ import epinioMgmtStore from './store/epinio-mgmt-store';
 import epinioRoutes from './routing/epinio-routing';
 import { MANAGEMENT } from '@shell/config/types';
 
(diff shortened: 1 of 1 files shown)


richard-cox (approved):
This kind of works, kind of doesn't

```
const junk = 'junk';
const semanticVersionRegex = /v(?:(\d+)\.)?(?:(\d+)\.)?(?:(\d+)\.\d+)/;

const version = junk.match(semanticVersionRegex) ? 'v1.5.1' : junk;

console.warn('version', version);

-->

version junk
```

If this is just to cut our their specific format, and it works for that, lets get it in

PR #7665, opened 2022-12-09
Namespaces filter, select on keydown Enter

### LINKED ISSUE #7048: Search box - Arrows up & down then Select on Enter
Rancher 2.6.8

Context Search box:

image

IMHO it would be great to select options with up & down arrows and select them by press Enter.
Currently Enter button has no effect.

torchiaf:

### Summary
Fixes #7048

### DIFF
--- a/shell/components/nav/NamespaceFilter.vue
+++ b/shell/components/nav/NamespaceFilter.vue
@@ -444,7 +444,7 @@ export default {
         e.preventDefault();
         e.stopPropagation();
         this.up();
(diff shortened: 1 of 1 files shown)


aalves08 (approved):
LGTM. One comment to add here (not related to the scope of this task) is that the focus (?) effect on the selected option doesn't change the font and icon colors, making them the same color as the bg and therefore invisible. This is noticeable when navigating through the namespace filter options with the arrow keys

Screenshot 2022-12-19 at 09 46 26
Screenshot 2022-12-19 at 09 46 29

PR #7670, opened 2022-12-09
clean steve fields from yaml view and download

mantis-toboggan-md:
This PR updates yaml viewing and downloading across the board, so it's worth checking a few different resources. It should also be tested as both an admin and a standard user with access to only downstream clusters. The following actions are potentially impacted:
* view yaml
* edit yaml (note that edit config->as yaml is different)
* download yaml
* download yaml in bulk

### DIFF
--- a/shell/components/ResourceDetail/index.vue
+++ b/shell/components/ResourceDetail/index.vue
@@ -26,7 +26,8 @@ function modeFor(route) {
   }
 }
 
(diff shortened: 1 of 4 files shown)


richard-cox (approved):
I've tested this against a legacy backend and it's safe to merge. Though QA should hold back on testing until the backend changes merge

Issue #7668, opened 2022-02-11
Chart Repository refresh does not actually refresh the chart repo


jamesongithub:
**Rancher Server Setup**
- Rancher version: v2.6.3
- Installation option (Docker install/Helm Chart):
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc):
- Proxy/Cert Details:

**Information about the Cluster**
- Kubernetes version: 1.21.7
- Cluster Type (Local/Downstream): local on AKS
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider):

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom) Admin
  - If custom, define the set of permissions:

   
**Describe the bug**

The refresh button 

![image](https://user-images.githubusercontent.com/4452738/153684059-bb67d22d-5def-4edb-a1b7-1798a244b08d.png)

![image](https://user-images.githubusercontent.com/4452738/153684067-ff6559f8-822d-4186-9df1-3454f5151b52.png)

does not actually sync the latest changes from the repo to rancher.

**To Reproduce**
 
1. create github repo with apps/marketplace rancher chart plus questions.yaml
1. create a branch on the repo
1. add repo url to rancher as a git repo pointing to branch
1. install chart
1. make a change to the chart and/or questions.yaml in the github repo branch
1. refresh the repo in rancher
1. attempt to deploy chart again

**Result**
Chart is deployed with old version

**Expected Result**

Chart to be deployed with latest changes on the branch after a refresh

**Screenshots**

**Additional context**

The only way is the create a new repo on rancher with a different name. Deleting and creating a repo with the same name does not fix the issue, you still have the old, cached version of the chart.

Is there even a automated periodic refresh at all?

catherineluse:
Related tech debt issue https://github.com/rancher/dashboard/issues/7963

PR #7672, opened 2022-12-09
This fixes an issue where catalog/charts were stale after a repo was updated

codyrancher:
### Summary
The catalog.cattle.io.clusterrepos/{repo} resource doesn't subscribe to updates. We now force a full request in the areas where it's appropriate to do so.

fixes rancher/dashboard#7668

### Testing
If you want me to modify the repo for you just let me know. If you want to test it yourself the following notes should be helpful.

To test this you'll need a chart repo that you can modify. I created a simple repo at `https://github.com/CodyrancherTest/charts/tree/master`.  
To modify do the following:
- fork that repo
- modify the values.yaml (I just change the logLevel)
- run `helm package  .` within the root of the directory
- Add, commit and push all of the changes to your github repo

### DIFF
--- a/shell/detail/catalog.cattle.io.app.vue
+++ b/shell/detail/catalog.cattle.io.app.vue
@@ -30,7 +30,7 @@ export default {
   },
 
   async fetch() {
-    await this.$store.dispatch('catalog/load');
+    await this.$store.dispatch('catalog/load', { force: true });
 
     this.allOperations = await this.$store.dispatch('cluster/findAll', { type: CATALOG.OPERATION });
   },
--- a/shell/mixins/chart.js
+++ b/shell/mixins/chart.js
@@ -242,7 +242,7 @@ export default {
 
   methods: {
     async fetchChart() {
-      await this.$store.dispatch('catalog/load');
+      await this.$store.dispatch('catalog/load', { force: true });
 
       if ( this.query.appNamespace && this.query.appName ) {
         // First check the URL query for an app name and namespace.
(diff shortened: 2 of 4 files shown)


mantis-toboggan-md (approved):

AndrewHoffmanQA:
@codyrancher I will gladly take you up on the offer to modify the repro, would you be interested in walking through it on a call? 

 I have a couple questions about the order of operations for this bug and repro steps on this work item vs the actual ticket and would benefit from the KT in general.

catherineluse:
@codyrancher I'm working on a related issue and have a question. Since we are now force fetching all chart repos on all app detail pages, does that make the "Refresh" button redundant? (The button in the linked issue on the git repositories list page) I'm thinking either the button should be removed, or we should do less force-fetching elsewhere to justify the existence of the button.

codyrancher:
> @codyrancher I'm working on a related issue and have a question. Since we are now force fetching all chart repos on most chart pages, does that make the "Refresh" button redundant? (The button in the linked issue on the git repositories list page) I'm thinking either the button should be removed, or we should do less force-fetching elsewhere to justify the existence of the button and improve performance for setups with a huge number of helm chart apps.

@catherineluse  I don't think the refresh button is redundant. I'm having to force refresh the data because the resource doesn't receive automatic updates over the socket. If you want to do less force fetching for performance reasons I think that's a reasonable goal but I think it will technically be wrong since the data could be stale.

Issue #7182, opened 2022-10-13
[RFE] Enable Azure accelerated networking - RKE2/K3s


gaktive:
Internal reference: SURE-3197
Reported in 2.5.9

Request description:
Node templates should have the option to create VMs in Azure public cloud with accelerated networking enabled.

Here's a Microsoft post about this feature: https://azure.microsoft.com/fr-fr/blog/maximize-your-vm-s-performance-with-accelerated-networking-now-generally-available-for-both-windows-and-linux/

Actual behavior:
It's not an option

Expected behavior:
Should be one

Additional notes from @HarrisonWAffel:

> I've taken a look at what it would take to implement this enhancement and it seems the following changes would be needed. Each change is accompanied by a story point estimate. Each point represents one day of development work, and does not include time needed for code review or QA verification.

### Dashboard Changes
- RKE1 UI would need a new check-box in the Azure node template, in the 'Networking' Section. Enabling this checkbox would enable accelerated networking. 
- RKE2 UI would need a new check box in the cluster creation screen, under the 'Advanced Options' for each node pool.
- The default machine config would need to be updated to include the new option for accelerated networking, it should be named 'acceleratedNetworking' and its default value should be 'false'

(There would be required Machine and Terraform changes too.)

> Some VM options do not support accelerated networking, though most do. The Azure API provides explanatory error messages which are bubbled up to the UI, so a user should be able to self diagnose their configuration if they select an incompatible VM. Azure does not provide a static list of VM's supporting accelerated networking, however they do have documentation on how to determine support using their CLI tool https://learn.microsoft.com/en-us/azure/virtual-network/accelerated-networking-overview#supported-vm-instances.

Connection to rancher/dashboard#7177 or in Zube, rancher/dashboard#7177

catherineluse:
When the UI gets the list of Azure VM sizes, it comes in the format of a list of about 600 strings originating from `https://github.com/rancher/rancher/blob/release/v2.7/pkg/api/norman/customization/aks/listers.go#L376`. Since we are adding the accelerated networking and availability zone options on 2.7.1, I wonder if it would be feasible to have the UI pass query options to that same endpoint in order to filter the VMs by the selected availability zone and whether they support accelerated networking or not, or if it would be better for the UI to bypass that endpoint and call the Azure API separately.

[...]

Issue #7180, opened 2022-10-13
[RFE] Rancher Azure Node Driver, add option to set Availability zone - RKE2/K3s


gaktive:
Internal reference: SURE-3319
Reported in 2.5.9

A user would like to be able to deploy nodes templates with an availability zone on Azure. Currently, the node templates handle only the Availability set.

Request description:
Add Availability zone to Rancher Azure Node Driver.

Actual behavior:
Rancher node driver does not support Azure Availability zones

Expected behavior:
Rancher node driver support for Azure Availability zones

Additional notes:
Rancher Azure node driver and docker node driver support availability set but not availability zones.

While availability sets are used to protect applications from hardware failures within an Azure data center, Availability zones, protect applications from complete Azure data center failures.

Docker Machine driver support option --azure-availability-set but has no option for availability zones. https://docs.docker.com/machine/drivers/

See also "Availability zones" description at: https://cloudacademy.com/course/getting-started-with-azure-virtual-machines-988/availability-sets-and-availability-zones/

Possible connection with rancher/dashboard#7164 or in Zube, rancher/dashboard#7164

gaktive:
More info from @HarrisonWAffel, both in context to proposed overall work and then Dashboard specific work:

> After looking into this issue a bit I believe the following changes will be necessary to support this new feature. All story point estimates represent one day of development work (including developer testing), but does not include time for code review or QA verification. 

### Rancher-machine (2-3pts)

- A new CLI argument, titled --azure-availability-zone 

Azure zones are denoted by integers, and there are a minimum of three zones per region. Each virtual machine can only be assigned to a single zone. As such, only a single integer should be passed to this flag.
Availability Sets cannot be used in conjunction with Availability Zones. The presence of --azure-availability-zone should cause rancher-machine to skip the configuration/creation of an Availability Set. If both are provided, an error should be thrown. 

- A new CLI argument, titled --azure-use-public-ip-standard-sku 

[Only Standard Public IP SKU's allow for the specification of Availability Zones.](https://learn.microsoft.com/en-us/azure/virtual-network/ip-services/public-ip-addresses#sku) Currently, rancher-machine only creates Public IP's with the Basic SKU. This flag must be passed in order for rancher-machine to use the Standard SKU so that Availability Zones work.
The Public IP Address must also be static, dynamic Public IP Addresses are not compatible with Availability Zones.
[...]

PR #7687, opened 2022-12-12
Add accelerated networking and availability zones to RKE2/K3s machine pools

catherineluse:
Addresses https://github.com/rancher/dashboard/issues/7182 and https://github.com/rancher/dashboard/issues/7180

Combining the two options into one PR to avoid merge conflicts

Summary:
 - Added  accelerated networking (AN) to Azure node options in the Advanced section for RKE2/K3s cluster provisioning
 - Added an Availability Zone option for Azure nodes
	 - You can have an Availability Set or Availability Zone, but not both
	 - The Availability Zone dropdown menu may or may not have any options, depending on the selected region. For example, if you select `westus` as the location, the Availability Zone dropdown should be disabled and you should see a warning because that region doesn't support AZs. If you select `westus2` as the location, the Availability Zone dropdown should have 1, 2 and 3 as options.

### Accelerated Networking
- When AN is enabled, the list of VM sizes should show that the VM sizes that don't support it are disabled.
- When AN is enabled, the VM size dropdown should have a tooltip explaining why some options are disabled.
- If you have AN disabled, then select a VM that does not support AN (such as `BasicA0`), then try enable AN, you should see an error that says the VM size is incompatible with AN. A warning also shows next to the VM size dropdown. The warning goes away if any other non-disabled VM is selected or if AN is turned off.

### Availability Zones
- Can select an AZ for a machine pool
- Warn the user when the selected VM size is not in the selected region
- Warn the user if they have selected availability zones, but that is not supported for the region or VM size

### DIFF
--- a/shell/machine-config/azure.vue
+++ b/shell/machine-config/azure.vue
@@ -13,6 +13,7 @@ import { randomStr } from '@shell/utils/string';
 import { addParam, addParams } from '@shell/utils/url';
 import KeyValue from '@shell/components/form/KeyValue';
 import { RadioGroup } from '@components/Form/Radio';
+import { _CREATE } from '@shell/config/query-params';
 
 export const azureEnvironments = [
   { value: 'AzurePublicCloud' },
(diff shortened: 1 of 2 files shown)


catherineluse:
@mantis-toboggan-md This is ready for review since it's going to be in the next medium sized release after all. Note: The accelerated networking boolean is only expected to be persisted in today's version of v2.7-head due to a backend change today

catherineluse:
Previously I had misunderstood - I thought this PR needed to use the original endpoint for VM sizes, but it needed to use V2. Sorry for the confusion on my part. I've updated it to use the V2 endpoint tested it locally using the latest 2.7-head. I believe it's working now.

shell/machine-config/azure.vue
```diff
@@ -177,19 +199,192 @@ export default {
...
+    },
+    selectedVmSizeSupportsAN() {
+      const selectedSizeIsValid = !!this.vmsWithAcceleratedNetworking.find((vmData) => {
+        return this.value.size === vmData.Name;
+      });
+
+      return selectedSizeIsValid;
+    },
+    vmSizeAcceleratedNetworkingWarning() {
+      if (!this.selectedVmSizeSupportsAN) {
```
mantis-toboggan-md, line 243: I am seeing the accelerated networking warning even when I haven't checked that box. I think this needs a `&& this.value.acceleratedNetworking`

shell/machine-config/azure.vue
```diff
@@ -177,19 +199,192 @@ export default {
...
+    },
+    vmAvailabilityZoneWarning() {
+      if (this.useAvailabilitySet) {
+        return '';
+      }
+      if (this.vmsWithAvailabilityZones.length === 0) {
+        return this.t('cluster.machineConfig.azure.size.regionDoesNotSupportAzs');
+      }
+
+      if (this.vmsWithAvailabilityZones > 0 && !this.selectedVmSizeHasZones) {
```
mantis-toboggan-md, line 286: I am never seeing the regionSupportsAzsButNotThisSize warning: it looks like this needs a `.length`

mantis-toboggan-md (changes requested):
2 minor fixes and but otherwise this PR looks good

mantis-toboggan-md (approved):

shell/machine-config/azure.vue
catherineluse: I've made this change and the error is showing properly now:
Screenshot 2022-12-16 at 2 53 06 PM

shell/machine-config/azure.vue
catherineluse: I think that's left over from when the checkbox was disabled to explain why it was disabled. But then I took out the disabled property to prevent the weird state where the checkbox could be both checked and disabled in edit mode. I've updated it so that the error appears when the box is checked but it's not supported

catherineluse:
I've updated it based on your comments and made a couple more small fixes as well. The VM size no longer disappears from the dropdown menu if the region is switched to a region where that size does not exist. And when the region is switched, a loading spinner shows instead of the VM size list while it is being fetched.

mantis-toboggan-md (dismissed):

mantis-toboggan-md (approved):

PR #7695, opened 2022-12-13
Add explore button to cluster detail view

### LINKED ISSUE #7694: Make it easier to explore a cluster from the cluster detail view
In Cluster Management, you can explore a cluster from the list view:

image

.. but if you click on a cluster, there is no quick way to explore - you have to go back.

We should add a button to this view to reduce clicks needed to explore a cluster.

nwmac:
Fixes #7694

This PR:

- Extends the detail masthead to support a single action that can be provided by the resource model to be shown as a button in the detail masthead
- Updates the cluster model to provide an action to be shown in the masthead which will take the user to the cluster

image

### DIFF
--- a/shell/components/ResourceDetail/Masthead.vue
+++ b/shell/components/ResourceDetail/Masthead.vue
@@ -125,6 +125,10 @@ export default {
       return null;
     },
 
+    detailsAction() {
+      return this.value?.detailsAction;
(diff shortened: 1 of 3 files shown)


aalves08 (approved):
LGTM. Testing looks fine. 👍

PR #7701, opened 2022-12-14
HARVESTER: Cluster field is not really *required* when creating RKE2/K3s Harvester

n313893254:
### Summary
Fixes Cluster field is not really *required* when creating RKE2/K3s Harvester

- https://github.com/harvester/harvester/issues/3165

### Screenshot/Video

image

### DIFF
--- a/shell/cloud-credential/harvester.vue
+++ b/shell/cloud-credential/harvester.vue
@@ -78,7 +78,28 @@ export default {
       this.$nuxt.$loading.finish();
 
       this.value.setData('kubeconfigContent', kubeconfigContent);
-    }
+    },
+
+    'value.decodedData.clusterId': {
+      handler() {
+        this.emitValidation();
+      },
+      immediate: true,
+    },
+
+    isImportCluster: {
+      handler() {
+        this.emitValidation();
+      },
+      immediate: true,
+    },
+
+    'value.decodedData.kubeconfigContent': {
+      handler() {
+        this.emitValidation();
+      },
+      immediate: true,
+    },
   },
 
   methods: {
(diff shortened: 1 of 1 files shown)


n313893254:
Hi, @mantis-toboggan-md, PR ready, PTAL, Thx. cc @richard-cox

shell/cloud-credential/harvester.vue
```diff
@@ -78,7 +78,40 @@ export default {
       this.$nuxt.$loading.finish();
 
       this.value.setData('kubeconfigContent', kubeconfigContent);
-    }
+    },
+
+    'value.decodedData.clusterId': {
+      handler() {
+        if (this.test() === true) {
```
mantis-toboggan-md, line 85: I think because these three handlers are identical they should probably refer to one function defined elsewhere eg a method 
```
emitValidation()
  if (this.test() === true) {
    this.$emit('validationChanged', true);
  } else {
    this.$emit('validationChanged', false);
  }
```

Then the handlers for `value.decodedData.clusterId` `isImportCluster` and `value.decodedData.kubeconfigContent` would look like:

```
 handler() {
   this.emitValidation();
 }
```

mantis-toboggan-md (changes requested):
Overall works and looks good, I just had one minor suggestion around code duplication.

shell/cloud-credential/harvester.vue
n313893254: Done. Thanks for your review. Good suggestion!

mantis-toboggan-md (approved):
LGTM

PR #7708, opened 2022-12-14
Fix: Also show RKE1/RKE2 toggle on cluster creation screen if all node drivers are deactivated

### LINKED ISSUE #7705: Cluster Creation: RKE1/RKE2 toggle is not visible if no node driver is active
**Setup**
- Rancher version: 2.7.0

**Describe the bug**
On the cluster creation page the RKE1/RKE2 toggle is not visible if there is no node driver activated (all node drivers deactivated.
This can be the case if a user only wants to allow custom clusters.

**To Reproduce**
* Go to Cluster Manger -> Drivers -> Node Drivers
* Deactivate all node drivers
* Create a new cluster

**Result**
There is a "Custom cluster" button, but no RKE1/RKE2 toggle.

**Expected Result**
There is a RKE1/RKE2 toggle visible.

**Screenshots**
Bildschirm­foto 2022-12-14 um 16 02 24

bashofmann:
### Summary
This fixes the cluster creation page to also show a RKE1/RKE2 toggle if all node drivers are deactivated.

Fixes #7705

This changes the `v-if` to decide if the toggle is shown on one of the headlines, to either show it at the 2nd item (node driver clusters or custom clusters) or at the 1st item (custom clusters, if both all node drivers and cluster drivers are deactivated).

Cluster creation screen with different variants of having cluster or node drivers activated. The toggle should always be visible.

Cluster creation screen.

### Screenshot/Video
Bildschirm­foto 2022-12-14 um 17 23 47
Bildschirm­foto 2022-12-14 um 17 23 38
Bildschirm­foto 2022-12-14 um 17 23 30
Bildschirm­foto 2022-12-14 um 17 23 18

### DIFF
--- a/shell/edit/provisioning.cattle.io.cluster/index.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/index.vue
@@ -205,10 +205,6 @@ export default {
 
     rke2Enabled: mapFeature(RKE2_FEATURE),
 
-    showRkeToggle() {
-      return this.rke2Enabled && !this.isImport;
-    },
-
     provisioner: {
       get() {
         // This can incorrectly return rke1 instead
@@ -352,9 +348,29 @@ export default {
 
       return sortBy(Object.values(out), 'sort');
     },
+
+    firstNodeDriverItem() {
+      return this.groupedSubTypes.findIndex(obj => [_RKE1, _RKE2].includes(obj.name));
+    },
+
+    firstCustomClusterItem() {
+      return this.groupedSubTypes.findIndex(obj => ['custom', 'custom1', 'custom2'].includes(obj.name));
+    },
   },
 
   methods: {
+    showRkeToggle(i) {
+      if (this.isImport || !this.rke2Enabled) {
+        return false;
+      }
+
+      if (this.firstNodeDriverItem >= 0) {
+        return i === this.firstNodeDriverItem;
+      }
+
+      return i === this.firstCustomClusterItem;
+    },
+
     loadStylesheet(url, id) {
       if ( !id ) {
         console.error('loadStylesheet called without an id'); // eslint-disable-line no-console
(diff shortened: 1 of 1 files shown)


rak-phillip (commented):
This change looks good but I'm unable to validate against v2.7-head because of #7762

shell/edit/provisioning.cattle.io.cluster/index.vue
```diff
@@ -355,6 +351,20 @@ export default {
   },
 
   methods: {
+    showRkeToggle(groupedSubTypes, i) {
```
rak-phillip, line 354: I think it would be better to rely on the `groupedSubTypes` computed prop that's declared above.

shell/edit/provisioning.cattle.io.cluster/index.vue
```diff
@@ -355,6 +351,20 @@ export default {
...
+      if (this.isImport || !this.rke2Enabled) {
+        return false;
+      }
+      const nodeDriverItem = groupedSubTypes.findIndex(obj => [_RKE1, _RKE2].includes(obj.name));
+      const customClusterItem = groupedSubTypes.findIndex(obj => ['custom', 'custom1', 'custom2'].includes(obj.name));
```
rak-phillip, line 359: This is more of a note for something to consider; we can make `nodeDriverItem` and `CustomClusterItem` computed props instead of finding the index each time `showRkeToggle` is invoked. I don't foresee the number of grouped subtypes growing to become exceedingly large, so the impact of making such a change is low.

rak-phillip (changes requested):
I think we can change the `showRkeToggle` method a little bit. Otherwise, LGTM

bashofmann:
I updated the method.

I was able to test this by deactivating each driver individually.

rak-phillip (approved):
Thanks for the changes, looks great 😄

PR #7733, opened 2022-12-16
Only show explore button when in detail view

nwmac:
Bug fix for #7695

Made spacing consistent:

image

or:

image

### DIFF
--- a/shell/components/ResourceDetail/Masthead.vue
+++ b/shell/components/ResourceDetail/Masthead.vue
@@ -556,7 +556,7 @@ export default {
   div.actions-container > div.actions {
     display: flex;
     flex-direction: row;
-    justify-content: end;
+    justify-content: flex-end;
   }
 
 </style>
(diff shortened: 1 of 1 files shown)


shell/components/ResourceDetail/Masthead.vue
```diff
@@ -435,7 +435,7 @@ export default {
         <div class="actions-container">
           <div class="actions">
             <button
-              v-if="detailsAction"
+              v-if="detailsAction && currentView === 'detail' && isView"
```
richard-cox, line 438: `'detail'` --> the const `_DETAIL` (need to pass it through to template via `data`)

richard-cox, line 438: Possibly controversial request, can we match the padding?

image

image

This would apply everywhere, but i think it looks alright?

image

richard-cox (changes requested):

shell/components/ResourceDetail/Masthead.vue
nwmac: Done

nwmac: Done

nwmac:
@richard-cox Updated as per your feedback

richard-cox (approved):

PR #7735, opened 2022-12-16
Fix issue with project name not editable

### LINKED ISSUE #7179: Unable to edit Project Name in Rancher 2.6 UI
Internal reference: SURE-5027
Reported in 2.6.4

Request description:
User is unable to edit Project name (new and existing Projects). This is supported in Rancher v2.5, but does not appear to be supported in Rancher v2.6. Looks to be a dropped or missing feature in terms of UI operations.

Actual behavior:
Project Name field in the form is greyed out, i.e. not available for edit.

Expected behavior:
Project name field should be editable by an authorized user.

Workaround:
A Project name can be modified using `kubectl` (projects.management.cattle.io) through the `spec.displayName` field.

View in 2.5.x:

View in 2.6.x:

nwmac:
Fixes #7179 

Original fix did not set the name, so I don't know how that every worked or was tested? @Shavindra 

This PR:

- Ensures the name is updated
- Forces a fetch of the updated resource, to avoid an issue where we update the Norman resource and when you return to the list page, the name does not reflect the change made - we force a fetch of the corresponding Steve resource

### DIFF
--- a/shell/edit/management.cattle.io.project.vue
+++ b/shell/edit/management.cattle.io.project.vue
@@ -152,7 +152,7 @@ export default {
               type: MANAGEMENT.PROJECT,
               id:   this.value.id,
               opt:  { force: true }
-             });
+            });
           }
 
           // // we allow users with permissions for projectroletemplatebindings to be able to manage members on projects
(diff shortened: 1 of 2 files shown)


aalves08 (approved):
testing and code LGTM

aalves08 (approved):
LGTM

PR #7764, opened 2022-12-19
re-add lost cluster.explore translation

### LINKED ISSUE #7763: Cluster management 'Explore' button is missing translation when the cluster is provisioning

mantis-toboggan-md:
Fixes #7763

We lost the `cluster.explore` translation string recently. Interestingly, the disabled button shown while a cluster is not ready was referencing a different key than the n-link used when the cluster can be explored. I opted to remove the `exploreHarvester` version because it's misleading when it is only applied to non-harvester clusters.


cnotv (approved):
LGTM

PR #7768, opened 2022-12-20
fix filter namespace

WuJun2016:
### Summary
1. There is no parameter to receive `getters` in mutations (`updateNamespaces`).  I think it may be unnecessary.
2. In `getActiveNamespaces`, I wonder if we should filter based on `allNamespaces`.    

Fixes #

### DIFF
--- a/shell/store/index.js
+++ b/shell/store/index.js
@@ -81,7 +81,13 @@ const getActiveNamespaces = (state, getters) => {
     return out;
   }
 
-  const namespaces = getters[`${ inStore }/all`](NAMESPACE);
+  let namespaces = [];
+
+  if (Array.isArray(state.allNamespaces) && state.allNamespaces.length > 0) {
+    namespaces = state.allNamespaces;
+  } else {
+    namespaces = getters[`${ inStore }/all`](NAMESPACE);
+  }
 
   const filters = state.namespaceFilters.filter(x => !!x && !`${ x }`.startsWith(NAMESPACED_PREFIX));
   const includeAll = getters.isAllNamespaces;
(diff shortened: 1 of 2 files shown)


shell/store/index.js
```diff
@@ -81,7 +81,13 @@ const getActiveNamespaces = (state, getters) => {
     return out;
   }
 
-  const namespaces = getters[`${ inStore }/all`](NAMESPACE);
+  let namespaces = [];
+
+  if (Array.isArray(state.allNamespaces) && state.allNamespaces.length > 0) {
```
mantis-toboggan-md, line 86: I think removing getters from updateNamespaces makes sense but I'm less sure about this change. Is there some benefit to using `allNamespaces` instead of `getters[`${ inStore }/all`](NAMESPACE)` ? It looks like it would work but I don't understand the motivation for the change.

I'm also curious as to why `allNamespaces` exists at all. I see how it is set, but (aside from the change in this PR) I don't see where it is accessed. @catherineluse and @richard-cox have more experience with namespace filtering, maybe they can clarify.

WuJun2016: `allNamespace` can be controlled by different products. For example, harvester does not need to display the system namespace.

If we use `getters[' ${inStore}/all '](NAMESPACE)`, then when the user clicks `All Namespaces`, `getters.isAllNamespaces` is always `true`, All namespaces (including the same namespace) will be displayed。

mantis-toboggan-md: Okay, thank you for explaining. I think this change is ok then

mantis-toboggan-md (approved):

PR #7771, opened 2022-12-20
Storybook: Ensure we install dependencies at top level

### LINKED ISSUE #7770: Solve Storybook deployment pipeline errors
# Description

Storybook deployment seems not working since a while.
https://github.com/rancher/dashboard/actions/workflows/storybook.yaml?page=3

## Context

For some reasons the check is passing. https://github.com/rancher/dashboard/actions/workflows/storybook-check.yaml
On top of solving the issue, we should ensure that the executed code is the same.

It may be convenient to:
- Merge [check](https://github.com/rancher/dashboard/blob/master/.github/workflows/storybook-check.yaml) and [publish](https://github.com/rancher/dashboard/blob/master/.github/workflows/storybook.yaml) workflows into one
- Skip publication if PR: `if: ${{ github.event_name == 'pull_request' }}` ([docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsif))
- Require build before publication
- Update publication script for Storybook

nwmac:
Fixes #7770 

We were not installing the dependencies at the top-level when publishing.

Merged build and publish pipelines into one, with publish only called when pushing to master.


.github/workflows/storybook.yaml
```diff
@@ -1,19 +1,28 @@
...
       uses: actions/setup-node@v1
       with:
         node-version: '14.x'
-    - name: script
+    - name: Install packages
+      run: yarn install
+    - name: Build Storybook
+      run: |
+        yarn build-storybook
+    - name: Publish Storybook
```
cnotv, line 26: Could we cache the build and publish conditionally this one, so we could merge the 2 jobs into one? It would be easier to track.

cnotv: Yes but in the issue I've written that it would be better to have both the workflows in the same file.
In that way we will have the same steps and cannot risk to fail on `master`.
The publish part can be turned into another job, that's why we could cache it and pass to the other job. It will also help to understand where it does fail.

Alternatively we could add the condition for the step, but the name for the workflow/job would be confusing (`storybook/publish`)

cnotv (approved):
LGTM

PR #7772, opened 2022-12-20
Prompt node pool

### LINKED ISSUE #6961: Prompt user to confirm node deletion when scaling down node pool in GUI
Enhancement

In a cluster created with a node template, navigate to the nodes page and click on the minus ("-") button next to a node pool.

**Actual Result:**
A node is immediately deleted upon clicking the button.

**Expected Result:**
There should be a confirmation prompt displayed to the user to prevent accidental deletion of nodes.
Note: When selecting "Delete" from the context menu of a node, such confirmation prompt is displayed indeed.

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
v2.1.5
- Installation option (single install/HA):
HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
Infrastructure Provider

bisht-richa:
### Summary
Fixes #6961

Added a confirmation dialog when the user pressed the minus icon on a cluster detail page 

### How to test

Screenshot 2022-12-20 at 16 06 13

Screenshot 2022-12-20 at 16 13 37

Go to Cluster Management -> [cluster] 
- When the user clicks on the minus icon, it should show a confirmation dialog.
- The dialog should also show a checkbox below the first line and it should be checked by default we need to remember this (only within the user's browser session ) 
- If the checkbox is checked by default, the next time when the user scales it down, the confirmation dialog will NOT prompt again.
- If the checkbox is not checked, the confirmation dialog will prompt again.
- Added confirmation prompt setting in pref page

Screenshot 2023-01-11 at 13 34 50

### DIFF
--- a/shell/detail/provisioning.cattle.io.cluster.vue
+++ b/shell/detail/provisioning.cattle.io.cluster.vue
@@ -693,7 +693,7 @@ export default {
                     :disabled="!group.ref.canScaleDownPool()"
                     type="button"
                     class="btn btn-sm role-secondary"
-                    @click="group.ref.scalePool(-1)"
+                    @click="group.ref.toggleScaleDownModal()"
                   >
                     <i class="icon icon-sm icon-minus" />
                   </button>
--- a/shell/detail/provisioning.cattle.io.cluster.vue
+++ b/shell/detail/provisioning.cattle.io.cluster.vue
@@ -180,9 +180,7 @@ export default {
     }
   },
 
-  data( { $cookies } ) {
-    const scalePoolPrompt = $cookies.get('scalePoolPrompt');
-
+  data() {
     return {
 
       allMachines:           [],
@@ -206,8 +204,7 @@ export default {
       logSocket: null,
       logs:      [],
 
-      showWindowsWarning:  false,
-      showScalePoolPrompt: scalePoolPrompt
+      showWindowsWarning: false
     };
   },
 
@@ -696,7 +693,7 @@ export default {
                     :disabled="!group.ref.canScaleDownPool()"
                     type="button"
                     class="btn btn-sm role-secondary"
-                    @click="group.ref.toggleScaleDownModal($event, showScalePoolPrompt)"
+                    @click="group.ref.toggleScaleDownModal($event)"
                   >
                     <i class="icon icon-sm icon-minus" />
                   </button>
--- a/shell/detail/provisioning.cattle.io.cluster.vue
+++ b/shell/detail/provisioning.cattle.io.cluster.vue
@@ -180,7 +180,9 @@ export default {
     }
   },
 
-  data() {
+  data( { $cookies } ) {
+    const scalePoolPrompt = $cookies.get('scalePoolPrompt');
+
     return {
 
       allMachines:           [],
@@ -204,7 +206,8 @@ export default {
       logSocket: null,
       logs:      [],
 
-      showWindowsWarning: false
+      showWindowsWarning:  false,
+      showScalePoolPrompt: scalePoolPrompt
     };
   },
 
@@ -693,7 +696,7 @@ export default {
                     :disabled="!group.ref.canScaleDownPool()"
                     type="button"
                     class="btn btn-sm role-secondary"
-                    @click="group.ref.toggleScaleDownModal()"
+                    @click="group.ref.toggleScaleDownModal($event, showScalePoolPrompt)"
                   >
                     <i class="icon icon-sm icon-minus" />
                   </button>
--- a/shell/dialog/ScalePoolDownDialog.vue
+++ b/shell/dialog/ScalePoolDownDialog.vue
@@ -24,7 +24,9 @@ export default {
   },
   computed: {
     machinenName() {
-      return this.resources[0].id.split('/')[1];
+      const name = this.resources.length > 0 ? this.resources[0].id.split('/')[1] : '';
+
+      return name;
     },
     protip() {
       return this.t('promptRemove.protip', { alternateLabel });
--- a/shell/dialog/ScalePoolDownDialog.vue
+++ b/shell/dialog/ScalePoolDownDialog.vue
@@ -30,9 +30,11 @@ export default {
       return this.t('promptRemove.protip', { alternateLabel });
     },
   },
-  mounted() {
+  created() {
     const showScalePoolPrompt = this.$cookies.get('scalePoolPrompt');
 
+    // Check for showScalePoolPrompt cookies
+    // If undefined set cookies and update promt checkbox
     if ( showScalePoolPrompt === undefined ) {
       this.promptConfirmation = true;
       this.$cookies.set('scalePoolPrompt', true);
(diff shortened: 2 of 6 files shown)


cnotv:
@rak-phillip Looks like we have restrictions about the budgets of the library bundle, which makes fail the build validation. Should we handle in a separated issue?
https://webpack.js.org/configuration/performance/

```
entrypoint size limit: The following entrypoint(s) combined asset size exceeds the recommended limit (244 KiB). This can impact web performance.
Entrypoints:
  @rancher/components.umd.min (338 KiB)
      @rancher/components.umd.min.js


 warning  

webpack performance recommendations: 
You can limit the size of your bundles by using import() or require.ensure to lazy load some parts of your application.
For more info visit https://webpack.js.org/guides/code-splitting/

  File                                   Size              Gzipped

  dist/@rancher/components.umd.min.js    338.45 KiB        107.65 KiB
  dist/@rancher/components.umd.js        1389.25 KiB       305.46 KiB
  dist/@rancher/components.common.js     1388.76 KiB       305.29 KiB

  Images and other types of assets omitted.
  ```

**Comment by rak-phillip** [comment]
@cnotv we have #5931 existing that we can add details to.

We've had the CI step for `@rancher/components` merged for 3 weeks. Why would the bundle size fail PRs now?

**Comment by cnotv** [comment]
> We've had the CI step for `@rancher/components` merged for 3 weeks (#7578). Why would the bundle size fail CI checks now?

That's definitely an interesting question. There must be something weird going on with the validation workflow. A second run seems to have been fixing the issue.

File: shell/dialog/ScalePoolDownDialog.vue [hunk 1]
```diff
@@ -0,0 +1,118 @@
...
+  },
+  computed: {
+    machinenName() {
+      return this.resources[0].id.split('/')[1];
+    },
+    protip() {
+      return this.t('promptRemove.protip', { alternateLabel });
+    },
+  },
+  mounted() {
```
cnotv, line 33: This can be done on `created` as well.

shell/dialog/ScalePoolDownDialog.vue
```diff
@@ -0,0 +1,118 @@
...
+    }
+  },
+
+  methods: {
+    close() {
+      this.$emit('close');
+    },
+
+    update(e) {
+      if (this.promptConfirmation) {
```
cnotv, line 50: The value `promptConfirmation` is already a boolean, so you dont need a condition logic. If the value may be nullish, then you can enforce it with `Boolean()` or double bangs (`!!`). 
Also you are not using the event, so you can avoid passing it.

E.g.:

```ts
update() {
  this.$cookies.set('scalePoolPrompt', !!this.promptConfirmation);
```

shell/dialog/ScalePoolDownDialog.vue
```diff
@@ -0,0 +1,118 @@
...
+        this.$cookies.set('scalePoolPrompt', true);
+      } else {
+        this.$cookies.set('scalePoolPrompt', false);
+      }
+    },
+
+    remove() {
+      // Delete pool
+      this.resources[0].scalePool(-1);
```
cnotv, line 59: The resource may be empty and therefore you cannot call the method.

shell/dialog/ScalePoolDownDialog.vue
```diff
@@ -0,0 +1,118 @@
...
+
+    return {
+      allToDelete,
+      type:               this.$store.getters['type-map/labelFor'](allToDelete[0].schema, allToDelete.length),
+      promptConfirmation: true
+    };
+  },
+  computed: {
+    machinenName() {
+      return this.resources[0].id.split('/')[1];
```
cnotv, line 27: The resource may be empty and therefore you cannot call the method.

shell/models/cluster.x-k8s.io.machinedeployment.js
```diff
@@ -100,6 +101,21 @@ export default class CapiMachineDeployment extends SteveModel {
     return machinePools.find(pool => pool.machineConfigRef.name === machineConfigName);
   }
 
+  toggleScaleDownModal( event, showScalePoolPrompt, resources = this ) {
+    const alt = isAlternate(event);
```
cnotv, line 105: Please add a comment about we check if the user keep pressed they key `Cmd/Ctrl`

The logic is already confusing as it is, since the function is called `Alt` but the key is not.

shell/models/cluster.x-k8s.io.machinedeployment.js
```diff
@@ -100,6 +101,21 @@ export default class CapiMachineDeployment extends SteveModel {
...
+
+    if (!alt && !showScalePoolPrompt) {
+      this.$dispatch('promptModal', {
+        component:  'ScalePoolDownDialog',
+        resources,
+        modalWidth: '450px'
+      });
+    } else {
+      // User held alt key, so don't prompt
+      this.scalePool(-1);
```
cnotv, line 115: Is there not a way to pass this function to the modal as callback? It's inconvenient to define the logic twice.

cnotv (changes requested):
Please add code documentation where you create functions.

shell/detail/provisioning.cattle.io.cluster.vue
```diff
@@ -180,7 +180,9 @@ export default {
     }
   },
 
-  data() {
+  data( { $cookies } ) {
+    const scalePoolPrompt = $cookies.get('scalePoolPrompt');
```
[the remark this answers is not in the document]
bisht-richa: Yes.

shell/models/cluster.x-k8s.io.machinedeployment.js
bisht-richa: This will check if the user has pressed `Alt` while deleting pod in that case we will not show model

cnotv: Yes I know, but this logic is repeated in another file. Is it not possible to pass in the modal the function and define it once only?

shell/detail/provisioning.cattle.io.cluster.vue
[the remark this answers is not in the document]
cnotv: Marking as unresolved, as we discussed that we rely on user preferences if we are logged in.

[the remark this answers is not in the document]
cnotv: You could replace the cookie method with 

```
// to get the value
this.$store.getters['prefs/get'](SCALE_POOL_PROMPT)`, 

// to set the value (for the code below)
this.$store.dispatch('prefs/set', { key: SCALE_POOL_PROMPT, value: !!this.promptConfirmation });
```

Keep in mind that you need to add `SCALE_POOL_PROMPT` as config inside `'@shell/store/prefs';`

shell/models/cluster.x-k8s.io.machinedeployment.js
cnotv: Since I've seen that it's a common patter, I set this to resolved.

cnotv (approved):
Prompt seems fine.

LGTM

PR #7787, opened 2022-12-21
Add support for Pod Security Admission in the Clusters and Workloads

### LINKED ISSUE #7628: Add support for Pod Security Admission in the Clusters and Workloads
# Description

As outcome of https://github.com/rancher/dashboard/issues/7119, add mentioned features:
5. Workload view (e.g. detail view for Pod, Deployment etc)
8. Cluster Creation RKE2

## Context

As second part for the PSA implementation, some updates are required from the server side.

## Requirements

The new versions will be listed with https://github.com/rancher/rancher/issues/39795.
This is required to access `serverArgs` and `agentArgs` responsible for all the inputs and tabs configuration.

cnotv:
### Summary
Fixes #7628,

- Navigation
  - Hidden PSP resource link in cluster manager
  - Added PSA resource link in cluster manager
- RKE2 cluster editor
  - Added dropdown for PSA for k8s 1.23+
  - Hidden PSP dropdown for k8s 1.25+
  - ~Replaced the label `RKE Default` with `(None)`~ Used `(None)` if no PSA is selected
  - Corrected reprovisioning banner hyperlink
- Workloads
  - Added Banner with icon for PSA errors, removed stacked view to balance with case lacking it

### Technical notes summary
- RKE2 cluster editor
  - Retrieved PSA templates and added in the select input
  - Added initial tests for RKE2 cluster editor
  - Added comments for each tab due collapse view
  - Added `haveArgInfo` i18n text
- Workloads
  - Extended `CruResource` component to map errors into objects with messages and icons
  - Added PSA RegEx computation to replace message and add icon to existing message

- Cluster creation for k8s 1.23.x, 1.24.x, 1.25.x with PSP and PSA
- Errors in Pod creation with namespace restriction
- Navigation

### Screenshot/Video

Cluster creation

https://user-images.githubusercontent.com/5009481/212337024-f487134a-8cbc-4efc-8c83-cf06e0a204b0.mov

Pod error banner

https://user-images.githubusercontent.com/5009481/212338223-f2ae67f6-39d7-492d-a980-ffc9d21ffdc4.mov

Switch to YAML

https://user-images.githubusercontent.com/5009481/212339519-f3d583c9-f3f6-4502-9f9e-41dfb25c316b.mov

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -1732,7 +1732,7 @@ cluster:
       header: Security
     defaultPodSecurityPolicyTemplateName:
       label: Default Pod Security Policy
-    defaultPodSecurityAdmissionConfigurationTemplateId:
+    defaultPodSecurityAdmissionConfigurationTemplateName:
       label: Default Pod Security Admission
       option: RKE2 Default
     enableNetworkPolicy:
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -1563,6 +1563,7 @@ cluster:
     rke2-k3-reprovisioning: 'Making changes to cluster configuration may result in nodes reprovisioning. For more information see the <a target="blank" href="{docsBase}/cluster-provisioning/rke-clusters/behavior-differences-between-rke1-and-rke2/" target="_blank" rel="noopener nofollow">documentation</a>.'
     desiredNodeGroupWarning: There are 0 nodes available to run the cluster agent. The cluster will not become active until at least one node is available.
     invalidPsps: You have one or more PodSecurityPolicy resource(s) in this cluster. Pod Security Policies are not available in Kubernetes v1.25.
+    haveArgInfo: Configuration information is not available for the selected Kubernetes version.  The options available in this screen will be limited, you may want to use the YAML editor.
 
   rkeTemplateUpgrade: Template revision {name} available for upgrade
 
@@ -1731,8 +1732,9 @@ cluster:
       header: Security
     defaultPodSecurityPolicyTemplateName:
       label: Default Pod Security Policy
-    defaultPodSecurityAdmissionTemplateName:
+    defaultPodSecurityAdmissionConfigurationTemplateId:
       label: Default Pod Security Admission
+      option: RKE2 Default
     enableNetworkPolicy:
       label: Project Network Isolation
       warning: Per default, the ingress controller will not be able to route requests to pods on other nodes.
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -1732,9 +1732,12 @@ cluster:
       header: Security
     defaultPodSecurityPolicyTemplateName:
       label: Default Pod Security Policy
+      option: (None)
     defaultPodSecurityAdmissionConfigurationTemplateName:
       label: Default Pod Security Admission
-      option: RKE2 Default
+      option: (None)
+    cisProfile:
+      option: (None)
     enableNetworkPolicy:
       label: Project Network Isolation
       warning: Per default, the ingress controller will not be able to route requests to pods on other nodes.
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -1732,7 +1732,7 @@ cluster:
       header: Security
     defaultPodSecurityPolicyTemplateName:
       label: Default Pod Security Policy
-      option: (None)
+      option: RKE Default
     defaultPodSecurityAdmissionConfigurationTemplateName:
       label: Default Pod Security Admission
       option: (None)
--- a/shell/edit/provisioning.cattle.io.cluster/rke2.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/rke2.vue
@@ -75,6 +75,8 @@ const ADVANCED = 'advanced';
 const HARVESTER = 'harvester';
 const HARVESTER_CLOUD_PROVIDER = 'harvester-cloud-provider';
 
+const noneOption = { label: '(None)', value: '' };
+
 export default {
   components: {
     ACE,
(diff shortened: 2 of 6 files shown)


shell/edit/provisioning.cattle.io.cluster/rke2.vue
```diff
@@ -1926,16 +2020,40 @@ export default {
           <h3>
             {{ t('cluster.rke2.security.header') }}
           </h3>
-          <div class="row">
+
+          <div
+            v-if="psaOptions && needsPSA"
```
mantis-toboggan-md, line 2025: It looks like `psaOptions` is always defined so I think this could just be `v-if="needsPSA"` (nitpicky of me though)

Should this dropdown be shown when the current user can't load PSAs? Standard users can provision clusters, but can't see PSAs, so they just get a dropdown with nothing but `(None)`. Maybe they should be allowed to add a user-defined value here? Idk

shell/assets/translations/en-us.yaml
```diff
@@ -1731,8 +1732,9 @@ cluster:
       header: Security
     defaultPodSecurityPolicyTemplateName:
       label: Default Pod Security Policy
-    defaultPodSecurityAdmissionTemplateName:
+    defaultPodSecurityAdmissionConfigurationTemplateName:
       label: Default Pod Security Admission
+      option: RKE2 Default
```
mantis-toboggan-md, line 1737: This doesn't appear to be used

shell/edit/provisioning.cattle.io.cluster/rke2.vue
```diff
@@ -75,6 +75,8 @@ const ADVANCED = 'advanced';
 const HARVESTER = 'harvester';
 const HARVESTER_CLOUD_PROVIDER = 'harvester-cloud-provider';
 
+const noneOption = { label: '(None)', value: '' };
```
mantis-toboggan-md, line 78: The label here should probably be localized

shell/edit/workload/index.vue
```diff
@@ -28,6 +28,40 @@ export default {
...
+        return {
+          message: `Pod "${ name }" Security Policy Violation "${ policy }"`,
+          icon:    'icon-pod_security'
+        };
+      }
+
+      default:
+        break;
+      }
+    },
```
mantis-toboggan-md, line 52: Is this dependent on https://github.com/rancher/dashboard/pull/7918? I don't see how to create a workload that will result in an error response due to a security policy violation; the most I've seen is a 201 response with an error header, which looks like what the other PR is for. The other changes to error formatting don't appear to cause issues, I just haven't verified that this part works as intended.

mantis-toboggan-md (changes requested):
I have some small pieces of feedback but overall this seems to work, I think?? I can see that the default PSA config field is set during cluster provisioning and the restrictions around kubernetes versions seem to work correctly, but I don't really know what this option does. I mean, I've set the `defaultPodSecurityAdmissionConfigurationTemplateName` to `restricted` and verified it's saved in the cluster spec but how do I know it's used within the cluster? I can see setting PSA on namespaces works (I tested by making workloads with `allowPrivilegeEscalation: true`), but I thought if I set the default psa config to `restricted` I'd get the same sort of errors on all namespaces that I haven't configured psa for. This is more of a confession of my own ignorance than criticism of the PR

shell/edit/provisioning.cattle.io.cluster/rke2.vue
```diff
@@ -1672,7 +1732,36 @@ export default {
...
-    }
+    },
+
+    /**
+     * Handle k8s changes side effects, like PSP and PSA resets
+     */
+    handleKubernetesChange(value) {
+      if (value) {
+        set(this.value.spec, 'defaultPodSecurityAdmissionConfigurationTemplateName', '');
+        set(this.value.spec, 'defaultPodSecurityPolicyTemplateName', '');
```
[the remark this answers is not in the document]
cnotv: We don't want to have cases where the PSP is not valid for the current version (1.25+) or if we set the PSA.
Since there were no better guidelines to match all the cases, I opted for the simplest and safest option.
Would you recommend to keep it if version is lower than 1.25?

shell/edit/provisioning.cattle.io.cluster/rke2.vue
cnotv: @kwwii was mentioning that we should keep `RKE Default`, although it's unclear when we have both the dropdowns if that's fine.

shell/assets/translations/en-us.yaml
cnotv: Good point. Let's wait for @kwwii feedback about this.

shell/edit/provisioning.cattle.io.cluster/rke2.vue
cnotv: Removed extra condition.

shell/edit/workload/index.vue
cnotv: No these are 2 different things. The error is for restricted, the banner is a growl for warning. In the videos and issue there's the steps to set the PSA to make it fail in the Pod creation.

shell/edit/provisioning.cattle.io.cluster/rke2.vue
cnotv: I restored the default label for PSA and PSP, also added i18n for `profileOptions`.

shell/assets/translations/en-us.yaml
cnotv: I restored the default label for PSA and PSP.

shell/edit/provisioning.cattle.io.cluster/rke2.vue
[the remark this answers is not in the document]
cnotv: This has been agreed with @kwwii to be ok.

shell/assets/translations/en-us.yaml
cnotv: Small update: we finally come out with a decision and went for the option labels `RKE Default` as it was for PSP and `(None)` for PSA.

PR #7848, opened 2023-01-05
ResourceTable, allow Enter event to be propagated

### LINKED ISSUE #7048: Search box - Arrows up & down then Select on Enter

torchiaf:

### Summary
Fixes #7048

### Technical notes summary

`Enter` shortcut is implemented in `SortableTable` component using the `v-shortkey` library.
Having a quick look to the library code, the event propagation is blocked.

This means when users navigate in a tab where the table is used, for example `Jobs` view, the event is captured by the table's shortcuts and it is not propagated to the Namespace's filter on the top.

Go to any tab with SortableTable inside, for example `Jobs` and try to use the Enter shortcut to select a namespace.

### DIFF
--- a/shell/components/ResourceTable.vue
+++ b/shell/components/ResourceTable.vue
@@ -150,6 +150,20 @@ export default {
     }
   },
 
+  mounted() {
+    /**
+     * v-shortkey prevents the event's propagation:
+     * https://github.com/fgr-araujo/vue-shortkey/blob/55d802ea305cadcc2ea970b55a3b8b86c7b44c05/src/index.js#L156-L157
+     *
+     * 'Enter' key press is handled via event listener in order to allow the event propagation
+     */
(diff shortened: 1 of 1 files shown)


shell/components/ResourceTable.vue
```diff
@@ -150,6 +150,14 @@ export default {
     }
   },
 
+  mounted() {
+    window.addEventListener('keyup', this.handleEnterKeyPress);
```
cnotv, line 154: Please add a note as reference to this issue/PR, where the plugin used for the shortcuts does not allow propagation to other components (e.g. Namespace) and we are forced to use the basic JS API.

PR #7852, opened 2023-01-05
Cancel XHR when environment is desktop

rak-phillip:
This allows for cancelling XHR objects when the environment is `desktop`. 

We need a strategy for cancelling XHR via axios. Without one, we end up with a scenario where the UI gets blocked because the connection will never timeout in the absence of a backend. While Dashboard is not designed to allow for rendering any content without a backend, this at least enables us to create a customized template for Rancher Desktop to use.  

I opted to use the deprecated `CancelToken` approach because we won't be able to make use of `AbortController` until we update node for Rancher Dashboard. 

#### Before
Always renders spinner, never progresses beyond registration of nuxt plugins

https://user-images.githubusercontent.com/835961/210877410-fede6c2a-c0d5-48df-a969-c8ca99d2d56e.mp4

#### After
Renders homepage, albeit with no content whatsoever

https://user-images.githubusercontent.com/835961/210877547-b6eec900-bea7-4d40-911d-91a4fcb44518.mp4

### DIFF
--- a/shell/plugins/steve/actions.js
+++ b/shell/plugins/steve/actions.js
@@ -116,8 +116,23 @@ export default {
       }
     }
 
+    function newCancelToken(cancelToken, timeout = 0) {
+      const source = cancelToken.source();
+
+      if (timeout) {
+        setTimeout(() => source.cancel('Operation cancelled by user'), timeout);
+      }
+
+      return source.token;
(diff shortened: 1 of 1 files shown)


rak-phillip:
Closing as stale

PR #7857, opened 2023-01-06
Use property forceYaml to create resources via yaml

DaiYuzeng:
### Summary

Harvester resources may display in other product, we hope user can only create harvester resources via yaml in that case.
Consequently, I exposed the forceYaml property from harvester models.

- https://github.com/harvester/harvester/issues/2184

### DIFF
--- a/shell/components/ResourceDetail/index.vue
+++ b/shell/components/ResourceDetail/index.vue
@@ -149,6 +149,11 @@ export default {
       initialModel = await store.dispatch(`${ inStore }/clone`, { resource: liveModel });
       model = await store.dispatch(`${ inStore }/clone`, { resource: liveModel });
 
+      if (model.forceYaml === true) {
+        as = _YAML;
+        this.as = as;
+      }
+
       if ( as === _YAML ) {
         yaml = createYaml(schemas, resource, data);
       }


mantis-toboggan-md:
I'm a little confused by this and the related harvester/dashboard PR. This issue https://github.com/harvester/harvester/issues/2184 is talking about Harvester clusters imported to Rancher, but the screenshot in https://github.com/harvester/dashboard/pull/559 looks like the standalone Harvester UI?

I don't think we can reasonably support hacking into the explorer product like this:
> Change harvester/harvesterhci.io.dashboard#vm => explorer in the url

In the case of an imported harvester cluster, if a user does that the Harvester plugin would no longer be loaded, right? I would expect that to cause an array of problems when trying to view/create/edit harvester resources.

Regardless I'm not sure this is the right solution: the yaml create/edit view should already load when there is no custom config component registered (a component in in `shell/edit/<type>` or `<loaded harvester plugin>/edit/<type>`); if that's not working right there is an underlying issue.

DaiYuzeng:
Hi @mantis-toboggan-md ! Go into explorer UI from Harvester demanded by some users. They want to observe extra resources in explorer product. It conduce to find out harvester resources issue easily.

In the case of an imported harvester cluster, Harvester plugin won't be loaded, everything works fine. But in the case of standalone Harvester, Harvester plugin will be loaded, custom edit page and custom detail page will be loaded too. Property `forceYaml` can force to create/view Harvester resources via yaml in that case.

mantis-toboggan-md:
> They want to observe extra resources in explorer product. It conduces to find out harvester resources issues easily.

I think there's room for better UX for this in a future release (I know deadlines for 2.7.next1 are close). Could this be accommodated by adding more resources to the Harvester side nav? If there is a technical limitation, something done in Explorer which can't be done in the Harvester extension, we should track that with an issue in rancher/dashboard so we can improve the extension api for a future release. We are intentionally filtering out imported Harvester clusters in the 'Explore Cluster' section of the top level menu (see [here](https://github.com/rancher/dashboard/blob/master/shell/components/nav/TopLevelMenu.vue#L62) and [here](https://github.com/rancher/dashboard/blob/master/shell/utils/cluster.js#L8)) so the explorer isn't developed with the expectation that it's being used for a Harvester cluster; any problems with this use-case are unlikely to be found in a timely manner. 


> But in the case of standalone Harvester, Harvester plugin will be loaded, custom edit page and custom detail page will be loaded too. Property forceYaml can force to create/view Harvester resources via yaml in that case.

Can this be achieved with the `canCustomEdit` model property? eg in model files:
```
get canCustomEdit(){
   return this.$rootGetters['currentProduct'] === HARVESTER
}
```

This property is used by the code that determines what is shown in the action menu and table actions so it would keep the list view from showing 'view config' option at all. More consistent with the other list views.

PR #7858, opened 2023-01-06
E2E Tests: Add logging for env vars

nwmac:
This PR adds logging to the Cypress config file that should notify users when they don't have the env vars set up for E2E tests correctly.

It also adds a convenience `cy:e2e` script that will open Cypress on the e2e tests, to save having to click each time - I'll update the docs for this in a later PR.

It also defaults the username to `admin` - which is what our docs say is the default, but we don't default to this (I think the docs mean that admin is the typical default name, rather than if you don't supply, we'll assume admin).


aalves08 (commented):
Codewise LGTM and after testing it fulfils the requirements. Quite helpful this PR. 👍 

There are some linting errors that need to be fixed before I can approve @nwmac . Thanks

aalves08 (approved):
LGTM

PR #7861, opened 2023-01-06
E2E: Improve reliability of the private label test

nwmac:
This PR improves the reliability of the Branding Private Label tests.

When running locally, the check for the title change would fail for me - I've changed this slightly to improve reliability.

### DIFF
--- a/cypress/e2e/po/components/async-button.po.ts
+++ b/cypress/e2e/po/components/async-button.po.ts
@@ -4,4 +4,19 @@ export default class AsyncButtonPo extends ComponentPo {
   click(): Cypress.Chainable {
     return this.self().click();
   }
+
+  label(name: string): Cypress.Chainable {
+    return this.self().contains(name);
+  }
+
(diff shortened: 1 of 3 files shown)


aalves08 (commented):
Codewise, looks good to me, however after running the tests i've found an issue with the `primaryColorPicker` (line 112) and the `Primary Color` test fails.

Screenshot 2023-01-06 at 15 18 07

I did a quick search on the code for the `data-testid` `primary-color-input` and can't find any reference to it apart from the e2e tests. If it existed it must have been erased.

nwmac:
@aalves08 Should be there  - it is generated though, so you won't find it - i.e. look for `-color-input` and you'll find it in `ColorInput.vue`.

aalves08 (approved):
LGTM. Thanks for the comment @nwmac . You were right and I could also run and pass the tests locally. I think the problem may have been on my side regarding the state the branding was currently at. I cleaned it up and it passed. 🙏

PR #7870, opened 2023-01-09
Cloud Provider, fix label for default value

### LINKED ISSUE #7851: Change rke2 cloud provider verbage
**Description:**
Currently, the Cloud Provider box for RKE2 defaults to `(None)`, which is a bit of a misnomer and has led to many users coming to Slack asking about it. This setting actually supplies no explicit cloud provider for rke2, which means that rke2 will use its own built in cloud provider stub. Then, if a user wants to install their own external cloud provider (vsphere, for example), they try and fail because there is actually already a cloud provider (the built in stub) in the cluster. 
image

**Desire:**
It would be helpful if we changed the language used here. I'm not sure what would be best, but some ideas we have are:
1. `None (RKE2 Embedded)`
2. `RKE2 Embedded`
3. `Default - RKE2 Embedded`

This would just need to change the value that is shown in the UI, not the value that is passed through the API.

torchiaf:

### Summary
Fixes #7851

### Screenshot/Video

image

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -1724,6 +1724,8 @@ cluster:
     cloudProvider:
       label: Cloud Provider
       header: Cloud Provider Config
+      defaultValue:
+        label: Default - RKE2 Embedded
     security:
       header: Security
     defaultPodSecurityPolicyTemplateName:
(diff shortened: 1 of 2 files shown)


cnotv (approved):
LGTM

PR #7903, opened 2023-01-12
Fix stuck loading indicator (auth)

### LINKED ISSUE #7818: Rancher is frozen when User click on 'Auth Provider' button once it is already connected to an Auth
**Setup**
- Rancher version: v2.7-c81b8b45136f3b5d26d41dca9ef0ddcabd2e026c-head
- Modes : Dark and Auto

**Describe the bug**
Rancher is frozen when User click on 'Auth Provider' button once it is already loaded. Window is frozen and it says loading but it never loads. User has to refresh the browser to perform other operations.

**To Reproduce**
1. Login to Rancher
2. Connect to one of the Auth Provider (AZUR)
3. Once it is connected 
4. Click on the 'Auth Provider' button again
5. Check for the system behavior

**Result**
1. Rancher window is grayed out and it says the page is loading but it never loads
2. No other buttons are working to perform any operation
3. User has to refresh the browser or hard refresh with keyboard to see rancher up.

**Expected Result**
Once user click on 'Auth Provider' button it should be on the same screen and it should not disable the screen. User should able to click on other buttons without a browser refresh.

**Note :** I m not seeing Loading page on any other screen navigations OR I am not seeing Loading Page after a page refresh

**Screenshots**

**Additional context**
Seeing the same issue on v2.6.9 also.

nwmac:
Fixes #7818 

The Nuxt loading indicator does not get removed when we prevent the navigation in this use case - this PR forces the loading indicator to be removed.

### DIFF
--- a/shell/pages/c/_cluster/auth/config/_id.vue
+++ b/shell/pages/c/_cluster/auth/config/_id.vue
@@ -19,6 +19,12 @@ export default {
       // Ensure we re-evaluate the redirect in case this auth provider has been disabled
       const authProvs = await authProvidersInfo(this.$store);
 
+      // Nuxt does not remove it's loading indicator - if we are not changing route, then hide it
+      // https://nuxtjs.org/docs/features/loading/
+      if (authProvs.enabledLocation) {
+        this.$nuxt.$loading.finish();
+      }
+
       next(!authProvs.enabledLocation);
     } else {
       next();


nwmac:
@mantis-toboggan-md Sorry about that - tidied up now

mantis-toboggan-md (approved):

PR #7910, opened 2023-01-12
Improve error messages when creating an Azure cloud credential

### LINKED ISSUE #7909: Error message from Azure Credential creation is not shown to user
When creating an Azure cloud credential, the error message is always generic, just saying that auth failed.

We should parse the response back and show an error message to help the user figure out what is wrong.

nwmac:
Fixes #7909 

Adds parsing of the error response when validating Azure cloud credential during creation.

There are two types of error response - one with a string message and the other with a json blob - this PR attempts to parse either.

With this in place, we now present more helpful messages to users to help them figure out what is not right with the data they have entered.

**Subscription ID problem:**

image

**Invalid Client ID:**

image

**Invalid Client Secret:**

image

Previous to this PR, in all cases you just get a generic message indicating that verification failed.


mantis-toboggan-md (approved):
LGTM

PR #7916, opened 2023-01-13
Fix storybook publish error

nwmac:
The Storybook publish task is broken at the moment - it references the `DIR` env var which is not set, so it thinks that the Storybook has not been built.

This PR fixes this - the script runs in the top-level folder by default, so we can just remove the `DIR` env var.


aalves08 (approved):
LGTM

PR #7918, opened 2023-01-13
Feature/7875 workload edit warning

### LINKED ISSUE #7875: PSA Workload edit screen warning
# Description

As outcome of https://github.com/rancher/dashboard/issues/7119 (point "6. Workload edit screen"), add warning inside Workload edit screen, if namespace has PSA as `warn` as `restricted` and security policies are infringed, e.g. set `Privilege Escalation` to `Yes`.

## Context

Add warning growl for workloads based on server specifications https://github.com/rancher/rancher/issues/39413
New design to be defined with https://github.com/rancher/dashboard/issues/7791

cnotv:
### Summary
Fixes #7875

- Added warning growl on PodSecurity violations with `warn:restricted`

### Technical notes summary
- Extended resource model with empty method `processSaveResponse()` to expose the response of the resource requests and allow manipulation; this has been discussed already with @nwmac FYI and peace of mind
- Integrated logic for method `processSaveResponse()`, allowing to parse header warnings and trigger growl if any

Pod creation with Namespace using PSA.

### Screenshot/Video

https://user-images.githubusercontent.com/5009481/212361757-aa0517ab-246e-44ae-9e19-df54f1c20644.mov

https://user-images.githubusercontent.com/5009481/212361785-bcc65cc3-84b0-497e-a39d-ae3853df3bdb.mov

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -2197,6 +2197,9 @@ growl:
   reconnected:
     message: "The connection to {url} was restored on attempt #{tries}."
     title: Websocket Reconnected
+  podSecurity:
+    message: "The creation of this Pod would violate existing restricted policies for the adopted Namespace"
+    title: PodSecurity violation
 
 hpa:
   detail:
--- a/shell/models/pod.js
+++ b/shell/models/pod.js
@@ -181,6 +181,20 @@ export default class Pod extends WorkloadService {
     return 0;
   }
 
+  processSaveResponse(res) {
+    if (res._headers && res._headers.warning) {
+      const warnings = res._headers.warning.split('299') || [];
+      const hasPsaWarnings = warnings.filter(warning => warning.includes('violate PodSecurity')).length;
+
+      if (hasPsaWarnings) {
+        this.$dispatch('growl/warning', {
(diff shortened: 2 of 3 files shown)


mantis-toboggan-md (changes requested):
This works well for pod creation but from the original issue, it sounds like it should apply to workload create/edit as well. It should probably be noted though in the case of workloads, if it is implemented in the same way, the growl will appear when a workload is scaled up/down - not sure if that's particularly helpful.

cnotv:
@mantis-toboggan-md @scures Could we have not fixed the only issue on Harvester??
This is a serious issue in styling to apply them globally, it's going to break somewhere else.

cnotv:
Talking with @nwmac it seems like we are currently sticking to the Pod, although I agree and had the same question now, about why in the issue is mentioning the whole workload.

mantis-toboggan-md (approved):
That's whacky but I believe you. Should there then be an additional issue to track adding a warning for psa violations during workload creation/editing?

cnotv:
> That's whacky but I believe you. Should there then be an additional issue to track adding a warning for psa violations during workload creation/editing?

This still needs to be defined and @nwmac mentioned to discuss after merging this.

PR #7923, opened 2023-01-14
Allow more retries to finish project creation

catherineluse:
This PR addresses this reopened issue https://github.com/rancher/dashboard/issues/6464#issuecomment-1378002783

The issue is only intermittently reproducible, and I believe the reason is that it depends on how many users and rolebindings are in your setup. With more users and rolebindings, it takes more time for a user to get permission to see a new project that they just created. So I was able to repro the issue more easily in Anu's setup than in my own setup.

To test this PR, I confirmed that project creation was successful even after it had to retry. I believe that the reason it didn't work before this PR was that it was failing after trying twice and needed to keep trying.
Screenshot 2023-01-13 at 5 44 47 PM

### DIFF
--- a/shell/models/management.cattle.io.project.js
+++ b/shell/models/management.cattle.io.project.js
@@ -99,18 +99,36 @@ export default class Project extends HybridModel {
     // and the PUT request should have a query param _replace=true.
     const newValue = await norman.save({ replace: forceReplaceOnReq });
 
-    try {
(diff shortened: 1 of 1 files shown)


aalves08 (approved):
Was able to reproduce the issue intermittently and all looks good. Codewise, same. LGTM.

PR #7924, opened 2023-01-14
In RKE2 cluster edit form, avoid showing YAML in form mode

catherineluse:
This PR addresses this bug https://github.com/rancher/dashboard/issues/7876 by only rendering the `ResourceYaml` component if `showYaml` is true.

I followed these steps both to reproduce the bug and to test that it was fixed:

1. Went to list of clusters in Cluster Management
2. Clicked Create
3. Clicked Azure
4. Clicked Edit as YAML
5. Changed `spec.rkeConfig.etcd.snapshotRetention` from 5 to 6
6. Clicked Back to Form
7. Clicked OK

Confirmed that this error is no longer seen:
Screenshot 2023-01-13 at 5 59 07 PM

### DIFF
--- a/shell/components/CruResource.vue
+++ b/shell/components/CruResource.vue
@@ -619,7 +619,7 @@ export default {
       </template>
       <!------ YAML ------>
       <section
-        v-else
+        v-else-if="showYaml"
         class="cru-resource-yaml-container resource-container cru__content"
       >
         <ResourceYaml


aalves08 (commented):
I can still reproduce this issue @catherineluse . Here are my steps:

Went to list of clusters in Cluster Management
Clicked Create
Clicked Digital Ocean
Clicked Edit as YAML
Changed `snapshotScheduleCron` (added some random numbers)
Clicked Back to Form
Clicked OK (no error on first try...)
Clicked again on Edit as YAML
Changed `snapshotScheduleCron` (added some random numbers again)
Clicked Back to Form
Clicked OK (second try I get error)

Screenshot 2023-01-17 at 10 52 01

I did a bit of digging around and found that the error comes from this method
https://github.com/rancher/dashboard/blob/master/shell/utils/create-yaml.js#L76

Which in turn is called from the `rke2.vue` file: 
https://github.com/rancher/dashboard/blob/master/shell/edit/provisioning.cattle.io.cluster/rke2.vue#L1607

Which might be the source of the problem. There are also other components like `CruResource` that call this method.

It needs more investigation 🙏

catherineluse:
@aalves08 You're right. I have made another change that fixes the bug in a different way, and I've updated the PR title and description to reflect the new change.

aalves08 (approved):
LGTM

PR #7934, opened 2023-01-17
Fixed problem with saving cloud credentials for pnap node driver

pajuga:
### Summary
Fixes #
Fixed problem with saving cloud credentials for pnap


richard-cox (approved):
Sorry, i missed this issue in the original review.

This could be simpler by using the `pnap/request` action, but for this release lets keep it simple.

There's some info on `meta/proxy` over at https://rancher.github.io/dashboard/code-base-works/machine-drivers#api-calls. FYI that talks about having to add the domain to a whitelist (<rancher url>/v1/management.cattle.io.settings/whitelist-domain).

Issue #7513, opened 2022-11-21
Adding cluster-id label to new Prometheus Recording rules created from the UI. 


geethub97:
When an alert fires it does not mention from which cluster it comes. I have made the change to include cluster_id to all existing alerts in the monitoring chart, but we need this change to be reflected on the UI for all new alerts. 

Related Ticket:  #https://github.com/rancher/rancher/issues/37128

Request: 
UI add another default label to all rules that to include the cluster_id. 

<img width="1243" alt="Screen Shot 2022-11-21 at 11 31 28 AM" src="https://user-images.githubusercontent.com/99695266/203142741-680997d4-d869-4f26-86cc-060dcc971f73.png">

catherineluse:
@gaktive This should be a quick and easy change.

prachidamle:
The backend ticket dependent on this addition has `2023-Q1-v2.7x` milestone, will it be possible to add in the UI change for the same?

prachidamle:
@gaktive should this be assigned to @catherineluse or anyone?

gaktive:
@prachidamle, we'll leave this open for the taking since if you need this in early, we'll need someone else to take care of this, unless it can wait.

ronhorton:
@Tejeev this ticket is to include the cluster-id as support for the backend work to inlcude the cluster-id. the question regarding cluster-name being included is being tracked in a rancher/rancher ticket: https://github.com/rancher/rancher/issues/37128
the disposition of that ticket will indicate if there's another UI ticket to include that as default in the UI. 
ie, your quesiton is being addressed in the other ticket - this ticket was specifically to include the cluster-id field in the UI.

ronhorton:
**Pass** verified in `v2.7-head commit ID 1f014b7`
`cluster-id` included as default included field in alert config UI

PR #7935, opened 2023-01-17
Add cluster ID label to new alerting rules & recording rules

catherineluse:
This PR addresses https://github.com/rancher/dashboard/issues/7513 by adding the cluster ID to the default labels for new PrometheusRules. It applies to both Recording Rules and Alerting Rules within the PrometheusRule form.

The default alerting rule fields now include cluster ID in the labels:

Screenshot 2023-01-17 at 1 49 11 AM

To test this PR,

1. I installed monitoring on a downstream cluster.
2. Went to the form to create a Prometheus rule
3. Within that form, clicked Add Record
4. Confirmed that the cluster ID label was there by default for the new rule
5. Deleted the recording rule
6. Clicked Add Alert
7. Confirmed that the cluster ID label was there by default for the new rule

### DIFF
--- a/shell/edit/monitoring.coreos.com.prometheusrule/GroupRules.vue
+++ b/shell/edit/monitoring.coreos.com.prometheusrule/GroupRules.vue
@@ -1,6 +1,5 @@
 <script>
 import has from 'lodash/has';
-
 import { Banner } from '@components/Banner';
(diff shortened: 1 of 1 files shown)


Sean-McQ (approved):
LGTM

PR #7948, opened 2023-01-17
re-add header tag global style to accommodate pre-existing extensions

### LINKED ISSUE #7942: [BUG] Older versions of the Harvester plugin have style compatibility issues with the latest version of Rancher
**Describe the bug**

Older versions of the Harvester plugin have style compatibility issues with the latest version of Rancher

**To Reproduce**

1. Install v2.7.0 Rancher and change the settings `ui-dashboard-index` to `https://releases.rancher.com/dashboard/latest/index.html` and `ui-offline-preferred ` to `Remote`
2. Install Harvester and change settings `ui-plugin-index` to `https://releases.rancher.com/harvester-ui/plugin/harvester-v1.1.1/harvester-v1.1.1.umd.min.js` and `ui-source` to `External`
3. Rancher Import Harvester cluster
4. Go the Harvester from Rancher UI and check the VM create page

**Result**

The "Create" button styles have a compatibility issue, displayed on the left.

**Expected Result**

The "Create" button display on the right

**Screenshots**

image

**Additional context**

mantis-toboggan-md:
### Summary
Fixes #7942 and re-opens https://github.com/rancher/dashboard/issues/7341 ... #7942 is a high priority for Harvester and #7341 is a tech debt issue so I think this is a fair trade.

Components in pre-existing Harvester extensions don't have the `header-layout` class and consequently the styles formerly applied to header tags are lost within those extensions. I think https://github.com/rancher/dashboard/issues/7341, the request to remove global styling of tags, will be more tricky to resolve with this limitation in mind. Honestly I can't think of a good solution and am open to suggestions. The only thing I thought of was using a selector like `HEADER:not([class])` in `__layout.scss` which would let us use header tags without those styles applied by adding any class, but it feels pretty hacky.

### DIFF
--- a/shell/assets/styles/global/_layout.scss
+++ b/shell/assets/styles/global/_layout.scss
@@ -11,7 +11,7 @@
   }
 }
 
-.header-layout {
+HEADER {
   display: grid;
   grid-template-areas:
     'type-banner type-banner'
--- a/shell/components/ResourceDetail/Masthead.vue
+++ b/shell/components/ResourceDetail/Masthead.vue
@@ -399,7 +399,7 @@ export default {
 
 <template>
   <div class="masthead">
-    <header class="header-layout">
+    <header>
       <div class="title">
         <div class="primaryheader">
           <h1>
--- a/shell/components/ResourceList/Masthead.vue
+++ b/shell/components/ResourceList/Masthead.vue
@@ -168,7 +168,7 @@ export default {
 </script>
 
 <template>
-  <header class="header-layout">
+  <header>
     <slot name="typeDescription">
       <TypeDescription :resource="resource" />
     </slot>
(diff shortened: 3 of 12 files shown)


mantis-toboggan-md:
> Considering this new information I think reverting the changes is ok. I'm also not sure of a better approach, not sure if @cnotv has some other ideas, we could create a follow-up card to investigate and deal with this scenario?

I reopened the original issue already though I think creating a new one would work just as well if preferred

cnotv:
@mantis-toboggan-md @scures Could we have not fixed the only issue on Harvester?
This is a serious issue in styling to apply them globally, it's going to break somewhere else.

cnotv:
I guess this is even worse if that's the situation. We need to understand what are the constrains and how to get rid of them.

PR #7949, opened 2023-01-17
Namespace filters aren't applied to the API call if force namespace off.

Sean-McQ:
If the option to force namespace filters above a threshold is not enabled, using a single namespace filter should still filter API requests as needed.

### Summary
Fixes #

resource-fetch-namespaced.js not checks if only a single namespace is picked when determining what to pass to __singleNamespaceFilter. nameSpaceFilter watch on the mixin and on the resourcelist now also perform a comparison of neu to old to see if the value has changed. previously, if the new value was false (the filter was cleared) and new API request would not be triggered even though it should be.

### Technical notes summary

resource-fetch-namespaced.js not checks if only a single namespace is picked when determining what to pass to __singleNamespaceFilter. nameSpaceFilter watch on the mixin and on the resourcelist now also perform a comparison of neu to old to see if the value has changed. previously, if the new value was false (the filter was cleared) and new API request would not be triggered even though it should be.

Load namespaced list page with a single namespace filter already in the namespace filter - API request should include a namespace filter
Load namespaced list page with multiple namespace filters already in the namespace filter - API request should not include a namespace filter and the table filtering should filter instead
Go from one namespace filter to none - new API request with no namespace filter
Go from no or multiple namespace filters to just one - No new API request since resources are already in the store

Chrome

### DIFF
--- a/shell/components/ResourceList/index.vue
+++ b/shell/components/ResourceList/index.vue
@@ -140,8 +140,12 @@ export default {
      *
      * This covers case 1
      */
-    namespaceFilter(neu) {
-      if (neu && !this.hasFetch) {
+    namespaceFilter(neu, old) {
+      if (
+        (neu || old) && // if neu or old have a value
+        neu !== old && // and there's a difference between them
+        !this.hasFetch // and there's a fetch passed into the component
+      ) {
         this.$fetchType(this.resource);
       }
     }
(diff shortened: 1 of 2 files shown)


richard-cox (changes requested):
> If the option to force namespace filters above a threshold is not enabled, using a single namespace filter should still filter API requests as needed.

If the feature is disabled we should not apply any of the feature, including applying the filtered namespace to the API request. This also may introduce flip flopping between fetching all resources (including the single namespace) followed by another  request to fetch resources for a single namespace  (which we already have)

PR #7979, opened 2023-01-20
Fix OPA Gatekeeper constraint inconsistencies

### LINKED ISSUE #7853: OPA violations count values is being reported differently
Internal reference: SURE-5753
Reported in 2.6.8

Issue description:
UI OPA violations count values is being reported differently.

For one label, if you compare Overview and Constraints under UI & OPA Gatekeeper, the example given shows 

Overview: 72 violations for a given label
Constraints, 20 Count (Red) for the same label
 
If you check from the constraint resource (`-o yaml`), it shows `totalViolations` as the first value: 72, so the question is where the 2nd number comes from.

Context:
> They created this constraint using a resource kind that is not a OPA default. They created custom policies as told by them, which for me this is a not-default OPA CRDs/APIs. And for that reason the UI is not watching/reporting the violations/counts(Red) accordingly. My question is, Do we have something to fix or improve for this env? Or that is the way UI was designed and expected to work (not reporting custom OPA CRDs/APIs)?

Workaround: none

Actual behavior:
UI OPA violations count values is being reported differently.

Expected behavior:
UI OPA violations count values reported accordingly.

nwmac:
Fixes #7853 

This PR Fixes an inconsistency in the OPA Gatekeeper violations view - on one page we show the count being the number of violations listed in the constraint resource, where as on the other we show the total violations form the count.

The correct value to use if from the `totalViolations` property - see: https://open-policy-agent.github.io/gatekeeper/website/docs/audit/#constraint-status - the constraint is limited to a number of violations that it will store detail for (default seems to be 20).

Also:
- Improved padding slightly to reduce it down
- Added a banner when the number shown is not the same as the number with detail
- Pulled out one more hard-coded string
- Fixed a bug where the kind in the constraint violations list was empty
- Added the count of violations to the title, so it is clear how many. there are
- 
Here's how it looks now:
image

Before:
image

For the overview, it now looks like this (22 is now shown correctly as the count):

image

before was like this (count of 20 shown is incorrect):

image

image

Easiest way to reproduce:

- Install OPA Gatekeeper into a cluster
- Using the global import yaml in the cluster, import this yaml:
```
apiVersion: constraints.gatekeeper.sh/v1beta1
kind: K8sRequiredLabels
metadata:
  name: all-must-have-owner
spec:
  match:
    kinds:
      - apiGroups: [""]
        kinds: ["Namespace"]
  parameters:
    message: "All namespaces must have an `owner` label that points to your company username"
    labels:
      - key: owner
        allowedRegex: "^[a-zA-Z]+.agilebank.demo$"
```

Go to OPA Gatekeeper and you should see this `all-must-have-owner` constraint and it will have violations for the existing namespaces.

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -2160,8 +2160,11 @@ gatekeeperConstraint:
       editAsYaml: Edit as YAML
       title: Parameters
   template: Template
+  enforcement:
+    action: Enforcement Action
   violations:
-    title: Violations
+    title: "Violations: {total}"
+    notAll: Not all are shown - the constraint only includes details for {shown}
 
 gatekeeperIndex:
   poweredBy: OPA Gatekeeper
--- a/shell/config/table-headers.js
+++ b/shell/config/table-headers.js
@@ -728,8 +728,8 @@ export const CONSTRAINT_VIOLATION_RESOURCE_LINK = {
 export const CONSTRAINT_VIOLATION_TYPE = {
   name:  'Type',
   label: 'Type',
-  value: `constraint.kind`,
-  sort:  `constraint.kind`
+  value: `kind`,
+  sort:  `kind`
 };
 
 export const CONSTRAINT_VIOLATION_MESSAGE = {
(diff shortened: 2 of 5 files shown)


jordojordo (approved):
lgtm - sorry for the long wait.

Issue #7523, opened 2022-11-22
DraggableZone, remove vuex store


torchiaf:
`DraggableZone` relies to the `vm` Vuex store. We want to remove it so we can use the component in others Rancher products without the Vuex requirement.

PR #7989, opened 2023-01-20
Remove DraggableZone from @rancher/components

rak-phillip:
This removes `DraggableZone` from `@rancher/components`. 

I don't think that `DraggableZone` belongs inside of `@rancher/components` until #7523 is completed. I'm moving `DraggableZone` under `@shell/components` for the time being to resolve issues encountered during development of rancher-sandbox/rancher-desktop#2822


torchiaf (approved):
LGTM

PR #7996, opened 2023-01-23
Fix locked project roles

### LINKED ISSUE #5952: Able to use Disabled/Locked Project Roles after upgrade
Internal reference: SURE-4203
Reported in 2.6.3 patch 1

Issue description:
After upgrade from Rancher 2.5.7 to 2.6.3-patch1, the Disabled/Locked Project Roles is able to being used.

Repro steps:

- Installed a fresh rancher 2.5.7
- Disabled/locked all Project template roles, less the Project Owner
- Upgraded Rancher to 2.6.3-patch1
- Was able to use the Project Namespace Creator to a local user

Additional info:

Someone else did the following with Rancher 2.5.7: 
- Disabled/Locked Project Roles, and there were not able to add a user to a project list/using the disabled/locked Project Roles
- Upgrade Rancher to 2.6.3 and they were able to see/list the disabled/locked Project Roles but after saving it, it does not work, no error showed up in UI, and when they tried to edit the Project again, the previous user added is not showed up. 

Belief therefore is it's necessary only to fix the listing of the Project Roles during adding a new user to a Project. Also in 2.6.3, the option to select ' Project Creator Default'  to 'Yes' doesn't work also.

@cbron indicated:
"If thats the case, then the UI just needs to filter when adding user and project role, to only the project roles that don't have `locked=true`.
 
On a side note, I went back and forth between old and new UI toggling both default-role and locked, and got very odd results. It's like some of the results were cached along the way, or some of the forms weren't accepting both options. Would be worth a look."

Actual behavior:
Previous disabled/locked Project roles in 2.5.7 are being able to use after Rancher upgraded to 2.6.3-patch1

Expected behavior:
Disabled/locked Project roles still locked for use after upgrades (2.5.7 -> 2.6.3-patch1)

bisht-richa:
### Summary
Fixes #5952

User should not be able to use Disabled/Locked Project Roles after the upgrade

### How to test
- Go to users and auth, > roles and cluster roles.
- Pick a role e.g. `Manage Cluster Backups` and select Edit Config from the action menu for it. Set 'Locked` to yes
- Go to the Cluster and Project Members and click Add > custom you'll get a list of custom roles
-  `Manage Cluster Backups` should be disabled 

Screenshot 2023-01-23 at 16 06 23

### DIFF
--- a/shell/components/form/Members/ClusterPermissionsEditor.vue
+++ b/shell/components/form/Members/ClusterPermissionsEditor.vue
@@ -279,7 +279,7 @@ export default {
             <i
               v-if="permission.locked"
               v-tooltip="permission.tooltip"
-              class="icon icon-lock icon-lg"
+              class="icon icon-lock icon-fw"
             />
           </div>
         </div>
--- a/shell/components/form/Members/ClusterPermissionsEditor.vue
+++ b/shell/components/form/Members/ClusterPermissionsEditor.vue
@@ -180,8 +180,8 @@ export default {
         const lockedExist = this.roleTemplates.find(roleTemplateItem => roleTemplateItem.displayName === customPermissionsItem.label);
 
         if (lockedExist.locked) {
-          customPermissionsItem['locked'] = true
-          customPermissionsItem['tooltip'] = 'This cluster role is locked.';
+          customPermissionsItem['locked'] = true;
+          customPermissionsItem['tooltip'] = 'This role is locked.';
         }
       });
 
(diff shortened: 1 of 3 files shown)


shell/components/form/Members/ClusterPermissionsEditor.vue
```diff
@@ -175,7 +175,20 @@ export default {
         opt:  { url: `/v3/principals/${ principalId }` }
       }, { root: true });
     },
+    customPermissionsUpdate() {
```
torchiaf, line 178: I think adding `locked` and `tooltip` properties to  `this.customPermissions` elements is not necessary here, since we are returning `customPermissionsUpdate` as a computed variable.
Suggested solution:
```
customPermissionsUpdate() {
  return this.customPermissions.reduce((acc, customPermissionsItem) => {
    const lockedExist = this.roleTemplates.find(roleTemplateItem => roleTemplateItem.displayName === customPermissionsItem.label);

    if (lockedExist.locked) {
      customPermissionsItem['locked'] = true;
      customPermissionsItem['tooltip'] = 'This role is locked.';
    }

    return [...acc, customPermissionsItem];
  }, []);
```

shell/components/form/ProjectMemberEditor.vue
```diff
@@ -188,6 +188,18 @@ export default {
           value:       'custom'
         }
       ];
+    },
+    customPermissionsUpdate() {
```
torchiaf, line 192: Same here

torchiaf (changes requested):
I made a test and it works fine. I would change the implementation of `customPermissionsUpdate`, we are updating another variable in there.

torchiaf (approved):
LGTM

nwmac:
@bisht-richa There are two comments on this PR not marked as resolved - have they been? I see its been approved, but we need to resolve conversations so it is clear, if that is the case.

bisht-richa:
> @bisht-richa There are two comments on this PR not marked as resolved - have they been? I see its been approved, but we need to resolve conversations so it is clear, if that is the case.



> @bisht-richa There are two comments on this PR not marked as resolved - have they been? I see its been approved, but we need to resolve conversations so it is clear, if that is the case.

Yes those are resolved.

PR #8000, opened 2023-01-23
Extract secondary nav into new component

rak-phillip:
### Summary

This is the first in a series of steps to allow for a customized layout to be used with the Rancher Desktop extension.  

- [ ] Create a secondary nav component to share between Default and Rancher Desktop layouts (shell)
- [ ] Create a new Rancher Desktop Layout (Shell)
- [ ] Create nuxt middleware that defaults to Rancher Desktop layout (Desktop)

Supports rancher-sandbox/rancher-desktop#2822

All behavior related to the secondary navigation in migrated away from the `Default` layout and into a `SecondaryNav` component.

- Secondary navigation
- Default layout
- Any components/pages that contain a nav element (global css to scoped css)

- The secondary nav component exclusively makes use of scoped CSS. There are concerns that some global CSS could experience regressions with this change. This is especially a concern for global styles that affect the `NAV` element
- Functionality of `Default` layout. A lot of code related to the secondary nav is not a concern of the Default layout. My concerns that some non-obvious regressions can occur with how tightly coupled this code is.
- Functionality of secondary navigation. Similar to `Default` layout concerns, there might be some non-obvious regressions related to this change  

#### Reference for new Rancher Desktop Layout

- https://xd.adobe.com/view/7b583c5c-c4d0-4a57-ac7a-78d97c4560a2-2e0b/

### DIFF
--- a/shell/components/NavigationSecondary.vue
+++ b/shell/components/NavigationSecondary.vue
@@ -9,7 +9,7 @@ export default {
   components: { Group },
   computed:   {
     ...mapState(['managementReady', 'clusterReady']),
-    ...mapGetters(['clusterId', 'isSingleProduct']),
+    ...mapGetters(['clusterId', 'isSingleProduct', 'isExplorer']),
     ...mapGetters({ locale: 'i18n/selectedLocaleLabel', availableLocales: 'i18n/availableLocales' }),
 
     showClusterTools() {
--- a/shell/components/NavigationSecondary.vue
+++ b/shell/components/NavigationSecondary.vue
@@ -1,9 +1,17 @@
 <script>
 import { mapState, mapGetters } from 'vuex';
-import { CATALOG } from '@shell/config/types';
+import { CATALOG, UI } from '@shell/config/types';
 import { HARVESTER_NAME as HARVESTER } from '@shell/config/features';
 import { getVersionInfo } from '@shell/utils/version';
 import Group from '@shell/components/nav/Group';
+import { replaceWith, clear, addObject, addObjects } from '@shell/utils/array';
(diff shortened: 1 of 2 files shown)


shell/components/NavigationSecondary.vue
```diff
@@ -0,0 +1,555 @@
...
+        this.getGroups();
+      }
+    },
+
+    product(a, b) {
+      if ( !isEqual(a, b) ) {
+        // Immediately update because you'll see it come in later
+        this.getGroups();
+      }
+    },
```
rak-phillip, line 165: What does the `product` watcher operate on? I don't see a corresponding `product` in `data` or `computed`.

rak-phillip:
I'm also considering if `SideNav` is a better name for this component. It doesn't seem that we refer to this as "secondary" in code.

cnotv:
@rak-phillip not to ruin the party, but would be possible to investigate and merge such refactory for this component after the PSA release, so we avoid issues? 
I'll have also to update the navigation as there's more work to be done.

If this component is required for this week, then we can try it out.

rak-phillip:
@cnotv I can continue developing other pieces if we anticipate that PSA will finish within the week. I might ask that you help review this change as I rebase to ensure that we stay synchronized with changes introduced into `master`. My concern is that we risk regressing new PSA work as we rebase.

I would also like for this to get some review and testing in the current state to address issues in this PR sooner rather than later.

cnotv:
@rak-phillip Ok then we can take a look into it now and be sure to avoid double work from any side.

rak-phillip:
Closing as stale

PR #8006, opened 2023-01-23
Allow to change minimum length requirement for user password

### LINKED ISSUE #6633: Allow to change minimum length requirement for user password

cnotv:
### Summary
Fixes #6633

New case with decimal using 0 as value, e.g. 1.0.

### DIFF
--- a/shell/utils/validators/formRules/__tests__/index.test.ts
+++ b/shell/utils/validators/formRules/__tests__/index.test.ts
@@ -1057,7 +1057,7 @@ describe('formRules', () => {
 
   describe.each([
     ['requiredInt', [2, 2.2], ['e']],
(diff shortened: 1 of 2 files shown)


torchiaf (approved):
LGTM

Issue #7999, opened 2023-01-23
PSA minor changes


kwwii:
Change `PodSecurityAdmissionConfigurationTemplates` to `Pod Security Admission`

Capitalise the checkbox labels

Change the text size of "Define the admission ctorl mode you want to use for pod security" AND check the text itself (@kwwii)

Change the cluster template labels to NOT have checkboxes as the cluster PSAs will always be set (the default setting is privileged). Namespace PSAs are still optional and require the checkbox

Add title "Pod Security Admission" to the tooltips in the namespace list

cnotv:
As mentioned in the meeting, there are the other 2 change requests:
https://github.com/rancher/dashboard/issues/7987
https://github.com/rancher/dashboard/issues/7988

kwwii:
Change the way Labels are shown in the namespace detail page.
Change show all labels to "Show System Labels" and when clicked, add the details of the PSA label to the overview shown "Audit restricted (v1.25)" becomes "Audit restricted (v1.25): longdetailsofthepsa"

kwwii:
Change the length of the growl warning message to 5s

cnotv:
@kwwii Change the text size of "Define the admission ctorl mode you want to use for pod security" AND check the text itself 

Both the text are 12px.

cnotv:
@kwwii I'm using the i18n of the term, as these will probably need to be changed as well.

cnotv:
The detailTop pages changes is going to take a while, since I've to bring back [my original code with the original design from the PR](https://github.com/rancher/dashboard/pull/7619).

liyimeng:
I think #8023 can be part of this one

markusewalker:
This has been tracked and marked as completed in the internal test K8s 1.25 test plan. Closing issue.

PR #8012, opened 2023-01-24
Update and refine PSP banners (RKE2)

### LINKED ISSUE #7791: Confirmation Warning when creating a v1.23-v1.24 RKE1/2 Cluster with PSPs enabled
Users running Kubernetes v1.23 - v1.24 should have a banner asking them to migrate to PSA/PSSs in the same places as we're doing for v1.25 with the phasing out of PSPs but it's more of a case of guidance before they upgrade.

An investigation is needed into the impact of maintaining PSPs in a cluster that no longer supports them. Preliminary findings show issues with orphaned PSP resources. It should be evaluated whether users are prevented from upgrading to 1.25 altogether with Rancher Launched Kubernetes and Distributions until PSPs are completely disabled in those clusters. 

(RKE1 upgrades to 1.25 will not be allowed if PSP is enabled, back-end will throw error; 
for RKE2, the upgrades are prevented when non-default RKE2 PSP is used -- by default RKE2 has PSP enabled and applies "allow all" default PSP)

nwmac:
Fixes #7791

PSPBanners

This PR refines the banners displayed in RKE2 cluster creation/edit. It also tweaks some of the interactions between the two:

Changes:

- PSAs and PSPs can be used independently, so this PR removes the logic of resetting one when you use the other - users should be able to configure both if they wish
- Moved the PSP deprecation banner shown at the top of the page to the Security section - otherwise it can not be seen as the user changes settings

Improvements:

When creating a new RKE2 cluster, we now show a banner in the security section:

When selected version is >= 1.25

image

otherwise we show:
 
image

When editing an existing RKE2 cluster, we will not show either of the above banners, however, if the user changes the Kubernetes version to >= 1.5, we show:

image

When we do this, we will hide the PSP drop-down - we now remember the previous setting, so if the user changes to 1.25 and then sees that PSPs are not supported and then changes the version back, the value of the PSP dropdown will go back to what it was before the version was changed to 1.25, rather than resetting to the default - this ensures users do not loose this setting in this case.

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -1607,6 +1607,8 @@ cluster:
     desiredNodeGroupWarning: There are 0 nodes available to run the cluster agent. The cluster will not become active until at least one node is available.
     invalidPsps: You have one or more PodSecurityPolicy resource(s) in this cluster. Pod Security Policies are not available in Kubernetes v1.25.
     haveArgInfo: Configuration information is not available for the selected Kubernetes version.  The options available in this screen will be limited, you may want to use the YAML editor.
+    deprecatedPsp: Pod Security Policies are deprecated as of Kubernetes v1.21, and have been removed in Kubernetes v1.25.
+    removedPsp: Pod Security Policies have been removed in Kubernetes v1.25, use PodSecurity Admission instead.
 
   rkeTemplateUpgrade: Template revision {name} available for upgrade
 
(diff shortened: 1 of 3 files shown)


nwmac:
@cnotv @kwwii I was reviewing banners in RKE2 cluster create/edit since this was the last bit to consider. I went ahead and made these tweaks which I think covers what we need to do for now for PSP removal banners.

cnotv:
@nwmac not to be ungrateful that you helped up, but usually we make a video to ensure that all the possible cases are covered by the PR. That's also something we mentioned in the retro.

cnotv:
I'll take time to review it after complete this issue, as we have 2 days :)
https://github.com/rancher/dashboard/issues/7999

nwmac:
@cnotv Thank you - video added

cnotv (approved):
LGTM

PR #8026, opened 2023-01-25
Added port input in receiver email form

### LINKED ISSUE #5453: [BUG] Form for configuring Receivers and Output should more clearly indicate where to input Port
Internal ref: SURE-3041

Bug/Parity

**Description**
The port should be input in a separate field when configuring receivers for alerts to keep parity with v1 and logging v2 outputs.  It's easy to miss that the port should be put in the URL field, especially when you've already pasted your URL endpoint in the field (which erases the example).  This would be consistent with Monitoring v1 Notifiers and Elastic Search outputs in Logging v2.
The grey on grey text of the examples on the light background is too easily missed by the user (the customer I was working with had trouble reading it and so do I) and the Logging output field does not even display an example.

**Result:**
Alerting Reciever:
image
image

Logging Output for an elasticsearch endpoint separates the port into its own box, like in v1
image

Note the example text in the "Description" field above is easily missed and for some will be illegible.
image

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.5.7

gz#15698

bisht-richa:
### Summary
Fixes #5453

Added port input in receiver email form

### How to test
- Goto > cluster > monitoring > route-receiver
- Click on `Create` > select Email tab > click on `Add email`, form should have a Port field

Screenshot 2023-01-25 at 12 05 43

### DIFF
--- a/shell/edit/monitoring.coreos.com.receiver/index.vue
+++ b/shell/edit/monitoring.coreos.com.receiver/index.vue
@@ -48,7 +48,6 @@ export default {
       for (let i = 0; i < this.value.spec.email_configs.length; i++) {
         if (this.value.spec.email_configs[i].smarthost) {
           const hostPort = this.value.spec.email_configs[i].smarthost.split(':');
-
           this.$set(this.value.spec.email_configs[i], 'host', hostPort[0] || '');
           this.$set(this.value.spec.email_configs[i], 'port', hostPort[1] || '');
           delete this.value.spec.email_configs[i]['smarthost'];
@@ -167,7 +166,8 @@ export default {
       if (this.value.spec.email_configs.length > 0) {
         this.value.spec.email_configs.forEach((email) => {
           if (email['port'] || email['host']) {
-            email.smarthost = email.port ? `${ email.host }:${ email.port }` : `${ email.host }`;
+            const hostValue =  email.host ? `${ email.host }` : '';
+            email.smarthost = email.port ? `${ hostValue }:${ email.port }` : `${ hostValue }`;
             delete email['port'];
             delete email['host'];
           }
--- a/shell/edit/monitoring.coreos.com.receiver/index.vue
+++ b/shell/edit/monitoring.coreos.com.receiver/index.vue
@@ -165,28 +165,19 @@ export default {
 
     createSmarthost() {
       if (this.value.spec.email_configs.length > 0) {
-        for (let i = 0; i < this.value.spec.email_configs.length; i++) {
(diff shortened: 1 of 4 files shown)


shell/edit/monitoring.coreos.com.receiver/index.vue
```diff
@@ -43,6 +44,18 @@ export default {
   data() {
     this.$set(this.value, 'spec', this.value.spec || {});
 
+    if (this.mode === _EDIT || this.mode === _VIEW) {
```
nwmac, line 47: The logic for this feels like it should be in one place - some is here, the rest is in the model.

I suggest using before and after save hooks here, so that it is all in this file.

As is, if there is an error saving, I think it will break, as the model will clean up port and host, but the view will still expect them to be present.

nwmac (changes requested):

shell/edit/monitoring.coreos.com.receiver/index.vue
```diff
@@ -161,8 +161,33 @@ export default {
...
+          }
+        }
+      }
     }
-  }
+  },
+
+  created() {
+    this.registerBeforeHook(this.createSmarthost, 'create-smarthost');
+    this.registerAfterHook(this.deletePortHost, 'delete-port-host');
```
nwmac, line 189: This sends host and port in the request - we only need to send `smarthost`.

nwmac (changes requested):

nwmac (changes requested):
Need to update to not send port and host

shell/edit/monitoring.coreos.com.receiver/index.vue
```diff
@@ -148,8 +161,24 @@ export default {
 
     createAddOptions(receiverType) {
       return receiverType.addOptions.map();
+    },
+
+    createSmarthost() {
+      if (this.value.spec.email_configs.length > 0) {
+        this.value.spec.email_configs.forEach((email) => {
+          if (email['port'] || email['host']) {
+            email.smarthost = `${ email.host }:${ email.port }`;
```
nwmac, line 170: If port is not specified, you get a colon at the end - we should only add the : and port if port is not empty

nwmac (changes requested):

shell/edit/monitoring.coreos.com.receiver/index.vue
bisht-richa: Fixed

PR #8036, opened 2023-01-26
Added placeholder in url input

### LINKED ISSUE #7657: Loki Outputs URL can accept invalid input
Internal reference: SURE-5577
Reported in 2.6.8-2.7.0

Issue description:
Configuring incomplete values for Outputs corresponding to Loki Server, resulted in configcheck failure. There is a requirement to enable validation while pressing the "create" button itself to avoid the error getting unnoticed.

Troubleshooting steps:
Create Outputs with Loki and enter a URL without scheme (http/https) in URL textbox. Rancher allows to create it without any error. Similar applicable for ClusterOutputs as well. However, Fluentd configcheck failed with a proper error. This would be helpful for those unfamiliar with the expected syntax based on other tools.

Repro steps:

- Rancher >> Cluster >> Logging >> Outputs/ClusterOutputs >> create >> Give a name >> Select Loki from the Output dropdown
- In the URL textbox, give Loki server name without http/https
- Click create 

Workaround:
Fill it out properly including the scheme in URL

Actual behavior:
Rancher allows creating the outputs without complaining. At the same time, Name textbox has proper validation and also it has a short description of the expected value

Expected behavior:
Either Rancher shouldn't allow creating the Output with incorrect URL format or URL textbox should have a sample value or hint of the expected format (similar to http(s)://<TARGET>:<PORT>)

bisht-richa:
### Summary
Fixes #7657

Added hint of the expected format in URL input

### How to test
- Rancher >> Cluster >> Logging >> Outputs/ClusterOutputs >> create >> Give a name >> Select Loki from the Output dropdown
- In the URL textbox, give Loki server name without http/https
- Click Create

### DIFF
--- a/shell/edit/logging.banzaicloud.io.output/index.vue
+++ b/shell/edit/logging.banzaicloud.io.output/index.vue
@@ -100,9 +100,6 @@ export default {
     }
   },
 
-  created() {
-    this.registerBeforeHook(this.willSave, 'willSave');
-  },
   methods: {
     getComponent(name) {
       return require(`./providers/${ name }`).default;
@@ -110,7 +107,18 @@ export default {
     launch(provider) {
       this.$refs.tabbed.select(provider.name);
     },
-    willSave() {
+    saveSettings(done) {
+      const t = this.$store.getters['i18n/t'];
+      const isLokiHttps = this.value.spec['loki'].url ? this.value.spec['loki'].url.toLowerCase().startsWith('https://') : undefined;
+
+      if (this.selectedProvider === 'loki' && !isLokiHttps) {
+        this.errors = [t('logging.loki.urlInvalid')];
+
+        return done(false);
+      }
+
+      this.errors = [];
+
       this.value.spec = { [this.selectedProvider]: this.value.spec[this.selectedProvider] };
 
       const bufferJson = jsyaml.load(this.bufferYaml);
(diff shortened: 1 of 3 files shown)


aalves08 (commented):
So, codewise and testing for `https://` works fine, but I was wondering if we shouldn't also support `http://`, as per https://github.com/rancher/dashboard/issues/7657?

One other thing, which might be out of scope for this one, is that there are other providers with URL inputs as well. Should those be covered in this go, or should we just create a follow-up issue?

bisht-richa:
> is that there are other providers with URL inputs as well. Should those be covered in this go, or should we just create a follow-up issue?

Neil said he will create a follow-up issue for this if needed.

shell/edit/logging.banzaicloud.io.output/index.vue
```diff
@@ -100,17 +100,26 @@ export default {
...
       return require(`./providers/${ name }`).default;
     },
     launch(provider) {
       this.$refs.tabbed.select(provider.name);
     },
-    willSave() {
+    saveSettings(done) {
+      const t = this.$store.getters['i18n/t'];
+      const urlCheck = ['https://', 'http://'].some(checkValue => this.value.spec['loki'].url.toLowerCase().startsWith(checkValue));
+      const isLokiHttps = this.value.spec['loki'].url ? urlCheck : undefined;
```
aalves08, line 113: By having `urlCheck` and `isLokiHttps` as constants, the operations that populate those run every time, therefore breaking the UI when were aren't in a `this.selectedProvider === 'loki'` scenario
Screenshot 2023-04-21 at 12 53 05
Screenshot 2023-04-21 at 12 53 09

aalves08 (commented):
Found an issue @bisht-richa 🙏

PR #8096, opened 2023-02-06
bumping shell package version for release

aalves08:
- Bumping shell package version for a new release

### DIFF
--- a/shell/package.json
+++ b/shell/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@rancher/shell",
-  "version": "0.2.5",
+  "version": "0.3.1",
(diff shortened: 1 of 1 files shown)


richard-cox (approved):
There's something odd about that package version. Previously it was `0.2.5` yet in npm it's `3.0.0`. At some point this will need to be published, but as there's already a misalignment this can merge

aalves08:
You are right @richard-cox . Probably that was my fault as I may have released the shell package without bumping the version on `package.json`, hence why I moved it to `0.3.1` so that they should be in sync. Thanks for the approval 🙇

PR #8133, opened 2023-02-09
PSA: Update PSA default

### LINKED ISSUE #8032: PSA: Update PSA default
**Changes required:**

The PSP default 'RKE Default' should be labeled 'Default - RKE2 Embedded'.

The PSA default is currently labeled '(None)' - this is fine for k8s <1.25 (because in k8s <1.25 RKE2 does not automatically apply any PSA).
When the version is >= 1.25, instead of "(None)" we should have "Default - RKE2 Embedded" (which "maps" to empty PSACT on Cluster YAML). Note that this does _not_ apply to k3s as it doesn't automatically apply PSA.

If the user changes the CIS Worker Profile to anything other than "None" **on k8s >= 1.25**, then we need to reset the PSA to "rancher-restricted" (it's guaranteed to exist) and disable the control to prevent the user changing it. We should show a banner or somehow indicate that we've done this.

In some cases, the user may create a PSA Configuration Template (PSACT) that is compatible with the "rancher-restricted" - they may want to use this with CIS Worker Profile set to something other than None - I suggest a checkbox 'Allow PSA Default Template to be overridden when using a CIS Profile". Checking this would enable the PSA template dropdown and allow them to select their own.

We should show a banner if they check this that they understand that they know what they are doing.

Note: The Rancher Default is different to the RKE2 Default, so we can't use their docs - Rancher should provide docs that we can link to when done.

cnotv:
### Summary
Fixes #8032

- Disabled/enabled PSA dropdown based on indications
- Added checkbox for hardening overriding logic
- Changed label based on versions

### Technical notes summary
- Corrected or added existing i18n for CIS UI
- Corrected unit test missing argument
- Prevented issues in case of lack of CIS profiles
- Added tests for the logic


aalves08 (commented):
Codewise, nothing stands out, so I would say all good. Ran the unit tests and all seem fine as well.

I got a couple of items that I caught:
There's an error on LabeledSelect when changing the kube version, which blocks the interface:

https://user-images.githubusercontent.com/97888974/218727861-92200e84-fa2d-46a3-86d7-a618bda4af75.mov

Also, the select for `Default Pod Security Admission` seems to have less options than yours:

https://user-images.githubusercontent.com/97888974/218728215-f1696bf6-8d30-4687-a6cf-abb7e740b82e.mov

Tested on rancher BE `v2.7.2-rc2`

cnotv:
> Codewise, nothing stands out, so I would say all good. Ran the unit tests and all seem fine as well.
> 
> I got a couple of items that I caught: There's an error on LabeledSelect when changing the kube version, which blocks the interface:

Good catch 👏 
Taking a look now.

> Also, the select for `Default Pod Security Admission` seems to have less options than yours:

This is because I created them. If you want you can give it a try as well at this path: `/c/_/manager/management.cattle.io.podsecurityadmissionconfigurationtemplate`

aalves08 (approved):
All good @cnotv . No more issues with the `LabeledSelect` for the kube version and the `Pod Security Admissions` created appear on the cluster creation UI. LGTM!

PR #8148, opened 2023-02-13
gate some critical methods that were introducing errors on elemental …

aalves08:
- gate some critical methods that were introducing errors on Elemental while running the Elemental extension on v2.7.0 dashboard

To reproduce the bug (cannot exactly test the solution because `@shell` needs to be update on Elemental):
- Provision a 2.7.0 dashboard
- build latest Elemental on `elemental-ui` repo using `yarn build-pkg elemental`
- Serve it using `yarn serve-pkgs` and copy the url for the latest version of Elemental
- Do a developer load for the new version of Elemental on the plugins area on dashboard
- Check errors on console (navigate to `local` cluster explorer/dashboard view as an example)

The gating should reflect the areas where the errors occur.

Note: Extensions API PR - https://github.com/rancher/dashboard/pull/7777/files. Also, file changes look somewhat extensive but it's basically only adding the if's in the correct places and the rest are indentation related changes

### DIFF
--- a/shell/core/plugin-helpers.js
+++ b/shell/core/plugin-helpers.js
@@ -94,78 +94,81 @@ function checkExtensionRouteBinding($route, locationConfig) {
 export function getApplicableExtensionEnhancements(pluginCtx, actionType, uiArea, currRoute, translationCtx = pluginCtx) {
   const extensionEnhancements = [];
 
-  const actions = pluginCtx.$plugin.getUIConfig(actionType, uiArea);
+  // gate it so that we prevent errors on older versions of dashboard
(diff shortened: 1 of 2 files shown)


jordojordo:
@aalves08 This seems to fix the issue I was [running into](https://github.com/rancher/dashboard/issues/8152) when I run this using `yarn link` and running the extension code locally. 

___Extension page with previous issue___
yarn-link

I am getting a few warnings in the console that seem unrelated but wanted your opinion before approving:
yarn-link-errors

jordojordo (approved):
lgtm - thanks for getting this fix in!

PR #8154, opened 2023-02-13
LabeledSelect: Displayed label does not update with options

### LINKED ISSUE #8149: LabeledSelect: Displayed label does not update with options
**Setup**
- Rancher version: any
- Browser type & version: any

**Describe the bug**
While changing the options for the `LabeledSelect` component, the selected displayed label does not update with them.

**To Reproduce**
- Identify any existing `LabeledSelect` component or create a new case with new options
- Update the options with an event (e.g. checkbox)

**Result**
Label stays the same.

**Expected Result**
New value is applied

**Screenshots**

https://user-images.githubusercontent.com/5009481/218461203-229738f8-0a6c-41b5-af7b-eec690024f1b.mov

cnotv:
### Summary
Fixes #8149

As in issue description.

To reproduce create/extend in the UI something like:

```ts
// ...
data: () {
  return {
    // ...
    toggleLabel:           false,
    myValue:               '',
  }
},


computed: {
  // ...

  myOptions() {
    return this.toggleLabel 
      ? [{ label: 'Old Label', value: '' }] 
      : [{ label: 'New Label', value: '' }];
  },
}
// ...
``` 

```html
      <LabeledSelect
        v-model="myValue"
        :options="myOptions"
      />
      <Checkbox
        v-model="toggleLabel"
        label="Toggle my label"
      />
```

### Technical notes summary
- Added unit tests for ensuring label updates
- Forced displayed label to match existing options with the same value

Any dropdown as we have no story nor tests.

Any dropdown as we have no story nor tests.

### Screenshot/Video

https://user-images.githubusercontent.com/5009481/218513012-fa6f3a03-8f89-4fee-80d5-7951aa01da7d.mov

### DIFF
--- a/shell/components/form/LabeledSelect.vue
+++ b/shell/components/form/LabeledSelect.vue
@@ -152,7 +152,10 @@ export default {
       const isOutdated = !this.options.find(opt => option[this.optionLabel] === opt[this.optionLabel]);
 
       if (isOutdated) {
-        return this.options.find(opt => option[this.optionKey] === opt[this.optionKey])?.[this.optionLabel];
+        const option = this.options.find(opt => option[this.optionKey] === opt[this.optionKey]);
+        const label = get(option, this.optionLabel);
+
+        return this.localizedLabel ? this.$store.getters['i18n/t'](label) || label : label;
       }
 
       if (this.$attrs['get-option-label']) {
--- a/shell/components/form/LabeledSelect.vue
+++ b/shell/components/form/LabeledSelect.vue
@@ -149,10 +149,10 @@ export default {
       }
 
       // Force to update the option label if prop has been changed
-      const isOutdated = !this.options.find(({ label }) => option.label === label);
+      const isOutdated = !this.options.find(opt => option[this.optionLabel] === opt[this.optionLabel]);
 
       if (isOutdated) {
-        return this.options.find(({ value }) => option.value === value)?.label;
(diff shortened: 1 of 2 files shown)


shell/components/form/LabeledSelect.vue
```diff
@@ -159,6 +148,13 @@ export default {
         return;
       }
 
+      // Force to update the option label if prop has been changed
+      const isOutdated = !this.options.find(({ label }) => option.label === label);
```
mantis-toboggan-md, line 152: It's possible to work with options that do not have `label` and `value` properties by setting the `optionLabel` and `optionValue` props, so I think this needs to use eg 
`!this.options.find((opt) => option[this.optionLabel]===opt[this.optionLabel]);`

shell/components/form/LabeledSelect.vue
```diff
@@ -159,6 +148,13 @@ export default {
...
+
+      if (isOutdated) {
+        return this.options.find(({ value }) => option.value === value)?.label;
```
mantis-toboggan-md, line 155: This should account for `this.localizedLabel` as lines 161-165 do. It should also reference `this.optionLabel` and `this.optionKey` (if defined)

mantis-toboggan-md (changes requested):
I suggested a couple small changes due to `LabeledSelect` being used 1,001 different ways, otherwise the pr looks good

mantis-toboggan-md (approved):

Issue #8175, opened 2023-02-14
`Custer Management > Cloud Credentials > Create > Amazon > default region` dropdown fails to populate on `2.7-head`


Josh-Diamond:
**Setup**
- Rancher version: `v2.7-ea94f0ad4fa168a597da1bc6df74a24faf28507c-head`
- Browser type & version: `Chrome`

**Describe the bug**
`default region` dropdown fails to populate when creating Amazon EC2 Cloud Credentials on rancher `v2.7-head`

**To Reproduce**
1. Fresh install of rancher `v2.7-head`
2. Navigate to `Custer Management > Cloud Credentials > Create > Amazon > default region`
3. Reproduced

**Result**
dropdown does not populate with regions to select

**Expected Result**
dropdown populates with regions to select

**Video Clip**

https://user-images.githubusercontent.com/46494969/218863218-1c9b5530-50be-4b00-9d09-73bb7126d5dd.mp4

Josh-Diamond:
Ticket #8175 - Test Results - ✅ 
---

With Docker on a single-node instance:

Verified on rancher `v2.7-d559c2f16d5a634d8e5188e242ee9a5b69c55908-head`:

Scenario 1:
1. Fresh install of rancher `v2.7-head`
2. As admin, navigate to `Cluster Management > Cloud Credentials > Create > Amazon > default Region`
3. Verified - `default region` dropdown is successfully populated and accurate
4. Create cloud credentials
5. Verified - cloud credentials successfully created

---

Scenario 2:
1. Fresh install of rancher `v2.7-head`
2. As admin, create a standard user
3. As standard user, navigate to `Cluster Management > Cloud Credentials > Create > Amazon > default Region`
4. Verified - `default region` dropdown is successfully populated and accurate
5. Create cloud credentials
6. Verified - cloud credentials successfully created

Screenshot:
<img width="1436" alt="Screenshot 2023-02-27 at 3 59 26 PM" src="https://user-images.githubusercontent.com/46494969/221713002-dee2f0f2-79d3-4644-97ac-6fe65504a3d2.png">

PR #8176, opened 2023-02-14
fix labeledselect getOptionLabel to use reduce prop when labels change

### LINKED ISSUE #8173: Azure RKE2 provisioning page location dropdown is broken
The location dropdown is empty and mis-styled:

The error in the browser console is `TypeError: Cannot read properties of undefined (reading 'name')
    at LabeledSelect.vue:155:1`
A non-default location must be selected to set an availability zone so this will block testing the RKE2 portion of https://github.com/rancher/dashboard/issues/8014

mantis-toboggan-md:
Fixes #8173, #8175
The changes introduced here in #8154 have caused some issues in other LabeledSelect implementations. The root of the problem seems to be that the code that runs when options update is using `optionKey`, which is defaulted to `null`. Sorry for the misinformation @cnotv my suggestion on your PR was way off. `optionKey` is used with [getOptionKey](https://vue-select.org/api/props.html#getoptionkey) to generate unique ids for the dropdown option elements. The customization of how the selected option is mutated to the `value` that is emitted is controlled by the `reduce` prop.

I went to update the associated tests and, frustratingly, couldn't write a valid test case where the existing code around `isOutdated` fails.  From what I could see that `isOutdated` is never `true` in the tests, I guess something is different about using `setProps` compared to how the options are updated for the psa default value? I hit a wall tbh maybe you have ideas.

### DIFF
--- a/shell/components/form/LabeledSelect.vue
+++ b/shell/components/form/LabeledSelect.vue
@@ -5,6 +5,7 @@ import { get } from '@shell/utils/object';
 import { LabeledTooltip } from '@components/LabeledTooltip';
 import VueSelectOverrides from '@shell/mixins/vue-select-overrides';
 import { onClickOption, calculatePosition } from '@shell/utils/select';
+import isEqual from 'lodash/isEqual';
 
 export default {
   name: 'LabeledSelect',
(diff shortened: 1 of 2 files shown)


cnotv:
Given 3 bugs (one I found in one of my issues) we rather need to write a test for each of them, as we cannot guess how a component works.

cnotv (approved):
LGTM

cnotv:
Just as a side note analyzing how the component work to keep for the records:
- If we use only options and these change, the value is not updated
- The reduce logic is fired at every change/event of the component, while we could simply provide them once

Issue #8074, opened 2023-01-31
Adding label to a pod causes error


gaktive:
Internal reference: SURE-5824
Reported in 2.7.0

Issue description:
Adding label to a pod via Rancher GUI causes error.

Repro steps:

step 1
Create a pod by kubectl: `$ kubectl run nginx-pod --image=nginx`
 
step 2
- Add a Label to the pod via Rancher GUI:
 - Cluster Manager -> local cluster -> Workload -> pods -> Edit Config -> Pod tab -> Labels & Annotations
- Push Add Label button: `key=foo, value=bar`
- Save
 
error occurs (error.png)
![error.png](https://zube.io/files/rancher/f6224c0ea57d0963bfb3944fcc2a2138-error.png)

```
Pod "nginx-pod" is invalid: spec: Forbidden: pod updates may not change fields other than `spec.containers[*].image`, `spec.initContainers[*].image`, `spec.activeDeadlineSeconds`, `spec.tolerations` (only additions to existing tolerations) or `spec.terminationGracePeriodSeconds` (allow it to be set to 1 if it was previously negative)
core.PodSpec{
... // 11 identical fields
NodeName: "35.74.247.85",
SecurityContext: &{},

ImagePullSecrets: nil,
+ ImagePullSecrets: []core.LocalObjectReference{},
Hostname: "",
Subdomain: "",
SetHostnameAsFQDN: nil,
Affinity: nil,
+ Affinity: &core.Affinity{},
SchedulerName: "default-scheduler",
Tolerations: { {Key: "node.kubernetes.io/not-ready", Operator: "Exists", Effect: "NoExecute", TolerationSeconds: &300}
, {Key: "node.kubernetes.io/unreachable", Operator: "Exists", Effect: "NoExecute", TolerationSeconds: &300}},
... // 11 identical fields
}
```

gaktive:
@scures saw this when trying to update environment variables in Epinio:
`Pod "epinio-server-5d5845745d-49ccj" is invalid: spec: Forbidden: pod updates may not change fields other than `spec.containers[*].image`, `spec.initContainers[*].image`, `spec.activeDeadlineSeconds`, `spec.tolerations` (only additions to existing tolerations) or `spec.terminationGracePeriodSeconds` (allow it to be set to 1 if it was previously negative)`

This may be fixed in 2.7.2 elsewhere via Epinio work but this could be tied to a Helm deployment done not in the UI originally.

Josh-Diamond:
[...]

PR #8192, opened 2023-02-16
Remove Pod edit fields that are not editabble from PUT request.

Shavindra:
### Summary
Fixes rancher/dashboard#8074 

When using kubectl shell to create a pod by default `imagePullSecrets` and `affinity` set to `null`. Once set these fields cannot be edited. 

When using UI fields are set to `[]` and `{}`. This create a conflict

### Technical notes summary
- Delete these properties if they are not set

- Enter Kubectl shell using keyboard command  `` ctrl+`  `` Or using the icon on top navigation
- Create a pod by kubectl: `$ kubectl run nginx-pod --image=nginx`
- Add a Label to the pod via Rancher GUI:
- Cluster Manager -> local cluster -> Workload -> pods -> Edit Config -> 
- Save (there is no need to update any fields. Although this can be tested)
- This should update the POD.

### DIFF
--- a/shell/edit/workload/mixins/workload.js
+++ b/shell/edit/workload/mixins/workload.js
@@ -775,7 +775,7 @@ export default {
 
         // Removing `affinity` fixes the issue with setting the `imagePullSecrets`
         // However, this field should not be set. Therefore this is explicitly removed.
-        if (template?.spec?.imagePullSecrets && Object.keys(template?.spec?.imagePullSecrets).length === 0) {
+        if (template?.spec?.imagePullSecrets && template?.spec?.imagePullSecrets.length === 0) {
           delete template.spec.imagePullSecrets;
         }
       }


shell/edit/workload/mixins/workload.js
```diff
@@ -766,6 +766,20 @@ export default {
...
+
+      // The fields are being removed because they are not allowed to be editabble
+      if (this.mode === _EDIT) {
+        if (template?.spec?.affinity && Object.keys(template?.spec?.affinity).length === 0) {
+          delete template.spec.affinity;
+        }
+
+        // Removing `affinity` fixes the issue with setting the `imagePullSecrets`
+        // However, this field should not be set. Therefore this is explicitly removed.
+        if (template?.spec?.imagePullSecrets && Object.keys(template?.spec?.imagePullSecrets).length === 0) {
```
nwmac, line 778: @Shavindra imagePullSecrets is an array - so I think we can do just `.length` without the Object.keys

nwmac (changes requested):

nwmac (approved):

PR #8198, opened 2023-02-17
Extensions: Add publish script to shell

nwmac:
This PR adds the publish script for extensions to shell and updates the creators to make this available for new extensions.

It also adds support for copying the annotations in the extension's `package.json` file to the generate chart file.


shell/creators/pkg/init
```diff
@@ -18,6 +18,7 @@ const files = [
 const topLevelScripts = {
   'build-pkg':  './node_modules/@rancher/shell/scripts/build-pkg.sh',
   'serve-pkgs': './node_modules/@rancher/shell/scripts/serve-pkgs',
+  'publish-pkgs': './node_modules/@rancher/shell/extension/scripts/publish',
```
aalves08, line 21: shouldn't this be `./node_modules/@rancher/shell/scripts/extension/publish` instead of `./node_modules/@rancher/shell/extension/scripts/publish` ?

aalves08 (commented):
Left a comment.

I've compared `helmpatch` with a current version and mainly detected changes on copying the annotations. As for the `publish` there are more changes, but I have a bit of a hard time following them, so I am going to trust you on those. 🙏

shell/creators/pkg/init
nwmac: @aalves08 Yes, you're right. Fixed

aalves08 (approved):
LGTM

PR #8199, opened 2023-02-17
PSA: Update PSA default

### LINKED ISSUE #8032: PSA: Update PSA default

cnotv:
### Summary
Fixes #8032

Missing requirement: Rename  PSP label in RKE2 cluster editor

### Screenshot/Video

Screenshot 2023-02-17 at 15 58 38

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -1796,7 +1796,7 @@ cluster:
       override: Allow PSA Default Template to be overridden when using a CIS Profile
     defaultPodSecurityPolicyTemplateName:
       label: Default Pod Security Policy
-      option: RKE Default
+      option: Default - RKE2 Embedded
     defaultPodSecurityAdmissionConfigurationTemplateName:
       label: Default Pod Security Admission
       option: 


aalves08 (approved):
LGTM

PR #8201, opened 2023-02-17
Extensions test script: ensure build fails when one of the build steps fails

### LINKED ISSUE #8179: Issue with `check-plugins-build` ->  `Validate Plugin Build System` github action
**Describe the bug**
The script on github action `check-plugins-build` ->  `Validate Plugin Build System` catches a problem with an import of a path of an extension to shell/dashboard code, but the validation does not fail, since the step is validated.
An example here: https://github.com/rancher/dashboard/actions/runs/4137433409/jobs/7152595631

**To Reproduce**
go to a page/component on the `/shell` folder and do an import of a method from an extension, example:
`import { something } from '@/pkg/harvester-manager/config/harvester-manager'`

Create a PR from those changes and check the github action `check-plugins-build` ->  `Validate Plugin Build System`

**Result**
validation does not fail

**Expected Result**
validation should fail so that the code won't enter shell/dashboard

nwmac:
Fixes #8179 

To improve logging, we pipe the output of some commands into `cat`. Unfortunately, this means if the command fails, the error does not cause the script to fail - we do `set -e` which does this, but this does not catch the errors lost in the pipe.

To fix this, we need to add:

```
set -o pipefail
```

so that errors in the pipe fail the script via `set -e`.

This PR fixes this.

### DIFF
--- a/shell/scripts/test-plugins-build.sh
+++ b/shell/scripts/test-plugins-build.sh
@@ -91,6 +91,10 @@ ${SHELL_DIR}/scripts/publish-shell.sh
 yarn build:lib
 yarn publish:lib
 
+# We pipe into cat for cleaner logging - we need to set pipefail
+# to ensure the build fails in these cases
+set -o pipefail
+
 if [ "${SKIP_STANDALONE}" == "false" ]; then
   DIR=$(mktemp -d)
   pushd $DIR > /dev/null


aalves08 (approved):
LGTM. Thanks for fixing this! 🙏

PR #8217, opened 2023-02-21
Projects/Namespaces, istio enabled icon

### LINKED ISSUE #7951: Namespaces no longer show if istio auto-injection is enabled
Internal ref: SURE-4624

In the Old UI, The Namespace list showed an icon next to the namespace in the list view if Istio auto-injection was enabled.

This is determined by the namespace having the label `istio-injection` set to `enabled`.

image

The new UI does not.

We should:

- Show the istio icon next to the namespace in the list view when the`istio-injection` set to `enabled`. This should have a hover tooltip that says `Istio automatic sidecar injection is enabled for this namespace`
- On the namespace detail view, we should show in the masthead if auto-injection is enabled

torchiaf:
### Summary
Fixes #7951

### Screenshot/Video

image

image

### DIFF
--- a/shell/models/namespace.js
+++ b/shell/models/namespace.js
@@ -260,4 +260,8 @@ export default class Namespace extends SteveModel {
   get hideDetailLocation() {
     return !!this.$rootGetters['currentProduct'].hideNamespaceLocation;
   }
+
+  get isIstioInjectionEnabled() {
+    return this.metadata.labels[ISTIO_LABELS.AUTO_INJECTION] === 'enabled';
+  }
 }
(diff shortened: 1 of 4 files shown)


shell/models/namespace.js
```diff
@@ -260,4 +260,8 @@ export default class Namespace extends SteveModel {
   get hideDetailLocation() {
     return !!this.$rootGetters['currentProduct'].hideNamespaceLocation;
   }
+
+  get isIstioInjectionEnabled() {
```
nwmac, line 264: @torchiaf There already is `injectionEnabled` defined on the model that we can use

nwmac (changes requested):

shell/models/namespace.js
torchiaf: Updated, Thanks.

nwmac (approved):

PR #8239, opened 2023-02-23
what's new for 2.7.2

### LINKED ISSUE #8050: PSP Removal: Update what's new document to mention PSP issues with installed Helm Charts
Need to update the what's new document and add a note to draw users' attention to the removal of PSPs in k8s 1.25+ and the need to check/update Helm applications.

gaktive:
### Summary
Fixes #8050 

Since PSAs will need some delicate work for users to navigate as they upgrade to Kubernetes 1.25, we're bringing back this file to ensure we warn folks early on that if they use PSPs, they'll need to migrate to PSAs if they go up to 1.25.

### DIFF
--- a/shell/content/docs/en-us/whats-new.md
+++ b/shell/content/docs/en-us/whats-new.md
@@ -2,6 +2,16 @@
 title: What's New in 2.7
 ---
 
+## New in 2.7.2
+
+### Kubernetes 1.25 and PSAs
+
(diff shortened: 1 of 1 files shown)


nwmac (approved):

nwmac:
LGTM to me @gaktive

ronhorton:
There's verbiage now indicating that psps have been removed in 2.7.2 which impacts upgrades.

PR #8244, opened 2023-02-23
HARVESTER: Fix Harvester v1.1.1 cluster members page is unavailable

n313893254:
### Summary
Fix Harvester v1.1.1 cluster members page is unavailable

- https://github.com/rancher/rancher/issues/40663
- https://github.com/harvester/harvester/issues/3515

### Technical notes summary

The root cause is [here](https://github.com/rancher/dashboard/pull/7245/files#diff-dd84197114a85e2c4f6c8f3af7938f1c603d9d82162f7aa6595c76a0437fc400), `steve` change to `clusterroletemplatebinding`, that makes v1.1.1 Harvester cannot obtain cluster RTB as expected. 

For v1.1.1 Harvester is 
```
    const steveBindings = await Promise.all(
      clusterRoleTemplateBindings.map(b => b.steve)
    );
 ```
 
 For v2.7-head Rancher is
 ```
     clusterRoleTemplateBindings() {
      return this.normanClusterRoleTemplateBindings.map(b => b.clusterroletemplatebinding) ;
    },
 ```

### DIFF
--- a/shell/models/clusterroletemplatebinding.js
+++ b/shell/models/clusterroletemplatebinding.js
@@ -21,4 +21,8 @@ export default class CRTB extends NormanModel {
   get clusterroletemplatebinding() {
     return this.$rootGetters[`management/byId`](MANAGEMENT.CLUSTER_ROLE_TEMPLATE_BINDING, this.id?.replace(':', '/'));
   }
+
+  get steve() {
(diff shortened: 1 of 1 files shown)


nwmac:
@n313893254 There was an issue with a new dependency getting picked up which cause the extensions check to fail - this is now fixed, but you will need to rebase this PR with master to get that fix and for the gate to then pass

mantis-toboggan-md (approved):
lgtm

PR #8245, opened 2023-02-23
Fix broken extensions check

nwmac:
`@types/node` got bumped today - this gets picked up by the extensions build which breaks it.

This PR pins it do the v6 node types.


aalves08 (approved):
LGTM! 🙇

PR #8246, opened 2023-02-23
Add a warning alert when migrating to 1.25 if they have PSPs. 

### LINKED ISSUE #7553: Add a warning alert when migrating to 1.25 if they have PSPs.
Internal reference: SURE-5392
Related backend work: https://github.com/rancher/rancher/issues/38875

From @snasovich:

When v1.25 is introduced in Rancher, the k3s and rke2 upgrades will break. I'm not 100% sure what will happen when upgrading provisioned clusters, but imported clusters will fail due to them using system-upgrade-controller under the hood. This is because hardened k3s/rke2 clusters use PSPs, and PSPs are removed in v1.25. Currently, system-upgrade-controller does not do any mitigation for these.

For k3s, see k3s-io/k3s#6121
For rke2, the steps to do a SUC upgrade to v1.25 are:
 - Change config.yaml on each node to use profile: cis-1.23 instead of profile: cis-1.6
- Run standard system-upgrade-controller steps. This step, from the Rancher side, just means selecting the v1.25 version desired from the dropdown when editing the cluster through the UI.

Describe alternatives you've considered

Lots of warnings everywhere
Update system-upgrade-controller itself or the Plans or the upgrade images to do any workarounds automagically.

Other notes:

For upgrading provisioned rke2, lock the only profile version option to cis-1.23. If there was previously a selection of cis-1.6 or cis-1.5, give the user a warning. For upgrading provisioned k3s, give the user a warning with instructions on which values to select. For upgrading imported rke2, have the user perform the steps mentioned above first on their nodes, then use the Rancher UI to upgrade. For upgrading imported k3s, provide a warning that the upgrade should be performed manually, and do not give any option to upgrade from a <v1.25 cluster to a >=v1.25.

cnotv:
### Summary
Fixes #7553

- Added requirement to have PSPs enabled for the current management cluster
- Check is done only for RKE2, editing mode of provisioned clusters (no `reconciling`)

### Technical notes summary
- Added PSPS type
- Corrected with optional chaining for provisioned cluster model while retrieving `mgmtClusterId`

- Create and complete provisioning of RKE2 cluster with version 1.23.x or 1.24.x
- PSP will be created within the cluster, you can explore it and search Pod Security Policies to ensure it
- Edit cluster configuration and select higher K8S version 1.25.x
- A banner will appear communicating the removal of PSPs

Note: It is NOT possible to create a cluster RKE2 without PSPs from the UI.

### Screenshot/Video

https://user-images.githubusercontent.com/5009481/220956862-de3414ca-aeca-4fdf-aac6-45af5924c715.mov

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -1619,7 +1619,7 @@ cluster:
     os: 'You are attemping to add a {newOS} worker node to a cluster with one or more {existingOS} worker nodes: some installed apps may need to be upgraded or removed.'
     rke2-k3-reprovisioning: 'Making changes to cluster configuration may result in nodes reprovisioning. For more information see the <a target="blank" href="{docsBase}/cluster-provisioning/rke-clusters/behavior-differences-between-rke1-and-rke2/" target="_blank" rel="noopener nofollow">documentation</a>.'
     desiredNodeGroupWarning: There are 0 nodes available to run the cluster agent. The cluster will not become active until at least one node is available.
-    invalidPsps: You have one or more PodSecurityPolicy resource(s) in this cluster. Pod Security Policies are not available in Kubernetes v1.25.
+    invalidPsps: You have one or more PodSecurityPolicy resource(s) in this cluster. Pod Security Policies are not available in Kubernetes v1.25 and will be automatically removed.
     haveArgInfo: Configuration information is not available for the selected Kubernetes version.  The options available in this screen will be limited, you may want to use the YAML editor.
     deprecatedPsp: Pod Security Policies are deprecated as of Kubernetes v1.21, and have been removed in Kubernetes v1.25.
     removedPsp: Pod Security Policies have been removed in Kubernetes v1.25, use PodSecurity Admission instead.
--- a/shell/edit/provisioning.cattle.io.cluster/rke2.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/rke2.vue
@@ -15,8 +15,7 @@ import {
   DEFAULT_WORKSPACE,
   SECRET,
   HCI,
-  PSP,
-  COUNT,
+  PSPS,
 } from '@shell/config/types';
 import { _CREATE, _EDIT, _VIEW } from '@shell/config/query-params';
 
@@ -131,8 +130,9 @@ export default {
   },
 
   async fetch() {
-    if (this.mode !== _CREATE) {
-      this.hasPsp = await this.checkPsp();
+    // Check presence of PSP in RKE2, which is where we show the templates
+    if (this.mode !== _CREATE && !this.isK3s) {
+      this.psps = await this.checkPsps();
     }
 
     if ( !this.rke2Versions ) {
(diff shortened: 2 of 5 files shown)


nwmac:
@cnotv  There was an issue with a new dependency getting picked up which cause the extensions check to fail - this is now fixed, but you will need to rebase this PR with master to get that fix and for the gate to then pass

cnotv:
Getting this error in E2E 🤔 
Screenshot 2023-02-24 at 10 52 26

shell/edit/provisioning.cattle.io.cluster/rke2.vue
```diff
@@ -2166,23 +2196,20 @@ export default {
             {{ t('cluster.rke2.security.header') }}
           </h3>
           <Banner
-            v-if="isEdit && displayInvalidPspsBanner"
+            v-if="isEdit && displayInvalidPspsBanner && hasPsps"
```
torchiaf, line 2199: If I remember correctly, we should display the warning banner (`invalidPsps`) only if there are PSPs in the Cluster, not the `templates`.
Given this, we should remove `displayInvalidPspsBanner` from the `v-if` statement, because it's checking for the PSPs templates; as it works now, the warning message is shown only when both PSPs `and` PSPs templates are in the Cluster

shell/assets/translations/en-us.yaml
```diff
@@ -1619,7 +1619,7 @@ cluster:
     os: 'You are attemping to add a {newOS} worker node to a cluster with one or more {existingOS} worker nodes: some installed apps may need to be upgraded or removed.'
     rke2-k3-reprovisioning: 'Making changes to cluster configuration may result in nodes reprovisioning. For more information see the <a target="blank" href="{docsBase}/cluster-provisioning/rke-clusters/behavior-differences-between-rke1-and-rke2/" target="_blank" rel="noopener nofollow">documentation</a>.'
     desiredNodeGroupWarning: There are 0 nodes available to run the cluster agent. The cluster will not become active until at least one node is available.
-    invalidPsps: You have one or more PodSecurityPolicy resource(s) in this cluster. Pod Security Policies are not available in Kubernetes v1.25.
+    invalidPsps: You have one or more PodSecurityPolicy resource(s) in this cluster. Pod Security Policies are not available in Kubernetes v1.25 and will be automatically removed.
```
torchiaf, line 1622: Should we re-phrase this warning message ? I'm not sure but did @nwmac propose a new message during the last meeting?

shell/edit/provisioning.cattle.io.cluster/rke2.vue
cnotv: We are checking the presence of PSPs, which include the use of templates.

Good point about the code deprecation.

shell/assets/translations/en-us.yaml
cnotv: Yes there was also a request to inform the user about the fact that the PSPs will be removed and not just deprecated.

torchiaf (approved):
LGTM

PR #8254, opened 2023-02-24
enhancement use dynamic store

WuJun2016:
### Summary
When in the harvester cluster, we should request `schema` from the harvester store.

Fixes #

- https://github.com/harvester/harvester/issues/3537

### DIFF
--- a/shell/store/index.js
+++ b/shell/store/index.js
@@ -241,6 +241,12 @@ export const getters = {
     return out;
   },
 
+  getStoreNameByProductId(state) {
+    const products = state['type-map']?.products;
+
+    return (products.find(p => p.name === state.productId) || {})?.inStore || 'cluster';
+  },
+
   currentStore(state, getters) {
     return (type) => {
       const product = getters['currentProduct'];
--- a/shell/utils/monitoring.js
+++ b/shell/utils/monitoring.js
@@ -26,9 +26,10 @@ export function haveV2Monitoring(getters) {
   if (haveV1Monitoring(getters)) {
     return false;
   }
+  const inStore = getters['getStoreNameByProductId'];
 
   // Just check for the pod monitors CRD
-  const schemas = getters[`cluster/all`](SCHEMA);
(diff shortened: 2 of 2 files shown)


WuJun2016:
Hi， @mantis-toboggan-md   This is a block issue, which requires merging in v2.7.2，  PTAL, thanks

richard-cox (commented):
There are a couple more places that `haveV2Monitoring` is called and don't have the custom store name passed through (`ifHave`, `monitoringStatus`). Is it not needed for them?

There is a helper that should be able to retrieve if the current context is `harvester` or `cluster`
```
const inStore = this.$store.getters['currentProduct'].inStore;
```
Would that help for these cases?

richard-cox (changes requested):
See questions/comment in https://github.com/rancher/dashboard/pull/8254#pullrequestreview-1315938206

WuJun2016:
> There are a couple more places that `haveV2Monitoring` is called and don't have the custom store name passed through (`ifHave`, `monitoringStatus`). Is it not needed for them?
> 
> There is a helper that should be able to retrieve if the current context is `harvester` or `cluster`
> 
> ```
> const inStore = this.$store.getters['currentProduct'].inStore;
> ```
> 
> Would that help for these cases?

@richard-cox  `this.$store.getters['currentProduct'].instore` cannot be called in `haveV2Monitoring`, which would result in nested calls, resulting in an endless loop。  So I use a new function to get the `storeName` of the current `activeProduct`.

Can you help to see if this will cause any other problems ？  thanks

richard-cox (approved):
Ok, I understand now, the reliance of `haveV2Monitoring `--> `currentProduct` / `currentStore` --> `activeProduct` --> `ifHave` --> `haveV2Monitoring`.

I had a quick look to see how easy it was to add all the plumbing in for `inStore` as a param and there would be a lot for `allDashboardsExist`.

This solution is more generic and will have less impact

PR #8263, opened 2023-02-24
invalid error handling creating workload

### LINKED ISSUE #8262: Broken error handling on create/edit workloads
**Setup**
- Rancher version: `v2.7-be585d369a71f4fbe88ef73865d159b0a2968f1c-head`
- UI version: `master e3a3e00b8`

**Describe the bug**
When trying to create a StatefulSet via YAML with errors, hitting `create` causes a blocking error on the interface

**To Reproduce**
- Go to local cluster -> Create -> From YAML
- Use an invalid YAML, such as: https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/ (has two resources `Service` and `StatefulSet`

**Result**
Blocking UI error

**Expected Result**
Should handle errors from server correctly

**Screenshots**
Screenshot 2023-02-24 at 16 47 20

aalves08:
Fixes #8262

- Fix error handling

To test:
- Repro steps on issue #8262

### DIFF
--- a/shell/edit/workload/index.vue
+++ b/shell/edit/workload/index.vue
@@ -37,15 +37,25 @@ export default {
      */
     mapError(error) {
       switch (true) {
-      case error.includes('violates PodSecurity'): {
-        const match = error.match(/\"(.*?)\"/gi);
-        const name = match[0];
-        const policy = match[1];
+      // for objects (ex: YAMLexceptions)
+      case typeof error === 'object' && !Array.isArray(error) && error !== null: {
+        return { message: error.message || '' };
+      }
+      // for strings
+      case typeof error === 'string': {
+        // PSPs related
+        if (error.includes('violates PodSecurity')) {
+          const match = error.match(/\"(.*?)\"/gi);
+          const name = match[0];
+          const policy = match[1];
 
-        return {
-          message: `Pod ${ name } Security Policy Violation ${ policy }`,
-          icon:    'icon-pod_security'
-        };
+          return {
+            message: `Pod ${ name } Security Policy Violation ${ policy }`,
+            icon:    'icon-pod_security'
+          };
+        }
+
+        return { message: error };
       }
 
       default:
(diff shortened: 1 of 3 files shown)


shell/edit/workload/index.vue
```diff
@@ -37,15 +37,25 @@ export default {
      */
     mapError(error) {
       switch (true) {
-      case error.includes('violates PodSecurity'): {
```
cnotv, line 40: This is meant to be a map, so it have certain cases to be mapped into something else. Although currently we have just one, the new design is aimed to map errors messages into something else and use icons.

We have just to prevent the object case to break and keep the rest of the logic. The `switch` needs to be kept for the substring check, till we have a better logic in place. The way to retrieve the message is the only difference we access to it.

Basically you can bring everything as it was before and just do the following ahead:

```ts
      // If we check for null, we can at this point extend it to undefined, so it does not break in case of undefined
      const isObject = error && typeof error === 'object' && !Array.isArray(error);
      // We have just 2 cases, so we'll set a ternary operation
      const errorMessage = isObject ? error.message || '' : error || '';

      switch (true) {
      case errorMessage.includes('violates PodSecurity'): {
        const match = errorMessage.match(/\"(.*?)\"/gi);
        const name = match[0];
        const policy = match[1];
        return {
          message: `Pod ${ name } Security Policy Violation ${ policy }`,
          icon:    'icon-pod_security'
        };
      }
      default:
        break;
      }
```

cnotv (changes requested):

shell/edit/workload/index.vue
```diff
@@ -37,15 +37,25 @@ export default {
      */
     mapError(error) {
```
cnotv, line 38: Please add a note related to the different types of errors, since we do not use Typescript, nor we have tests in place.

cnotv (approved):
LGTM

PR #8268, opened 2023-02-27
statefulsets pods list bug

### LINKED ISSUE #7555: UI list pods from multiple statefulset if they share the same label
Internal reference: SURE-5576
Reported in 2.6.9

If there are two StatefulSets with the same label. The UI will list the pods from both StatefulSet.
As you can see in the screenshots below, when selecting StatefulSet busybox-1, the user can list the pods of both busybox-1 and busybox-2 because they share the same label (app: busybox). (This is from 2.6.9)

And when selecting any of the two StatefulSets, the user can see pods from both StatefulSets

It seems even kubectl doesn’t have the ability to list pods for specific StatefulSet. We can only use labels. Any guidance here will be appreciated.

Issue also affects old version 2.5

Business impact:

UI list pods from multiple StatefulSet if they share the same label

Troubleshooting steps:

Repro steps:

- Create a Rancher cluster
- Deploy two Statefulset with the same label
- Go to the Statefulset page in the UI, and it will list pods from both StatefulSets

No workaround

Expected behavior:

UI should not list pods from multiple StatefulSets if they share the same label

aalves08:
Fixes #7555 

- Fix issue with incorrect list of `pods` on the `statefulset` details view

## To test
- Go to local cluster -> `Workloads` -> `statefulset` -> create
- Create two `statefulset` via YAML (just change the name field):

```yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: web1
  namespace: default
spec:
  selector:
    matchLabels:
      app: nginx # has to match .spec.template.metadata.labels
  serviceName: “nginx”
  replicas: 3 # by default is 1
  minReadySeconds: 10 # by default is 0
  template:
    metadata:
      labels:
        app: nginx # has to match .spec.selector.matchLabels
    spec:
      terminationGracePeriodSeconds: 10
      containers:
      - name: nginx
        image: [registry.k8s.io/nginx-slim:0.8](http://registry.k8s.io/nginx-slim:0.8)
        ports:
        - containerPort: 80
          name: web
        volumeMounts:
        - name: www
          mountPath: /usr/share/nginx/html
  volumeClaimTemplates:
  - metadata:
      name: www
    spec:
      accessModes: [ “ReadWriteOnce” ]
      storageClassName: “my-storage-class”
      resources:
        requests:
          storage: 1Gi
```
- enter the details view of each `statefulset` and make sure only one `pod` appears per `statefulset`

### DIFF
--- a/shell/detail/workload/index.vue
+++ b/shell/detail/workload/index.vue
@@ -108,8 +108,8 @@ export default {
   computed: {
     ...mapGetters(['currentCluster']),
 
-    isActive() {
-      return this.value.metadata.state.name === 'active';
+    isScalable() {
+      return this.value?.canUpdate;
     },
 
     isJob() {
@@ -335,7 +335,7 @@ export default {
         class="text-right"
         :label="t('tableHeaders.scale')"
         :value="value.spec.replicas"
-        :disabled="!isActive"
+        :disabled="!isScalable"
         @minus="scaleDown"
         @plus="scaleUp"
       />
--- a/shell/detail/workload/index.vue
+++ b/shell/detail/workload/index.vue
@@ -108,8 +108,8 @@ export default {
   computed: {
     ...mapGetters(['currentCluster']),
 
-    isScalable() {
-      return this.value?.canUpdate;
+    isActive() {
+      return this.value.metadata.state.name === 'active';
     },
 
     isJob() {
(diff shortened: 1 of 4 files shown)


torchiaf (changes requested):
It seems that it breaks the pods filter, because I can see also the pods from other workloads resources, like `Deployments`:

image

torchiaf (approved):
LGTM

PR #8286, opened 2023-02-28
Reserved memory not updating correctly from the nodes view on Rancher UI

### LINKED ISSUE #7598: Reserved memory not updating correctly from the nodes view on Rancher UI
Internal reference: SURE-5598
Reported in 2.6.9

Issue description:
The user set up the configuration for reserving memory and CPU for the operational system and for Kubernetes resources. In the stress tests, everything worked as expected. Also, describing the nodes shows the difference that they configured between the capacity and the allocatable values. However, their question is related to how this information is displayed in the webUI. When they click on nodes >> Select the node, the memory information still shows the node's total value, but the CPU info was updated.

Business impact:
This can cause the information to be displayed that does not match the correct settings.

Troubleshooting steps:
No troubleshooting was done as it seems to be a webUI bug

Repro steps:

This was the configuration on rancher in order to reserve the compute resources:
```
 kubelet:
      extra_args:
        enforce-node-allocatable: 'pods,kube-reserved,system-reserved'
        eviction-hard: 'memory.available<100Mi,nodefs.available<10%'
        kube-reserved: 'cpu=500m,memory=1Gi'
        kube-reserved-cgroup: /kube-reserved.slice
        system-reserved: 'cpu=1000m,memory=2Gi'
        system-reserved-cgroup: /system.slice
```

Also, they created these groups that are not created by default in the server machines, both from kubernetes (mykube) and for the OS (system.slice):

```
/sys/fs/cgroup/hugetlb/mykube
/sys/fs/cgroup/cpu/mykube
/sys/fs/cgroup/cpuacct/mykube
/sys/fs/cgroup/cpuset/mykube
/sys/fs/cgroup/memory/mykube
/sys/fs/cgroup/systemd/mykube
/sys/fs/cgroup/pids/mykube
/sys/fs/cgroup/hugetlb/system.slice
/sys/fs/cgroup/cpu/system.slice
/sys/fs/cgroup/cpuacct/system.slice
/sys/fs/cgroup/cpuset/system.slice
/sys/fs/cgroup/memory/system.slice
/sys/fs/cgroup/systemd/system.slice
/sys/fs/cgroup/pids/system.slice
```

Workaround:
None

Actual behavior:
The reserved memory and total memory reserved were updated correctly, but from clicking on nodes under the webUI it shows the old information.

Expected behavior:
The webUI needs to update the correct value for the nodes

From the page that shows overall the usage of resources on the cluster, the reserved memory and total memory were updated correctly. 

However, when clicking on a node, only the CPU was updated with the new allocatable memory. The total memory value should appear with the decrease in reserved memory, as it does with the CPU.

Here's what `kubectl describe node` shows in this example:
```
Capacity:
    cpu: 8
    ephemeral-storage: 18422Mi hugepages-1Gi: 0 hugepages-2Mi: 0
    memory: 9766740Ki
    pods: 110
Allocatable:
    cpu: 6500m
    ephemeral-storage: 17385180337 hugepages-1Gi: 0 hugepages-2Mi: 0
    memory: 6518612Ki
    pods: 110
```

So they would expect that the memory value for the node would also be set to 6.5GB and not 9.31 as shown in the node in the GUI.

The ask: show the reserved memory for the node. It appears that the UI shows the node's total memory instead of what is allocatable when looking at the node view. It sounds like a missing feature or a bug in calculation with a confusing label.

cnotv:
### Summary
Fixes #7598

- Added reserved bar for RAM in Node details

### Screenshot/Video
Screenshot 2023-02-28 at 09 25 54

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -3499,6 +3499,7 @@ node:
     glance:
       consumptionGauge:
         used: Used
+        reserved: Reserved
         amount: "{used} of {total} {unit}"
         cpu: CPU
         memory: MEMORY
--- a/shell/detail/node.vue
+++ b/shell/detail/node.vue
@@ -229,13 +229,24 @@ export default {
         :capacity="value.cpuCapacity"
         :used="value.cpuUsage"
       />
-      <ConsumptionGauge
-        :resource-name="t('node.detail.glance.consumptionGauge.memory')"
-        :capacity="value.ramCapacity"
-        :used="value.ramUsage"
-        :units="memoryUnits"
-        :number-formatter="memoryFormatter"
-      />
+      <div>
+        <ConsumptionGauge
+          :resource-name="t('node.detail.glance.consumptionGauge.memory')"
+          :capacity="value.ramCapacity"
+          :used="value.ramUsage"
+          :units="memoryUnits"
+          :number-formatter="memoryFormatter"
+        />
+        <ConsumptionGauge
+          class="mt-20"
+          :resource-name="t('node.detail.glance.consumptionGauge.reserved')"
+          :used-as-resource-name="true"
+          :capacity="value.ramReserved"
+          :used="value.ramUsage"
+          :units="memoryUnits"
+          :number-formatter="memoryFormatter"
+        />
+      </div>
       <ConsumptionGauge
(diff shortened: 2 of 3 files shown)


cnotv:
PA not anymore relevant to the issue.

PR #8288, opened 2023-02-28
fix(RegistryConfigs): fix missing TLS Secret options issue

### LINKED ISSUE #8287: Create a RKE2 cluster, TLS Secret only shows the None option only
**Setup**
- Rancher version: v2.7.1
- Browser type & version: Chrome 110.0.5481.100

**Describe the bug**

The TLS Certification was successfully created in the Local cluster, but when creating the RKE2 cluster, it is impossible to select the TLS Certification of created in the `registry authentication`

**To Reproduce**

1. Create TLS Certification in the Local cluster and select the `fleet-default` namespace.
2. Create a RKE2 cluster and set the `registries` parameter
3. Select `Configure advanced container mirroring and registry authentication options`, select` TLS Secret`

**Result**
The drop-down box only has `None` as an option, and the TLS Secret created in the local cluster is not displayed

**Expected Result**

The TLS Secret created in the local cluster can be displayed in the drop-down box
**Screenshots**

secret

ly5156:
### Summary
Fixes #8287

fix(RegistryConfigs): fix missing TLS Secret options issue

### DIFF
--- a/shell/components/form/SecretSelector.vue
+++ b/shell/components/form/SecretSelector.vue
@@ -55,6 +55,10 @@ export default {
     mode: {
       type:    String,
       default: _EDIT
+    },
+    inStore: {
+      type:    String,
+      default: 'cluster',
     }
   },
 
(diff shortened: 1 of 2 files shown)


nwmac (approved):
LGTM

PR #7010, opened 2022-09-27
HARVESTER: add harvester feature

WuJun2016:
### Summary
Fixes # 
- https://github.com/harvester/harvester/issues/2816
- harvester feature

PR #6302, opened 2022-07-08
Can select secrets for receivers in AlertmanagerConfig

catherineluse:
Addresses a bug which caused https://github.com/rancher/dashboard/issues/5875 to be reopened by QA.

The TL;DR was that after https://github.com/rancher/dashboard/issues/5875 was reviewed and merged, I made another change https://github.com/rancher/dashboard/pull/5999 that broke the secret selector for the AlertmanagerConfig receivers. This PR fixes that.

The more detailed reason for this PR is that the original event listener that I had, which is `option:selecting`, never actually worked:

```
@option:selecting="$emit('input', $event)"
```
I just thought it worked because the `input` event was being emitted by something else anyway. So when I started listening for an event with a different name in this PR https://github.com/rancher/dashboard/pull/5999, the event that I wanted was not emitted anymore. But since I changed it from `@option:selecting` to `@option:selected`, now the event is actually emitted.

PR #8327, opened 2023-03-02
Fix alertmanager receiver secret inputs

### LINKED ISSUE #8324: AlertmanagerConfig doesn't appear to save dropdown fields on Rancher v2.6
**Setup**
- Rancher version: `v2.6-head`, and also `v2.6.11-rc9`
- Browser type & version: Google Chrome `110.0.5481.177`, Mozilla Firefox `109.0.1`
- Monitoring Chart version: `100.2.0+up40.1.2`

This bug was tested also in `v2.7-head` and is **not** present in the newer version; dropdowns behave correctly in v2.7.

**Describe the bug**
When editing email receivers under Monitoring->AlertManagerConfig, it seems that all of the form fields are saving correctly except for dropdown elements, which affects secret selection. The form _appears_ to save correctly, but the contents of the dropdown do not make it into the resulting `.yaml` configuration.

**To Reproduce**
First, create a secret under Storage->Secrets. Create a new AlertManageConfig, then add an email receiver. Along with the other required information, attempt to select the secret you created, along with the key. Then save the receiver. 

**Result**
To verify the buggy behavior, open the created receiver back up and inspect the Email configuration; now the dropdown is empty.

This bug also occurs when editing existing receivers. If the `.yaml` already contains valid secret configuration then this is correctly populated into the dropdown, but any _changes_ to the dropdowns still do not save. In this case, it's impossible to switch to a different secret using the form alone. To workaround, one must hand-edit the YAML config instead.

**Expected Result**
The value the user enters into the form should be the value that is saved in the underlying `.yaml`

**Screenshots**
email_receiver_broken_dropdowns

**Additional context**
I bumped into this while investigating SURE-4788 and believe it to be related.

mantis-toboggan-md:
### Summary
Fixes #8324

This appears to be a regression, it was fixed here https://github.com/rancher/dashboard/pull/6302 and re-broken here: https://github.com/rancher/dashboard/pull/7010/files#diff-5316c4da698262a356a5f70f02f4d94b95c66b780cedf9cb04515deb011017d8R281

### DIFF
--- a/shell/components/form/LabeledSelect.vue
+++ b/shell/components/form/LabeledSelect.vue
@@ -278,7 +278,7 @@ export default {
       @search:focus="onFocus"
       @search="onSearch"
       @open="onOpen"
-      @option:selecting="$emit('selecting', $event)"
+      @option:selected="$emit('selecting', $event)"
       @option:deselecting="$emit('deselecting', $event)"
     >
       <template #option="option">


rak-phillip (approved):
Thanks @mantis-toboggan-md, great find 🚀

LGTM - tested in development environment against v2.6.11-rc9. Works as expected.

rak-phillip:
No associated change is required against `master`. This is not an issue in 2.7.x.

Issue #8346, opened 2023-03-06
Wrong URL for embedded Grafana dashboards and links in local cluster


bashofmann:
**Setup**
- Rancher version: 2.7.2-rc5
- Browser type & version: Chrome
- Rancher-Monitoring: 102.0.0+up40.1.2

**Describe the bug**
With the new version of Rancher-Monitoring, the embedded Grafana dashboards and links in the local cluster do not work anymore, but return a 404 from Grafana.

In all other clusters the links are correct.

**To Reproduce**
* Install the newest monitoring version in the local cluster
* Go to any page that embeds a Grafana dashboard, e.g. the Cluster Explorer overview page.

**Result**
A 404 page is displayed.

**Expected Result**
The correct dashboard is embedded.

**Screenshots**
![Bildschirm­foto 2023-03-06 um 09 39 46](https://user-images.githubusercontent.com/243056/223060387-4c5305dc-4f74-4c04-9b67-134400f978e0.png)

**Additional context**
The used iframe URL in the local cluster has this pattern:

```
https://192.168.178.21:8005/api/v1/namespaces/cattle-monitoring-system/services/http:rancher-monitoring-grafana:80/proxy/api/dashboards/uid/rancher-etcd-1
```

The correct URL for the new monitoring version is

```
https://192.168.178.21:8005/k8s/clusters/local/api/v1/namespaces/cattle-monitoring-system/services/http:rancher-monitoring-grafana:80/proxy/?orgId=1
```

The dashboard currently has an exception to create this URL differently for the local cluster.

SURE-5995

rancherbot:
@richard-cox, Not creating `backport` issue for issue `8346` in repository `dashboard` because milestone `2.7.2` does not exist or is not an open milestone

PR #8347, opened 2023-03-06
Fix Grafana URLs in local cluster for new monitoring version

bashofmann:
### Summary
Fixes https://github.com/rancher/dashboard/issues/8346

With the new version of Rancher-Monitoring, the embedded Grafana dashboards and links in the local cluster do not work anymore, but return a 404 from Grafana.

In all other clusters the links are correct.

This changes an exception in the creation of these links for the local cluster, which is not needed anymore in the new monitoring version >=102.0.0+up40.1.2.

Embedded Grafana dashboards and Grafana links in the local cluster.

* Cluster Overview page
* Workload detail page
* Pod detail page
* Node detail page
* Grafana link on Monitoring detail page

### DIFF
--- a/shell/components/EtcdInfoBanner.vue
+++ b/shell/components/EtcdInfoBanner.vue
@@ -8,7 +8,8 @@ import { CATALOG } from '@shell/config/types';
 export default {
   components: { Banner, Loading },
   async fetch() {
-    const res = await this.$store.dispatch(`cluster/find`, { type: CATALOG.APP, id: 'cattle-monitoring-system/rancher-monitoring' });
+    const inStore = this.$store.getters['currentProduct'].inStore;
+    const res = await this.$store.dispatch(`${ inStore }/find`, { type: CATALOG.APP, id: 'cattle-monitoring-system/rancher-monitoring' });
     const monitoringVersion = res?.currentVersion;
     const leader = await hasLeader(monitoringVersion, this.$store.dispatch, this.currentCluster.id);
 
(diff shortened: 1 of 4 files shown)


bashofmann:
This also fixes project monitoring dashboards in the local cluster.
Bildschirm­foto 2023-03-06 um 16 54 42

richard-cox:
@bashofmann Similar to the other PR, i've updated this and issue to 2.7 Q2. I think we need a fix for 2.7.2 as well, can you create the backported issue and PR?

In terms of the fix, we need to be able to support previous versions of the monitoring chart as well. Can you confirm the fix works for both old and new charts?

bashofmann:
@richard-cox Unfortunately, the old URL only works with the old version and the new one only with the new version :-(. I could check which version is installed there, but this looked a bit more complicated. I'll try to think about this a bit more, maybe there's something simpler. Do you have any idea of the top of your head?

richard-cox:
@bashofmann I've created https://github.com/rancher/dashboard/issues/8380 to backport the fix to 2.7.2. We're on a pretty tight deadline there.

To check which version of monitoring is installed you should be able to grab the specific installed app. This ONLY works if the user cannot change the namespace at install time (can't check out the moment, i haven't got a system that hasn't got monitoring installed, or previously installed)

So something like....
```
   const res = await this.$store.dispatch(`cluster/request`, { type: CATALOG.APP, id: 'cattle-monitoring-system/rancher-monitoring' });
    const currentVersion = res?.currentVersion 
```
You can use shell/utils/version.js `compare` to check if the currentVersion is before `102.0.0+up40.1.2`

richard-cox (changes requested):
No comments on top of those in the PR, just aligning PR state with them

bashofmann:
@richard-cox I added the version compare in all places and also wrote a test to check if the comparison works correctly.

richard-cox (approved):

PR #8354, opened 2023-03-06
GitLab integration

torchiaf:

- add gitLab APIs
- convert UI components to support generic Git repos
- code refactoring

### Summary
Fixes https://github.com/epinio/ui/issues/173,
Fixes https://github.com/epinio/ui/issues/199


pkg/epinio/detail/applications.vue
```diff
@@ -178,42 +141,80 @@ export default Vue.extend<Data, any, any, any>({
...
-            commitId:  cur?.sha,
-            author:    cur.author,
-            isChecked: false,
-            date:      cur?.commit.committer.date
-          });
-
-          return acc;
-        }, []);
+    gitType() {
+      const envs = this.value?.envDetails;
```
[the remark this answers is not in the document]
torchiaf: Updated :+1:

torchiaf:
Thanks for the review @scures , I just added the git Icon based on the git source of the applications, in the `Deployment details` page.
I have opened an issue to request a new icon for GtiLab.

pkg/epinio/utils/git.ts
```diff
@@ -0,0 +1,119 @@
+import { EPINIO_APP_ENV_VAR_GIT } from '../types';
+import startCase from 'lodash/startCase';
```
richard-cox, line 2: Think we're trying t o back out of using lodash. There's a file with some basic string utils `/shell/utils/string.js`. Would `ucFirst` work instead?

shell/components/form/GitPicker.vue
```diff
@@ -1,13 +1,18 @@
...
 import debounce from 'lodash/debounce';
 import { isArray } from '@shell/utils/array';
-import LabeledSelect from '@shell/components/form/LabeledSelect';
-import LabeledInput from '@components/Form/LabeledInput/LabeledInput';
+import LabeledSelect from '@shell/components/form/LabeledSelect.vue';
+import LabeledInput from '@components/Form/LabeledInput/LabeledInput.vue';
 import some from 'lodash/some';
+import { GitUtils } from '@/pkg/epinio/utils/git';
+
+// TODO remove any types
```
richard-cox, line 12: TODO

shell/components/form/GitPicker.vue
```diff
@@ -140,13 +164,13 @@ export default {
           if (this.oldUsername !== this.selectedAccOrOrg) {
             this.oldUsername = this.selectedAccOrOrg;
 
-            // Resets state, just in case.
+            // Resets state, just in case. TODO to be removed because is called in reset()
```
richard-cox, line 167: TODO

pkg/epinio/utils/git.ts
```diff
@@ -0,0 +1,119 @@
...
+        htmlUrl:     data.html_url,
+        name:        data.name
+      }),
+      commit: (data: any) => ({
+        message:  data.commit?.message,
+        htmlUrl:  data.html_url,
+        sha:      getShortHash(data.sha),
+        commitId: data.sha,
+        author:   {
+          name:      data.author.login,
```
richard-cox, line 80: Bug - create app --> github --> richard-cox --> icons --> master caused a crash. There's a commit that we've fetched that has no auther, this fails.

If this happens on `epinio-dev` (i've got a feeling it does) can you create an issue? If not can you fix in this PR (we should fall back on something sensible)?

pkg/epinio/components/application/AppSource.vue
```diff
@@ -396,8 +397,15 @@ export default Vue.extend<Data, any, any, any>({
...
+    sourceValue() {
+      return {
+        ...this.source.git,
+        type: toLabel(this.type),
+      };
+    },
+
     type() {
-      // There's a bug in the select component which fires off the option ({ value, label}) instead of the value
+      // TODO There's a bug in the select component which fires off the option ({ value, label}) instead of the value
```
richard-cox, line 408: TODO

richard-cox (changes requested):
- Sorin called out an issue that might be tracked somewhere else. When the user selects gitlab and then org, branch, etc.... then changes the source type to github... the org/user field remains populated
- This is capture elsewhere - the icon for gitlab is missing

torchiaf:
> - Sorin called out an issue that might be tracked somewhere else. When the user selects gitlab and then org, branch, etc.... then changes the source type to github... the org/user field remains populated
> - This is capture elsewhere - the icon for gitlab is missing

Thanks for the review Richard! Yes, i tracked last week those bugs and I wanted to resolve the  in separated PRs, but since there are other changes to do, I will rather include all in this PR.

pkg/epinio/utils/git.ts
torchiaf: It happens only in this PR, just fixed.

pkg/epinio/components/application/AppSource.vue
torchiaf: Removed `unSafeType` variable. Do you remember in what case the selected option is of type` ({ value, label})`. I did several tests but I am not able to replicate the bug in this PR.

torchiaf:
@richard-cox I made some code changes to fulfill the requested changes except for the GitLab icon, since we don't have a GitLab icon yet: https://github.com/rancher/dashboard/issues/8407.
Could we open a separate issue for this and postpone the fix to when Manuele will provide the new icon?

pkg/epinio/models/application-action.js
```diff
@@ -165,9 +165,18 @@ export default class ApplicationActionResource extends Resource {
...
         git:  {
-          revision:   source.github.commit,
-          repository: source.github.url,
-          branch:     source.github.branch
+          revision:   source.git.commit,
+          repository: source.git.url,
+          branch:     source.git.branch
+        },
+      };
+    case APPLICATION_SOURCE_TYPE.GIT_LAB:
```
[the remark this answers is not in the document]
richard-cox: I think this has introduced a bug. Deploy step fails with `Json: cannot unmarshal string into Go struct field ApplicationOrigin.origin.Kind of type int`. The payload origin's `kind` is `git_lab`, which is an unsupported value (can only be git, see comment for APPLICATION_MANIFEST_SOURCE_TYPE). Looking at the original code it does seem like github was accepted, so not sure what's happening. needs investigation
```
kind: APPLICATION_MANIFEST_SOURCE_TYPE.GIT_HUB,
```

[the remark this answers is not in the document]
richard-cox: The same happens when trying to deploy github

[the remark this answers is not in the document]
richard-cox: We should set it to GIT and the inner type to source.type

Looking at 
```
    originLabel() {
      return this.value.sourceInfo.kind === APPLICATION_MANIFEST_SOURCE_TYPE.GIT ? ucFirst(this.gitType) : this.value.sourceInfo.label;
    }

    gitType() {
      if (this.envs[APPLICATION_ENV_VAR]) {
        const { type } = JSON.parse(this.envs[APPLICATION_ENV_VAR]) as EPINIO_APP_ENV_VAR_GIT;

        if (type) {
          return type.toLowerCase();
        }
      }

      return 'github';
    },
```

pkg/epinio/components/application/AppSource.vue
richard-cox: Could have been fixed in upstream. If you've tested each source then we can keep your changes

richard-cox (changes requested):
- applications that were previously deployed as `git` have app deployment details as `github` 
- see inline comments for other issues

PR #8386, opened 2023-03-08
Fixing build errors in drone

codyrancher:
Removed typescript from the vue.config files and resolved additional drone build errors.

Verified the following commands work: 

- `docker build --no-cache -t "test:Dockerfile" .`
- `./scripts/build-hosted`


richard-cox (approved):
The e2e gate was failing, I gave it a bump and now it's good. Will merge to see what happens with the builds

PR #8390, opened 2023-03-09
Added create name space button on flat list view

### LINKED ISSUE #8216: Create Namespaces Button not visible when using "Flat List" view
Internal Reference: SURE-5787

Some work was done here: https://github.com/rancher/dashboard/issues/5120

.. but as an admin, in the flat view on Projects/Namespaces, there is no button to create a namespace.

If you go to the grouped view, you can add a namespace in the 'Not in a Project' section.

You should be able to create a namespace in the flat view.

bisht-richa:
### Summary
Fixes #8216

Create Namespaces Button is not visible when using the "Flat List" view

### How to test

> Go to local cluster > Projects/Namespaces > click on `Flat list` view
> Create Namespaces Button should be there > click on `Create Namespaces Button`
> Create screen should have a dropdown of projects.

### DIFF
--- a/shell/components/ExplorerProjectsNamespaces.vue
+++ b/shell/components/ExplorerProjectsNamespaces.vue
@@ -292,7 +292,7 @@ export default {
       const location = this.createNamespaceLocationOverride ? { ...this.createNamespaceLocationOverride } : {
         name:   'c-cluster-product-resource-create',
         params: {
-          product:  this.$store.getters['currentProduct'].name,
+          product:  this.$store.getters['currentProduct']?.name,
           resource: NAMESPACE
         },
       };
@@ -369,7 +369,6 @@ export default {
         <n-link
           :to="createNamespaceLocationFlatList()"
           class="btn role-primary mr-10"
-          data-testid="cluster-manager-list-import"
         >
           {{ t('projectNamespaces.createNamespace') }}
         </n-link>
--- a/shell/components/ExplorerProjectsNamespaces.vue
+++ b/shell/components/ExplorerProjectsNamespaces.vue
@@ -4,7 +4,7 @@ import ResourceTable from '@shell/components/ResourceTable';
 import { STATE, AGE, NAME } from '@shell/config/table-headers';
 import { uniq } from '@shell/utils/array';
 import { MANAGEMENT, NAMESPACE, VIRTUAL_TYPES } from '@shell/config/types';
-import { PROJECT_ID } from '@shell/config/query-params';
+import { PROJECT_ID, FLAT_VIEW } from '@shell/config/query-params';
 import Masthead from '@shell/components/ResourceList/Masthead';
 import { mapPref, GROUP_RESOURCES, ALL_NAMESPACES } from '@shell/store/prefs';
 import MoveModal from '@shell/components/MoveModal';
@@ -236,12 +236,16 @@ export default {
 
     notInProjectKey() {
       return this.$store.getters['i18n/t']('resourceTable.groupLabel.notInAProject');
+    },
+    showCreateNsButton() {
+      return this.groupPreference !== 'namespace';
     }
   },
   methods: {
     /**
      * Get PSA HTML to be displayed in the tooltips
      */
+
     getPsaTooltip(row) {
       const dictionary = row.psaTooltipsDescription;
       const list = Object.values(dictionary)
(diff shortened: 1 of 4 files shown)


shell/components/ExplorerProjectsNamespaces.vue
```diff
@@ -284,6 +287,21 @@ export default {
 
       return location;
     },
+
+    createNamespaceLocationFlatList() {
+      const location = this.createNamespaceLocationOverride ? { ...this.createNamespaceLocationOverride } : {
+        name:   'c-cluster-product-resource-create',
+        params: {
+          product:  this.$store.getters['currentProduct'].name,
```
cnotv, line 295: Are we sure to always have a product and a name? We should add always optional chaining as the object may not exists.

shell/edit/namespace.vue
```diff
@@ -50,7 +50,7 @@ export default {
       originalQuotaId = `${ this.liveValue.metadata.name }/default-quota`;
     }
 
-    const projectName = this.value?.metadata?.labels?.[PROJECT] || this.$route.query[PROJECT_ID];
+    const projectName = this.value?.metadata?.labels?.[PROJECT] || this.$route.query[PROJECT_ID] || '(None)';
```
cnotv, line 53: Please restrain from hardcode i18n values.

cnotv (changes requested):
Could you please rebase the project so I can QA it?

I think we need also a video of the changes.

shell/edit/namespace.vue
```diff
@@ -169,9 +174,9 @@ export default {
       :value="value"
       :namespaced="false"
       :mode="mode"
+      :extra-columns="['project-col']"
     >
       <template
-        v-if="project"
```
cnotv, line 174: If you remove the condition, it will always display the project dropdown. This was meant while editing namespaces created within a project.

You will have then to extend to 2 conditions `v-if="flatView || project"` so it will match both the cases.

shell/edit/namespace.vue
```diff
@@ -181,7 +186,18 @@ export default {
         />
       </template>
     </NameNsDescription>
-
+    <div
```
cnotv, line 189: Remove this markup after fixing the condition.

shell/edit/namespace.vue
```diff
@@ -115,6 +118,9 @@ export default {
     },
 
     projectName(newProjectName) {
+      if (newProjectName === null) {
```
cnotv, line 121: Projects set as `null` or undefined will automatically turn into `(None)` option, so this is not necessary.

shell/components/ExplorerProjectsNamespaces.vue
```diff
@@ -343,7 +361,20 @@ export default {
...
-    />
+    >
+      <template
+        v-if="showCreateNsButton"
+        slot="extraActions"
+      >
+        <n-link
+          :to="createNamespaceLocationFlatList()"
+          class="btn role-primary mr-10"
+          data-testid="cluster-manager-list-import"
```
cnotv, line 372: Please remove the data attribute 😛

cnotv (changes requested):
It is unclear to me why we cannot set the project as `null`, so we can simply have empty string.
The project name will come out from the options. For some reasons while editing I get this for some cases. We need also to investigate about it.

Screenshot 2023-03-13 at 17 48 57

shell/edit/namespace.vue
bisht-richa: If I remove this. It shows an empty option on selection on None.

https://user-images.githubusercontent.com/18264463/225609011-e92f734f-a90a-4c63-92ee-a3c37dc49694.mov

cnotv: No, you need to set a value different from `null`, empty string or undefined.
There's some logic in this dropdown which does not let you do it.
We should not submit the label as value.

shell/edit/namespace.vue
```diff
@@ -50,7 +50,7 @@ export default {
       originalQuotaId = `${ this.liveValue.metadata.name }/default-quota`;
     }
 
-    const projectName = this.value?.metadata?.labels?.[PROJECT] || this.$route.query[PROJECT_ID];
+    const projectName = this.value?.metadata?.labels?.[PROJECT] || this.$route.query[PROJECT_ID] || this.t('namespace.project.none');
```
cnotv, line 53: We should not add this logic, as we don't want to have the label as value.

cnotv (changes requested):
Please let's try to correct the `(None)` case.

cnotv (approved):
LGTM

PR #8391, opened 2023-03-09
Dev server: Improve ctrl+c handling and suppress proxy setup messages

nwmac:
For `yarn dev`, this PR:

Adds a CTRL+C handler to shut down the dev server and exit the process - this fixes and issue where once the build starts, you can not CTRL+C out of it.

Hides the logging that we get for the proxies being set up - we currently get multiple lines of the form '[HPM] Creating proxy"


richard-cox (approved):
On linux this does stop the build but doesn't return the user to the terminal prompt.

Big improvement on what's there, though

nwmac:
Interesting. I’ll look in Linux later.

PR #8397, opened 2023-03-10
Added check box for truncate Hostnames

### LINKED ISSUE #8223: Ability to truncate RKE2 hostnames to 15 characters or less
Internal reference: SURE-5923, SURE-5811

Request description:
Offer users the ability to truncate existing hostname logic to 15 characters during provisioning

Actual behaviour:
Unix style name generation is too large for the 15-character limit of NetBIOS. This prevents machines from provisioning correctly

Changing existing logic for all cluster naming for all users is too disruptive; it should be an option that is available at cluster creation

Expected behaviour:
Customers can create netbios-compliant clusters for domain joining

Workaround:
None

Changing naming conventions which is not acceptable or sustainable

Target platforms:
Provisioned RKE2, Windows RKE2 & k3s clusters

Additional notes:
This will affect all Windows setups eventually and really reduces our ability to deliver Windows Kubernetes as a solution

There is already a standard pattern for truncation in Azure CAPI for this purpose which we could emulate:

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/scope/machine.go#L429

_*What needs to be done*_

- Add a checkbox at the cluster-level 'Truncate hostnames to 15 characters for NetBIOS compatibility'
- Checkbox can ONLY be set on create - it should be disabled on edit
- Note: Although we set a boolean, this is stored as an integer on the cluster object - true should be set to 15, false should remove the field
- On edit, ONLY if the value is 15, should we show the checkbox as checked
- On edit, if the value is present but NOT 15, we should show an info banner beneath the checkbox saying "Hostname truncation has been manually configured for this cluster to N" (where N is the value specified)
- Truncation can be set per machine pool. We do NOT allow this to be edited in the UI, but as a bonus, we do want to show per-machine pool if the value is set (to any value) - we should show a similar banner: "Hostname truncation has been manually configured for this pool to N" (where N is the value specified for the pool)

bisht-richa:
### Summary
Fixes #8223

Checkbox added to the right of 'NodePort Service Port Range' in the 'Networking' tab for hostnames Truncate

### How to test

 As of now the new param per machine pool named `instanceNameLimit`  is not available from the backend, so for the testing plead add this code:
 
out.find((pool) => {
       pool.pool.instanceNameLimit = 15;
});
> Go to Cluster management > create > select DO 

> In Cluster Configuration section > click 'Networking' tab 

> It should have a check box with a label for hostnames Truncate

**Some other use cases to test**
> Checkbox can ONLY be set on create - it should be disabled on edit
> On edit, ONLY if the value is 15, should we show the checkbox as checked
> On edit, if the value is present but NOT 15, we should show an info banner beneath the checkbox
> Show banner per-machine pool if the value is set.

### DIFF
--- a/shell/edit/provisioning.cattle.io.cluster/MachinePool.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/MachinePool.vue
@@ -178,7 +178,7 @@ export default {
       color="info"
     >
       <div class="text">
-        {{ t('cluster.machinePool.truncationPool') }} {{ value.pool.hostnameLengthLimit }}
+        {{ t('cluster.machinePool.truncationPool', { limit: value.pool.hostnameLengthLimit }) }}
       </div>
     </Banner>
     <div class="row">
--- a/shell/edit/provisioning.cattle.io.cluster/MachinePool.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/MachinePool.vue
@@ -174,11 +174,11 @@ export default {
 <template>
   <div>
     <Banner
-      v-if="value.pool.instanceNameLimit"
+      v-if="value.pool.hostnameLengthLimit"
       color="info"
     >
       <div class="text">
-        {{ t('cluster.machinePool.truncationPool') }} {{ value.pool.instanceNameLimit }}
+        {{ t('cluster.machinePool.truncationPool') }} {{ value.pool.hostnameLengthLimit }}
       </div>
     </Banner>
     <div class="row">
--- a/shell/edit/provisioning.cattle.io.cluster/MachinePool.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/MachinePool.vue
@@ -173,6 +173,14 @@ export default {
 
 <template>
   <div>
+    <Banner
+      v-if="value.pool.instanceNameLimit"
+      color="info"
+    >
+      <div class="text">
+        {{ t('cluster.machinePool.truncationPool') }} {{ value.pool.instanceNameLimit }}
+      </div>
+    </Banner>
     <div class="row">
       <div class="col span-4">
         <LabeledInput
@@ -294,7 +302,6 @@ export default {
         :read-allowed="false"
         :value-can-be-empty="true"
       />
-
       <div class="spacer" />
 
       <Taints
--- a/shell/edit/provisioning.cattle.io.cluster/rke2.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/rke2.vue
@@ -340,6 +340,15 @@ export default {
       cisPsaChangeBanner:    false,
       psps:                  null, // List of policies if any
       truncateHostnames:     false,
+      pp:                    [{
+        id: 0, instanceNameLimit: ''
+      },
+      {
+        id: 1, instanceNameLimit: ''
+      },
+      {
+        id: 2, instanceNameLimit: ''
+      }],
     };
   },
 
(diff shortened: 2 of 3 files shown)


nwmac:
@bisht-richa Backend will add a new param per machine pool named `instanceNameLimit` to support this.

We are only allowing this to be set once per cluster, rather than per machine pool - so you will need to update this PR to add the above field to each machine pool if the user checks the box. It should be set to `15 and remove it from each pool if not.

For edit, you will need to look at each machine pool and see if the value is set, if it is set on any pool, then I would translate that to the checkbox being checked.

Backend PR for reference: https://github.com/rancher/rancher/compare/release/v2.7...jakefhyde:rancher:15-character-limit#diff-e05181e81036cad4014bb4843ea91837fa7dd238dc3434bc4fcc598c56db6428R31

kwwii:
Does this mean that we want to put the checkbox back into the Networking tab of the cluster creation UI?

Screenshot 2023-03-24 at 10 12 18

Oats87:
@bisht-richa @nwmac the design for this has not been finalized, has it?

I just gave a preliminary review to the design document and notably we'll be making some pretty major changes to what we expect. Notably, the backend will allow the hostname limit to be something > 6 characters, and we'll provide the ability to set a cluster-wide default on `hostnameLimit` and the corresponding other field. The field names have not yet been finalized either.

nwmac:
@bisht-richa I think we can be unblocked now on the backend.

For the cluster-level setting, the setting will be `machinePoolDefaults.hostnameLengthLimit` - this will be the one that we need to set to 15, if the checkbox is checked.

Each machine pool can also have a `hostnameLengthLimit` field set - we aren't setting this, but if it is set, we will show a banner.

shell/edit/provisioning.cattle.io.cluster/MachinePool.vue
```diff
@@ -173,6 +173,14 @@ export default {
 
 <template>
   <div>
+    <Banner
+      v-if="value.pool.instanceNameLimit"
```
nwmac, line 177: `instanceNameLimit` should now be `hostnameLengthLimit`

shell/edit/provisioning.cattle.io.cluster/rke2.vue
```diff
@@ -1061,6 +1066,20 @@ export default {
     nlToBr,
     set,
 
+    /**
+     * set instanceNameLimit to 15 to all pool machine if truncateHostnames checkbox is clicked
+     */
+    truncateName() {
+      if (this.truncateHostnames) {
```
nwmac, line 1073: This is incorrect - we no longer have to set on each pool.

We should now be setting `machinePoolDefaults.hostnameLengthLimit` on the cluster.

Note that `machinePoolDefaults` may not exist, so we need to add that object first.

if truncation not set, we should remove the `machinePoolDefaults` object if it is empty

shell/edit/provisioning.cattle.io.cluster/rke2.vue
```diff
@@ -1117,6 +1136,19 @@ export default {
             config: config ? await this.$store.dispatch('management/clone', { resource: config }) : null,
             configMissing
           });
+
+          // Added instanceNameLimit manually to each pool for testing purpose
+          // out.map((pool) => {
+          //   pool.pool.instanceNameLimit = 17;
+          // });
+
+          out.filter((pool) => {
```
nwmac, line 1145: We should be picking this up from the cluster, not the pools

nwmac (changes requested):
@bisht-richa Can you update this - field names are now updated.

The behaviour needs updating to set this on the cluster, not the individual pools.

nwmac (approved):

bisht-richa:
@nwmac I will revert it.

gaktive:
Backend changes have made it now so we can ...unrevert.

PR #8400, opened 2023-03-10
Network Policy, combine `podSelector` and `namespaceSelector` in the same rule

### LINKED ISSUE #7660: Network policy showing multiple sources improperly as separate items in an array
Internal reference: SURE-5159
Reported in 2.6.7

Issue description:
Adding multiple sources in a single rule through rancher UI results in two separate items in the same array. 

Troubleshooting steps:

After adding the network policy, check it by "Edit YAML". We can see a hyphen (with a red dash icon) added to both `namespaceSelector` and `podSelector`, which means those are two items of the array. In this case, if either of the conditions are true, it will allow the ingress traffic.

```
  ingress:
  - from:
    - namespaceSelector:
        matchExpressions:
        - key: kubernetes.io/metadata.name
          operator: In
          values:
          - kube-system
        matchLabels:
          kubernetes.io/metadata.name: kube-system
    - podSelector:
        matchExpressions:
        - key: k8s-app
          operator: In
          values:
          - kube-dns
        matchLabels:
          k8s-app: kube-dns
```

Repro steps:

- Go to Rancher >> Explorer >> Service Discovery >> Network Policies >> Create
- Name the policy and check "configure ingress rules to restrict traffic"
- Create a rule by pressing green plus button
- Add the first source by clicking "Add allowed traffic source" and select "Namespace Selector" from the dropdown
- Enter "kubernetes.io/metadata.name: kube-system" as Key/Value and "in list" as operator 
- Add the second source by clicking "Add allowed traffic source" again and select "Pod Selector" this time
- Enter "k8s-app: kube-dns" as Key/Value and "in list" as operator

Workaround:
Use "Edit As YAML" after adding required sources and make it part of the same item (remove `-`)

Actual behavior:
It adds namespaceSelector, podSelector as separate items of the array, resulted in matching either of it allows traffic

Expected behavior:
Should be a mechanism (AND button in the same rule) to add it as part of the same item so that it will allow traffic when both the conditions (matches both namespaceSelector and podSelector) are true. Preferably an OR button to add a second item for the existing source.

torchiaf:
### Summary
Fixes #7660

This PR will introduce the possibility to define a single `to/from` entry that specifies both `namespaceSelector` _and_ `podSelector` that selects particular Pods within particular namespaces.

see k8s doc: https://kubernetes.io/docs/concepts/services-networking/network-policies/#behavior-of-to-and-from-selectors

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -3490,6 +3490,7 @@ networkpolicy:
           other {Matches {matched, number} of {total, number} existing namespaces, including "{sample}"}
         }
     matchingNamespacesAndPods:
+      tooltip: A single entry rule that specifies both Namespace Selector and Pod Selector that selects particular Pods within particular namespaces
       matchesSome: |-
         {matchedPods, plural,
           =0 {Matches 0 of {totalPods, number} pods}
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -3440,6 +3440,8 @@ networkpolicy:
   labelsAnnotations:
     label: Labels & Annotations
   rules:
+    pod: Pod
+    namespace: Namespace
     ruleLabel: Rule {index}
     addPort: Add allowed port
     type: Rule type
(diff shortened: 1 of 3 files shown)


cnotv (dismissed):
LGTM

New form for Pod and Namespaces combined and related outcome (same list without the `-`)

Screenshot 2023-03-14 at 11 45 36

Screenshot 2023-03-14 at 11 46 56

Screenshot 2023-03-14 at 11 47 04

cnotv (approved):

cnotv (approved):

PR #8401, opened 2023-03-10
Add missing '/' to the ROUTER_BASE in build-hosted

codyrancher:
The missing ending '/' was causing the router to parse routes incorrectly

Verified using this build `https://releases.rancher.com/dashboard/prod-dashboard-dev/index.html`


richard-cox (approved):
There's a good chance the same change needs to be applied to the hosted dashboard build process `scripts/build-embedded`. We will only be able to confirm this though when the first 2.7 Q2 RC is created, so heads up for @mantis-toboggan-md just in case

PR #8403, opened 2023-03-10
Correct base path with absolute value for RKE2 CIS hardening documentation and avoid banner on creation

### LINKED ISSUE #8123: Alert users when selected CIS Profile is no longer available for newly selected k8s version on Edit Cluster for RKE2/k3s clusters
Following the implementation of https://github.com/rancher/rancher/issues/40435, `cis-1.5` and `cis-1.6` CIS Profile options have been replaced with `cis-1.23` for k8s 1.25 versions.
However, when editing cluster with k8s <=1.24 with `cis-1.5` or `cis-1.6` selected and setting k8s version to 1.25+, the selection in "CIS Profile" dropdown stays at the previous value:
image

When k8s versions is switched, check if currently selected CIS Profile is still available. If not:
* If the selected k8s version has a single profile, present the user with a modal (?) confirmation stating something like `The selected Kubernetes Version no longer supports CIS Profile "{current CIS Profile}"; the selection will be automatically updated to "{new k8s version only supported CIS profile}" as the only profile supported by the selected Kubernetes version. It's recommended to review RKE2 documentation to evaluate the impact of changing CIS Profile. Proceed?` with the options "Yes" and "No".
* If the selected k8s version has more than one available profile, _(this is to be discussed - the only case where it would come up currently is if user first switches to 1.25 and then switches back to 1.24)_.

N/A

**Additional context**
N/A

cnotv:
### Summary
Fixes #8123

- Corrected URL with full path as documentation is located outside of Rancher official documentation
- Prevent to display banner on creation

### Screenshot/Video

https://user-images.githubusercontent.com/5009481/224400541-4d7bdd68-2cb1-40fd-b36d-8cb441c4c704.mov

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -1774,7 +1774,7 @@ cluster:
     banner:
       psaChange: PSACT is now set to Rancher default automatically
       cisOverride: Changing this setting may affect cluster security as it overrides default CIS settings
-      cisUnsupported: The selected Kubernetes Version no longer supports CIS Profile "{cisProfile}". Please select a supported profile. We recommend reviewing the <a href="{docsBase}/security/hardening_guide" target="_blank" rel="noopener noreferrer nofollow">documentation</a> to evaluate the impact of changing the CIS Profile.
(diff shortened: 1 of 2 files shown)


torchiaf (approved):
LGTM

Issue #8310, opened 2023-03-01
Update missing parts of extension docs


aalves08:
There are still some areas that need tackling in terms of documentation to cover all areas:
- [x] translations https://github.com/rancher/dashboard/pull/8417 @jordojordo 
- [x] routing (after POC https://github.com/rancher/dashboard/issues/8311) @aalves08 
- [x] publishing extensions (POC https://github.com/rancher/dashboard/issues/8312) @jordojordo 
- [x] product registration (after POC https://github.com/rancher/dashboard/issues/8311) @aalves08

PR #8417, opened 2023-03-13
Add extension translation docs

jordojordo:
### Summary
Reference #8310 

This adds information for extensions regarding translations. Since the existing documentation for localization is rather extensive and applies, I felt we could reuse this as much as possible.
Also updated the title for Localization to `l10n` from `i10n`.


aalves08 (approved):
LGTM

PR #8419, opened 2023-03-13
Fixed selected rows counter

### LINKED ISSUE #8273: Selected rows counter, overflow
**Setup**
- Rancher version: `2.7-head`
- Browser type & version: any

**Describe the bug**

We have a selection counter on the top of the each table where the number of selected rows are displayed.
If the user hits the  `shift+j` combination, the counter never stops, even after all rows are selected;
Same for `shift+k`, which selects the rows from bottom to the top.

**To Reproduce**

Hit `shift+j` until all rows are selected, then continue to see the counter overflow.

**Result**
The counter never stops.

**Expected Result**

The counter should display the number of selected rows.

**Screenshots**

selection-counter-error-2023-1-27

bisht-richa:
### Summary
Fixes #8273

Fixed selected row counter

### How to test

> Go to any table with rows
> Hit shift+j until all rows are selected, the counter should not overflow.

### DIFF
--- a/shell/components/SortableTable/index.vue
+++ b/shell/components/SortableTable/index.vue
@@ -787,7 +787,7 @@ export default {
     toggleExpand(row) {
       const key = row[this.keyField];
       const val = !this.expanded[key];
-      
+
       this.expanded[key] = val;
       this.expanded = { ...this.expanded };
 
@@ -818,6 +818,7 @@ export default {
 
     nearestCheckbox() {
       const $cur = $(document.activeElement).closest('tr.main-row').find('.checkbox-custom');
+
       return $cur[0];
     },
 
--- a/shell/components/SortableTable/index.vue
+++ b/shell/components/SortableTable/index.vue
@@ -436,14 +436,6 @@ export default {
       return this.$store.getters['activeNamespaceCache'];
     },
 
-    selectedRowsLength() {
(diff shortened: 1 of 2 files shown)


richard-cox (changes requested):
This looks like it masks the issue. We should ensure `this.selectedRows` contains valid entries (i.e understand why bad entries are added and fix there) and that the bulk actions that use `selectedRows` work fine

bisht-richa:
> This looks like it masks the issue. We should ensure `this.selectedRows` contains valid entries (i.e understand why bad entries are added and fix there) and that the bulk actions that use `selectedRows` work fine

I have done changes in the fix.

richard-cox (changes requested):
This is masking the issue again. If you put a breakpoint on `selection.js` `update` you'll see it gets called by `index.vue` `focusNext`... which finds the row by `focusAdjacent`. 

That fn 'focusAdjacent' is returning the last element, when it should return null and the caller then counts null as a no-op.


Edit:  it looks like the select next / previous is pretty broken. starting from the middle and alternating between select next / previous causes all sorts of issues (this is in current master). we should also resolve the issue of selecting already selected rows

richard-cox (approved):
LGTM. The fix to skip the selection if it's already selected works really nicely

PR #8441, opened 2023-03-15
Migrate all of the files within the nuxt directory to the shell directory

codyrancher:
I broke this up into several commits to make it a little more obvious about when I'm moving files vs editing files so it's easier to review.

This PR is dependent on https://github.com/rancher/dashboard/pull/8436


richard-cox (dismissed):
There's a conflict, but the rebase is clean and tested locally with it.

richard-cox (approved):

PR #8483, opened 2023-03-21
[2.7.2] Update caniuselite

nwmac:
Updates `caniuselite` so the nag messages don't appear during build/dev.


aalves08 (approved):
LGTM! no errors doing yarn install and running the app. ✅

PR #8466, opened 2023-03-17
Disallow `v-html` usage

rak-phillip:
### Summary

This replaces the use of the `v-html` directive with a custom `v-clean-html` directive that uses the `DOMPurify` library to sanitize HTML input before rendering it. `v-html` usage is discouraged[^1] in the Vue documentation, and should only be used on trusted content. Our widespread usage of `v-html` makes this difficult to guarantee safe usage across the board, so it's better to disallow `v-html` and sanitize existing usages in Rancher Dashboard.

### Occurred changes and/or fixed issues

- Disallows use of `v-html` with `eslint`
- Adds new `v-clean-html` directive that only allows existing tags used in `en-us.yaml`
- Replaces all existing instances of `v-html` with `v-clean-html` 

### Technical notes summary

The current implementations contains a list of allowed tags for use with `v-clean-html`:

```javascript
const ALLOWED_TAGS = [
  'code',
  'li',
  'a',
  'p',
  'b',
  'br',
  'ul',
  'pre',
  'span',
  'div',
  'i',
  'em',
  'strong',
];
```

These tags are derived from a list of unique tags obtained from `en-us.yaml`. An alternative approach might be to replace this list of allowed tags with the `DOMPurify` `html` profile[^2], which should allow all safe HTML (e.g. `DOMPurify.sanitize(dirty, {USE_PROFILES: {html: true}});`)

### Areas or cases that should be tested

We need to identify at least one instance for each of the allowed tags to ensure that content is rendered properly.

[^1]: https://v2.vuejs.org/v2/guide/syntax.html#Raw-HTML
[^2]: https://github.com/cure53/DOMPurify#can-i-configure-dompurify

PR #8520, opened 2023-03-24
Backport 2.7.2: Disallow v-html usage

rak-phillip:
### Summary

This replaces the use of the `v-html` directive with a custom `v-clean-html` directive that uses the `DOMPurify` library to sanitize HTML input before rendering it. `v-html` usage is discouraged[^1] in the Vue documentation, and should only be used on trusted content. Our widespread usage of `v-html` makes this difficult to guarantee safe usage across the board, so it's better to disallow `v-html` and sanitize existing usages in Rancher Dashboard.

- Disallows use of `v-html` with `eslint`
- Adds new `v-clean-html` directive that only allows existing tags used in `en-us.yaml`
- Replaces all existing instances of `v-html` with `v-clean-html` 

### Technical notes summary

The current implementations contains a list of allowed tags for use with `v-clean-html`:

```javascript
const ALLOWED_TAGS = [
  'code',
  'li',
  'a',
  'p',
  'b',
  'br',
  'ul',
  'pre',
  'span',
  'div',
  'i',
  'em',
  'strong',
];
```

These tags are derived from a list of unique tags obtained from `en-us.yaml`. An alternative approach might be to replace this list of allowed tags with the `DOMPurify` `html` profile[^2], which should allow all safe HTML (e.g. `DOMPurify.sanitize(dirty, {USE_PROFILES: {html: true}});`)

We need to identify at least one instance for each of the allowed tags to ensure that content is rendered properly.

Backports #8466

[^1]: https://v2.vuejs.org/v2/guide/syntax.html#Raw-HTML
[^2]: https://github.com/cure53/DOMPurify#can-i-configure-dompurify

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -166,7 +166,7 @@ module.exports = {
     // Vue/HTML Formatting kept for HARVESTER ONLY!
     // TODO: Move back here the HTML formatting
     'vue/html-self-closing': 'off',
-    'vue/no-v-html':         'off',
+    'vue/no-v-html':         'error',
   },
   overrides: [
     {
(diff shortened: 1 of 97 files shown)


rak-phillip:
Converting to draft, we need a proper release of 2.7.2 before merging.

rak-phillip:
Updated milestone to `v2.6.patch2` will update target branch when this becomes available.

PR #8539, opened 2023-03-27
fix /meta/proxy requests when running locally

mantis-toboggan-md:
Similar to the fix here: https://github.com/rancher/dashboard/pull/8386/commits/724d952e962243b1d663dd2f34b0c89ad88b5d71 I'm seeing requests to cloud providers eg  `.../meta/proxy/api.digitalocean.com/v2/regions?per_page=1000` fail when running the dash locally. Given that no one else has brought this up I have to think that like the other issue this isn't happening on all rancher instances; I can share my setup to repro if needed. 

Screen Shot 2023-03-27 at 2 38 03 PM

### DIFF
--- a/shell/vue.config.js
+++ b/shell/vue.config.js
@@ -576,6 +576,7 @@ function proxyMetaOpts(target) {
     target,
     followRedirects: true,
     secure:          !dev,
+    changeOrigin:    true,
     onProxyReq,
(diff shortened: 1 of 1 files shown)


codyrancher (approved):
I'm guessing very few people actually launch it in the browser via localhost.

PR #8546, opened 2023-03-28
Fix Link Icon margins

Shavindra:
### Summary
Fixes rancher/dashboard#7976 

- Updated link component styles

- Navigate to Continuous delivery
- Create a gitRepo

- `before-icon-class` and `after-icon-class` doesn't seem to be in use. 
- In gitRepo its applied via formatter and through following getters which takes `before-icon-key` and  `after-icon-key`. Haven't really found usage outside of gitRepo. So I assume the impact would be minimal
```    beforeIconClass() {
      if ( this.beforeIconKey ) {
        return get(this.row, this.beforeIconKey);
      }

      return this.beforeIcon;
    },

    afterIconClass() {
      if ( this.afterIconKey ) {
        return get(this.row, this.afterIconKey);
      }

      return this.afterIcon;
    },
```

### DIFF
--- a/shell/components/formatter/Link.vue
+++ b/shell/components/formatter/Link.vue
@@ -139,7 +139,6 @@ export default {
     v-if="isInternal && href"
     :to="href"
     class="link-text-icon"
-
   >
     <i
       v-if="beforeIconClass"


cnotv (approved):
LGTM

PR #8551, opened 2023-03-29
CruResource, make preventSubmit optional

### LINKED ISSUE #8470: Pressing 'n' in `textarea` elements opens the `namespaces` filter
**Setup**
- Rancher version: `master-528d7d94c`
- Browser type & version: any

**To Reproduce**

- Go to local -> Storage -> ConfigMaps
- Create a new ConfigMap
- Add a new entry in Data form
- Select the `value` field and press 'n'

**Additional Info**
This is a regression of https://github.com/rancher/dashboard/pull/7267

### LINKED ISSUE #8550: Pressing `Enter` in `textarea` elements not working
**Setup**
- Rancher version: `master-528d7d94c`
- Browser type & version: any

**To Reproduce**

- Go to local -> Storage -> ConfigMaps
- Create a new ConfigMap
- Add a new entry in Data form
- Select `value` field and press `Enter`

**Result**

Nothing happens

**Expected Result**

A new line should be added to the value.

**Additional Info**
This is a regression of https://github.com/rancher/dashboard/pull/7267
Possible related issue: https://github.com/rancher/dashboard/issues/8470

torchiaf:
Signed-off-by: Francesco Torchia [francesco.torchia@suse.com](mailto:francesco.torchia@suse.com)

### Summary
Fixes #8550

### DIFF
--- a/shell/components/CruResource.vue
+++ b/shell/components/CruResource.vue
@@ -101,6 +101,11 @@ export default {
       default: null,
     },
 
+    preventEnterSubmit: {
+      type:    Boolean,
+      default: false,
+    },
+
     applyHooks: {
       type:    Function,
       default: null,
@@ -380,6 +385,12 @@ export default {
           throw new Error(`Could not create the new namespace. ${ e.message }`);
         }
       }
+    },
+
+    onPressEnter(event) {
+      if (this.preventEnterSubmit) {
+        event.preventDefault();
+      }
     }
   }
 };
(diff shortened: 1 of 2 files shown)


cnotv (approved):
LGTM

torchiaf:
As we discussed during the daily meetings, there is a related issue https://github.com/rancher/dashboard/pull/7267, where we disabled the `enter` button from submit forms, - every forms in the UI -.
I did several tests to verify if this current change do not break this behavior, but I could't find anything wrong.
@cnotv @nwmac  do you have any specific cases I should check to for regressions ?

cnotv:
@torchiaf just as for you, I am not able to cross any issue in preventing to submit the form. From the other hand, changing the value to false does not make me submit it either, so the whole logic must not work at all.
Considering it does not work in both the cases, plus we solved the issue with the value input and the other scope is not related to this issue, I would say that everything is fine in the end.

PR #8558, opened 2023-03-29
Bug fix for custom cloud credentials from extensions

nwmac:
There is a copy and paste error that means custom cloud credentials in extensions don't work.

We should probably back-port this to 2.7.2.


aalves08 (approved):
Good catch. LGTM

Issue #8567, opened 2023-03-30
Build & Deploy DevKit only on master


scures:
**Is your feature request related to a problem? Please describe.**

Seems that now we build & deploy our DevKit based on multiples branches which can lead to the deployment of an older version of the documentation.

**Describe the solution you'd like**

Only Build & Deploy on merge into `master` which should always have the latest documentation.

PR #8568, opened 2023-03-30
fix: triggers publishing docusaurus on merge into master

scures:
### Summary
Fixes rancher/dashboard#8567

We should only build & deploy the devkit when merging into the `master` branch in order to keep the documentation up-to-date

EDIT: I also changed the file for the storybook which has the same behaviour

### DIFF
--- a/.github/workflows/storybook.yaml
+++ b/.github/workflows/storybook.yaml
@@ -23,5 +23,5 @@ jobs:
       run: |
         yarn build-storybook
     - name: Publish Storybook
-      if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' && github.repository_owner == 'rancher' }}
+      if: ${{ github.event_name == 'push' && github.repository_owner == 'rancher'}}
       run: .github/workflows/scripts/publish-storybook


.github/workflows/storybook.yaml
```diff
@@ -1,14 +1,11 @@
...
-env: 
+env:
     STORYBOOK_TOKEN: ${{ secrets.STORYBOOK_TOKEN }}
 on:
   push:
     branches:
       - master
-  pull_request:
-    branches:
-      - master
```
nwmac, line 10: I think we can leave the PR for master - the publish step should only run if its a push rather than a pull.

nwmac (dismissed):

cnotv:
If you exclude the PR, we would not be able to know if the process works or not in case of a PR.
Also, this removal will risk to break everyone merge runs as we had before, this is why we added it.

I would suggest to remove the `'release-*'` as I cannot picture a scenario to require it.

We should rather investigate on why the conditions are actually not working. I am not entirely sure of the nature of this issue/PR.

cnotv (approved):
LGTM

Issue #8238, opened 2023-02-13
[RFE] Support overriding CRD values on main chart install (for Gatekeeper)


vardhaman22:
Earlier there were not much values in crd charts but as a part of kubernetes 1.25 support now there are some more chart values are added. So support on UI is needed for these charts so that user can change the default values.

there were additional variables added for the following CRD charts:
1. rancher-gatekeeper-crd:
 https://github.com/rancher/charts/blob/dev-v2.7/charts/rancher-gatekeeper-crd/102.0.0%2Bup3.10.0/values.yaml#L15-L21
 added for the fix for issue: https://github.com/rancher/rancher/issues/40108

gaktive:
@vardhaman22 for our context, can OPI Gatekeeper be installed without this? I see the related backend ticket that was blocked has been closed so we need to know if this can be pushed to Q2 2.7.x or if this needs to go in for 2.7.2.

gaktive:
Focus on this ticket is purely for Gatekeeper. Extending the CRD overrides should be a separate ticket to discuss with Product for Q2.

vardhaman22:
@gaktive chart installation with default values will work fine on the OS which supports **seccomp**. but in case someone uses an image with OS which does not support **seccomp**  then it may be a problem.
seccomp is supported by most modern operating system. so i think we should be fine.
@prachidamle please share your opinion on this.
cc: @MKlimuszka

rancherbot:
@mantis-toboggan-md, Not creating `backport` issue for issue `8238` in repository `dashboard` because milestone `2.7.2` does not exist or is not an open milestone

ronhorton:
**Pass** verified in rancher `v2.7-head commit ID 95edd02`
1. created downstream 1.24 rke1 cluster
2. installed rancher-gatekeeper:102.1.0+up3.12.0
3. enabled psps (override yaml values)
`chart installed successfully`
4. created downstream 1.25 rke1 cluster
5. installed rancher-gatekeeper:102.1.0+up3.12.0
6. enable psps not avail as 1.25 has no psps
`chart installed successfully`

PR #8573, opened 2023-03-30
Fix overwritten chart global values

### LINKED ISSUE #8566: PSP value in custom charts is not kept on submission
**Setup**
- Rancher version: 2.7.2rc7
- Browser type & version: Any

**Describe the bug**
While editing cluster charts the PSP enabled value is not kept while it's still visible in the YAML editor or data.

**To Reproduce**
- Provision cluster K8S 1.24.x
- Install custom chart (e.g. monitoring) 
- Enable PSP checkbox
- Save

**Result**
PSP is disabled.

**Expected Result**
PSP value is still enabled.

**Screenshots**

image
image
image

**Additional context**

mantis-toboggan-md:
### Summary
Fixes #8566

### Technical notes summary
The bug was introduced here: https://github.com/rancher/dashboard/compare/v2.7.2-rc4...v2.7.2-rc7#diff-39e1e049c631a635017305ddbeb1944cff1559d39f32b31efea2e38cd142d611R1196

In the above, `versionInfo.values === crd chart values` and `values === chart values`

This was meant to allow custom chart UIs to modify crd chart values not found in the main chart, specifically OPA gatekeeper crd enableRuntimeDefaultSeccompProfile : https://github.com/rancher/dashboard/issues/8238. Previously, CRD charts were installed with _only_ the values of the main chart. The change I linked above was intended to install CRD charts with their own values plus whatever chart values were customized in the UI. Problem appears to be that I merged crd chart values w/ main chart values the wrong direction. We want the chart's global values (eg enable PSP) to overwrite the values of the crd chart, only saving crd values not present on the main chart. I did the opposite and overwrote the user-configured PSP value with the crd chart's default (`false`)

 Installing monitoring or gatekeeper with/without psp enabled

The OPA gatekeeper chart install has an 'enable runtime default seccomp profile' checkbox that controls a value in the gatekeeper-crd chart - this should control the gatekeeper-crd chart `values.enableRuntimeDefaultSeccompProfile` still

### DIFF
--- a/shell/pages/c/_cluster/apps/charts/install.vue
+++ b/shell/pages/c/_cluster/apps/charts/install.vue
@@ -1200,7 +1200,7 @@ export default {
           version:     versionInfo.chart.version,
           releaseName: versionInfo.chart.annotations[CATALOG_ANNOTATIONS.RELEASE_NAME] || chart.name,
           projectId:   this.project,
-          values:      merge(this.addGlobalValuesTo({ global: values.global }), versionInfo.values)
+          values:      merge(versionInfo.values, this.addGlobalValuesTo({ global: values.global }))
         });
       }
       /*


cnotv (approved):
LGTM

Issue #8531, opened 2023-03-27
[Backport v2.7] Continuous Delivery - User with custom role with access to only list GitRepos can see the Add repo UI elements


rancherbot:
This is a backport issue for https://github.com/rancher/dashboard/issues/8467, automatically created via rancherbot by @mantis-toboggan-md

Original issue description:

**Setup**
- Rancher version: `v2.7-head` `a49d36c`
- Fleet version: `v0.6.0-rc.5`
- Browser type & version: `Firefox` `Chrome`

**Describe the bug**

A user with permissions to only list GitRepos is shown the Add Repository UI elements.

**To Reproduce**

- Add a new user role with all permissions (create, delete, list, etc) on `fleetworkspaces` resources.
- Add access to `gitrepos` resources but only to `list`

**Result**

The user see the add repositories UI elements.

Both cases:
- GitRepos exist
- GitRepos do not exist

The user can't add the GitRepo as expected. With error: `<user> is not able to add the repository as expected. ("Method POST not supported")`

**Expected Result**
Access to the existing GitRepos list, but no add repository elements.

izaac:
Dashboard version: `release-2.7.2 d1ea2d47b`
Rancher version: `v2.7-head-4a9b7ba`
Fleet version: `fleet-agent:v0.6.0-rc.5`

The user custom role has these permissions:

<img width="625" alt="Screen Shot 2023-03-31 at 8 14 38 AM" src="https://user-images.githubusercontent.com/27659/229161531-dce26595-83e9-4c18-aa06-2e7e62cb9b54.png">


But a user with that custom role assigned still see the `Get Started button/page`


<img width="573" alt="Screen Shot 2023-03-31 at 8 14 16 AM" src="https://user-images.githubusercontent.com/27659/229161865-b752cd94-d6d6-4703-90ac-5036df280b74.png">

Resulting in:


<img width="269" alt="Screen Shot 2023-03-31 at 8 19 58 AM" src="https://user-images.githubusercontent.com/27659/229161935-fbd88da0-d364-456f-937e-49a6f7744001.png">

izaac:
@mantis-toboggan-md Closing as validated.

I've opened a separate issue for the workspaces elements I am still seeing if a user doesn't have permissions on `fleetworkspaces`
https://github.com/rancher/dashboard/issues/8624

PR #8590, opened 2023-03-31
hide continuous delivery get started btn if user cant create gitrepos

### LINKED ISSUE #8467: Continuous Delivery - User with custom role with access to only list GitRepos can see the Add repo UI elements
**Setup**
- Rancher version: `v2.7-head` `a49d36c`
- Fleet version: `v0.6.0-rc.5`
- Browser type & version: `Firefox` `Chrome`

**Describe the bug**

A user with permissions to only list GitRepos is shown the Add Repository UI elements.

**To Reproduce**

- Add a new user role with all permissions (create, delete, list, etc) on `fleetworkspaces` resources.
- Add access to `gitrepos` resources but only to `list`

**Result**

The user see the add repositories UI elements.

Both cases:
- GitRepos exist
- GitRepos do not exist

The user can't add the GitRepo as expected. With error: `<user> is not able to add the repository as expected. ("Method POST not supported")`

**Expected Result**
Access to the existing GitRepos list, but no add repository elements.

mantis-toboggan-md:
### Summary
Fixes #8467 see comment on backport [here](https://github.com/rancher/dashboard/issues/8531#issuecomment-1492112791)

Updated the fleet landing page shown when a user has no git repos to hide the 'get started' button when the user can't create git repos.

### Technical notes summary
I've updated the `permissions` property to check available methods on the git repo schema rather than just checking for its existence - though git repos were added to this permission check previously, I don't see them used in the component elsewhere, so this change looks fine to me. 

1. Users with permission to list but not create/edit gitrepos should not see 'get started' button on the fleet landing page

### DIFF
--- a/shell/pages/c/_cluster/fleet/index.vue
+++ b/shell/pages/c/_cluster/fleet/index.vue
@@ -46,7 +46,7 @@ export default {
     this.fleetWorkspacesData = hash.fleetWorkspaces || [];
 
     try {
-      const permissions = await checkPermissions({ workspaces: { type: FLEET.WORKSPACE }, gitRepos: { type: FLEET.GIT_REPO } }, this.$store.getters);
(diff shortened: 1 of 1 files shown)


rak-phillip:
LGTM - confirmed this change by following repro steps as defined in issue.

Issue #6763, opened 2022-08-26
RKE2 cluster create/edit screen allows saving invalid configuration


snasovich:
**Setup**
- Rancher version: 2.6.7
- Browser type & version: Chrome

**Describe the bug**
When creating or editing RKE2 clusters the UI allows saving the clusters with node pools having invalid configuration, e.g. zero cp/etcd/worker nodes. 

**To Reproduce**
Locate any RKE2 node provisioned cluster and set node pools that have etcd nodes to 0 node count.

**Result**
The UI shows what appears to be error stating the reason configuration is invalid, however "Save" button is active and calls back end to create a cluster with invalid configuration:
<img width="303" alt="image" src="https://user-images.githubusercontent.com/85187633/186986905-18ea026d-f7c0-4688-9aac-e5b2ec6d11ed.png">

**Expected Result**
UI should either disable Save button in this case or show an error when Save is clicked (without calling back-end). The UI for RKE1 clusters does the latter.

**Screenshots**
N/A

**Additional context**
Back-end issue is logged (https://github.com/rancher/rancher/issues/36984), however it would still be beneficial to prevent invalid configurations to be saved (unless there was a good reason why it's done this way).

gaktive:
Internal reference: SURE-4279

This may require a 2.6.x back port though I'm collecting more details.

rancherbot:
@gaktive, Not creating `backport` issue for issue `6763` in repository `dashboard` because milestone `v.2.6.next2` does not exist or is not an open milestone

cnotv:
Issue seems to have been already resolved with https://github.com/rancher/dashboard/issues/5454 as mentioned [here](https://github.com/rancher/dashboard/issues/8489#issuecomment-1479872661) and seen in the screenshot.

<img width="961" alt="Screenshot 2023-04-03 at 13 45 34" src="https://user-images.githubusercontent.com/5009481/229500183-4f1fc3c4-9011-4be9-bf98-25b4c2a13da1.png">

cnotv:
Sorry for messing it up but I just realized that the button was disabled only due the lack of cluster name 🙃 

![Screenshot 2023-04-03 at 18 05 15](https://user-images.githubusercontent.com/5009481/229566154-aa986983-aa32-4cc7-a70e-2a7f597b11db.png)

Josh-Diamond:
Self-assigning, as I have validated the associated backport issue

Josh-Diamond:
[...]

PR #8608, opened 2023-04-03
RKE2 cluster create/edit screen allows saving invalid configuration

cnotv:
### Summary
Fixes rancher/dashboard#6763

- Prevent to continue if any node pool error is displayed
- Allows to create cluster without machine pool requirement

### Technical notes summary
- Adopted existing logic used for the badge to evaluate errors
- Replaced hardcoded i18n value
- Replaced hardcoded CSS classes with constants

Note: Although string check may be not the optimal choice, a change of logic or duplicated logic with a lack of tests would lead to an high risk of regression.

As defined in the issue.

Cluster creation with and without pool cases.

### Screenshot/Video

https://user-images.githubusercontent.com/5009481/231116961-1799effb-da89-41d8-bf27-d1fd404e0c18.mp4

### DIFF
--- a/shell/edit/provisioning.cattle.io.cluster/rke2.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/rke2.vue
@@ -1295,7 +1295,9 @@ export default {
     },
 
     validationPassed() {
-      return (this.provider === 'custom' || this.isElementalCluster || !!this.credentialId) && this.hasRequiredNodes();
+      const validMachinePools = this.hasMachinePools ? this.hasRequiredNodes() : true;
+
+      return (this.provider === 'custom' || this.isElementalCluster || !!this.credentialId) && validMachinePools;
     },
 
     cancelCredential() {


aalves08 (approved):
Gave it a test and works as expected. Nice touch on moving the colors to a constant. Much neater.

The only remark is it looks a bit odd (to me) to have the logic for `hasRequiredNodes` bound to colors instead of the checkbox values... But it works. Approved ✅

cnotv:
> Gave it a test and works as expected. Nice touch on moving the colors to a constant. Much neater.
> 
> The only remark is it looks a bit odd (to me) to have the logic for `hasRequiredNodes` bound to colors instead of the checkbox values... But it works. Approved ✅

Yeah I totally agree, but to refactor such thing is too much effort, considering the E2E is failing 😄

cnotv:
The issue was not with E2E, rather it always required machine pools to be checked.
Updated video with both cases.

aalves08 (approved):
Same comment as before on the `hasRequiredNodes`, otherwise testing and code LGTM 🙏

Issue #8615, opened 2023-04-04
Cluster and Project Members shows "loading..." next to principle name


Shavindra:
**Setup**
- Rancher version:
- Browser type & version:

**Describe the bug**
Cluster and Project Members shows "loading..." next to principle name rancher/dashboard#8615

**To Reproduce**

https://user-images.githubusercontent.com/1387263/229741149-290f1f67-b979-41c2-b713-9020fa95aa65.mov

**Result**

**Expected Result**
`Loading...` shouldn't show up

**Screenshots**

**Additional context**
This could be down `v-clean-html`

gaktive:
Engineering to write an automated test in #8804 to help with QA.

PR #8616, opened 2023-04-04
Fix v-clean-html rendering for cluster and project members.

Shavindra:
### Summary
Fixes rancher/dashboard#8615

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -4088,7 +4088,7 @@ prefs:
     scalingDownPrompt: Do not ask for confirmation when scaling down node pools.
 
 principal:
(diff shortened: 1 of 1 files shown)


nwmac (changes requested):
This is a hack.

There is a weird issue here that v-clean-html is being used within a template block that has a conditional.

It is the v-clean-html directive the has broken this here.

Looking at the Vue code for `v-html`, it does not update `innerHtml` directly - I believe there is an issue doing that when its part of a template.

We should fix this properly.

nwmac:
@Shavindra PR template needs filling out

nwmac:
See https://github.com/rancher/dashboard/pull/8617 for a non-hacky fix

PR #8617, opened 2023-04-04
Fix the v-clean-html directive

### LINKED ISSUE #8615: Cluster and Project Members shows "loading..." next to principle name

nwmac:
Fixes #8615 

The `v-clean-html` directive does not work when used inside a `<template>` section.

Looking at https://github.com/LeSuisse/vue-dompurify-html/blob/main/packages/vue-dompurify-html/src/dompurify-html.ts the trick seems to be to add an `unbind` to the directive.

This PR updates our directive to do this - this resolves the original issue.

### DIFF
--- a/shell/plugins/clean-html-directive.js
+++ b/shell/plugins/clean-html-directive.js
@@ -25,6 +25,9 @@ export const cleanHtmlDirective = {
   },
   componentUpdated(el, binding) {
     el.innerHTML = purifyHTML(binding.value);
+  },
+  unbind(el) {
+    el.innerHTML = '';
   }
 };
 


rak-phillip (approved):
LGTM

NOTE: This will need to be bundled with our backports for 2.6 & 2.7

PR #8625, opened 2023-04-04
Separate label slot in `RadioButton.vue`

### LINKED ISSUE #8619: [@rancher/components] It's not possible to use the `label` template in `RadioButton` component
**Describe the bug**

`RadioButton.vue` component exposes 2 slots, `description` and `label`, but only `description` works with custom templates:

```
    <radio-button
    ...
    >
      <template #description>
        <span v-html="option.description" />
      </template>
    </radio-button>
```
There are 2 problem with the `label` slot:

1. The label `slot` would be displayed only if the label prop is not null. We should split the logic to display label prop / label slot separately.
2. The label slot is always overwritten by the v-html property of the `label` tag:
image (8)

This is blocking some new features for Rancher Desktop.

**To Reproduce**

Define a custom label template and see that is not shown as label for RadioButtons:
```
<template #label>
    <span v-html="option.label" />
  </template>
```

rak-phillip:
### Summary
This separates `label` slot from the `label` prop content in `RadioButton.vue`. Doing this will allow developers to provide custom content to the `label` slot that would otherwise get overwritten by the `v-html` directive. 

Fixes #8619

### Technical notes summary

I did a search for existing `RadioButton` instances that make used of the `label` slot an none came up. Please double-check that this is the case. 

I suspect that we won't find any instances in our code because this use case was never supported with the existing implementation.

Anything that uses a radio button.

Anything that uses a radio button.

### DIFF
--- a/pkg/rancher-components/package.json
+++ b/pkg/rancher-components/package.json
@@ -2,7 +2,7 @@
   "name": "@rancher/components",
   "repository": "git://github.com:rancher/dashboard.git",
   "license": "Apache-2.0",
-  "version": "0.1.0",
+  "version": "0.1.1",
   "module": "./main.js",
   "main": "./dist/@rancher/components.umd.min.js",
   "files": [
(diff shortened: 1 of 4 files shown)


rak-phillip:
@s0nea I attempted to add you as a reviewer but you aren't showing up on my list.

torchiaf (approved):
LGTM, I did a test and it works fine:

image

torchiaf (approved):

s0nea:
> @s0nea I attempted to add you as a reviewer but you aren't showing up on my list.

Thanks for pinging me :slightly_smiling_face: I guess I don't have the required privileges.

Issue #8062, opened 2023-01-28
CRD charts should receive same global fields as main chart


aiyengar2:
As part of the larger issue in https://github.com/rancher/rancher/issues/40331 and specifically for monitoring in https://github.com/rancher/rancher/issues/40350, we need to ensure that on an install any CRD chart is provided the same settings under `global` that are being used to upgrade the main chart.

Specifically, for `rancher-monitoring-crd`, the `global.cattle.psp.enabled` field introduced in 2.7.2 needs to match what the user provided in the main chart.

stormqueen1990:
This may also apply to the following charts:
* rancher-gatekeeper-crd
* rancher-logging-crd

mantis-toboggan-md:
This will tie into https://github.com/rancher/dashboard/issues/8062 - the `global.cattle.psp.enabled` field needs to be added to the monitoring custom component

gaktive:
The 6 charts (from #8053) are:
- Logging
- Alerting Drivers
- Istio
- Monitoring
- Rancher Backup
- OPA Gatekeeper

mantis-toboggan-md:
The dashboard already configures crd charts with the same globals, see https://github.com/rancher/dashboard/blob/master/shell/pages/c/_cluster/apps/charts/install.vue#L1245
and in our docs: https://rancher.github.io/dashboard/code-base-works/helm-chart-apps#injected-global-values

nickwsuse:
on v2.7.2-rc8 I see that the Monitoring CRD does have the global.cattle.psp.enabled field, but it's also not a 1:1 match with the main chart's global settings (I'm not sure if it's supposed to or if we mostly just care about the psp field)

That said, I checked the Logging chart and Logging CRD and there is NOT a global.cattle.psp.enabled field in the CRD's yaml, but there is for the main chart. 

@mantis-toboggan-md @gaktive @stormqueen1990

mantis-toboggan-md:
> That said, I checked the Logging chart and Logging CRD and there is NOT a global.cattle.psp.enabled field in the CRD's yaml, but there is for the main chart. I'm currently checking the rest of the charts mentioned above to be sure if it's just logging or not.

@nickwsuse I'm not able to reproduce that behavior when installing logging. When I set `gloabl.cattle.psp.enabled: true` during install, then navigate to 'installed apps' and view the yaml of the crd app I see `spec.values.global.cattle.psp: true`...I see this in the Values YAML tab of the detail view as well

[...]

PR #8636, opened 2023-04-05
Ensure CRDs are always installed with the same global values as the main chart

mantis-toboggan-md:
This PR addresses the bug described here specifically: https://github.com/rancher/dashboard/issues/8612#issuecomment-1496222854

This PR updates the logic around CRD chart values in a few ways. There are two requests for 2.7.2 we need to accomodate in this area:
* CRDs should be installed with the same global values as the main chart https://github.com/rancher/dashboard/issues/8062
* Custom UIs should be able to configure CRD chart values: https://github.com/rancher/dashboard/issues/8238

#### Background info
In the install/upgrade request the UI passes only values that differ from the chart's default values. This means that when looking at the API call made,  if 'enable PSPs' is not checked, we shouldn't see `values.global.cattle.psp.enabled` in the request body of either chart at all. When it is checked, it should be there. Similarly, in #8238 we should only `values. enableRuntimeDefaultSeccompProfile` in the request when it has set to false in the UI. 

When viewing an installed app's detail view 'Values YAML' the user is shown all values the chart has been installed with, which is the combination of the default values and any customized values sent during installation/upgrading.

#### Changes in this PR
Previously we were mutating the CRD `values` object returned from `getVersionInfo` and sending that whole modified object in the request. We need to: 
1. use a new object each install to avoid polluting cached charts
2. only send values that are different from default
3. if upgrading, retrieve previously configured values and populate the UI form with them 

#### Monitoring
1. Install monitoring with PSP enabled and ensure both the main and CRD charts have `values.global.cattle.psp.enabled: true`
2. Wait for installation to finish.
3. Upgrade monitoring and uncheck the PSP option, then ensure both main chart and CRD *do not send* `values.global.cattle.psp.enabled` in the request - this should result in the 'values YAML' tab of detail view showing `psp.enabled: false` for both charts.

#### Gatekeeper
1. Install gatekeeper with enableRuntimeDefaultSeccompProfile set to false (not the default) and ensure the CRD chart int he request (but not main chart) has `values.enableRuntimeDefaultSeccompProfile: false`
2. Navigate back to the gatekeeper install/upgrade form and verify that the checkbox is still unchecked
4. Re-check the box; upgrade and verify the request does not contain `enableRuntimeDefaultSeccompProfile`
5. Navigate back to the gatekeeper install page and upgrade with PSP enabled; verify the request shows both main chart and CRD have this configured.

### DIFF
--- a/shell/chart/gatekeeper.vue
+++ b/shell/chart/gatekeeper.vue
@@ -27,12 +27,13 @@ export default {
       get() {
         const crdInfo = this.autoInstallInfo.find(info => info.chart.name.includes('crd'));
 
-        return crdInfo ? crdInfo.values : null;
+        // values are defaults from the chart; allValues are defaults + anything set on previous install if this is an update/upgrade
+        return crdInfo ? crdInfo.allValues : null;
       },
       set(values) {
         const crdInfo = this.autoInstallInfo.find(info => info.chart.name.includes('crd'));
 
-        this.$set(crdInfo, 'values', values);
+        this.$set(crdInfo, 'allValues', values);
       }
     }
   }
--- a/shell/mixins/chart.js
+++ b/shell/mixins/chart.js
@@ -14,6 +14,8 @@ import { CAPI, CATALOG } from '@shell/config/types';
 import { isPrerelease } from '@shell/utils/version';
 import difference from 'lodash/difference';
 import { LINUX } from '@shell/store/catalog';
+import { clone } from 'utils/object';
+import { merge } from 'lodash';
 
 export default {
   data() {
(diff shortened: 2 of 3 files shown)


cnotv (dismissed):
LGTM


https://user-images.githubusercontent.com/5009481/230208785-b236f5e6-9c06-4650-a93a-b7fa1212ba7f.mp4


https://user-images.githubusercontent.com/5009481/230208836-2e1ac3a5-09a0-455e-a3da-e30d0c807013.mp4

rak-phillip (approved):
LGTM

Issue #8582, opened 2023-03-31
Create the routing APIs


codyrancher:
Add the following APIs to the extension API object:

- `api.routing.addRoute(RouteRecordRaw)`
   -  [RouteRecordRaw](https://router.vuejs.org/api/#routerecordraw) 
   - _We need to check to see how this interacts with the [RouterLink](https://router.vuejs.org/api/#Variables-RouterLink) component. It's possible we can't support addRoute because our routes will need to be created before the rest of the app is initialized. If this is problematic, we may need to allow the extension registration hook to return a configuration object to support configurations that need to be done before app initialization._
- `api.routing.push(RouteLocationRaw)`
  - [RouteLocationRaw](https://router.vuejs.org/api/#routelocationraw) 
- `api.routing.replace(RouteLocationRaw)`
  - [RouteLocationRaw](https://router.vuejs.org/api/#routelocationraw)  
- `api.routing.currentRoute`

codyrancher:
I decided that `api.routing.addRoute(RouteRecordRaw)` is the only relevant API here. We're only exposing APIs that should be used to configure the extension at the initialization phase.

codyrancher:
Closing since we took a different direction.

Issue #8580, opened 2023-03-31
Create the first API which demonstrates the structure and how the API will be initialized


codyrancher:
- Create a namespaces APIobject which is organized/looks like `api.isInitialized`
  - Ensure the creation of this object is done though the invocation of some method or constructor because later APIs will likely need to be initialized before getting passed to the extension.
- Give access to this API object to the extension by passing it to the extension's registration hook

codyrancher:
Closing due to traction

PR #8647, opened 2023-04-07
Create the base for a new stripped down version of extensions

codyrancher:
- fixes rancher/dashboard#8580
- fixes rancher/dashboard#8582

### What
Adds support for a new way to register extensions without interfering with the existing method. The new method still closely resembles the old method.

Example extension configuration
```ts
// pkg/chat/configure.ts 

import { IExtension, RegisterFn } from '@shell/core/extensions/types';
import Chat from './components/Chat/index.vue';

// Configure the extension
export default function(register: RegisterFn) {
  const extension: IExtension = register(require('./package.json'));

  extension.router.addRoute({
    name:      'chat',
    path:      '/chat',
    component: Chat
  });
}
```

### Extension bootstrapping steps
1. In `shell/vue.config.js` we create a new virtual module / import named `@rancher/extensionsConfiguration` which imports the `config.ts` of each pkg if it exists.
1. In `shell/nuxt/index.js` we will execute the `shell/core/extensions/loader` as though it were a nuxt plugin in order to bootstrap the new version of extensions.
1. In `shell/core/extensions/loader.js` we import and execute `@rancher/extensionsConfiguration` by passing it an object with a `configure` method that the import is aware of.
1. The `configure` method in the previous step will create the object which implements the `IExtension` interface and pass the object to the default method defined in the `config.ts` of each pkg.

### DIFF
--- a/shell/core/extensions/extension/index.ts
+++ b/shell/core/extensions/extension/index.ts
@@ -1,0 +1,82 @@
+import Router, { IRouter } from './router';
+
+export interface ExtensionMetadata {
+  name: string;
+  version: string;
+  description: string;
+  icon: string;
+}
+
+export interface IExtension {
+  id: string;
+  name: string;
+  description: string;
+  metadata: ExtensionMetadata;
+
+  isInitialized: boolean;
+  router: IRouter;
+}
+
+export class Extension implements IExtension {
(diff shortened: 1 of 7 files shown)


codyrancher:
No longer using

Issue #8319, opened 2023-03-02
[Backport v2.6] Rancher is frozen when User click on 'Auth Provider' button once it is already connected to an Auth


dasarinaidu:
**Setup:**

Rancher version: `v2.6-a55f968896341f311fc8302b3ed6247da3c13900-head`
Browser type & version: Chrome
Modes : Dark and Auto

**Describe the bug:**
Rancher is frozen when User click on 'Auth Provider' button once it is already loaded. Window is frozen and it says loading but it never loads. User has to refresh the browser to perform other operations.

**To Reproduce:**

1. Login to Rancher
2. Connect to one of the Auth Provider (AZUR)
3. Once it is connected
4. Click on the 'Auth Provider' button again
5. Check for the system behavior

**Result:**

1. Rancher window is grayed out and it says the page is loading but it never loads
2. No other buttons are working to perform any operation
3. User has to refresh the browser or hard refresh with keyboard to see rancher up.

**Expected Result**
Once user click on 'Auth Provider' button it should be on the same screen and it should not disable the screen. User should able to click on other buttons without a browser refresh.

Note : I m not seeing Loading page on any other screen navigations OR I am not seeing Loading Page after a page refresh

This issue is fixed on v2.7 but not on v2.6

Failed on v2.6-head and v2.6.11-rc9

Same issue fixed on v2.7 : https://github.com/rancher/dashboard/issues/7818

gaktive:
Backport PR: https://github.com/rancher/dashboard/pull/8660

dasarinaidu:
I have validated this issue on v`2.6-a52143b1fa790176f1db4d314e8d5a1f4f648309-head`. The issue got fixed and closing this card.

**Validation details as below:**

1. Chrome and Firefox browsers
2. Auth (Azure) Users
3. Dark and Lite mode

**Validation Steps**:

1. Login to Rancher
2. Connect to one of the Auth Provider (AZUR)
3. Once it is connected
4. Click on the 'Auth Provider' button again
5. Check for the system behavior

**Actual Result**
Once user click on 'Auth Provider' button user is  on the same screen, No disable the screen. User able to click on other buttons without a browser refresh.

PR #8660, opened 2023-04-13
Backport - Fix stuck loading indicator (auth)

Shavindra:
Fixes: https://github.com/rancher/dashboard/issues/8319
Original issue https://github.com/rancher/dashboard/issues/7818

The Nuxt loading indicator does not get removed when we prevent the navigation in this use case - this PR forces the loading indicator to be removed.


gaktive:
I think the actual fix is for #8319

nwmac (approved):
LGTM

PR #8662, opened 2023-04-13
Fix repeated style definitions from _mixins.scss

### LINKED ISSUE #8663: _mixins.scss contains style definitions
`_mixins.scss` defines not only mixins but also a few styles. This file in included in every Vue component's styles, so that they can access the mixins. This results in the styles be defined within every component.

e.g.

image

nwmac:
Fixes #8663 

This PR fixes an issue with our CSS.

`_mixins.scss` defines not only mixins but also a few styles. This file in included in every Vue component's styles, so that they can access the mixins. This results in the styles be defined within every component.

This PR moves the style declrations out of `_mixins.css`, so that they are only delcared once. Mixins now only declares actual scss mixins.

Also changed the comment in `_variables.scss` to use `//`, so that it is removed - the `/*` style of comments is preserved - meaning we get the comment included at the top of every style generated for every Vue component.

### DIFF
--- a/shell/assets/styles/base/_basic.scss
+++ b/shell/assets/styles/base/_basic.scss
@@ -1,3 +1,5 @@
+@import "./mixins";
+
 // -----------------------------------------------------------------------------
 // This file contains very basic styles.
 // -----------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


aalves08 (approved):
went around the UI for a couple of these classes and haven't observed any side-effect, so LGTM

PR #8679, opened 2023-04-17
Auth Provider: Support Okta with LDAP search

### LINKED ISSUE #8675: Ability to select groups within the Organizational Unit (OU) with Okta integration
We need a UI that enables users with OKTA integration to search all groups in the OU and not just the groups they are a part of.

We can leverage what we have already in place with Shibboleth, both for the SAML behaviour and the appearance, though @nwmac and @kwwii may have a design in mind already.

Related backend: https://github.com/rancher/rancher/issues/38029

nwmac:
Fixes #8675 

Adds the option to the OKTA Auth Provider to specify LDAP configuration for advanced group search. This follow the same pattern as the Shibboleth auth provider.

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -455,7 +455,13 @@ authConfig:
     objectClass: Object Class
     password: Password
     port: Port
+    protocol: Protocol
+    protocols:
+      starttls: Start TLS
+      ldap: LDAP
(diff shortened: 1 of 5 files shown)


aalves08 (approved):
LGTM although I've haven't tested it locally, but we ran through the demo together. I've seen you fixed that missing key/translation as well. 🙏

Issue #8455, opened 2023-03-16
Fleet: new checkbox to enable GitRepos


gaktive:
Internal reference: SURE-6037

Quoting @mattfarina:

> The fleet change to accommodate the request is to put a flag on GitRepo to keep the resource when deleted. By default they will be deleted as they are today. One opts in to keeping them. Fleet will remove the Helm chart without uninstalling any resources. If the Helm chart is reinstalled into the cluster with resources in the same namespace with the same names, the new chart instance will pick them up. If the name/namespace has a collision there will be an error to correct but the workloads will keep running.

> This new flag will be exposed in the Rancher UI screen for GitRepos. GitRepos are an existing screen in the dashboard.

gaktive:
Related Fleet PR: https://github.com/rancher/fleet/pull/1382

ronhorton:
**Pass** verified in rancher `v2.7-head commit ID 1f014b7`
1. navigate to local cluster
2. search for 'git repos'
3. create repo
4. new section `resource handling` has checkbox `keep resources after deletion`

![2023-05-09_11-32-49.png](https://zube.io/files/rancher/42f9c2434e639dde41a9c1a5a94fb9b3-2023-05-09_11-32-49.png)

PR #8684, opened 2023-04-18
Add keepResources property to GitRepo creation and edit.

Shavindra:
### Summary
Fixes rancher/dashboard#8455 

- Add a checkbox to toggle boolean value of  `value.spec.keepResources`
- Field is updatable via YAML as well.

### Technical notes summary
Related Fleet work:
- Issue: https://github.com/rancher/fleet/issues/680
- PR: https://github.com/rancher/fleet/pull/1382

- Ensure that Backend is using latest rancher head with the commit with PR: https://github.com/rancher/fleet/pull/1382
- Navigate to Continuous Delivery > Git Repos
- Click on `Add Repository` or Edit Existing repository
- There is a new section `Resource Handling` which allows to set the field.
- It can also be set via YAML.

- GitRepo creation and edit

### Screenshot/Video
image

### DIFF
--- a/shell/edit/fleet.cattle.io.gitrepo.vue
+++ b/shell/edit/fleet.cattle.io.gitrepo.vue
@@ -604,12 +604,6 @@ export default {
       </template>
       <div class="spacer" />
       <h2 v-t="'fleet.gitRepo.resources.label'" />
-
-      <Banner
-        color="info"
-      >
-        {{ t('fleet.gitRepo.resources.resourceBanner') }}
-      </Banner>
       <Checkbox
         v-model="value.spec.keepResources"
         class="check"
@@ -617,6 +611,11 @@ export default {
         label-key="fleet.gitRepo.resources.keepResources"
         :mode="mode"
       />
+      <Banner
+        color="info"
+      >
+        {{ t('fleet.gitRepo.resources.resourceBanner') }}
+      </Banner>
       <div class="spacer" />
       <h2 v-t="'fleet.gitRepo.paths.label'" />
       <ArrayList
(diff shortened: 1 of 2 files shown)


shell/edit/fleet.cattle.io.gitrepo.vue
```diff
@@ -601,7 +603,21 @@ export default {
         </div>
       </template>
       <div class="spacer" />
+      <h2 v-t="'fleet.gitRepo.resources.label'" />
 
+      <Banner
```
nwmac, line 608: I think we should modify the Banner text - I think the text should be clearer that the resource handling only happens when the checkbox is checked- suggest:

`When enabled below, resources will be kept when deleting a GitRepo or Bundle - only Helm release secrets will be deleted.`

nwmac (dismissed):
Text change, otherwise looks good to me

shell/edit/fleet.cattle.io.gitrepo.vue
nwmac: @kwwii see comment above - at present it looks like this:

image

The usual pattern is to have the content as part of the checkbox, under the checkbox label - but at least being clearer that this is the behaviour when checked seems important.

kwwii: Agreed, the text should read:
"When enabled, resources will be kept when deleting a GitRepo or Bundle - only Helm release secrets will be deleted"
and the info banner should be placed below the checkbox as it is elsewhere in the UI

nwmac:
@Shavindra Has the backend for this been added? If not, we should not merge, otherwise we're exposing UI that does not have the required backend support.

PR #8689, opened 2023-04-18
Revert "Added check box for truncate Hostnames (#8397)"

bisht-richa:
This reverts commit b161d67d7e5aba76e7e77778b86eb21c9d8f73f0.

### Summary
Fixes #

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -1689,9 +1689,6 @@ cluster:
     noAccessBanner: "You do not have access to see this machine pool's configuration."
     configNotFound: "We could not find this machine pool's configuration. We recommend that you create a new cluster with the desired configuration."
     noPoolsDisclaimer: You do not have any machine pools defined, click the plus to add one.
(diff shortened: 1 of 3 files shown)


nwmac:
@bisht-richa We should be able to look at the schema and see if it has the field we need - if we do this, we could merge the PR and the checkbox would only show when the field was added by the backend

nwmac:
@bisht-richa 

I'd suggest using this in the `data` method:

```
    // Is hostname truncation supported by the backend?
    const provSchema = this.$store.getters['management/schemaFor'](CAPI.RANCHER_CLUSTER);
    const specSchemaName = provSchema?.resourceFields?.spec?.type;
    const specSchema = specSchemaName ? this.$store.getters['management/schemaFor'](specSchemaName) : {};
    const supportsTruncation = !!specSchema?.resourceFields?.machinePoolDefaults;
```

and store `supportsTruncation` on the component data and use that in the `v-if` for showing the checkbox for hostname truncation.

bisht-richa:
Created a fix for this revert.

PR #8693, opened 2023-04-18
Fix issue where project owner does not have permission to update PSP template bindings for a project

### LINKED ISSUE #8508: Project owners/cluster members get permissions errors when adding members/groups when PSPs are enabled
It appears that there needs to be a global permission missing when project/cluster owners add users or groups to clusters/projects when PSPs are enabled

error creating binding: Forbidden 403: podsecuritypolicytemplateprojectbindings.management.cattle.io is forbidden: User "u-b4xv6mhhxu" cannot create resource "podsecuritypolicytemplateprojectbindings" in API group "management.cattle.io" in the namespace "c-7ldp6

Steps to reproduce:

1. enable PSPs on a cluster
2. create project owner and project member users
3. Create a project and set the PSP on the project to 'unrestricted'
4. add the project owner user as an owner of a project
5. log in as the project owner and edit the project to add the project member

You'll be faced with that error, but the project member will be added. It doesn't seem to do this when you navigate to the project and use the members page, only when editing the project to add users that way.

This same behavior seems to also happen when cluster members create projects when PSPs are enabled. Not sure if this should be considered a separate issue.

gz#16350
gz#16365

nwmac:
Fixes #8508 

This PR addresses the issue above.

Detailed steps to test are in the issue.

This PR:

- Disables the editing of the PSP on a project if the user does not have permissions to access PSP template bindings
- Updates the model so we only attempt to save the PSP Template binding if the value changes

UI when the user does not have permissions:

image

### DIFF
--- a/shell/config/types.js
+++ b/shell/config/types.js
@@ -177,6 +177,7 @@ export const MANAGEMENT = {
   GLOBAL_ROLE:                   'management.cattle.io.globalrole',
   GLOBAL_ROLE_BINDING:           'management.cattle.io.globalrolebinding',
   POD_SECURITY_POLICY_TEMPLATE:  'management.cattle.io.podsecuritypolicytemplate',
+  PSP_TEMPLATE_BINDING:          'management.cattle.io.podsecuritypolicytemplateprojectbinding',
   PSA:                           'management.cattle.io.podsecurityadmissionconfigurationtemplate',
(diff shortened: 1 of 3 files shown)


aalves08 (approved):
LGTM! just ran your branch and the PSP control is disabled and no errors on the console/network. 👍

PR #8654, opened 2023-04-11
Upgrade to Vue 2.7

cnotv:
### Summary
Fixes #8078

### Technical notes summary
Followed instruction as defined in the migration guide, adding them one commit at the time to ensure passing tests.

- `eslint-plugin-vue` upgraded to 9.10.0
- Pinned existing libraries which fail on automatic upgrade
- Added linting rules to prevent the build to fail
- Corrected some linting which cannot be overruled
- Updated `yarn.lock` to ensure dependencies to work properly
- Updated test app and package configuration
- Added resolution to enforce use of Webpack 4

PR #8695, opened 2023-04-18
Migrate `@rancher/components` to Vite

rak-phillip:
### Summary
Raising the draft now while we wait for the outcome of #8654.

I would like to move away from Vue CLI for a few reasons, including but not limited to:

* Vue CLI is in Maintenance Mode, Vite-based projects are now the default
* Build times with Vite are considerably faster, even with a small library like this

Fixes #


rak-phillip:
closing in favor of #8766

PR #8706, opened 2023-04-20
Fixes Advanced Socket Worker after Nuxt removal

Sean-McQ:
### Summary
Fixes #
It doesn't look like window.__NUXT__ actually exists anymore which makes sense since we removed NUXT but there's still a lot of references for it in code.

Removed a check for the window.__NUXT__ object which caused store initialization to return early.

### Technical notes summary
I didn't remove all references to the window.__NUXT__ object in code since that'd require pretty exhaustive testing but that's something we should look at in the near future.

Tested on chrome latest
Tester should see that with the advanced webworker enabled, the cluster socket never appears in the "ws" tab of chrome's dev tools indicating that either the socket was never created or the worker was never created (the latter being the actual cause).

Unknown

### DIFF
--- a/shell/plugins/dashboard-store/index.js
+++ b/shell/plugins/dashboard-store/index.js
@@ -38,7 +38,7 @@ export default (vuexModule, config, init) => {
     store.registerModule(namespace, vuexModule);
     store.commit(`${ namespace }/applyConfig`, config);
 
-    if ( !process.client || !window.__NUXT__ ) {
+    if ( !process.client ) {
       return;
     }
 


shell/plugins/dashboard-store/index.js
```diff
@@ -38,7 +38,7 @@ export default (vuexModule, config, init) => {
     store.registerModule(namespace, vuexModule);
     store.commit(`${ namespace }/applyConfig`, config);
 
-    if ( !process.client || !window.__NUXT__ ) {
+    if ( !process.client ) {
```
codyrancher, line 41: This check shouldn't be needed at all anymore since we don't do SSR.

richard-cox (approved):
Before this fix there were two parts that were skipped

- store `init`, which is basically just for web worker creation
- client store hydration from SSR page (not needed)

So before the only thing that was broken was something advanced worker based, which was probably how we missed it.

Lets get this one in and create an issue to understand the impact of the missing __NUXT__ in other places (looks like only in the shell/nuxt folder, hopefully only for SSR things)

PR #8707, opened 2023-04-20
Space between elements inside the Banner

scures:
### Summary
Fixes rancher/dashboard#8185

### Technical notes summary

Adds a `gap` propriety to the content element, this will ensure that all the children will evenly spread between them

### Screenshot/Video

Pasted Graphic

Pasted Graphic 1

### DIFF
--- a/pkg/rancher-components/src/components/Banner/Banner.vue
+++ b/pkg/rancher-components/src/components/Banner/Banner.vue
@@ -164,6 +164,7 @@ $icon-size: 24px;
     width: 100%;
     border-left: solid $left-border-size transparent;
     display: flex;
+    gap: 3px;
 
     .primary & {
       background: var(--primary);


pkg/rancher-components/src/components/Banner/Banner.vue
```diff
@@ -164,6 +164,7 @@ $icon-size: 24px;
     width: 100%;
     border-left: solid $left-border-size transparent;
     display: flex;
```
cnotv, line 166: Is there a reason that we converts these elements into Flex boxes? Considering this is text `inline-block` is usually the standard display type of all the text-like elements.
If we are not able to find out why, let's just approve the fix and review it another time.

cnotv: Yeah, but why would you inject a `paragraph` instead of a `span`? 😬 
That's ABC of the HTML and I would rather flag a PR which does that.

cnotv: Anyhow, since we do not want to make an inquisition about this, I'll rather create an issue for this case and remove all the injected values of this type, unless we can set any type of automation.

cnotv: Added https://github.com/rancher/dashboard/issues/8708

cnotv (approved):
Global issue with the component to be reviewed with https://github.com/rancher/dashboard/issues/8708
Temporary fix seems addressing the specific case.

LGTM

PR #8742, opened 2023-04-26
Update vue options to preserve whitespace

nwmac:
Addresses #8389

This PR fixes an issue with Nuxt removal where whitespace is not preserved.

This PR updates the webpack configuration to set whitespace to 'preserve' - in Vue it defaults to 'compact'.


bisht-richa (approved):
KGTM

Issue #8744, opened 2023-04-26
Zero to Hero Fixes


codyrancher:
- [ ] https://rancher.github.io/dashboard/extensions/extensions-getting-started#creating-an-extension
   - Make it clear where to put the options
   - Maybe add the options by default
- [x] https://rancher.github.io/dashboard/extensions/extensions-getting-started#configuring-an-extension
   - Fix the TS error regarding implicit any on the $plugin and store arguments
- [ ] Fix the missing `API=` in each of the getting started locations
- [ ] https://rancher.github.io/dashboard/extensions/extensions-getting-started#running-the-app
   - Link to the docs or explain how to edit the page that reads `Default dashboard for myProductName `.
- [ ]  https://rancher.github.io/dashboard/extensions/extensions-getting-started#loading-into-rancher
    - Look into moving dynamic loading for development into the advanced section?
    - Mention having to enable/install extension support
    - `EXCLUDES_PKG='test'` isn't working. The extension is still getting loaded.
- [ ] https://rancher.github.io/dashboard/extensions/extensions-getting-started#loading-into-another-rancher-instance
    - Give a friendly error message when running `yarn serve-pkgs` if `yarn build-pkg test` wasn't run? Or maybe run it for them?
    - Include the commands here too and possibly screenshots or a gif in both places.
    - Check to see why this was failing.
- [ ] https://rancher.github.io/dashboard/extensions/extensions-getting-started#release-prerequisites
    - It would be nice to link to gh-pages docs or provide screenshots

codyrancher:
Closing. We already did zero to hero.

PR #8745, opened 2023-04-26
Provide the relevant types for the product.ts and update the appropriate docs

codyrancher:
https://github.com/rancher/dashboard/issues/8744 - `Fix the TS error regarding implicit any on the $plugin and store arguments`

### Note
The `Leaving these here for completeness but I don't think these should be advertised as usable to plugin creators.` comment sections should probably be removed. I just wanted the both of you to get a chance to comment if you think they should be a part of the interfaces or not.

### DIFF
--- a/docusaurus/docs/extensions/api/nav/custom-page.md
+++ b/docusaurus/docs/extensions/api/nav/custom-page.md
@@ -4,12 +4,14 @@
 As we've seen from the previous chapter, a developer can register a top-level product with the `product` function. How about adding a custom page to your extension product? To do that, we can use the function `virtualType` coming from `$plugin.DSL`. As an example usage of that method, one could do the following:
 
 ```ts
+import { IPlugin } from '@shell/core/types';
+
 // this is the definition of a "blank cluster" for Rancher Dashboard
 // definition of a "blank cluster" in Rancher Dashboard
 const BLANK_CLUSTER = '_';
 
 
-export function init($plugin, store) {
+export function init($plugin: IPlugin, store: any) {
   const YOUR_PRODUCT_NAME = 'myProductName';
   const CUSTOM_PAGE_NAME = 'page1';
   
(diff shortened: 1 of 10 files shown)


docusaurus/docs/extensions/api/nav/custom-page.md
```diff
@@ -9,7 +9,7 @@ As we've seen from the previous chapter, a developer can register a top-level pr
 const BLANK_CLUSTER = '_';
 
 
-export function init($plugin, store) {
+export function init($plugin: IPlugin, store: any) {
```
aalves08, line 12: I believe we need to `import { IPlugin } from '@shell/core/types'`. We should add it to all places, imho.

shell/types/rancher/index.d.ts
```diff
@@ -5,3 +5,5 @@ declare module '@rancher/auto-import' {
 declare module '@shell/store/type-map' {
   export function DSL(store: any, name: string): any;
 }
+
+declare module '@shell/plugins/dashboard-store';
```
aalves08, line 9: Why do we need to declare this module?

aalves08 (commented):
Added a couple of comments @codyrancher . Thanks for taking care of this mate! 🙏  Let me know if you need any help with it.

docusaurus/docs/extensions/api/nav/custom-page.md
codyrancher: Good catch. I fixed it.

jordojordo (commented):
lgtm - just one minor comment about header options.

shell/core/types.ts
```diff
@@ -136,6 +137,248 @@ export type LocationConfig = {
...
+   */
+  // ifHaveVerb: string | RegExp;
+  // removable: string;
+  // showWorkspaceSwitcher: boolean;
+  // supportRoute: string;
+  // to: string;
+  // typeStoreMap: string;
+}
+
+export interface HeaderOptions {
```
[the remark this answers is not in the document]
codyrancher: Added

jordojordo (approved):

PR #8754, opened 2023-04-28
HARVESTER: add bare-metal container and mcm support

WuJun2016:
### Summary
- https://github.com/rancher/dashboard/issues/8800

change: 

- [x] Replace the correct harvester icon in the cluster list and Top Level Menu.
image
- [x] Enable the 'set as default page' feature on the Harvester page.
- [x] Add the ability to navigate to the Harvester Dashboard from the Harvester cluster on the Home page.
image
- [x] Disable & Filter Harvester-related actions in the Explorer page
     1.  The 'Upgrade Rancher' button will be hidden in the 'Install App' page."
     2.  In the 'Cluster Tools' page, hide all action buttons related to Harvester tools (e.g. Longhorn, Monitoring) 
     image
     image

     3.   CRDs prefixed with [harvesterhci.io](http://harvesterhci.io/).* and [kubevirt.io](http://kubevirt.io/).* will preferably not be displayed since they are Harvester system-level resources.
     image
     4. Editing/upgrading/deleting Harvester managed system services, such as monitoring, logging, and Harvester, will be disabled since these are Harvester system components that should be managed only by Harvester upgrade only.  （Note: 
 In order to prevent the user from operating the service related to harvester, we hide the resources belonging to the following namespace (harvester-system and cattle-system), because there is no way for the backend to add a specific annotation to a service like harvester）
      
image

Fixes #

### Screenshot/Video

Fix the issue of duplicate rendering of old route components when navigating from the Harvester page to the Explorer page within the same cluster.

https://user-images.githubusercontent.com/24985926/233048695-bfcfc5f0-501a-4a6a-88a2-df9f6b7a3d8f.mov

### DIFF
--- a/shell/components/nav/TopLevelMenu.vue
+++ b/shell/components/nav/TopLevelMenu.vue
@@ -303,59 +303,12 @@ export default {
             <div>
               <a
                 v-if="isRancherInHarvester"
-                class="option cluster selector home"
+                class="option"
                 @click="goToHarvesterCluster()"
               >
-                <!-- <i
+                <i
                   class="icon icon-dashboard"
-                /> -->
-                <svg
-                  class="harvester-dashboard"
-                  width="24"
-                  height="24"
-                  xmlns="http://www.w3.org/2000/svg"
-                >
-                  <g>
-                    <title>Layer 1</title>
-                    <path
-                      id="svg_1"
-                      stroke="null"
-                      d="m12.84736,9.44495l0,-7.66518l9.32087,0l0,7.66518l-9.32087,0zm-11.01558,3.40674l0,-11.07193l9.32087,0l0,11.07193l-9.32087,0zm11.01558,9.36856l0,-11.07193l9.32087,0l0,11.07193l-9.32087,0zm-11.01558,0l0,-7.66518l9.32087,0l0,7.66518l-9.32087,0zm1.6947,-11.07193l5.93146,0l0,-7.66518l-5.93146,0l0,7.66518zm11.01558,9.36856l5.93146,0l0,-7.66518l-5.93146,0l0,7.66518zm0,-12.77531l5.93146,0l0,-4.25844l-5.93146,0l0,4.25844zm-11.01558,12.77531l5.93146,0l0,-4.25844l-5.93146,0l0,4.25844zm5.93146,-9.36856zm5.08411,-3.40674zm0,5.11012zm-5.08411,3.40674z"
-                    />
-                    <rect
-                      id="svg_6"
-                      height="8.72054"
-                      width="6.9697"
-                      y="2.81481"
-                      x="3.04377"
-                      stroke="null"
-                    />
-                    <rect
-                      id="svg_7"
-                      height="5.65656"
-                      width="7.07071"
-                      y="15.60942"
-                      x="2.87542"
-                      stroke="null"
-                    />
-                    <rect
-                      id="svg_8"
-                      height="5.45454"
-                      width="7.03704"
-                      y="2.61279"
-                      x="13.78451"
-                      stroke="null"
-                    />
-                    <rect
-                      id="svg_9"
-                      height="8.88889"
-                      width="7.00337"
-                      y="12.3771"
-                      x="13.6835"
-                      stroke="null"
-                    />
-                  </g>
-                </svg>
+                />
                 <div>
                   {{ t('nav.harvesterDashboard') }}
                 </div>
--- a/shell/components/nav/TopLevelMenu.vue
+++ b/shell/components/nav/TopLevelMenu.vue
@@ -111,7 +111,16 @@ export default {
     multiClusterApps() {
       const options = this.options;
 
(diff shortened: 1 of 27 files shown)


WuJun2016:
Hi，@nwmac ，  Can this PR be merged into the 2.7.4 branch ?  Or will there be a new branch to release this feature, planned for release in May ?

Hi， @mantis-toboggan-md  PR can be reviewed now.  PTAL, thanks

mantis-toboggan-md:
@WuJun2016 there will be a new branch; I think this should target `master`

WuJun2016:
> @WuJun2016 there will be a new branch; I don't think we want this going into the v2.7.4 release

I will wait for the rancher team to switch to a new branch, then I will do cherry-pick

WuJun2016:
Hi,  @mantis-toboggan-md   Can you take some time to review this PR first?  Although this feature may still change, I don't think there will be any significant changes. If you could help me first check if the current changes are reasonable (whether in UX design or code), I think it would be very meaningful. This can help me identify problems in a timely manner (Because I think this PR needs to merge into the rancher repo as soon as possible, otherwise our QA may not have time to test it)

shell/pages/c/_cluster/explorer/index.vue
```diff
@@ -400,6 +407,21 @@ export default {
     // Events/Alerts tab changed
     tabChange(neu) {
       this.selectedTab = neu?.selectedName;
+    },
+
+    async goToHarvesterCluster() {
+      try {
```
mantis-toboggan-md, line 413: Could this be a method on the provisioning cluster type? This would be easier to maintain than having it defined in 3 places as it is now

shell/store/index.js
```diff
@@ -157,25 +159,28 @@ const updateActiveNamespaceCache = (state, activeNamespaceCache) => {
...
+    workspace:                 null,
+    error:                     null,
+    cameFromError:             false,
+    pageActions:               [],
+    serverVersion:             null,
+    systemNamespaces:          [],
+    isSingleProduct:           undefined,
+    namespaceFilterMode:       null,
+    supportRancherManage:      true,
+    openRancherManagerSupport: false, // Use the harvester embedded rancher
```
mantis-toboggan-md, line 182: This is a little nitpicky of me but it's not super clear what this variable is for in some of the other places its used and might be confusing for devs not familiar with the Harvester feature set. Something with Harvester in the name like `isRancherInHarvester`  would be clearer. Also it looks like `supportRancherManage` is not being used.

shell/list/service.vue
```diff
@@ -36,14 +36,26 @@ export default {
...
 
     await this.$fetchType(this.resource);
+  },
+
+  computed: {
+    filterRows() {
+      const currentCluster = this.$store.getters['currentCluster'];
+
+      if (currentCluster.isHarvester) {
+        return this.rows.filter(row => !['harvester-system', 'cattle-system'].includes(row.namespace));
```
mantis-toboggan-md, line 46: I think the requirement to filter these services was dropped from the PD&O

mantis-toboggan-md (changes requested):
I had a couple small suggestions and one question about requirements but overall this looks really good :)

gaktive:
Related backend for the feature: https://github.com/rancher/rancher/pull/41710

shell/pages/c/_cluster/explorer/index.vue
WuJun2016: Done

WuJun2016:
new commit change:  
1. fix comment
2. Display the harvester version on the explorer page
image
3. For the isRancherInHarvester mode cluster, we put the Virtualization Management menu into the hci category. If it is an external rancher, the Virtualization Management position remains the same (because we think if it is an external rancher, putting Virtualization Management at the top will feel more obtruse. This is not appropriate)
image
image
4. In the namespace edit page, I added some display logic related to harvester

shell/edit/namespace.vue
```diff
@@ -110,7 +110,15 @@ export default {
...
       return (this.$route.query[FLAT_VIEW] || false);
-    }
+    },
+
+    showPodSecurityAdmission() {
+      return !this.isStandaloneHarvester;
+    },
+
+    showHarvesterHelpText() {
+      return !this.isStandaloneHarvester && this.$store.getters['currentProduct'].inStore === HARVESTER;
```
mantis-toboggan-md, line 120: It looks like `isStandaloneHarvester` isn't defined here

shell/components/nav/TopLevelMenu.vue
```diff
@@ -269,10 +295,96 @@ export default {
...
+            <div>
+              <a
+                v-if="isRancherInHarvester"
+                class="option cluster selector home"
+                @click="goToHarvesterCluster()"
+              >
+                <!-- <i
+                  class="icon icon-dashboard"
+                /> -->
+                <svg
```
mantis-toboggan-md, line 312: Could this SVG be defined in /assets or something instead of in the component?

mantis-toboggan-md (changes requested):

WuJun2016:
Hi,  @mantis-toboggan-md   PTAL,  thanks

mantis-toboggan-md:
This is waiting for the backend PR to add the relevant feature flag: https://github.com/rancher/rancher/pull/41710

PR #8775, opened 2023-05-02
Feat/inactivity configuration

scures:
### Summary
Fixes rancher/dashboard#8322 
Fixes rancher/dashboard#8321

### Technical notes summary

Adds configuration into https://localhost:8005/c/local/settings/performance; enable & the inactivity modal display time.

## To test:
- Set the timer to 1 min
- After 56 seconds (90% of the total time), the modal will show and it will trigger the countdown which will go for the remaining 10% (6 seconds)
- Once it reaches 0, it will unsubscribe from all the stores (checking the console log will display:
image

### Screenshot/Video

image
image
image
image

### DIFF
--- a/shell/components/Inactivity.vue
+++ b/shell/components/Inactivity.vue
@@ -43,14 +43,14 @@ export default {
     // Total amount of time before the user's session is lost
     const thresholdToSeconds = settings?.inactivity?.threshold * 60;
 
-    // Amount of time the user sees the inactivty warning
+    // Amount of time the user sees the inactivity warning
     this.courtesyTimer = Math.floor(thresholdToSeconds * 0.1);
     this.courtesyTimer = Math.min(this.courtesyTimer, 60 * 5); // Never show the modal more than 5 minutes
     // Amount of time before the user sees the inactivity warning
     // Note - time before warning is shown + time warning is shown = settings threshold (total amount of time)
     this.showModalAfter = thresholdToSeconds - this.courtesyTimer;
 
-    console.debug(`Inactivty modal will show after ${ this.showModalAfter / 60 }(m) and be shown for ${ this.courtesyTimer / 60 }(m)`); // eslint-disable-line no-console
+    console.debug(`Inactivity modal will show after ${ this.showModalAfter / 60 }(m) and be shown for ${ this.courtesyTimer / 60 }(m)`); // eslint-disable-line no-console
 
     this.courtesyCountdown = this.courtesyTimer;
 


cnotv:
@scures could you please fix the issue found in E2E?

shell/components/Inactivity.vue
```diff
@@ -0,0 +1,208 @@
...
+      courtesyTimer:       null,
+      courtesyTimerId:     null,
+      courtesyCountdown:   null,
+      trackInactivity:     throttle(this._trackInactivity, 1000),
+    };
+  },
+  async mounted() {
+    const _settings = await this.$store.dispatch('management/find', { type: MANAGEMENT.SETTING, id: SETTING.UI_PERFORMANCE });
+
+    const settings = JSON.parse(_settings.value).inactivity;
```
cnotv, line 30: I get `expected end of JSON input` error.

cnotv: We should probably set a fallback and a chained condition for all the users which do not have any setting. I would recommend a unit test here, as it breaks the page.

cnotv: Yeah I've seen that too, but it did not work out 😬 
Looking forward to the updates 🍺

shell/assets/translations/en-us.yaml
```diff
@@ -2313,6 +2313,23 @@ growl:
...
+  bannerExpired: Your session has expired in this tab due to inactivity.
+  content: Click “Resume Session” to keep the session in this tab active or refresh the browser after the session has expired.
+  contentExpired: To return to this page click “Refresh” below or refresh the browser.
+  cta: Resume Session
+  ctaExpired: Refresh
+  settings:
+    title: Inactivity
+    checkboxLabel: Enable inactivity session expiration
+    inputLabel: Inactivity timeout (minutes)
+    information: To change the automatic logout behaviour, edit the authorisation and/or session token timeout values (auth-user-session-ttl-minutes and auth-token-max-ttl-minutes) in the general settings.
```
richard-cox, line 2329: `To change the automatic logout behaviour, edit the authorisation and/or session token timeout values ('auth-user-session-ttl-minutes' and 'auth-token-max-ttl-minutes') in the Settings page.`

The settings could be in `code` elements if they looks nice

shell/pages/c/_cluster/settings/performance.vue
```diff
@@ -23,6 +23,7 @@ export default {
   async fetch() {
     try {
       this.uiPerfSetting = await this.$store.dispatch('management/find', { type: MANAGEMENT.SETTING, id: SETTING.UI_PERFORMANCE });
+      this.authUserTTL = await this.$store.dispatch(`management/find`, { type: MANAGEMENT.SETTING, id: SETTING.AUTH_USER_SESSION_TTL_MINUTES });
```
richard-cox, line 26: This try/catch is specific to the UI_PERFORMANCE setting request. If the request to fetch AUTH_USER_SESSION_TTL_MINUTES fails then the former gets reset to default. Can it be in it's own try/catch with a fall back value?

shell/pages/c/_cluster/settings/performance.vue
```diff
@@ -89,8 +101,39 @@ export default {
...
+            <LabeledInput
+              v-model="value.inactivity.threshold"
+              :mode="mode"
+              :label="t('inactivity.settings.inputLabel')"
+              :disabled="!value.inactivity.enabled"
+              class="input mb-10"
+              type="number"
+              min="0"
+            />
+            <Banner
```
richard-cox, line 124: This should probably use the existing validation feature. There should be a prop that accepts a fn which if it fails will return an error message. There should be some docs of it in our dev kit somewhere, though if in doubt reach out to Sean

richard-cox: That might also help with the visually valid `-1` (don't think `LabeledInput` emits a value for minus figures but doesn't block the user from entering it)

shell/components/Inactivity.vue
```diff
@@ -0,0 +1,215 @@
...
+    if (!settings || !settings?.inactivity || !settings?.inactivity.enabled) {
+      return;
+    }
+
+    this.enabled = settings?.inactivity?.enabled || false;
+
+    const thresholdToMinutes = settings?.inactivity?.threshold * 60;
+
+    this.showModalAfter = parseInt(thresholdToMinutes * 0.9); // convert minutes, triggers at 90% completion
+    this.courtesyTimer = Math.floor(thresholdToMinutes * 0.1); // 10% of the threshold in minutes, goes by for the remaining 10% of the threshold
```
richard-cox, line 41: Was this the end point decision on ratio? If this is high then that's a lot of time. For example 60 minute inactivity means 6 minutes of the modal on screen

richard-cox (changes requested):
- The timeouts are never cleared when the component is destroyed. So navigating to pages with different layouts (instances of Inactivity) results in multiple instances of checkInactivityTimer running

shell/pages/c/_cluster/settings/performance.vue
cnotv: We had a really wide range of validation cases in https://github.com/rancher/dashboard/issues/6633 which may be worthy to reuse if the requirements are the same.
Specifically [these](https://github.com/rancher/dashboard/blob/master/shell/config/settings.ts#L102-L113) rules:
>
```
      {
        name: 'isInteger',
        key:  'Password', <-- term to be replaced for the error
      },
      {
        name: 'isPositive',
        key:  'Password',
      },
      {
        name: 'isOctal',
        key:  'Password',
      }
```

cnotv: For the records the prop is `rules` and the method of the mixin is `fvGetAndReportPathRules`.
It would be required to add the whole validation to the CRUD component in order to have a real meaning.

Some info about it are defined in the devkit [here](https://rancher.github.io/dashboard/code-base-works/forms-and-validation#form-validation).

cnotv (dismissed):
LGTM

If users has never saved performances settings, it's required to do so, otherwise it's disabled.

shell/pages/c/_cluster/settings/performance.vue
cnotv: @richard-cox that page is full of inputs which have no validation at all.
I would recommend to create a validation for the whole page in a separated issue, otherwise we'll have issues with the consistency.

shell/assets/translations/en-us.yaml
richard-cox: `in the general settings.` --> `in the Settings page.`

shell/assets/translations/en-us.yaml
```diff
@@ -2313,6 +2313,23 @@ growl:
...
+inactivity:
+  title: Session expiring
+  titleExpired: Session expired
+  banner: Your session is about to expire due to inactivity. Any unsaved changes will be lost.
+  bannerExpired: Your session has expired in this tab due to inactivity.
+  content: Click “Resume Session” to keep the session in this tab active or refresh the browser after the session has expired.
+  contentExpired: To return to this page click “Refresh” below or refresh the browser.
+  cta: Resume Session
+  ctaExpired: Refresh
+  settings:
```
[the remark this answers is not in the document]
richard-cox: The entire inactivity section was moved to the performance page's translations, which include translated not related to it. Which is the reversal of the original issue.

The `performance` section covers translations in, and only of, the performance page. So the current contents of `performance.inactivity.settings` section should move to `performance.inactivty`

The `Components & Pages` section would be a good place for the inactivty modal translations. So the current `performance.inactivtiy` section (minus `settings`) should move to a new `inactivity` section there.

shell/pages/c/_cluster/settings/performance.vue
richard-cox: Correct. Though thinking about it if we can't fetch a `authUserTTL` value we shouldn't set a default and then not validate for it.

shell/pages/c/_cluster/settings/performance.vue
richard-cox: This was a small task, and to avoid introducing technical debt and a different validation ux i've pushed a commit with the fix in

shell/components/Inactivity.vue
richard-cox: Any news on this?

shell/pages/c/_cluster/settings/performance.vue
```diff
@@ -30,6 +30,12 @@ export default {
       this.uiPerfSetting.metadata = { name: SETTING.UI_PERFORMANCE };
     }
 
+    try {
+      this.authUserTTL = await this.$store.dispatch(`management/find`, { type: MANAGEMENT.SETTING, id: SETTING.AUTH_USER_SESSION_TTL_MINUTES });
+    } catch (error) {
+      return;
```
richard-cox, line 36: This shouldn't return, just swallow the error `} catch { }`, or alternatively log the error.

shell/pages/c/_cluster/settings/performance.vue
```diff
@@ -56,9 +64,26 @@ export default {
...
+
+    canSave() {
+      return this.value.inactivity.enabled ? this.isInactivityThresholdValid : true;
+    }
   },
 
   methods: {
+    validateInactivityThreshold(value) {
+      if (!this.authUserTTL?.value) {
+        return;
```
richard-cox, line 76: If there's no authUserTTL validation has basically passed, so need to set `this.isInactivityThresholdValid = false; ` before returning. Without this, currently, when inactivity is enabled the field is always invalid and the changes can never be saved

shell/assets/translations/en-us.yaml
richard-cox: ` in the General settings page.` --> `in the Settings page.`

If the user looks for a General Settings page they won't find one, but they will find a `Settings` page that's in current context.

image

If there really 100% needs to be general in there, it should be `in the general Settings page`.

shell/components/Inactivity.vue
richard-cox: I've pushed a fix with this in

shell/pages/c/_cluster/settings/performance.vue
richard-cox: i made a mistake with the suggestion, it should be set to true, updating in incoming commit

richard-cox (approved):
- Make sure to double check my last commit. It contains some important changes, primarily fixing an issue that caused a white page crash if the UI_PERFORMANCE setting was missing
- This PR also only links to the config issue, link to the other issue with the usual `Fixes: #` text

richard-cox (approved):
note - the issue linking needs to be in separate fixes, so 

```
Fixes https://github.com/rancher/dashboard/issues/8322
Fixes https://github.com/rancher/dashboard/issues/8321
```

PR #8788, opened 2023-05-04
HARVESTER: expose suppress_type_name for opensearch provider

DaiYuzeng:

- https://github.com/harvester/harvester/issues/3589

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -2749,6 +2749,7 @@ logging:
     clientKeyPass: Client Key Pass from Secret
     verifySsl: Verify SSL
     sslVersion: SSL Version
+    suppressTypeName: Suppress Type Name
   redis:
     host: Host
     port: Port
--- a/shell/edit/logging.banzaicloud.io.output/providers/opensearch.vue
+++ b/shell/edit/logging.banzaicloud.io.output/providers/opensearch.vue
@@ -45,6 +45,11 @@ export default {
       if (typeof this.value.ssl_version === 'undefined') {
         this.$set(this.value, 'ssl_version', sslVersions[0]);
       }
+
+      // Suppress type name by default
+      if (typeof this.value.suppress_type_name === 'undefined') {
+        this.$set(this.value, 'suppress_type_name', true);
+      }
     }
 
     return { protocolOptions: protocol, sslVersions };
@@ -196,6 +201,12 @@ export default {
         <Checkbox
           v-model="value.ssl_verify"
           :label="t('logging.elasticsearch.verifySsl')"
+          :disabled="disabled"
+          :mode="mode"
+        />
+        <Checkbox
+          v-model="value.suppress_type_name"
+          :label="t('logging.elasticsearch.suppressTypeName')"
           :disabled="disabled"
           :mode="mode"
         />


shell/edit/logging.banzaicloud.io.output/providers/opensearch.vue
```diff
@@ -45,6 +45,11 @@ export default {
       if (typeof this.value.ssl_version === 'undefined') {
         this.$set(this.value, 'ssl_version', sslVersions[0]);
       }
+
+      // Suppress type name by default
+      if (typeof this.value.suppress_type_name === 'undefined') {
+        this.$set(this.value, 'suppress_type_name', true);
```
richard-cox, line 51: Enabling this by default for an optional parameter is an interesting mix. This would also enable the property when editing an existing resource. @bashofmann i think you've had experience with this provider, does the change make sense?

DaiYuzeng: Hi @richard-cox ! Thanks for your review!

I think it won't affect the behavior of exisiting resource since we have the prerequisite at line 34.

```
if (this.mode === _CREATE) {
```

bashofmann: According to the docs (https://docs.fluentbit.io/manual/pipeline/outputs/opensearch#action-metadata-contains-an-unknown-parameter-type) this flag should be `true` for OpenSearch >= 2.0.0.

I think it's fine to have this on for new outputs by default

It would be good to add the info what this flag is about and when it must be on as a help text in the UI as well.

Also this should be added to the ElasticSearch output as well, there it is mandatory to be on for ElasticSearch >= 8.0.0 (https://docs.fluentbit.io/manual/pipeline/outputs/elasticsearch#action-metadata-contains-an-unknown-parameter-type).

DaiYuzeng: Thanks for your review! I'll push the changes today.

DaiYuzeng: Help text is `Hide type names by default to improve readability.`

Does it make sense? Can you please take a re-review? @bashofmann Thanks!

image

bashofmann: There should be a space between `server.` and `You`, so `server. You`. Otherwise looks great.

shell/assets/translations/en-us.yaml
```diff
@@ -2749,6 +2749,10 @@ logging:
     clientKeyPass: Client Key Pass from Secret
     verifySsl: Verify SSL
     sslVersion: SSL Version
+    suppressTypeName:
+      label: Suppress Type Names
+      elasticSearchTips: Suppress type names by default to be compatiable with log server.You should enable it for Elasticsearch >= 8.0.0
```
bashofmann, line 2754:
```suggestion
      elasticSearchTips: Suppress type names by default to be compatible with log server. You should enable it for Elasticsearch >= 8.0.0
```

shell/assets/translations/en-us.yaml
```diff
@@ -2749,6 +2749,10 @@ logging:
...
+      openSearchTips: Suppress type names by default to be compatiable with log server.You should enable it for OpenSearch >= 2.0.0
```
bashofmann, line 2755:
```suggestion
      openSearchTips: Suppress type names by default to be compatible with log server. You should enable it for OpenSearch >= 2.0.0
```

bashofmann (changes requested):

DaiYuzeng:
@bashofmann  Thanks for your suggestions! It has been merged into this PR.

bashofmann (approved):

PR #8790, opened 2023-05-04
Hostname truncation: Update to sync with backend

nwmac:
Updates so property matches backend - machinePools was under spec.rk3eConfig, not at the top level

Removed check for schema, as property will always be in backend now.


bisht-richa (approved):
LGTM - Tested locally

PR #8792, opened 2023-05-04
minor updates for extensions

aalves08:
- docs minor update 
- add gitignore file to creator npm package


shell/creators/app/files/.gitignore
jordojordo, line 1: There's a lot in here that I don't think we should ignoring automatically. Maybe we should only ignore leftover files that are generated from developing/building extensions and let the user decide for further integrations.

jordojordo (changes requested):

shell/creators/app/files/.gitignore
aalves08: I've lightened it up quite a bit. I think the remaining items are "industry standards" and should be kept as a basic config.
Let me know if you feel it needs further changes. 🙏

jordojordo: Looks good, thanks for the change 🧙

jordojordo (approved):

PR #8809, opened 2023-05-06
fix object overwrite & rerender

WuJun2016:
### Summary
Using the management/find method here will cause the value in the store to overwrite the value in the machine pool (because the value of `machinePool[*].config` has not been cloned at this point).

Fixes #
- https://github.com/harvester/harvester/issues/3786


WuJun2016:
Hi @mantis-toboggan-md   Can this issue fix be merged into `2.7.4` ?

richard-cox (changes requested):
The 2.7 Q2 release, currently `rancher/dashboard` `master`, will be the next release this can be included in

mantis-toboggan-md (dismissed):

mantis-toboggan-md (approved):

mantis-toboggan-md:
I added a comment to explain the change, otherwise LGTM

PR #8832, opened 2023-05-09
Sanitize project label `v-html`

rak-phillip:
### Summary
This removes the need to use `v-html` when displaying project labels by shifting the `span` tag outside of the translations inside of `en-us.yaml` and into the vue template.

Listing projects and namespaces.

#### Before

image

#### After

image


mantis-toboggan-md (approved):
lgtm

shell/assets/translations/en-us.yaml
```diff
@@ -4587,7 +4587,7 @@ resourceTable:
     machinePool: "<span>Pool:</span> {name}"
     notInANamespace: Not Namespaced
     notInAProject: Not in a Project
-    project: "<span>Project:</span> {name}"
+    project: "Project:"
```
nwmac, line 4590: The name param is here so that for rtl locales, - can't always assume that the name should go after the label.

nwmac (changes requested):

rak-phillip:
@mantis-toboggan-md @nwmac I've updated this PR to utilize `DOMPurify`. Can you take another look?

PR #8836, opened 2023-05-10
Add transformer for require using Babel

### LINKED ISSUE #5631: Configure Jest to mock require.context
**Detailed Description**
Configure Jest so it is possible to create test for ~SSR code~ logic using `require` Node function.

**Context**
As we use SSR require dynamic imports, it is not possible to create test for components which use the function `require.context`, as it needs to be mocked or returns error with message

```
TypeError: require.context is not a function.
```

Test example: [Storage component](https://github.com/rancher/dashboard/blob/master/shell/edit/workload/storage/__tests__/Storage.test.ts)

cnotv:
### Summary
Fixes #5631

### Technical notes summary
- Added `babel-plugin-transform-require-context` library
- Converted `babel.config.js` into a function and added conditionally logic if test

Try to remove comment to the example file and error should change (test is still incomplete 😅)

### DIFF
--- a/package.json
+++ b/package.json
@@ -159,6 +159,7 @@
     "babel-jest": "27.5.1",
     "babel-loader": "8.2.4",
     "babel-plugin-istanbul": "6.1.1",
+    "babel-plugin-transform-require-context": "0.1.1",
     "babel-preset-vue": "2.0.2",
     "core-js": "3.21.1",
     "css-loader": "4.3.0",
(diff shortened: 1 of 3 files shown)


codyrancher (approved):
I think this is fine. We should double check our extensions won't need to be updated. I don't think they will.

rak-phillip (approved):
LGTM - I'm noticing some CI steps failing. Attempting to rerun those.

cnotv:
> I think this is fine. We should double check our extensions won't need to be updated. I don't think they will.

So far I have never been running unit tests for the extensions, so I don't know even for all the other configurations we have how do they behave.
This change is aimed to the dashboard only, as we use no `require` in `rancher-components`.

PR #8838, opened 2023-05-10
Fix import for clone from utils/object - Bump 0.3.9 

jordojordo:
### Summary

This fixes the import for the `clone` method as this is causing build failures with extensions using the chart mixin. Example [here](https://github.com/kubewarden/ui/actions/runs/4938978264/jobs/8829388513#step:9:145)

Also bumps `0.3.9` of `@rancher/shell`


aalves08 (approved):
LGTM

PR #8272, opened 2023-02-27
Remove jquery from dependencies

torchiaf:
Signed-off-by: Francesco Torchia [francesco.torchia@suse.com](mailto:francesco.torchia@suse.com)

### Summary
Contributes to #5070
Contributes to https://github.com/rancher/dashboard/issues/4534

### Areas which could experience regressions

`UI Extensions`, `shell` package

PR #8869, opened 2023-05-12
Revert "Remove jquery from dependencies"

mantis-toboggan-md:
Reverts rancher/dashboard#8272
Seen as part of: https://github.com/rancher/dashboard/issues/8686
Removing jQuery has broken some extensions: I tried harvester 1.0.3, harvester 1.1.0, elemental 1.0.0 and elemental 1.1.0...it's likely others are broken too. 
Screen Shot 2023-05-12 at 11 50 54 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,6 +84,7 @@
     "is-base64": "^1.1.0",
     "is-url": "1.2.4",
     "jexl": "2.2.2",
+    "jquery": "3.5.1",
     "js-cookie": "2.2.1",
     "js-yaml": "4.1.0",
     "js-yaml-loader": "1.2.2",
--- a/shell/core/plugins-loader.js
+++ b/shell/core/plugins-loader.js
@@ -1,4 +1,5 @@
 import Vue from 'vue';
+import $ from 'jquery';
 import JSZip from 'jszip';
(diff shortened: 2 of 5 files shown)


nwmac (approved):

cnotv:
Should extension not be independent packages?

cnotv:
Also we should probably extend this test to cover this https://github.com/rancher/dashboard/pull/8128 @aalves08

cnotv:
Just another point: I got the same error when accessing the pods:

image

nwmac:
@cnotv The extensions only use jquery becuase it is used in dashboard - they are standalone, but to reduce bundle size, we externalise jquery and provide if from the shell - since the old extension required jquery, it broke as we were no longer providing it.

cnotv:
> @cnotv The extensions only use jquery becuase it is used in dashboard - they are standalone, but to reduce bundle size, we externalise jquery and provide if from the shell - since the old extension required jquery, it broke as we were no longer providing it.

But would not that force us to keep it forever to support the old extension versions?
In any case, I've create [this issue](https://github.com/rancher/dashboard/issues/8874) exactly to investigate this layer of complexity.

Issue #8723, opened 2023-04-24
prometheus.prometheusSpec.storageSpec.volumeClaimTemplate.spec.selector created in monitor values YAML, when installing rancher-monitoring chart as YAML


nwmac:
Internal reference: SURE-2997

**Repo steps:**

When installing momnitoring:

- Deploy local-path provisioned in downstream https://github.com/rancher/local-path-provisioner
- Go to Cluster Explorer > Apps & Marketplace -> Charts -> Monitoring > Prometheus and 'Enable Persistent Storage for Prometheus', selecting local-path StorageClass
- Click 'Edit as YAML'
- Click 'Install'

Install will fail:

prometheus-rancher-monitoring-prometheus-db-prometheus-rancher-monitoring-prometheus-0 volume in cattle-monitoring-system Namespace fails to successfully provision with the following error:

failed to provision volume with StorageClass "local-path": claim.Spec.Selector is not supported

The error that needs fixing is that when you click 'Edit as YAML', we populate an empty selector: `volumeClaimTemplate.spec.selector` - this should be removed before being sent to the backend.

aalves08:
Pattern can be found here:
https://github.com/rancher/dashboard/pull/8643

PR #8889, opened 2023-05-16
Remove empty selector from prometheus.

Shavindra:
### Summary
Fixes rancher/dashboard#8723

When installing monitoring chart and setting up prometheus. If selector is empty then remove it before the request is sent.

### Technical notes summary
```
      if (
        selector &&
        isEmpty(selector?.matchExpressions) &&
        isEmpty(selector?.matchLabels)
      ) {
        delete values.prometheus.prometheusSpec.storageSpec
          .volumeClaimTemplate.spec.selector;
      }

```
- Deploy local-path provisioned in downstream https://github.com/rancher/local-path-provisioner using kubectl shell commands
- Go to Cluster Explorer > Apps & Marketplace -> Charts -> Monitoring > Prometheus and 'Enable Persistent Storage for Prometheus', selecting local-path StorageClass
- Click 'Edit as YAML'
- Click 'Install'

- Creating monitoring chart

### DIFF
--- a/cypress/e2e/tests/pages/explorer/workloads/pods.spec.ts
+++ b/cypress/e2e/tests/pages/explorer/workloads/pods.spec.ts
@@ -48,7 +48,7 @@ describe('Cluster Explorer', () => {
 
           const clonedPodPage = new WorkLoadsPodDetailsPagePo(clonePodName);
 
-          clonedPodPage.goTo();
+          clonedPodPage.goTo().wait(10000);
 
           cy.wait('@clonedPod')
             .then(({ response }) => {


shell/pages/c/_cluster/apps/charts/install.vue
```diff
@@ -1029,6 +1030,19 @@ export default {
         setIfNotSet(cattle, 'windows.enabled', true);
       }
 
+      // Global prometheus storage values get added to the values
```
mantis-toboggan-md, line 1033: This doesn't look like the right place to perform this cleanup; everything modified in `addGlobalValuesTo` is in `<rancher-monitroing chart>.values.global` while the prometheus configuration is within `<rancher-monitoring chart>.values.prometheus`...this should probably be addressed in a monitoring-scoped file instead of this general one

shell/chart/monitoring/index.vue
```diff
@@ -176,6 +176,10 @@ export default {
       const selector =
         prometheusSpec?.storageSpec?.volumeClaimTemplate?.spec?.selector;
 
+      // This works for UI editor installation
+      // However, it doesn't work for yaml editor installation
+      // Global values later merged again in charts/install.vue addGlobalValuesTo()
+      // We still need to remove the global values from charts/install.vue addGlobalValuesTo()
```
mantis-toboggan-md, line 182: After trawling through the monitoring code, I think this original solution is just not great. That's hardly your fault but I think https://github.com/rancher/dashboard/issues/8723 can be fixed without adding further complexity. The empty selector is initialized when the 'persistent storage for prometheus' checkbox is clicked:
https://github.com/rancher/dashboard/blob/d91591ceb70e22e5de497c2eb8db5f0a284f9ede/shell/chart/monitoring/prometheus/index.vue#L149-L160

Rather than the original solution of always adding something that breaks some storage providers then removing it before save, I think we could just stop initializing the `volumeClaimTemplate` this way (ie delete line 157 linked above) and update `matchChanged` in the same file to delete `volumeClaimTemplate.spec.selector` if all selectors are cleared from the UI https://github.com/rancher/dashboard/blob/d91591ceb70e22e5de497c2eb8db5f0a284f9ede/shell/chart/monitoring/prometheus/index.vue#L173-L184

Would become something like:
```
    matchChanged(expressions) {
      const { matchLabels, matchExpressions } = simplify(expressions);
      const storageSpec = this.value.prometheus.prometheusSpec.storageSpec.volumeClaimTemplate.spec;

      if (storageSpec?.selector && isEmpty(matchLabels) && isEmpty(matchExpressions)) {
        delete storageSpec.selector;
      } else {
        this.$set(storageSpec, 'selector', { matchLabels, matchExpressions });
      }
    },
```

mantis-toboggan-md (changes requested):

mantis-toboggan-md (approved):
lgtm

PR #8896, opened 2023-05-16
fix issue with Tolerations component

### LINKED ISSUE #8886: `all` toleration value throws an error in the cluster for cluster / fleet agent customization setting
**Setup**
- Rancher version: v2.7-head (b04c942)
- Rancher UI Extensions: n/a

**Describe the bug**

when using the `all` option for rke2 `appendTolerations` through the UI, fleet will throw an error

**To Reproduce**

* provision an rke2 cluster
  * add a fleet agent customization -> toleration using the `all` value

**Result**
cluster will provision, but fleet won't deploy to the cluster

**Expected Result**

only valid options should be in the dropdown

**Screenshots**

Screen Shot 2023-05-15 at 2 02 23 PM

### LINKED ISSUE #8910: toleration `key` has header: `label key` for cluster / fleet agent customization fields. Expected only `key` or possibly `taint key`
**Setup**
- Rancher version: v2.7-head(42f3b50)
- Rancher UI Extensions: n/a

**Describe the bug**

the text for a toleration key is currently set to `label key` which may be confusing, since tolerations are for taints, and affinity are for labels. 

**To Reproduce**

* deploy an rke1 or rke2 cluster
 * go to cluster or fleet agent customization and add a toleration

**Result**
key reads `label key`

**Expected Result**

`key` or possibly `taint key` 

**Screenshots**

Screen Shot 2023-05-16 at 10 27 58 PM

**Additional context**

when adding a taint, the field reads as just `key`

aalves08:
Fixes #8886 

- fixes issue with `Tolerations` component not covering the spec correct correctly when talking about effect `All`. It should not send string `effect: All`, but `effect: null` as per the kubernetes spec https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#toleration-v1-core

Fixes #8910 
- Change label in Tolerations from `Label Key` to `Key`

Follow reproduction steps on issue

### DIFF
--- a/cypress/e2e/tests/pages/data/agent-configuration-rke2-payload.ts
+++ b/cypress/e2e/tests/pages/data/agent-configuration-rke2-payload.ts
@@ -264,13 +264,11 @@ export const payloadComparisonData = {
       {
         key:      'key1',
         operator: 'Exists',
-        effect:   null
       },
       {
         key:      'key2',
         operator: 'Equal',
         value:    'val2',
-        effect:   null
       },
       {
         key:      'key3',
@@ -568,13 +566,11 @@ export const payloadComparisonData = {
       {
         key:      'key1',
         operator: 'Exists',
-        effect:   null
       },
       {
         key:      'key2',
         operator: 'Equal',
         value:    'val2',
-        effect:   null
       },
       {
         key:      'key3',
--- a/cypress/e2e/tests/pages/data/agent-configuration-rke2-payload.ts
+++ b/cypress/e2e/tests/pages/data/agent-configuration-rke2-payload.ts
@@ -264,13 +264,13 @@ export const payloadComparisonData = {
       {
         key:      'key1',
         operator: 'Exists',
-        effect:   'All'
+        effect:   null
       },
       {
         key:      'key2',
         operator: 'Equal',
         value:    'val2',
-        effect:   'All'
+        effect:   null
       },
       {
         key:      'key3',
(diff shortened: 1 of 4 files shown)


mantis-toboggan-md (changes requested):
When 'All' is selected, nothing shows in the dropdown:
https://github.com/rancher/dashboard/assets/42977925/53571659-f17f-45e8-a59b-55e17cbe22b5

I think this component doesn't handle `null` options well. One possible workaround would be to set the 'All' option to something else: 
```
 effectOpts() {
      return [
        {
          label: this.t('workload.scheduling.tolerations.effectOptions.all'),
          value: ''
        },
        ...
```

and then strip this value in the `update` method 
```
        if (newRule.effect === '') {
          delete newRule.effect;
        }
    },
```

aalves08:
> When 'All' is selected, nothing shows in the dropdown:
> 
>  Screen.Recording.2023-05-17.at.1.28.05.PM.mov 
> I think the select component doesn't handle `null` options well. One possible workaround would be to set the 'All' option to something else:
> 
> ```
>  effectOpts() {
>       return [
>         {
>           label: this.t('workload.scheduling.tolerations.effectOptions.all'),
>           value: ''
>         },
>         ...
> ```
> 
> and then strip this value in the `update` method
> 
> ```
>         if (newRule.effect === '') {
>           delete newRule.effect;
>         }
>     },
> ```

You're right @mantis-toboggan-md . I don't know how I missed that. Thanks for pointing it out to me. 🙏

mantis-toboggan-md (approved):

PR #8905, opened 2023-05-17
Fix some warnings from unit tests

nwmac:
Fixes some of the warnings during the unit test run


jest.setup.js
```diff
@@ -3,17 +3,24 @@ import { directiveSsr as t } from '@shell/plugins/i18n';
...
+import ClientOnly from 'vue-client-only';
 
 import Vue from 'vue';
 
 Vue.config.productionTip = false;
 Vue.use(VTooltip).use(VModal);
 Vue.component('v-select', vSelect);
 
+// Component: <ClientOnly> (originally from Nuxt)
+Vue.component(ClientOnly.name, ClientOnly);
```
cnotv, line 15: We had no issue related to Nuxt, why are we adding this?

jest.setup.js
```diff
@@ -3,17 +3,24 @@ import { directiveSsr as t } from '@shell/plugins/i18n';
...
+
+Vue.component('codemirror', () => {});
```
cnotv, line 17: Code mirror issues have to be handled within it's own test.

jest.setup.js
```diff
@@ -24,7 +31,13 @@ beforeEach(() => {
   // jest.clearAllMocks(); // Use this function inside your test if you need to reset mocks
   jest.restoreAllMocks(); // Use this function inside your test if you need to reset mocks and restore existing functionality
 
-  config.mocks['$store'] = { getters: { 'i18n/t': jest.fn() } };
+  config.mocks['$store'] = {
+    getters: {
+      'i18n/t':    jest.fn(),
+      'prefs/get': jest.fn()
```
cnotv, line 37: Global mocks are just required for the i18n as it's an uncommon pattern.
We should not add anything here, unless is strictly required by every component, which is not the case.

cnotv (changes requested):
Please let's revert merged tests which create issues or fix the tests. The setup has no issues.

cnotv (changes requested):
Please let's revert merged tests which create issues or fix the tests. The setup has no issues.

jest.setup.js
nwmac: I'd argue prefs is common too, its just we don't have enough tests that use it - as we write more tests we're going to end up mocking it in every tests rather that using the same pattern as i18n/t

jest.setup.js
nwmac: Its a global component that is normally loaded by the code mirror provider - if we add the component in the keyvalue test, it will then be available globally, so it seems to make more sense to add it globally - which is effectively what we do in the app.

Under your argument, v-select should not be added here either.

PR #8919, opened 2023-05-18
Purify tooltip content with `v-clean-tooltip` directive

rak-phillip:
### Summary

This replaces the use of the `v-tooltip` directive with a custom `v-clean-tooltip` directive that uses the `DOMPurify` library to sanitize HTML input before rendering. `html` usage is discouraged by `v-tooltip`[^1], and should only be used on trusted content. Similar to #8466, we disallow `html` by default for `v-tooltip` and sanitize existing usages across the board throughout Rancher Dashboard.

- Set `defaultHtml: false` in `v-tooltip`
- Adds new `v-clean-tooltip` directive that sanitizes tooltip content based on rules defined in `shell/plugins/clean-html-directive.js`
- Replaces all existing instances of `v-tooltip` with `v-clean-tooltip` 

### Technical notes summary

In essence, we are monkey patching the `bind` method that `v-tooltip` utilizes to configure the directive. By taking this approach, we can ensure that we are able to sanitize tooltip content without altering the original behavior of the `v-tooltip` directive.

We need to investigate tooltips throughout Rancher Dashboard and ensure that there are no notable regressions after this change has been applied. 

[^1]: https://floating-vue.starpad.dev/legacy/v2/#directive


mantis-toboggan-md (approved):
LGTM: I verified that tooltips with allowed tags still render correctly, and tooltips with other v-tooltip options (`placement`, `classes`) look right.

rak-phillip:
@jordojordo can you please do a quick validate of Kubewarden?

rak-phillip:
@aalves08 can you review and validate extensions?

mantis-toboggan-md (changes requested):
The Harvester 1.0.3 plugin's tooltips load just fine. 

However I did notice a few places in /shell where we're using `v-tooltip.bottom` - should these be updated to use `v-clean-tooltip` as well? 
* `shell/components/CompoundStatusBadge`
* `shell/components/fleet/FleetRepos`
* `shell/components/formatter/PodImages`

rak-phillip:
@mantis-toboggan-md that's a good catch, thank you. Ready for another review.

mantis-toboggan-md (approved):

aalves08 (approved):
@rak-phillip this is not something straightforward to test on an extension as it's dependent on the shell package, however I did install Elemental (using your branch) and all looks good. I don't think it's the proper test, to be fair, as I believe that we would need to get this in first, then bump the shell package in order for it to be used on extensions.
As far as Elemental goes, there is no "custom" usage of `v-tooltip`, meaning that I'll be having tooltips from whatever `@shell/components` provides to me (where v-tooltip is used from there).
If those were already tested and passed, then Elemental should be fine as well.

jordojordo (approved):
@rak-phillip as Alex mentioned, it's not easily testable for Kubewarden. I attempted to test using `yarn link` with your fork but Kubewarden is still using Nuxt (I haven't updated the `@rancher/shell` version), so I was unable to run the test. But I don't think this will cause issues. - Everything lgtm otherwise

PR #8948, opened 2023-05-22
Backport/2.7.4 harvester/chore/publish docusaurus

rak-phillip:
### Summary
This backports changes made to our publish docusaurus step so that it will only run when changes are made to the master branch.


rak-phillip:
BAD PR - wrong target branch

PR #8951, opened 2023-05-22
add cluster-name label to prometheus recording and alerting rules

### LINKED ISSUE #8902: Adding cluster-name label to new Prometheus Recording rules created from the UI.
When an alert fires it does not mention from which cluster it comes. I have made the change to include cluster_name to all existing service monitors in the monitoring chart, but we need this change to be reflected on the UI for all new alerts. 

Related Tickets:  https://github.com/rancher/dashboard/issues/7513
Backend Ticket: https://github.com/rancher/rancher/issues/37128

Request: 
UI add another default label to all rules that to include the cluster_name. (Repeat steps for cluster_id label.) 

Screen Shot 2022-11-21 at 11 31 28 AM

mantis-toboggan-md:
### Summary
Fixes #8902 

This PR adds another default label to prometheus alerting rules, as we do with the cluster id (#7935):
Screen Shot 2023-05-22 at 3 00 42 PM

### DIFF
--- a/shell/edit/monitoring.coreos.com.prometheusrule/GroupRules.vue
+++ b/shell/edit/monitoring.coreos.com.prometheusrule/GroupRules.vue
@@ -33,9 +33,10 @@ export default {
       expr:   '',
       for:    '0s',
       labels: {
-        severity:   'none',
-        namespace:  'default',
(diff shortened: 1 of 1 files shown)


rak-phillip (approved):
LGTM

PR #8991, opened 2023-05-25
Updating the types for extensions

codyrancher:
Found that these types were necessary for some of our docs and example projects


aalves08 (approved):
lgtm

PR #8992, opened 2023-05-25
Add the ability to execute a container shell command immediately after the shell connects.

codyrancher:
Example usage:
image

Note:
I'm using `paste` instead of `write` because `write` either wasn't getting a response or wasn't processing the response. It would just render a message in the terminal.


jordojordo (approved):
This is great, thanks for adding this!

Issue #5529, opened 2020-11-17
Cluster Explorer fails with HTTP Error 500 when cluster agent is down


dnoland1:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
1. Create a custom cluster on Rancher v2.5.2.
2. Attach a one node RKE cluster with all roles by using the docker run... command.
3. Pause the cattle cluster agent on the node by running a command similar to `docker pause k8s_cluster-register_cattle-cluster-agent-7c9b787668-r5mz9_cattle-system_c012e40d-b5fe-4b1b-857f-a36d1f678b11_1`
4. For the newly created cluster, click the Cluster Explorer button.

**Result:**
Get 500 error:
<img width="1679" alt="image" src="https://user-images.githubusercontent.com/468462/99451181-aa49ae80-28d6-11eb-9ee0-f2dc6695d93a.png">

**Other details that may be helpful:**
Would expect Rancher to failover to using the cattle node agent if the cattle cluster agent is not functional.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
v2.5.2
- Installation option (single install/HA):
HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
Custom 1.18.8
- Machine type (cloud/VM/metal) and specifications (CPU/memory):
cloud aws t3.medium (2 CPU, 4GB RAM)
- Kubernetes version (use `kubectl version`):

```
 kubectl version
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-16T11:56:40Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.8", GitCommit:"9f2892aab98fe339f3bd70e3c470144299398ace", GitTreeState:"clean", BuildDate:"2020-08-13T16:04:18Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
```

- Docker version (use `docker version`):

```
docker version
Client: Docker Engine - Community
 Version:           19.03.13
 API version:       1.40
 Go version:        go1.13.15
 Git commit:        4484c46d9d
[...]

PR #8994, opened 2023-05-26
Display a clear error message.

Shavindra:
### Summary
Fixes rancher/dashboard#5529 

Render the message.

- Set server to trigger an error 500
- NB: original instructions in rancher/dashboard#5529 on triggering 500 can work as well. However, redirect may happen very quickly.
- Load page that would trigger 500

### Screenshot/Video
Screenshot 2023-05-26 at 09 16 30

### DIFF
--- a/shell/pages/fail-whale.vue
+++ b/shell/pages/fail-whale.vue
@@ -35,7 +35,7 @@ export default {
     },
 
     displayError() {
-      return stringify(this.error);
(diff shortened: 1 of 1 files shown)


cnotv (approved):
As QA has already be done for this issue since years, I guess it makes more sense for them to check it again.
We should write a ticket for the automation at a certain point, more generic than just this fragment though.

PR #9007, opened 2023-05-30
Add pod clone E2E

Shavindra:
### Summary
Fixes #https://github.com/rancher/dashboard/pull/8733 - E2E only
Original functional Fix: https://github.com/rancher/dashboard/pull/8733

Add E2E

### Technical notes summary
- Create initial pod via Kubectl
- Clone the pod
- No functional changes.

### DIFF
--- a/cypress/e2e/tests/pages/explorer/workloads/pods.spec.ts
+++ b/cypress/e2e/tests/pages/explorer/workloads/pods.spec.ts
@@ -15,7 +15,6 @@ describe('Cluster Explorer', () => {
         const { name: origPodName, namespace } = createPodBluerint.metadata;
         const { name: clonePodName } = clonePodBlueprint.metadata;
 
-
         beforeEach(() => {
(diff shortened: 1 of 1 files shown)


cypress/e2e/blueprints/explorer/workload-pods.ts
```diff
@@ -0,0 +1,29 @@
+export const createPodBluerint = {
```
cnotv, line 1: Fixture have already a specific section.

Screenshot 2023-05-30 at 17 14 16

cnotv (changes requested):
As you may see from the linter, it's not allowed to use arbitrary `wait()` time.
We should try to rely on `intercept()` of the single requests.

If the case with the CLI does not generate any requests, we should think about how to tackle that down, probably better with a command to reuse?

cnotv (approved):
LGTM

PR #9008, opened 2023-05-30
Move /epinio/about to /:product/c/:cluster/about

torchiaf:
### Summary
Fixes https://github.com/epinio/ui/issues/241


richard-cox (changes requested):
This works, but it would be nice to ensure the route name follows convention. 

`${ EPINIO_PRODUCT_NAME }-about` --> `${ EPINIO_PRODUCT_NAME }-c-cluster-about`

Which means

```
aboutPage:           createEpinioRoute('about', { cluster: EPINIO_STANDALONE_CLUSTER_NAME }),
-->
aboutPage:           createEpinioRoute('c-cluster-about', { cluster: EPINIO_STANDALONE_CLUSTER_NAME })
```

In theory this should clash with the resource route, but given the order they load we're fine

```
{
  name:      `${ EPINIO_PRODUCT_NAME }-c-cluster-resource`,
  path:      `/:product/c/:cluster/:resource`,
  component: ListEpinioResource,
}
```

richard-cox (approved):

PR #9011, opened 2023-05-30
Fix malformed pod affinity and node affinity terms

mantis-toboggan-md:
Fixes https://github.com/rancher/dashboard/issues/9012 - this PR is for specifically the first issue described in this comment https://github.com/rancher/dashboard/issues/9012#issuecomment-1569155864 . As noted in slack discussion, the other two potential causes of the 502 error may be more complex to address so we're starting here.

The dashboard is sending an extra weight field within node/pod preferred terms,  `nodeAffinity.preferredDuringSchedulingIgnoredDuringExecution[].preference.weight` we should only be sending `nodeAffinity.preferredDuringSchedulingIgnoredDuringExecution[].weight` Similarly the UI is sending `podAffinity.preferredDuringSchedulingIgnoredDuringExecution[].podAffinityTerm.weight` and should only send `podAffinity.preferredDuringSchedulingIgnoredDuringExecution[].weight`

### DIFF
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -97,6 +97,7 @@ export default defineConfig({
     },
     experimentalSessionAndOrigin: true,
     specPattern:                  getSpecPattern(),
-    baseUrl
+    baseUrl,
+    numTestsKeptInMemory:         0
   },
 });
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -97,7 +97,6 @@ export default defineConfig({
     },
     experimentalSessionAndOrigin: true,
     specPattern:                  getSpecPattern(),
-    baseUrl,
-    numTestsKeptInMemory:         0
+    baseUrl
   },
 });
--- a/cypress/e2e/tests/pages/data/agent-configuration-rke2-payload.ts
+++ b/cypress/e2e/tests/pages/data/agent-configuration-rke2-payload.ts
@@ -350,7 +350,6 @@ export const payloadComparisonData = {
                   ]
                 }
               ],
-              weight: 10
             }
           }
         ],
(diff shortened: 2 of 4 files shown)


mantis-toboggan-md:
I've updated this PR to also remove the `matching` field from `matchExpression` and `matchField` terms, for the same reason the `weight` field needs to be removed. I also found an issue (fix included here) where if a node selector term has a matchExpression defined and a matchField defined, then the matchField is deleted, it's gone from the UI but preserved in the spec (as seen when switching to yaml view)

aalves08 (dismissed):
LGTM @mantis-toboggan-md . Thanks for the fix 🙏

aalves08:
adding @cnotv to this review, just for the config changes in Cypress you've made. Not sure if it has any relevant impact. 

@cnotv is this ok to enter?

cnotv:
> adding @cnotv to this review, just for the config changes in Cypress you've made. Not sure if it has any relevant impact. Looks fine to me.
> 
> @cnotv is this ok to enter?

I have no idea why we are setting it to 0 globally, but I guess it makes sense on CI though, as you cannot navigate through the snapshots. May I ask you what's the benefit of changing it in this context?

mantis-toboggan-md:
I've removed that change now -browser was running out of memory running tests locally but it doesn't need to be configured globally

aalves08 (approved):
lgtm!

cnotv:
> I've removed that change now -browser was running out of memory running tests locally but it doesn't need to be configured globally

Oh ok that's something to keep in mind then. I was not able to run the tests at all recently, so maybe it works for me too 😅

PR #9017, opened 2023-05-31
Fix pod clone e2e wait time

Shavindra:
Cloned pod page take too long to load in CICD


richard-cox (approved):

cnotv (approved):
LGTM

It still fails sometimes though, so we should check it later again anyway

Issue #8885, opened 2023-05-15
Embedded Grafana dashboards don't load in the local cluster with the latest version of monitoring


mantis-toboggan-md:
This came out of an investigation into https://github.com/rancher/rancher/issues/41562 and ties back in to recent work here: https://github.com/rancher/dashboard/pull/8347/files

To reproduce:
1. Install rancher monitoring 102.0.1+up40.1.2 in the local cluster
2. Navigate to the local cluster's explorer dashboard  (`/dashboard/c/local/explorer`)
Result:
Embedded grafana pages fail to load

Additionally, the grafana link on the monitoring index page is not adding this url exception for the local cluster as described in https://github.com/rancher/dashboard/issues/8346. We need to update the hardcoded link [here](https://github.com/rancher/dashboard/blob/master/shell/pages/c/_cluster/monitoring/index.vue#L61) to use the `getClusterPrefix` utility.

The assumption in https://github.com/rancher/dashboard/pull/8347 was that for monitoring versions >=102.0.0+up40.1.2 the grafana url would not need to be different for the local cluster - this isn't the case. See more details from @nflynt  [here](https://github.com/rancher/rancher/issues/41562#issuecomment-1548556453)

We need to update the grafana url-constructing utility [here](https://github.com/rancher/dashboard/blob/master/shell/utils/grafana.js#L9) to include the `/k8s/clusters/${ clusterId }` prefix for local clusters **only when monitoring `100.2.0+up40.1.2` or `102.0.0+up40.1.2` is installed** and continue using this prefix for all downstream clusters regardless of version.

w13915984028:
As described in https://github.com/harvester/harvester-installer/pull/492

In https://charts.rancher.io/index.yaml, there is `assets/rancher-monitoring/rancher-monitoring-102.0.0+up40.1.2.tgz`, but no `102.0.1+up40.1.2` , which is not tformally released yet ? **who can help confirm if `102.0.1+up40.1.2` is released?**  thanks.

And, in 102.0.0+up40.1.2, there is such a bug, that is fixed in 102.0.1+up40.1.2, this is an very important bug fix for us.

https://github.com/rancher/charts/blob/dev-v2.7/charts/rancher-monitoring/102.0.1%2Bup40.1.2/charts/grafana/templates/nginx-config.yaml
```
          {{- if eq .Values.global.cattle.clusterId "local" -}}
[...]

PR #9019, opened 2023-05-31
fix grafana url prefix for local clusters

mantis-toboggan-md:
### Summary
Fixes #https://github.com/rancher/dashboard/issues/8885

* updated grafana url-construction utility to only apply the `/k8s/clusters/${ clusterId }` prefix for monitoring versions `100.2.0+up40.1.2` and `102.0.0+up40.1.2`
* updated the grafana link on the monitoring index page to use the above utility
* updated the embedded grafana dashboard component to use the correct url for iframe src

### Technical notes summary
The GrafanaDashboard component was using a url generated before fetching the installed monitoring version, so it was wrong some of the time. I've updated it to use the `currentUrl` computed property which properly accounts for the monitoring version.

Test both `100.2.0+up40.1.2` and `102.0.0+up40.1.2` monitoring each on a local and downstream cluster. Verify that the embedded grafana views and grafana link on the cluster dashboard work, and verify that the grafana link on the monitoring index works.

I hit issues intermittently when runing this code locally - sometimes requests to the grafana service proxy returned 401. To get around this I built and uploaded the dashboard, so rather than running this code locally you can also test this PR by changing your ui-dashboard-index setting to https://releases.rancher.com/dashboard/nancy-dev/index.html

### DIFF
--- a/shell/components/GrafanaDashboard.vue
+++ b/shell/components/GrafanaDashboard.vue
@@ -40,7 +40,7 @@ export default {
   },
   data() {
     return {
-      loading: false, error: false, interval: null, initialUrl: this.computeUrl(), errorTimer: null, monitoringVersion: null
+      loading: false, error: false, interval: null, errorTimer: null, monitoringVersion: null
     };
   },
   computed: {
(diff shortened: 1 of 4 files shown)


jordojordo (approved):
lgtm

PR #9036, opened 2023-06-05
Handle String Error Messages.

Shavindra:
### Summary
Fixes rancher/dashboard#5529

- Handle strigified error message

### Technical notes summary
- BE could send error stringified messages. 
- Check if the message is a string and if it can be parsed to JSON.
- If not possible just return the message
- Make sure the JSON response still handled

1. Fresh install of Rancher `v2.7-head`
2. Provision a downstream cluster w/ 1 node (all roles)
5. Back in Rancher UI, navigate to `Cluster Explorer > Workloads > Pods > All Namespaces` for the downstream RKE1 cluster
6. Back in terminal, scale down `cattle-cluster-agent` replicas to `0`:
```kubectl
kubectl -n cattle-system scale deploy cattle-cluster-agent --replicas=0
```
7. Observe `cattle-cluster-agent` begin termination via Rancher UI
8. Hard refresh the page
9. There are multiple errors and first error you'd get is related to namespace. There is a specific error `Lost Connection to cluster...` This is the one that need to be tested and comes after the namespace error. So need to wait about a second or so to refresh. If refresh is too quick then you'd get Namespace error. Too late? Redirect to Home page.

### DIFF
--- a/shell/utils/error.js
+++ b/shell/utils/error.js
@@ -43,8 +43,9 @@ export function stringify(err) {
       if (typeof str === 'string') {
         try {
           const data = JSON.parse(str).data;
+
           if (data) {
-            return data
+            return data;
           }
         } catch {}
       }


richard-cox (changes requested):
Dived in a bit more and there's a change of behaviour / fall back mostly revolving around bypassing `stringify`.


1) previously if there's no error.data we fall back on the purpose built error parser `stringify`. If that had error.message.detail it'd get used. With the new code it won't (if error.message exists but fails to parse only error.message is used).
  Can we move the try block from fail-whale `displayError` into `stringify` before the part we check for err.detail? something like

```
      str = err.message;
      if (typeof str === 'string') {
        try {
          const data = JSON.parse(str).data;
          if (data) {
            return data
          }
        } catch {}
      }

      if ( err.detail ) {
        if ( str ) {
          str += ` (${ err.detail })`;
        } else {
          str = err.detail;
        }
      }

```

I gave this is quick test and it seemed to work

image

2) is the early exit given `error?.data` needed anymore? 

FWIW just rebooting the node can give the same error without borking deployments

`this.error` `toString` --> `{"data":"lost connection to cluster: failed to find Session for client stv-cluster-c-gzf6w"}`
`this.error` error object --> 
```
{
    "type": "error",
    "status": 500,
    "statusText": "Internal Server Error",
    "url": "/k8s/clusters/c-gzf6w/v1/schemas",
    "message":  "{\"data\":\"lost connection to cluster: failed to find Session for client stv-cluster-c-gzf6w\"}"
}
```

PR #9041, opened 2023-06-05
Modify zube auto-labelling action

nwmac:
Updates the auto-labelling GH Action workflow to ignore all labels starting `[zube]:` when deciding whether to auto-label an issue with the `Triage` label.

Have to do a bit of work to check the labels - ended up using a small bash script to do this.


nwmac:
@richard-cox I have a test repo I can give you access to, if you want to test this.

richard-cox (approved):

PR #9051, opened 2023-06-06
Remove child yarn.lock files (docusaurus and storybook)

nwmac:
This PR removes the two `yarn.lock` files that we have for Docusaurus and Storybook.


aalves08 (approved):
lgtm

PR #9054, opened 2023-06-06
add e2e tests for users/roles list view for standard user

### LINKED ISSUE #8777: Automation: List global roles/users pages for standard users when they have custom permissions assigned
Taken from work done in https://github.com/rancher/dashboard/issues/8344, we should write an automated test to ensure that we do list the appropriate material in Page Configuration >> Users & Authentication pages for a standard user who has the roles or users.

E2E feels like the goal

To reproduce:
1. As an admin, create a global role by navigating to Users and Auth >> Roles >> Global Roles >> Verbs: List, Get & Resources: Global Roles
2. Create a standard user as an admin and assign the role
3. Log in as standard user

Expected result:
The user to be able to list/get the pages `Users & Auth`

Factors:
- We need to add a test suite around non-admin users.

aalves08:
Fixes #8777 

- Adds e2e test


aalves08:
closed in favor of https://github.com/rancher/dashboard/pull/9097

PR #9056, opened 2023-06-06
Automation: App breaks when creating a new namespace from a workload

### LINKED ISSUE #8806: Automation: App breaks when creating a new namespace from a workload
Based on work done in https://github.com/rancher/dashboard/issues/8632, let's write an automated test to confirm that nothing breaks when creating a new namespace from a workload.

@cnotv can offer guidance (or maybe he has a test written somewhere)

To Reproduce: 
- Access to Cluster explorer
- Open any Workload section list, e.g. Jobs
- Create a new Job
- Create a new namespace instead of picking one

Expected behaviour:
- Namespace is created and the UI doesn't break

aalves08:
Fixes #8806 

- Adds e2e tests

### DIFF
--- a/cypress/e2e/po/components/labeled-input.po.ts
+++ b/cypress/e2e/po/components/labeled-input.po.ts
@@ -11,6 +11,13 @@ export default class LabeledInputPo extends ComponentPo {
     );
   }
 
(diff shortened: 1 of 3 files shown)


jordojordo (approved):
lgtm

PR #9063, opened 2023-06-07
e2e: Add built UI to docker container rather than run a node process

nwmac:
This PR changes the way e2e CI tests run the dashboard - instead of running a node process to serve the UI, we run the docker container and map a volume to the built code - this has the advantage that this is closer to how things work in production.

We know the GitHub test runners are resource-constrained, so hopefully having one less process consuming resources may help make the tests run a tiny bit quicker and possibly slightly more reliably, if things are more responsive.


richard-cox (dismissed):
As discussed, at some point we should replace scripts/build-e2e with script/build-embedded.. but not now though

richard-cox (approved):
LGTM. I'm going to merge this to assist with CI straightaway

nwmac:
Thanks @richard-cox

PR #5850, opened 2022-05-05
Add more secret selectors for receiver configs

catherineluse:
# Changes in this PR

## New secret selectors
This PR adds the last 7 secrets for the receivers in AlertmanagerConfigs. These are required to configure the `httpConfig` section defined here: https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#httpconfig

The API docs lacked details about how to configure the `tlsConfig` part (it didn't have a blue link to its definition) so I found its type def in the Prometheus operator code and went off of that:

```go
type SafeTLSConfig struct {
    // Struct containing the CA cert to use for the targets.
    CA SecretOrConfigMap `json:"ca,omitempty"`
    // Struct containing the client cert file for the targets.
    Cert SecretOrConfigMap `json:"cert,omitempty"`
    // Secret containing the client key file for the targets.
    KeySecret *v1.SecretKeySelector `json:"keySecret,omitempty"`
    // Used to verify the hostname for the targets.
    ServerName string `json:"serverName,omitempty"`
    // Disable target certificate validation.
    InsecureSkipVerify bool `json:"insecureSkipVerify,omitempty"`
}
```

- in the `httpConfig` section:
	- `bearerTokenSecret`
	- `basicAuth.username`
	- `basicAuth.password`
- in the `tlsConfig` section which is grouped under httpConfig:
	- CA (`CA`) - SecretOrConfigMap (the same API as the other secrets, but nested under a `secret` field)
	- Client Cert (`Cert`) - SecretOrConfigMap (the same API as the other secrets, but nested under a `secret` field)
	- Client Key (`KeySecret`) - SecretKeySelector
	- Server Name (`ServerName`) - String

The new secret selectors are available from the webhook config form:
Screen Shot 2022-05-05 at 1 49 58 PM

When all secrets are selected, the webhook config yaml looks something like this:

```yaml
webhookConfigs:
        - httpConfig:
            tlsConfig:
              ca:
                secret:
                  name: test-secret
                  key: username
              cert:
                secret:
                  name: test-secret
                  key: username
              keySecret:
                name: test-secret
                key: username
            basicAuth:
              username:
                key: username
                name: test-secret
              password:
                key: password
                name: test-secret
```
Note that either `basicAuth` or `bearerTokenSecret` should be configured, but not both. So without basic auth filled in, it should look like this:

```yaml
webhookConfigs:
        - httpConfig:
            tlsConfig:
              ca:
                secret:
                  name: test-secret
                  key: username
              cert:
[...]

PR #9065, opened 2023-06-07
Fix `selecting` event in LabeledSelect.vue

torchiaf:
### Summary
Fixes https://github.com/epinio/ui/issues/205

We need this to fix a wrong behavior on Epinio namespace selectors, where we have the option to create a new namespace.

### Technical notes summary

The 'Create a new namespace' option is on top of the namespaces list, in Epinio UI.
When the user edit the new namespace's name and then clicks on 'X' icon, the creation option is no more selectable.
The Namespace dropdown element is made on top of `LabeledSelect.vue` that send a selecting event when a new option is selected by the user:

```
@option:selected="$emit('selecting', $event)"
```
When the creation option is clicked for the second time, the `@option:selected` is not fired since v-select already have this value internally.
This is fixed by replace @option:selected with @option:selecting, which is fired every time an option is selected, regardless of what is the value stored by `v-select`.
Difference between selected and selecting for `v-select`: https://vue-select.org/api/events.html#open

### DIFF
--- a/shell/components/form/LabeledSelect.vue
+++ b/shell/components/form/LabeledSelect.vue
@@ -285,7 +285,7 @@ export default {
       @search="onSearch"
       @open="onOpen"
       @close="onClose"
-      @option:selected="$emit('selecting', $event)"
+      @option:selected="$emit('selected', $event)"
     >
       <template #option="option">
         <template v-if="option.kind === 'group'">
--- a/shell/components/form/LabeledSelect.vue
+++ b/shell/components/form/LabeledSelect.vue
@@ -285,6 +285,7 @@ export default {
       @search="onSearch"
       @open="onOpen"
       @close="onClose"
+      @option:selecting="$emit('selecting', $event)"
       @option:selected="$emit('selected', $event)"
     >
       <template #option="option">
--- a/shell/components/SortableTable/index.vue
+++ b/shell/components/SortableTable/index.vue
@@ -1062,7 +1062,7 @@ export default {
                   :multiple="false"
                   :taggable="false"
                   :placeholder="t('sortableTable.selectCol')"
-                  @selecting="(col) => advFilterSelectedLabel = col.label"
+                  @selected="(col) => advFilterSelectedLabel = col.label"
                 />
               </div>
               <div class="bottom-block">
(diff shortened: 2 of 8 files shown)


shell/components/form/LabeledSelect.vue
```diff
@@ -285,7 +285,7 @@ export default {
       @search="onSearch"
       @open="onOpen"
       @close="onClose"
-      @option:selected="$emit('selecting', $event)"
+      @option:selecting="$emit('selecting', $event)"
```
cnotv, line 288: I searched in the code if we have any of these 2 and while we don't use `@selected`, there's a bunch of `@selecting` already.
I guess this should change the behaviors, not sure if it's in good or bad 😬 

Also the component has also both, so not so sure if we are overriding the other one, which in any case we don't use, but we'll change the behaviors at that point.

Screenshot 2023-06-07 at 15 37 35

torchiaf: I am reconsidering this change, since the `option:selecting` event is fired _before_ updating the value field; This can cause a lot of regressions.
We would specify both `selecting` and `selected`, leave the `selected` for other components and use selecting inside `NameNsDescription` where I am pretty much sure that updating the value after the selecting event doesn't cause any error.
What do you think Richard ?

torchiaf: Please have a look to the new commits, this way the event handling is much more safer.

richard-cox: From memory this area caused a large regression in the past. There's a root component that used selecting or selected that clashed with what we were using. Can you check the git blame on the original bindings and see if you can find the history? i have a feeling it might be doing something that's required

richard-cox: Which are, as always, subject to the nature of the business. Lets investigate if this was an area of previous regression first and if the events emitted need to stay the same. If so comments and unit test issue to follow. Either way there's probably a less impactfull fix that can be found.

The PR also points to `master`, so needs to wait until we release 2.7.5 or branch for 2.7q3

richard-cox: Please look through 
https://github.com/rancher/dashboard/pull/5850
https://github.com/rancher/dashboard/pull/5999
https://github.com/rancher/dashboard/pull/6302

richard-cox:
Superseded by https://github.com/rancher/dashboard/pull/9112

PR #9066, opened 2023-06-07
Automation: Namespace cannot inherit the Resource Limit default value from the project

### LINKED ISSUE #8786: Automation: Namespace cannot inherit the Resource Limit default value from the project
Based on work done in https://github.com/rancher/dashboard/issues/8205, let's write an automated test to check that namespaces are inheriting values from the project.

@mantis-toboggan-md should be able to provide guidance upon request.

To Reproduce

- Create a project and set the default value of Container Default Resource Limit
- Create a namespace in this project, check the default value of Container Resource Limit

Expected
- Inherited default values from project are in the namespace.

cnotv:
### Summary
Fixes #8786

### Technical notes summary
- Add test to ensure that resource limit configuration is inherited by the project
- Prevent errors in case of missing `annotations`

### DIFF
--- a/shell/edit/__tests__/namespace.test.ts
+++ b/shell/edit/__tests__/namespace.test.ts
@@ -1,0 +1,46 @@
+import { mount } from '@vue/test-utils';
+import Namespace from '@shell/edit/namespace.vue';
+import { DefaultProps } from 'vue/types/options';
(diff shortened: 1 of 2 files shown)


torchiaf (approved):
LGTM

PR #9071, opened 2023-06-08
Fix/key value upload file

torchiaf:
### Summary
Fixes https://github.com/epinio/ui/issues/243

### Technical notes summary

The upload file functionality for value fields is implemented; We should discuss about the UX, I have prepared some solutions, contained in the single commits of this PR.
Please, check the last 2 commits, we will manage the merging branch in a second step.

0- Current Version
image

1- inline button
image

2- Small buttons
small-buttons

3- Large buttons
large-buttons

4- Label buttons
upload-span-style

### DIFF
--- a/shell/components/form/KeyValue.vue
+++ b/shell/components/form/KeyValue.vue
@@ -550,7 +550,12 @@ export default {
       this.$set(this.codeMirrorFocus, idx, value);
     },
     onValueFileSelected(idx, file) {
-      this.rows[idx][this.valueName] = file;
+      const { name, value } = file;
+
+      if (!this.rows[idx][this.keyName]) {
+        this.rows[idx][this.keyName] = name;
+      }
+      this.rows[idx][this.valueName] = value;
     },
     isValueFieldEmpty(value) {
       return !value || value.trim().length === 0;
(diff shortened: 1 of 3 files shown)


torchiaf:
Thanks for feedback @scures, I also like the button inside the value field.
Below the new version, we can discuss about more improvements.

image

shell/components/form/KeyValue.vue
```diff
@@ -662,39 +672,51 @@ export default {
...
+                :disabled="isView || disabled || isProtected(row.key)"
+                :type="valueConcealed ? 'password' : 'text'"
+                :placeholder="valuePlaceholder"
+                autocorrect="off"
+                autocapitalize="off"
+                spellcheck="false"
+                @input="queueUpdate"
+              >
+              <FileSelector
+                v-if="parseValueFromFile && readAllowed && !isView && row[keyName] && isValueFieldEmpty(row[valueName])"
```
richard-cox, line 714: I had a play around and I'm not sure we need to hide the upload button if there's no name. Hiding the `Upload` when there's content makes a bit more sense (just from the perspective of how the button is positioned and avoiding overlap).

We should raise how this behaves in the UI Demo on Thursday

torchiaf:
> We've discussed how this should work and have decided...
> 
> * Generic value `Read from File` in KeyValue component dual mode, needs to be clearer
> * `upload` button - uses filename as key (if no key already present), multine value added to value field
> * multiline values should show in whole (not a single line)

This is how it looks like,

- Upload button need to be always visible to be consistent with point 2
- Value field has a max height ATM


image

richard-cox (approved):
I take it back about not using code mirror. It does look better. Can we enable it in this PR? If not i'll do it in a separate one

image

richard-cox (approved):

PR #9075, opened 2023-06-08
Bump shell - Clean up temporary charts - fix chart url path

jordojordo:
### Summary

This fixes an issue where initial extension packages will have a duplicated folder name in the `url` path for the package chart. Also removed unnecessary directory lingering in artifact build.

Bumps shell to `0.3.13`


nwmac (approved):
LGTM

PR #9103, opened 2023-06-12
VZ-0014 Machines screen does not render in Rancher unless you do a reload of the page

knelson7:


knelson7:
created PR by mistake

PR #9116, opened 2023-06-14
Onenter newline textarea

### LINKED ISSUE #8817: Automation: Pressing Enter in textarea elements not working
Based on work done in https://github.com/rancher/dashboard/issues/8550, let's write an automated test to ensure the Enter works in text areas. @torchiaf can offer guidance.

To reproduce:
- Go to local -> Storage -> ConfigMaps
- Create a new ConfigMap
- Add a new entry in Data form
- Select value field and press Enter

Expected behaviour:
A new line is added to the value.

bisht-richa:
### Summary
Fixes #8817

Test a new line is added to the value.

### DIFF
--- a/pkg/rancher-components/src/components/Form/TextArea/TextAreaAutoGrow.vue
+++ b/pkg/rancher-components/src/components/Form/TextArea/TextAreaAutoGrow.vue
@@ -155,6 +155,7 @@ export default Vue.extend({
 <template>
   <textarea
     ref="ta"
+    data-testid="text-area-auto-grow"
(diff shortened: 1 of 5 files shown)


torchiaf (approved):
After some research, the problem with `keydown.enter` is that we cannot set the value using a keyboard `event`;
For this reason I used the `setValue()` to test if the textarea receives the new line correctly and moved the actual text case into `CruResource`, where we verify that Enter button works when textarea is focused, which was the original issue.
Please have a look to my commit and if all is fine, I would approve and merge it.

PR #9121, opened 2023-06-14
fix issue where extension versions appeared as duplicated

### LINKED ISSUE #9117: Extension versions are getting added twice
**Describe the bug**

Extension versions are getting added twice to each extension, causing some problems on the UI

**To Reproduce**

- Make sure you have the official Rancher Extensions repo installed (https://github.com/rancher/ui-plugin-charts -> `main` branch)
- Go to Extensions main screen and click on an extension card
- Check slide in panel the versions available

**Result**
- Extension versions are duplicated

**Expected Result**

- It should only exist one of each version

**Screenshots**

Screenshot 2023-06-14 at 09 03 45
Screenshot 2023-06-14 at 09 07 52

**Additional context**

It seems that the action `catalog/load` get's triggered twice on the main extension's screen https://github.com/rancher/dashboard/blob/master/shell/pages/c/_cluster/uiplugins/index.vue via the `this.refreshCharts` method and `async fetch` on screen load

aalves08:
Fixes #9117 

- fix issue where extension versions appeared as duplicated

### DIFF
--- a/shell/pages/c/_cluster/uiplugins/index.vue
+++ b/shell/pages/c/_cluster/uiplugins/index.vue
@@ -87,7 +87,7 @@ export default {
       }
     }
 
-    hash.load = await this.$store.dispatch('catalog/load', { force: true, reset: true });
+    hash.load = await this.$store.dispatch('catalog/load', { reset: true });
 
     if (this.$store.getters['management/schemaFor'](CATALOG.OPERATION)) {
       hash.helmOps = await this.$store.dispatch('management/findAll', { type: CATALOG.OPERATION });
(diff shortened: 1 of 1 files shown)


richard-cox (approved):
This is a good workaround for where we are in the release process. The downside is we still recreate the charts list twice when the extension page loads (but importantly don't re-fetch charts).

That double up of `catalog/load` comes from the `fetch` calls to `updateInstallStatus` and then `dispatch('catalog/load'`. We should be able to remove that later dispatch (and make a few things async/await). That has a much bigger impact though, and where we are this change is lot safer (and doesn't need airgap testing)

PR #9131, opened 2023-06-15
Import `clean-tooltip-directive` in `jest.setup`

rak-phillip:
### Summary
Fixes #


nwmac:
@rak-phillip Think this can be closed as Nancy added this already in another PR

PR #9140, opened 2023-06-16
fixed duplicate warnings on PrometheusRules create page

### LINKED ISSUE #7925: Trying to close a PrometheusRule warning creates duplicate warning
Seen in v2.7-head.

To reproduce,

1. Install monitoring on a cluster.
2. In Cluster Explorer, go to Monitoring > Advanced > PrometheusRules.
3. Click Create.
4. On the form that appears, there is a warning: `At least one alert rule or one recording rule is required in rule group 1.`
5. Click X on the warning.
6. See that every time you click an X, a duplicate warning appears.
Screenshot 2023-01-14 at 4 06 15 PM

Ideally, we wouldn't see any warnings until the form is touched. But at minimum, the warnings should be able to be dismissed.

momesgin:
### Summary
Fixes #7925 

Created a property for closed errors in order to filter the errors shown on the page. Since `fvUnreportedValidationErrors` and `CruResource.vue` are used in many places this approach makes sure for no side effects for the rest of the app.

### Screenshot/Video

https://github.com/rancher/dashboard/assets/135728925/9c0fea0a-f1c4-4b58-bff3-a59359523a9d


rak-phillip (commented):
This approach definitely works, but I don't think this solution addresses the root cause of the problem or that that tracking our closed error messages is the right approach.

Based on what I'm seeing, `fvUnreportedValidationErrors` is the source of the two errors displayed by default in this form. I see an issue with the way this method merges `modelErrors` with `errors` and how we deal with dismissing errors for a `CruResource`. 

When we dismiss an error for a `CruResource`, we update the `errors` collection via `@error="(e) => (errors = e)"`. `errors` is initially an empty array for `monitoring.coreos.com.prometheusrule/index.vue`.

https://github.com/rancher/dashboard/blob/b1078bd6c5aae5b06d98d420cbf84f13fb8622bf/shell/edit/monitoring.coreos.com.prometheusrule/index.vue#L127-L129

Since we update errors on close, this means that we up merging an ever-growing array of errors with the original `modelErrors` in `fvUnreportedValidationErrors`.

https://github.com/rancher/dashboard/blob/8c36ed5f5dbb4448c2b6caa48850ee0bd2f0e6b0/shell/mixins/form-validation.js#L166-L177

Furthermore, I don't think that we can ever dismiss an error properly in any forms where `modelErrors` are present and `fvUnreportedValidationErrors` is used.

@Sean-McQ are `modelErrors` meant to be dismissed? I'm trying to think about a way to resolve this issue in a clean way but I don't want to suggest anything that goes against the form validation design.

Sean-McQ (dismissed):
LGTM!

Sean-McQ (approved):
My name is Sean-McQ and I approve of these changes 👍

PR #9154, opened 2023-06-20
fix: missing scroll on login/setup page

scures:
### Summary
Fixes rancher/dashboard#9153

Solution:
Adds vertical scroll on the left panel  for both Login and Setup pages.

Setup:
image

Login:
image

### Note:
We should look into a new layout for those pages which is more responsive.

### DIFF
--- a/shell/pages/auth/login.vue
+++ b/shell/pages/auth/login.vue
@@ -539,6 +539,16 @@ export default {
     }
   }
 
+  .gutless {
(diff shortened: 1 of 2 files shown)


richard-cox (approved):
It would be nice to not have the gap between scroll bar and logo on right in the initial view of the login page in the setup flow. However, this isn't shown on the normal login page after setup and these changes are a clear improvement

PR #9174, opened 2023-06-21
use projecthelmchart dashboardValues for grafana url

### LINKED ISSUE #9099: bug- Project monitoring grafana link and embedded metrics are throwing 404 errors.
**Setup**
- Rancher version: 2.7.5-rc4
- monitoring 102.0.1+up40.1.2
- prom federator 2.1.0+up0.3.0 

**Describe the bug**
in project monitoring, when the user views a project monitoring page, neither the embedded grafana metrics nor the link to grafana behave as expected, and instead they respond with 404s

**To Reproduce**
scenario 1

1. install monitoring
2. install prom federator
3. navigation - side menu > monitoring > project monitoring
4. create a project monitor
5. once green and active, click on the project monitor name (user is taken to proj monitor page)
6. click the grafana link between Prometheus and alertmanager

scenario 2

1. install monitoring
2. install prom federator
3. navigation - side menu > monitoring > project monitoring
4. create a project monitor
5. once green and active, click on the project monitor name (user is taken to proj monitor page)
6. the embedded grafana is not displayed correctly- throws 404

**Result**
grafana metrics not displayed

**Expected Result**
grafana metrics displayed

**Screenshots**

**Additional context**
video 

https://github.com/rancher/dashboard/assets/89030218/662d58e0-bae6-4975-84f7-7dd341d4ceeb

mantis-toboggan-md:
### Summary
Fixes #9099 

On most page that the component for embedded grafana dashboards is used, a hardcoded url is passed in, which is then modified to apply a cluster id for downstream clusters, and sometimes the local cluster depending on the version of monitoring installed. The project monitoring detail page, however, does not use hardcoded urls: the urls are pulled from `<projecthelmchart>.status.dashboardValues`. Previously we accounted for the url prefixing the `GrafanaDashboard` component performs by stripping those prefixes from the url that the project monitoring page passes in. This was ..._okay_, until the logic for project monitoring diverged from other monitoring grafana dashboards. 

The solution implemented here is to just trust that the url the backend has provided in the projecthelmchart is going to work whether the cluster is local or downstream (from my testing, it does). This PR removes the url-modifying logic from the project monitor detail page, and updates the embedded grafana dashboard component to optionally bypass all prefix modification so we can avoid messing with the urls provided by the projecthelmchart resource.

The modification of `prometheusURL` and `alertmanagerURL` didn't appear to do anything so that's removed too. While investigating we noticed that if the grafana url provided by `status.dashboardValues` was missing the trailing slash, the request would be re-routed to an invalid url. That can be avoided by ensuring there's always a trailing slash in the grafana link on the project monitor detail page (BE is aware of this issue and will implement a fix, but we should catch it here so links made by existing versions of prom federator work).

Monitoring 102.0.1+up40.1.2 (local cluster has a different grafana url prefix) and 102.0.0+up40.1.2 (local cluster does not have a different grafana url prefix) should each be installed on the local cluster and the following areas checked:
- Project monitor detail view embedded graphs and grafana link (involves ins talling prometheus federator too)
- Main monitoring embedded graphs, eg those on the cluster dashboard

### DIFF
--- a/shell/detail/helm.cattle.io.projecthelmchart.vue
+++ b/shell/detail/helm.cattle.io.projecthelmchart.vue
@@ -5,7 +5,6 @@ import Tab from '@shell/components/Tabbed/Tab';
 import DashboardMetrics from '@shell/components/DashboardMetrics';
 import AlertTable from '@shell/components/AlertTable';
 import { Banner } from '@components/Banner';
-import { parse as parseUrl } from '@shell/utils/url';
 import { ENDPOINTS } from '@shell/config/types';
 import { allHash } from '@shell/utils/promise';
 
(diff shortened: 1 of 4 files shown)


shell/detail/helm.cattle.io.projecthelmchart.vue
```diff
@@ -211,7 +198,7 @@ export default {
           </div>
           <a
             :class="{disabled: !grafanaServiceEndpointEnabled}"
-            :href="value.status.dashboardValues.grafanaURL"
+            :href="grafanaURL"
```
richard-cox, line 201: In both flavours mentioned in review description... this lead to what looks like a default home page.

image

The dashboard links do end with a `proxy/` but is redirected by something grafana side to a url ending with `proxy/?orgId=1`. Is that waiting a BE fix?

richard-cox (approved):
I haven't covered all angles, but did look at downstream clusters and regressions. I also tested with this branch but built (ui-dashboard-index `richard-dev`)

- Env 1 - Newish world
  - Downstream RKE2 1.26
  - prometheus-federator:2.1.0+up0.3.0
  - rancher-monitoring:102.0.0+up40.1.2
- Env 2 Latest world
  - As per env 1, with upgrade to rancher-monitoring:102.0.1+up40.1.2
- Env 3 - Oldish world
  - Downstream v1.24.14 +rke2r1
  - prometheus-federator:2.0.0+up0.2.1
  - rancher-monitoring:101.0.0+up19.0.3 (pre grafana bump)

Apart from inline comment, everything looks good. Cluster, Node and Project monitoring graphs displayed and links worked.

I'm going to merge this. If the inline comment requires additional work, we should create a separate issue and i'd vote to resolve in the next version

PR #9244, opened 2023-06-30
Fix role template title

### LINKED ISSUE #8576: Role template management not labeled as such
internal reference: SURE-6091
Reported in 2.6.11 & 2.7.1

Issue description:
Role Templates and Roles are two very different things and yet the UI calls Templates "Roles" except when actually creating one.  These are Role Templates, NOT Roles.  Roles are configured at the cluster level.  If you edit any of the objects, their `kind` will correctly indicate what type of object they are and this should be correctly reflected in the UI to avoid confusion.

Impact:
Users are already struggling to work with Roles and RoleBindings and this just increases the barrier to entry for inexperienced users of Rancher (whether they are experienced with k8s or not).  This means that Rancher actually makes using K8s MORE difficult in this regard.

Troubleshooting steps:
Global -> Users & Authentication -> Roles

Actual behavior:
There is no indication in the UI that these are Rancher cattle CRVs or what they do

Expected behavior:
Since it says Roles on the tin, these are expected to be objects from the rbac.authorization.k8s.io apiVersion of Kinds: Role or ClusterRole

Actual expected behaviour would be to have the UI say these are Templates and the docs to outline how these work

bisht-richa:
### Summary
Fixes #8576

Fixed title in Aurth > role template

Screenshot 2023-06-30 at 9 29 55

### DIFF
--- a/cypress/e2e/tests/pages/users-and-auth/users-and-auth.spec.ts
+++ b/cypress/e2e/tests/pages/users-and-auth/users-and-auth.spec.ts
@@ -65,7 +65,7 @@ describe('Users and Authentication', () => {
 
       // navigate to the roles page and make sure user can see it
       userRoles.goTo();
-      userRoles.listTitle().should('contain', 'Roles');
+      userRoles.listTitle().should('contain', 'Role Templates');
       userRoles.listElements().should('have.length.of.at.least', 1);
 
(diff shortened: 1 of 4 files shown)


aalves08 (approved):
lgtm

PR #9257, opened 2023-07-03
Add wrap styling to image column

### LINKED ISSUE #9235: Image table column should wrap
Example on Pod table:

image

The image column does not wrap/break correctly, so long image names expand the column size.

The `image` column should use `word-break: break-all`.

momesgin:
### Summary
Fixes rancher/dashboard#9235 

### Screenshot/Video
https://github.com/rancher/dashboard/assets/135728925/0f753843-8bda-4adb-8866-18478cb04ecd

### DIFF
--- a/shell/components/SortableTable/index.vue
+++ b/shell/components/SortableTable/index.vue
@@ -1665,6 +1665,10 @@ export default {
         border-bottom: 1px solid var(--sortable-table-top-divider);
         background-color: var(--sortable-table-row-bg);
 
+        td.col-pod-images  {
+          word-break: break-all;
+        }
+
         &.main-row.has-sub-row {
           border-bottom: 0;
         }
--- a/shell/components/SortableTable/index.vue
+++ b/shell/components/SortableTable/index.vue
@@ -1665,10 +1665,6 @@ export default {
         border-bottom: 1px solid var(--sortable-table-top-divider);
         background-color: var(--sortable-table-row-bg);
 
-        td.col-pod-images  {
-          word-break: break-all;
-        }
-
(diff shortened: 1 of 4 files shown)


rak-phillip (commented):
Does this style belong in the css for sortable table? I would expect styles for `col-pod-images` to be closer to the consumer of this component.

momesgin:
> Does this style belong in the css for sortable table? I would expect styles for `col-pod-images` to be closer to the consumer of this component.

@rak-phillip That's a good point, my understanding from the ticket's description was that we should apply `word-break: break-all` to all the occurrences of image column but I might be wrong

rak-phillip:
@momesgin My main concern is that we're introducing implementation details about consumers of Sortable Table into the component itself. In my opinion, it would be better if this component focused solely on generic concepts related to tables, such as paging, grouping, table headers, table actions, etc...

I believe we can find a better place for this specific style to live. One possible solution could be to utilize styles within our formatters. I think we might have an example of this approach in `shell/components/formatter/KeyValue.vue` and we have `shell/components/formatter/PodImages.vue` available to us. Could you please investigate to see if we can make use of a formatter to resolve this issue?

momesgin:
@rak-phillip Thanks for the explanation, yeah you're right. I'll check the formatter you mentioned above

rak-phillip (approved):
The formatter implementation looks much better. Thanks for taking the time to follow-up on this approach 🙂

momesgin:
> The formatter implementation looks much better. Thanks for taking the time to follow-up on this approach 🙂

Thanks for the suggestion and your help @rak-phillip

PR #9260, opened 2023-07-04
working on partners repo support for extensions

### LINKED ISSUE #9214: Extensions: Add partner repository
We are introducing an additional repository:

- https://github.com/rancher/partner-extensions

We currently have the Rancher extensions repo - we now need a Partner repository.

We need to cover 3 cases:

(1) Fresh install of Extensions (i.e. enabling the Extensions operator) - we need to add this repository as part of dong this and allow the user to chose when installing (and uninstalling) - e.g. for installing:

image

(We should add a checkbox and also do the same for uninstall)

(2) Upgrade

If a user upgrades and already has the extensions operator installed, we need to allow them to add the new repository.

I'd suggest we add a new setting - in case 1 above, we set this setting, so we know the user has done a fresh install.

On the extensions page, we check for this setting - if not set, we show a banner saying that new repository is available - closing the banner will set the setting, so we don't show it again. We also provide a button/link in the banner that when clicked will add the new repository and set the new setting - this should open a dialog to do this - same as (3) below.

(3) Manual

In step (2), the user might dismiss the banner but then decide they want to add the new repository - we should provide a new option in the menu from the three dots:

image

Suggest adding a new option 'Add Rancher Repositories' below 'Manage Repositories'.

This would bring up a dialog same as for enabling extensions, but only with 2 checkboxes for each of the Rancher repositories.

aalves08:
Fixes #9214


shell/pages/c/_cluster/uiplugins/AddExtensionRepos.vue
```diff
@@ -0,0 +1,175 @@
...
+        }
+      }
+
+      const res = await Promise.allSettled(promises);
+
+      res.forEach((result) => {
+        if (result.status === 'rejected') {
+          console.error(result.reason); // eslint-disable-line no-console
+
+          this.errors.push(result.reason);
```
codyrancher, line 97: Don't you need tocall $emit('done') and btnCb(false) here? Otherwise I think the asyncbutton will look like it succeeded.

shell/pages/c/_cluster/uiplugins/SetupUIPlugins.vue
```diff
@@ -43,7 +49,7 @@ export default {
       this.installCharts = found;
 
       if (this.$store.getters['management/schemaFor'](CATALOG.CLUSTER_REPO)) {
-        await this.$store.dispatch('management/findAll', { type: CATALOG.CLUSTER_REPO });
+        await this.$store.dispatch('management/findAll', { type: CATALOG.CLUSTER_REPO, opt: { force: true } });
```
codyrancher, line 52: I noticed we're doing force: true here and elsewhere. It might be reasonable to comment why we need to do a full request in these areas.

jordojordo (changes requested):
Two minor things - otherwise looks good to me.

- The banner should also be gated on the fact that the repositories exist - so if you add both repositories, there is no reason to show the banner
- There should be a button within the banner that also opens the dialog, rather than telling the user to open the action menu.

richard-cox:
@aalves08 can you make that banner centre aligned? given the rest of the screen it looks jarring at the top

image

vs

image

richard-cox:
ahhh sorry, i must have missed the build

PR #9266, opened 2023-07-05
HARVESTER: only change priority by select component

DaiYuzeng:
### Summary

Accroding to the comment https://github.com/harvester/harvester/issues/2317#issuecomment-1617940290, we would like to keep the weight input component when we edit it.

- https://github.com/harvester/harvester/issues/2317#issuecomment-1617940290


mantis-toboggan-md (dismissed):
Updated to add the same fix for node affinity, and include unit tests for this bug.

mantis-toboggan-md (approved):

PR #9272, opened 2023-07-05
ui automation: update api env variable

yonasberhe23:
### Summary

updated api env variable:
- handles cases where TEST_BASE_URL may or may not contain `/dashboard` suffix.
- this is done by checking if '/dashboard' exists in baseUrl, if so remove it, else do nothing. that way if `localhost:8005` is used as TEST_BASE_URL we will still be able to send api requests.

### DIFF
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -87,7 +87,7 @@ export default defineConfig({
         'pkg/rancher-components/src/components/**/*.{vue,ts,js}',
       ]
     },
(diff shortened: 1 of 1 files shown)


richard-cox (approved):
Tweaked slightly (see commit comment). If you're happy with those feel free to merge

There was a suggestion elsewhere to update API to TEST_API. That would require some plumbing changes though, so left as is for the moment.

PR #9278, opened 2023-07-06
Added validation to 'Rancher backups' chart installer

### LINKED ISSUE #6181: "Rancher Backups" helm chart -- during installation, on the 'values' page none of the required fields are marked as such.
**Rancher Server Setup**

- Rancher version v2.6.- head `f4d09c7`

- Kubernetes version: v1.21.12
- Cluster Type (Local/Downstream): Downstream node driver 1 worker, 1 etcd, 1 cp RKE1

**Bug Description**
When  installing the helm chart "Rancher Backups" and selecting S3 storage, the bucket name and endpoint are required fields, however the UI does not display this information and users are allowed to attempt to install the chart without providing the required fields. 

**Reproduction Steps**
1. stand up Rancher instance and a downstream cluster
2. In Rancher UI Navigate to Local cluster > apps and marketplace > charts > Rancher Backups 
3. click to install latest Rancher Backup
4. select to install to 'system' project. 
5. click next
6. for storage options, select the S3 compatible object store. 

**Expected Result** 
Required fields are clearly designated and the user is not allowed to click 'install' before providing them.

**Actual Result**
Neither endpoint or bucket name appear to be required fields, users can click install without providing either value

**Screenshots**

error thrown when attempting to install without a bucket name 

screenshot showing lack of UI designation that the fields are required

eva-vashkevich:
### Summary
Fixes #6181
Marked bucket name and endpoint fields as required and disabled "install" and "update" buttons if these fields are empty.

- Marked fields as "required"
- Disabled default behaviour for "values" step for backup chart

### Technical notes summary
No default behaviour has changed. Added a "valid" event listener to chart installer that should be used by other charts if fields require validation.

Tested in Chrome
1. Go to "Apps"-> "Charts"
2. Select "Rancher backups"
3. There are 2 scenarios here. New chart creation and updating already existing chart. This fix applies to both.
4. Click "Install" (for new) or "Update" (for existing)
5. Click "Next"
6. Select "Use an S3-compatible object store"
7. Validate that "Bucket Name" and "Endpoint" fields are marked as required
8. Validate that "Install" or "Update" button is only enabled when both fields have non-empty values. Try removing each and both and validate the behaviour.

- Chart installer

### Screenshot/Video
Screenshot 2023-07-06 at 1 09 08 PM

### DIFF
--- a/shell/chart/__tests__/S3.test.ts
+++ b/shell/chart/__tests__/S3.test.ts
@@ -1,0 +1,50 @@
+/* eslint-disable jest/no-hooks */
+import { mount, createLocalVue } from '@vue/test-utils';
+
+import S3 from '@shell/chart/rancher-backup/S3';
+import Vuex from 'vuex';
+
+const localVue = createLocalVue();
+
+localVue.use(Vuex);
+
+describe('rancher-backup: S3', () => {
+  const mockStore = { getters: { 'i18n/t': (text: string) => text, t: (text: string) => text } };
+  const wrapper = mount(S3, { mocks: { $store: mockStore } });
+
+  it('should emit invalid when form is not filled', () => {
+    expect(wrapper.emitted('valid')).toHaveLength(1);
+    expect(wrapper.emitted('valid')![0][0]).toBeFalsy();
+  });
+
+  it('should emit valid when required fields are filled', async() => {
+    const bucketName = wrapper.find('[data-testid="S3-bucketName"]').find('input');
+    const endpoint = wrapper.find('[data-testid="S3-endpoint"]').find('input');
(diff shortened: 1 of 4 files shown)


mantis-toboggan-md (changes requested):
When I leave the required fields blank and toggle from 'Use an S3 compatible object store' to 'No default storage location' the 'Install' button stays grayed out. I think this could be fixed by having the S3 component emit an event signaling its valid when its no longer needed eg 
```
  beforeDestroy() {
    this.$emit('valid', true);
  },
  ```

mantis-toboggan-md (approved):
lgtm

PR #9284, opened 2023-07-07
5991 workloads crud operations e2e

Shavindra:
### Summary
Fixes rancher/dashboard#5991 

No functional changes.

### Technical notes summary
Workload Deployments CRUD operations only

E2E only

E2E only


cypress/globals.d.ts
```diff
@@ -1,5 +1,7 @@
 type Matcher = '$' | '^' | '~' | '*' | '';
 
+type Verbs = 'GET' | 'POST' | 'PUT' | 'DELETE' | 'PATCH' | 'WATCH' | 'LIST' | 'CREATE' | 'UP'
```
cnotv, line 3: This may probably be convenient to have globally in the `@shell/types` maybe as `api.d.ts`.

cypress/support/commands.ts
```diff
@@ -183,6 +183,28 @@ Cypress.Commands.add('keyboardControls', (triggerKeys: any = {}, count = 1) => {
...
 });
 
+/**
+ * Intercept all requests and return
+ * @param {array} intercepts - Array of intercepts to return
+ * return {array} - Array of intercepted request strings
+ * return {string} - Intercepted request string
+ */
+Cypress.Commands.add('interceptAny', (method = '/GET/POST/PUT/PATCH/', url = '/v1/*', wait = false, timeout = 10000) => {
+  if (Array.isArray(url)) {
```
cnotv, line 193: Let's avoid logic in the test, especially commands. It's important to be clear in an assertion that if you want a single command or more commands, there's 2 different cases.
You man still use the single one in the array case.
Also the name is kinda confusing, because you are clearly intercepting the request, so maybe something like `interceptRequest` and `interceptAllRequests` (I added the `all` because just a `s` it's easy to miss).

cypress/support/commands.ts
```diff
@@ -183,6 +183,28 @@ Cypress.Commands.add('keyboardControls', (triggerKeys: any = {}, count = 1) => {
   }
 });
 
+/**
+ * Intercept all requests and return
+ * @param {array} intercepts - Array of intercepts to return
+ * return {array} - Array of intercepted request strings
+ * return {string} - Intercepted request string
+ */
+Cypress.Commands.add('interceptAny', (method = '/GET/POST/PUT/PATCH/', url = '/v1/*', wait = false, timeout = 10000) => {
```
cnotv, line 192: 1. We need to add the types, wonder why you don't get error about the Array
2. Timeout is not used here, what's it for?

cypress/support/commands.ts
```diff
@@ -183,6 +183,28 @@ Cypress.Commands.add('keyboardControls', (triggerKeys: any = {}, count = 1) => {
...
+  if (Array.isArray(url)) {
+    const interceptedUrls: string[] = [];
```
cnotv, line 194: This can be directly a map.
Assuming we are going to refactor this for an array case, we use `urls` plural.

```ts
const interceptedUrls: string[] = urls.map((url, i) => {
  cy.intercept(method, url).as(`interceptReq${ i }`);

  return `@interceptReq${ i }`
});
```

cypress/support/commands.ts
```diff
@@ -183,6 +183,28 @@ Cypress.Commands.add('keyboardControls', (triggerKeys: any = {}, count = 1) => {
   }
 });
 
+/**
+ * Intercept all requests and return
```
cnotv, line 187: In case of multiple intercepted requests is the wrapper going to be overlapping if using the same term?
An alternative may be to parse and chain the method and URL in a camel case with `@intercept` as base?

shell/components/ResourceDetail/Masthead.vue
```diff
@@ -416,7 +416,7 @@ export default {
               {{ parent.displayName }}:
             </nuxt-link>
             <span v-else>{{ parent.displayName }}:</span>
-            <span v-if="value.detailPageHeaderActionOverride && value.detailPageHeaderActionOverride(realMode)">{{ value.detailPageHeaderActionOverride(realMode) }}</span>
+            <span v-if="value?.detailPageHeaderActionOverride && value?.detailPageHeaderActionOverride(realMode)">{{ value.detailPageHeaderActionOverride(realMode) }}</span>
```
cnotv, line 419: Conditional operator does not work in the Vue template (differently from React/Angular) afaik.

cypress/e2e/blueprints/explorer/workloads/deployments/deployment-create.ts
```diff
@@ -0,0 +1,240 @@
...
+    },
+    selector: { matchLabels: { 'workload.user.cattle.io/workloadselector': 'apps.deployment-default-test-deployment' } }
+  }
+};
+
+export const deploymentCreateResponse = {
+  id:    'default/test-deployment',
+  type:  'apps.deployment',
+  links: {
+    remove: 'https://localhost:8005/v1/apps.deployments/default/test-deployment',
```
cnotv, line 85: Is the URL affecting in somehow the functionality? Because in that case we may want to replace with the `baseUrl`.

cypress/e2e/po/components/kubectl.po.ts
```diff
@@ -36,4 +38,18 @@ export default class Kubectl extends ComponentPo {
 
     return this;
   }
+
+  openAndExecuteCommand(command: string, wait = 3000) {
```
cnotv, line 42: Did we not say to do not use time last time or?

cypress/e2e/po/pages/explorer/workloads/workloads.po.ts
```diff
@@ -0,0 +1,182 @@
...
+import WorkloadPo from '@/cypress/e2e/po/pages/explorer/workloads.po';
+import PromptRemove from '~/cypress/e2e/po/prompts/promptRemove.po';
+
+export class workloadDetailsPageBasePo extends PagePo {
+  static url: string;
+
+  private static createPath(
+    workloadId: string,
+    clusterId: string,
+    workloadType: string,
```
cnotv, line 15: Sounds like a type 😛

cypress/e2e/po/pages/explorer/workloads/workloads.po.ts
```diff
@@ -0,0 +1,182 @@
...
+    namespaceId: string,
+    queryParams?: { [key: string]: string }
```
cnotv, line 17: If you want it can be `queryParams?: Record<string, string>`.

cypress/e2e/po/pages/explorer/workloads/workloads.po.ts
```diff
@@ -0,0 +1,182 @@
...
+    const params = new URLSearchParams(queryParams);
+
+    return `${ urlStr }?${ params.toString() }`;
+  }
+
+  static goTo(): Cypress.Chainable<Cypress.AUTWindow> {
+    return super.goTo(this.url);
+  }
+
+  constructor(workloadId: string, clusterId: string, workloadType: string, queryParams?: { [key: string]: string }, namespaceId = 'default') {
```
cnotv, line 34: Optional parameters should be always after required ones. The linter should throw an error here I think.

cypress/e2e/po/pages/explorer/workloads/workloads.po.ts
```diff
@@ -0,0 +1,182 @@
...
+    super(workloadDetailsPageBasePo.createPath(workloadId, clusterId, workloadType, namespaceId, queryParams));
+
+    workloadDetailsPageBasePo.url = workloadDetailsPageBasePo.createPath(
```
cnotv, line 37: Is there not a way to pass it as a whole object, so you don't have to define it like 4 times the keys and 2 times the type?
You may create a type out of that as well.

cypress/e2e/po/pages/explorer/workloads/workloads.po.ts
```diff
@@ -0,0 +1,182 @@
...
+  deleteWithKubectl(name: string, namespace = 'default') {
+    this.workload().deleteWithKubectl(name, namespace);
+  }
+
+  createWithKubectl(blueprintJson: string | Object, wait = 6000) {
+    this.workload().createWithKubectl(blueprintJson, wait);
+  }
+}
+
+export class WorkloadsListPageBasePo extends PagePo {
```
cnotv, line 59: One Po per page is way more readable and easy to maintain.

cypress/e2e/po/pages/explorer/workloads/workloads.po.ts
```diff
@@ -0,0 +1,182 @@
...
+    const baseResourceList = new BaseResourceList(this.self());
+
+    return baseResourceList.masthead().actions().eq(0).click();
+  }
+
+  resourcesList() {
+    return new BaseResourceList(this.self());
+  }
+
+  sortableTable() {
```
cnotv, line 86: Is this returning the element?

cypress/e2e/po/pages/explorer/workloads/workloads.po.ts
```diff
@@ -0,0 +1,182 @@
...
+    return new WorkloadPo();
+  }
+
+  deleteWithKubectl(name: string, namespace = 'default') {
+    this.workload().deleteWithKubectl(name, namespace);
+  }
+}
+
+export class WorkloadsCreatePageBasePo extends PagePo {
+  static createPath(clusterId: string, workloadType: string, queryParams?: { [key: string]: string }) {
```
cnotv, line 120: `createPath()` is defined multiple times, so you may avoid code repetition by creating an util or a base class to extend since we are in the OOP world.

cypress/e2e/tests/pages/explorer/workloads/deployments.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+        workloadDetailsPage.goTo();
+        workloadDetailsPage.mastheadTitle().should('contain', workloadName);
+      });
+    });
+
+    describe('List: Deployments', () => {
+      // To reduce test runtime, will use the same workload for all the tests
+      it('Should list the workloads', () => {
+        deploymentsListPage.goTo();
+        e2eWorkloads.forEach(({ name }) => {
```
cnotv, line 70: Can't we store the workloads within the PO at this point, taking advantage of the OOP architecture?

cypress/e2e/tests/pages/explorer/workloads/deployments.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+      });
+    });
+
+    describe('Delete: Deployments', () => {
+      const deploymentsToDeleteWithUI = [deploymentCreateRequest.metadata.name];
+
+      // To reduce test runtime, will use the same workload for all the tests
+      it('Should be able to delete a workload', () => {
+        deploymentsListPage.goTo();
+        deploymentsToDeleteWithUI.forEach(( name ) => {
```
cnotv, line 82: Why do we loop a single request?

cnotv (changes requested):

cypress/e2e/po/pages/explorer/workloads/workloads.po.ts
cnotv: It's really difficult without the types, but k.

cnotv (approved):
LGTM

PR #9289, opened 2023-07-07
Add conditions for updating root index of extension

### LINKED ISSUE #9277: Assets `urls` property for extensions being overwritten in index.yaml during build
**Setup**
- Rancher version: v2.7.5
- Rancher UI Extensions: N/A

**Describe the bug**

During a github build of an extension, the `urls` property within the `index.yaml` is being overwritten ([example](https://github.com/kubewarden/ui/blob/85a87b10997164b42f6d2898b77b3a9b12265686/index.yaml#L21-L22)) in the script [here](https://github.com/rancher/dashboard/blob/90377f035ebdf0da35563cc7692ab2d89d6e2b92/shell/scripts/extension/publish#L302-L309) - This property defines the location of the extension assets which the chart uses during the installation process.

**Expected Result**

The url for an extension should be a path with the `assets` directory prefixed to the location. For instance:
```yaml
    urls:
    - assets/kubewarden/kubewarden-1.1.0-rc1.tgz
```

**Additional context**

The existence of these commands is necessary for container builds, so adding additional conditions which prevent the overwriting of this file based whether  `GITHUB_BUILD` is false and if the `ROOT_INDEX` does not exist should suffice.

jordojordo:
### Summary
Fixes #9277 

The `urls` property was being correctly populated with the `BASE_URL` (`assets/`) directory initially but was being overwritten when the script reached lines 302 or 307.  This would only occur if the Root index existed, so in the case of the issue, the `index.yaml` was found and placed at the root level - As it existed the index was being updated with the assets within the `./assets` directory while not specifying the `--url` property, thus overwriting what had been previously applied on line 293.

These conditions exist for container image build, as this could be an initial build (i.e. no root index.yaml exists yet) or it could be an updated build where the index.yaml would most likely exist.

Added two conditions:
1. If the root index.yaml exists && the build is not classified as a Github build, then create an index.yaml within the assets directory by merging the root index.
2. If the assets/index.yaml file exists && the root index.yaml does not exist, then copy the assets/index.yaml into the root directory for the image build.  

- Extension builds that both contain and do not contain an index.yaml (migrating extensions from a `main` branch to a `gh-pages` branch could be a good example)
- Extension builds that are not built within a Git repository
- Extension container builds

### DIFF
--- a/shell/scripts/extension/publish
+++ b/shell/scripts/extension/publish
@@ -299,12 +299,12 @@ for d in pkg/*/ ; do
   fi
 done
 
-if [ -f ${ROOT_INDEX} ]; then
+if [ -f ${ROOT_INDEX} ] && [ "${GITHUB_BUILD}" == "false" ]; then
   UPDATE="--merge ${ROOT_INDEX}"
   helm repo index ${ASSETS} ${UPDATE}
 fi
 
-if [ -f ${ASSETS}/index.yaml ]; then
+if [ -f ${ASSETS}/index.yaml ] && ! [ -e "${ROOT_INDEX}" ]; then
   cp ${ASSETS}/index.yaml ${BASE_DIR}
 fi
 


aalves08 (commented):
Looks good to me, but I'm not the best person to review this... Will wait for @codyrancher review. Let me know if you're in a hurry to get this in mate, then I can approve just to unblock your work. 🙏

aalves08 (approved):
lgtm

PR #9291, opened 2023-07-07
Remove overlay loading from pod details

### LINKED ISSUE #9236: Pod detail view shows full-view loading indicator briefly
Go to the detail view of a pod:

image

Observe that the page changes and then you briefly get a loading indicator on a grey panel.

This is because the page is using `fetch` with.a loading indicator - we should show the content as soon as possible and not use a loading indicator - other bits of the UI that need data should update when the data has been loaded.

momesgin:
### Summary
The overlay loading was only used to show/hide one of the tabs(pod-metrics) in the pod details page. After discussing this with the UX/UI team the final decision was to remove the loading.

Fixes #9236


rak-phillip (approved):
LGTM

PR #9297, opened 2023-07-11
demo

codyrancher:
Here you can see that we're able to redirect from /floof to /floofy the example I added to the server-middleware.

https://github.com/rancher/dashboard/assets/55104481/d5477cf0-39ce-4ad6-9fcc-68c19d914a31

### DIFF
--- a/shell/server/server-middleware.js
+++ b/shell/server/server-middleware.js
@@ -1,11 +1,9 @@
-import { URL } from 'url';
+module.exports = function(req, res, next) {
+  if (req.url === '/floof') {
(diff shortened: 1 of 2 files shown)


shell/server/server-middleware.js
```diff
@@ -1,11 +1,9 @@
-import { URL } from 'url';
-
-export default function(req, res, next) {
-  const parsed = new URL(req.url, 'https://localhost');
+module.exports = function(req, res, next) {
```
codyrancher, line 1: We have to switch to this module system to make `require()` work in vue.config.js

PR #9309, opened 2023-07-12
remove error icon when the condition is Ready and Waiting

### LINKED ISSUE #9227: Don't show error icon when a cluster is updating
When you provision a cluster and it is in the updating state, we show the error iconL:

image

This is because  one of the conditions is not ready - you'll see this if you look at the cluster and its conditions:

image

We should update this so that if the cluster is transitioning/updating, we don't show the error icon, since the user knows the cluster is in the process of updating - the purpose of showing the error icon is to alert users to an error condition for a cluster that they expect to be in a stable state.

momesgin:
### Summary
Filtering the "Ready and Waiting" status condition from the list of errors where we don't want to show an error icon based on that.

Fixes #9227 

### Screenshot/Video
https://github.com/rancher/dashboard/assets/135728925/790ae89e-ad17-4b6e-8a19-af18eb203a42


rak-phillip:
Rerunning failed e2e test. The error looks to be unrelated to this change

```
Running:  account-api-keys.spec.ts                                                      (14 of 24)


  User can make changes to their account
    (Attempt 1 of 3) Can navigate to Account and API Keys page and change their password
    (Attempt 2 of 3) Can navigate to Account and API Keys page and change their password
    1) Can navigate to Account and API Keys page and change their password
    ✓ Can create and delete API keys (14733ms)


  1 passing (4m)
  1 failing

  1) User can make changes to their account
       Can navigate to Account and API Keys page and change their password:
     AssertionError: Timed out retrying after 60000ms: Expected to find element: `.tooltip.popover.vue-popover-theme`, but never found it.
      at UserMenuPo.isOpen (https://127.0.0.1/__cypress/tests?p=cypress/e2e/tests/pages/user_menu/account-api-keys.spec.ts:1326:34)
      at Context.eval (https://127.0.0.1/__cypress/tests?p=cypress/e2e/tests/pages/user_menu/account-api-keys.spec.ts:1349:30)

```

momesgin:
@rak-phillip yeah, I am not able to re-run the failed jobs due the temporary permission issue that I have.

shell/components/formatter/ClusterLink.vue
```diff
@@ -43,10 +44,10 @@ export default {
             formattedTooltip.push(`<p>${ [c.type] } (${ c.status })</p>`);
           });
 
-        return formattedTooltip.toString().replaceAll(',', '');
+        return formattedTooltip.toString().replace(/,/g, '');
```
rak-phillip, line 47: What is the reason for this change, is there an issue with `replaceAll`?

rak-phillip (commented):
I think this change looks great. I only have one question and then there's the question of the making sure that the failing e2e test passes.

shell/components/formatter/ClusterLink.vue
momesgin: that's because `replaceAll()` is Node 16^ only but the GH actions are running the unit tests in Node 14

momesgin: looks like we have a polyfill for replaceAll but it's imported in the app's initialization. I can import it in the test itself to fix the issue and revert the change above

rak-phillip (approved):
LGTM - I created issue #9334 so we hopefully won't need to worry about polyfills in order to successfully run tests in the future.

momesgin:
> LGTM - I created issue #9334 so we hopefully won't need to worry about polyfills in order to successfully run tests in the future.

Thanks for creating the issue and taking the time to review this @rak-phillip

PR #9327, opened 2023-07-13
Remove what's new page and link to release notes instead

mantis-toboggan-md:
fixes https://github.com/rancher/dashboard/issues/9332 - migrating 'getting started' will require coordination with docs team; this PR only updates the release note links. 

Testing should include -rc and -head versions as well as a regular release


cypress/e2e/tests/pages/home.spec.ts
```diff
@@ -17,13 +15,12 @@ describe('Home Page', () => {
     HomePagePo.goToAndWaitForGet();
   });
 
-  it('Can navigate to What\'s new page', { tags: ['@adminUser', '@standardUser'] }, () => {
+  it.only('Can navigate to What\'s new page', { tags: ['@adminUser', '@standardUser'] }, () => {
```
cnotv, line 18: I think we should not use only 😅

cypress/e2e/tests/pages/home.spec.ts
```diff
@@ -17,13 +15,12 @@ describe('Home Page', () => {
...
     /**
-     * Click link on home page and verify user lands on What's new page
-     * Verify contents of What's new page
-     * Verify changlog banner is hidden after navigating to the page
+     * Verify changlog banner is hidden after clicking link
+     * Verify release notes link is valid github page
      */
-    const burgerMenuPo = new BurgerMenuPo();
+    // const burgerMenuPo = new BurgerMenuPo();
```
cnotv, line 23: we can probably remove this

shell/store/index.js
```diff
@@ -573,6 +573,18 @@ export const getters = {
     return state.targetRoute;
   },
 
+  releaseNotesUrl(state, getters) {
+    const version = getters['management/byId'](MANAGEMENT.SETTING, 'server-version')?.value;
+
+    const base = 'https://github.com/rancher/rancher/releases';
+
+    if (version && version.match(/v[0-9]+\.[0-9]+\.[0-9]$/)) {
```
cnotv, line 581: Is there anything we can get from the util? If not maybe we can extend it?

https://github.com/rancher/dashboard/blob/master/shell/utils/version.js

cnotv (changes requested):

cypress/e2e/tests/pages/home.spec.ts
mantis-toboggan-md: Updated to remove only 😅

shell/store/index.js
mantis-toboggan-md: Good call - I've updated the getter to use the existing function `isDevBuild`

cnotv (approved):
LGTM

PR #9329, opened 2023-07-14
HARVESTER: support harvester product icon & fix bug

WuJun2016:
### Summary
1. Supports custom product logos and product names in a single  harvester cluster
2. The harvester PCI device needs to support other types of group and hide the namespace group.
3. Fix single cluster menu alignment issue
image
4. fix set defaultStorage race case.   https://github.com/harvester/dashboard/pull/459

Fixes #


shell/components/BrandImage.vue
```diff
@@ -49,7 +53,7 @@ export default {
     },
 
     pathToBrandedImage() {
-      if (this.fileName === 'rancher-logo.svg') {
+      if (this.fileName === 'rancher-logo.svg' || this.supportCustomLogo) {
```
mantis-toboggan-md, line 56: Why is this needed instead of using the `brand` mechanism?

mantis-toboggan-md (changes requested):

richard-cox:
I'm not sure when this change is required, so putting in to 2.8q4. @WuJun2016 @mantis-toboggan-md please update if this is incorrect.

shell/components/BrandImage.vue
WuJun2016: Because we need to support custom logos at the top of the menu, such as this
image

WuJun2016:
@richard-cox   I want to know when `2.8q4` starts and ends ？ 

I found that rancher is working on several releases at the same time, of which `2.7.6` and `2.7.7` are emergency and safe releases, but when will `q3 2.7.x` end?

This change will be required in harvester v1.3.0 （After this pull request is merged, it will be necessary to release a corresponding @rancher/shell package, as harvester v1.3.0 will start using @rancher/shell to provide support for the rancher core.）.

WuJun2016:
Hi @mantis-toboggan-md   PTAL again, thanks

mantis-toboggan-md (changes requested):

torchiaf:
`productName !== 'Harvester'` is always false so it will not work for non-harvester product.
I also think we should replace `isSingleProduct.supportCustomLogo` with isSingleProduct.logo but it will require a fix on Harvester Ui side.

torchiaf:
Closing as superseded.

PR #9339, opened 2023-07-17
HARVESTER: Changes from Harvester

DaiYuzeng:
### Summary
There is some changes from harvester repo.

### DIFF
--- a/shell/components/ExplorerMembers.vue
+++ b/shell/components/ExplorerMembers.vue
@@ -11,6 +11,7 @@ import SortableTable from '@shell/components/SortableTable';
 import { mapGetters } from 'vuex';
 import { canViewProjectMembershipEditor } from '@shell/components/form/Members/ProjectMembershipEditor.vue';
 import { allHash } from '@shell/utils/promise';
+import { HARVESTER_NAME as HARVESTER } from '@shell/config/features';
 
 /**
  * Explorer members page.
@@ -227,6 +228,9 @@ export default {
     canEditClusterMembers() {
       return this.normanClusterRTBSchema?.collectionMethods.find((x) => x.toLowerCase() === 'post');
     },
+    isHarvester() {
+      return this.$store.getters['currentProduct'].inStore === HARVESTER;
+    },
   },
   methods: {
     getMgmtProjectId(group) {
@@ -324,7 +328,7 @@ export default {
         />
       </Tab>
       <Tab
-        v-if="canManageProjectMembers"
+        v-if="canManageProjectMembers && !isHarvester"
         name="project-membership"
         :label="t('members.projectMembership')"
       >
--- a/shell/components/form/SecretSelector.vue
+++ b/shell/components/form/SecretSelector.vue
@@ -90,7 +90,8 @@ export default {
       }
     },
     secrets() {
-      const allSecrets = this.$store.getters[`${ this.inStore }/all`](SECRET);
+      const inStore = this.$store.getters['currentProduct'].inStore;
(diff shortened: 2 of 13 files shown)


shell/components/ExplorerMembers.vue
```diff
@@ -324,7 +328,7 @@ export default {
         />
       </Tab>
       <Tab
-        v-if="canManageProjectMembers"
+        v-if="canManageProjectMembers && !isHarvester"
```
DaiYuzeng, line 331: Harvester want to display `Cluster Membership` only

Harvester:
image

Rancher:
image

DaiYuzeng:
@WuJun2016 PTAL! Thanks!

richard-cox (changes requested):
We need to find a better way to support disabling / hiding functionality without `if harvesters` in the dashboard code. 

@mantis-toboggan-md can you check with @aalves08 / @jordojordo if a way exists already for these places (PR probably good for 2.7q3), or how to get something introduced to support them (PR would probably be 2.8q4)?

richard-cox:
I'm not sure when this change is required, so putting in to 2.8q4. @WuJun2016 @mantis-toboggan-md please update if this is incorrect.

shell/components/form/SecretSelector.vue
```diff
@@ -90,7 +90,8 @@ export default {
       }
     },
     secrets() {
-      const allSecrets = this.$store.getters[`${ this.inStore }/all`](SECRET);
+      const inStore = this.$store.getters['currentProduct'].inStore;
+      const allSecrets = this.$store.getters[`${ inStore }/all`](SECRET);
```
mantis-toboggan-md, line 94: The `inStore` component property no longer works with this change

mantis-toboggan-md (changes requested):
There are a handful of things in this PR which I think would require additions to the extensions API to implement in a less Harvester-specific way. 
This patterns appears repeatedly in this PR:

```
const currentCluster = this.$store.getters['currentCluster'];
const inStore = currentCluster.isHarvester ? 'harvester' : 'cluster';
```

If `this.$store.getters['currentStore']` doesn't work here I think we need to revisit this in Q4 to come up with a more extensions-generic solution. 

We may be able to accomodate the Harvester-specific logging functionality by [overwriting the logging forms](https://rancher.github.io/dashboard/extensions/api/components/resources#editconfig-view) - as is we can add tabs to the logging forms but hiding existing inputs is trickier.

I think we would need to expand the extensions API to facilitate hiding the project membership tab.

aalves08:
@mantis-toboggan-md could you create an issue labeled `area/plugins` with the areas you wished to be addressed by the extensions API?

shell/edit/logging-flow/index.vue
```diff
@@ -51,15 +57,17 @@ export default {
   mixins: [CreateEditView],
 
   async fetch() {
-    const hasAccessToClusterOutputs = this.$store.getters[`cluster/schemaFor`](LOGGING.CLUSTER_OUTPUT);
-    const hasAccessToOutputs = this.$store.getters[`cluster/schemaFor`](LOGGING.OUTPUT);
+    const currentCluster = this.$store.getters['currentCluster'];
+    const inStore = currentCluster.isHarvester ? VIRTUAL : 'cluster';
```
mantis-toboggan-md, line 61: Could this be switched to use `this.$store.getters['currentProduct'].inStore`?

WuJun2016: I think here we can switch to `this.$store.getters['currentProduct'].inStore`

torchiaf:
Closing as superseded.

PR #9382, opened 2023-07-20
fix gating issue with node version

aalves08:
- fixes gating issue with node version (ex: https://github.com/rancher/dashboard/actions/runs/5609178000/jobs/10262526658?pr=9342)


richard-cox (approved):
As you called out on slack, we might have an issue where a dep has bumped it's version.

This would be apparent in the create-app test given we don't use a lock file there.

Lets get the gate passing either way

PR #9400, opened 2023-07-21
Update button component example

### LINKED ISSUE #9398: Create Button component or merge with existing AsyncButton
# Description

We currently do not have a button component for Vue and we simply use HTML markup. E.g.:

```html
          <button
            class="btn role-primary"
            @click="buttonAction"
          >
            {{ buttonText }}
          </button>
```

rak-phillip:
### Summary
Fixes #9398

### DIFF
--- a/pkg/rancher-components/src/components/Button/Button.vue
+++ b/pkg/rancher-components/src/components/Button/Button.vue
@@ -1,0 +1,50 @@
+<script lang="ts">
+import { defineComponent, computed } from 'vue';
+
+type Props = {
+  primary: boolean;
+  secondary: boolean;
+  tertiary: boolean;
+};
+
+export default defineComponent({
+  props: {
+    primary:   Boolean,
+    secondary: Boolean,
+    tertiary:  Boolean,
+  },
+  setup(props) {
+    const buttonRoles = [
+      { role: 'primary', className: 'role-primary' },
+      { role: 'secondary', className: 'role-secondary' },
+      { role: 'tertiary', className: 'role-tertiary' },
+    ];
+
+    const buttonClass = computed(() => {
+      const activeRole = buttonRoles.find(({ role }) => props[role as keyof Props]);
+
+      return {
+        btn: true,
+
+        [activeRole?.className || 'role-primary']: true,
+      };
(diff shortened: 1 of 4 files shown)


pkg/rancher-components/src/components/Button/Button.vue
```diff
@@ -0,0 +1,50 @@
...
+import { defineComponent, computed } from 'vue';
+
+type Props = {
+  primary: boolean;
+  secondary: boolean;
+  tertiary: boolean;
+};
+
+export default defineComponent({
+  props: {
```
cnotv, line 11: Would not be better a type/eum?

pkg/rancher-components/src/components/Button/Button.vue
```diff
@@ -0,0 +1,50 @@
...
+    primary:   Boolean,
+    secondary: Boolean,
+    tertiary:  Boolean,
+  },
+  setup(props) {
+    const buttonRoles = [
+      { role: 'primary', className: 'role-primary' },
```
cnotv, line 18: would it not be more convenient a dictionary?

pkg/rancher-components/src/components/Button/Button.vue
rak-phillip: It looks like Vue 2 does expose `script setup`, so we can make use of `defineProps` to produce something like this

```
type Props = {
  primary?: boolean;
  secondary?: boolean;
  tertiary?: boolean;
}

const props = defineProps<Props>();
```

pkg/rancher-components/src/components/Button/Button.vue
rak-phillip: Maybe..

I chose an array of objects because I find order to be really important here. I want to assert that we maintain a hierarchy of styles from Primary => Secondary => Tertiary. 

Perhaps `Map`[^1] would better fit the bill here? Although, I find the ease of access using array methods to be quite convenient, but this is easily mitigated. 

[^1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

Issue #9137, opened 2023-06-16
Continuous Delivery dashboard shows 0/0 Bundles Ready


nwmac:
Internal Ref: SURE-6398

Add a Git repo that has a `fleet.yaml` file with a `name` set - this causes the bundle(s) generated to use that name.

The logic we use to link GitRepos to Bundles in that case - we assume the name of the GitRepo is the prefix of the Bundle name.

We should be using the fleet.cattle.io/repo-name label of the Bundle to get the name of the associated GitRepo

![image-2023-05-24-16-19-49-681.png](https://zube.io/files/rancher/3a7dd77d0f216eb851bd350d4c31e405-image-2023-05-24-16-19-49-681.png)

Troubleshooting steps:
running `kubectl get bundles -n fleet-default fleet-name-here -o=jsonpath={.status}` to get the bundle status shows:

 ```
{"conditions":
[
{"lastUpdateTime":"2023-05-24T23:17:58Z","status":"True","type":"Ready"},
{"lastUpdateTime":"2023-05-24T23:17:59Z","status":"True","type":"Processed"}
],
"display":{"readyClusters":"1/1"},
"maxNew":50,
"maxUnavailable":1,
"maxUnavailablePartitions":0,
"observedGeneration":1,
"partitions":
[
{"count":1,"maxUnavailable":1,"name":"All","summary":{"desiredReady":1,"ready":1}}
],
"resourceKey":
[
{"apiVersion":"apps/v1","kind":"Deployment","name":"frontend","namespace":"fleet-name-here-ns"},
{"apiVersion":"apps/v1","kind":"Deployment","name":"redis-master","namespace":"fleet-name-here-ns"},
{"apiVersion":"apps/v1","kind":"Deployment","name":"redis-slave","namespace":"fleet-name-here-ns"},
{"apiVersion":"rbac.authorization.k8s.io/v1","kind":"ClusterRoleBinding","name":"fleet-name-here-crb"},
{"apiVersion":"v1","kind":"Service","name":"frontend","namespace":"fleet-name-here-ns"},
{"apiVersion":"v1","kind":"Service","name":"redis-master","namespace":"fleet-name-here-ns"},
{"apiVersion":"v1","kind":"Service","name":"redis-slave","namespace":"fleet-name-here-ns"}
],
"summary":{"desiredReady":1,"ready":1},
"unavailable":0,
"unavailablePartitions":0}>
```

Format that JSON, and it shows:

```
   "summary":{
[...]

PR #9408, opened 2023-07-24
Filter the bundles by label repo-name.

Shavindra:
### Summary
Fixes rancher/dashboard#9137 

### Technical notes summary

- If the yaml file is created on the root then it would have a different name compared to yaml in a nested folder. 
- If use the following gitrepo https://github.com/rancher/fleet-examples.git with the `path` is set in dashboard UI to be `simple` (/simple/fleet.yaml) it would work.
- If yaml is under the root of the repo as below, then it would not work
image
 

`
- Create a git repository with the following yaml in the root of the git repo.
```
apiVersion: v1
kind: Service
metadata:
  name: redis-slave
  labels:
    app: redis
    role: slave
    tier: backend
spec:
  ports:
  - port: 6379
  selector:
    app: redis
    role: slave
    tier: backend
```
- Add a new GitRepo from Dashboard UI
- Add second git repo from https://github.com/rancher/fleet-examples.git
- Set the path to `simple`, so that it would point the `/simple` folder in Git Repo
- Navigate back to Continuous Delivery dashboard.

- Continuous Delivery dashboard

### Screenshot/Video
image

### DIFF
--- a/cypress/e2e/tests/pages/home.spec.ts
+++ b/cypress/e2e/tests/pages/home.spec.ts
@@ -163,12 +163,12 @@ describe('Home Page', () => {
 
     homePage.supportLinks().should('have.length', 6);
 
-    // click Docs
-    homePage.clickSupportLink(0, true);
-    cy.origin('https://rancher.com', () => {
-      cy.on('uncaught:exception', () => false);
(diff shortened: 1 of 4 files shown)


mantis-toboggan-md (approved):
LGTM

PR #9448, opened 2023-07-28
fix issue with height on containerShell

### LINKED ISSUE #9422: Navigating to cluster explorer page returns Cannot read properties of undefined sometimes
Setup

Rancher version: 2.7-head `c00569d`

**Describe the bug**
When we have a kubectl shell open and then navigate to user Accounts & API keys, create a token and navigate back to the downstream explorer, the explorer page wont load and returns `Cannot read properties of undefined`

**To Reproduce**
1. Create an rke1 and rke2 node driver cluster
2. Open kubectl shell from the explorer of the rke1 and k3s node driver cluster
3. Navigate to profile >> Accounts & Api keys
4. Create a token with the downstream cluster scope.
5. Navigate back to the downstream cluster explorer 

**Result**
The page wont load and returns a blank page with `Cannot read properties of undefined (reading 'rows')`

**Expected Result**
Expected the page to load without errors

**Screenshots**

**Additional context**
Console Errors:
```
console.js:31 Cannot read properties of undefined (reading 'rows')
```

aalves08:
### Summary
Fixes #9422 
 
 - Update `height` prop definition to not be required, like `width` in `ContainerShell`
 - Update `proposeDimensions` return to be empty object in the case it can return `undefined`
 
 **NOTE:** with UI 2.7.5 (remote instance) I couldn't reproduce the bug
 
 ## To test
 - run latest UI code
 - go to `local` cluster explorer
 - Open a shell (header button)
 - Go to `profile` -> `Account & api keys`
 - Navigate back to local cluster explorer (back button on `Account & api keys` page)
 - Check that there are no logs on the console

### DIFF
--- a/shell/components/nav/WindowManager/ContainerShell.vue
+++ b/shell/components/nav/WindowManager/ContainerShell.vue
@@ -316,7 +316,7 @@ export default {
 
       this.fitAddon.fit();
 
-      const { rows, cols } = this.fitAddon.proposeDimensions();
+      const { rows, cols } = this.fitAddon.proposeDimensions() || {};
 
       if (!this.isOpen) {
         return;
--- a/shell/components/nav/WindowManager/ContainerShell.vue
+++ b/shell/components/nav/WindowManager/ContainerShell.vue
@@ -40,8 +40,8 @@ export default {
 
     // The height of the window
     height: {
-      type:     Number,
-      required: true,
+      type:    Number,
+      default: undefined,
     },
 
     // The width of the window


mantis-toboggan-md:
I found that I was able to prevent the error in the above screenshot by altering this line to allow for the possibility that `proposeDimensions` returns `undefined`
https://github.com/rancher/dashboard/blob/aac790e29a6b3a7e717f2e283841171e5504101b/shell/components/nav/WindowManager/ContainerShell.vue#L318-L320

`const { rows, cols } = this.fitAddon.proposeDimensions() || {};`

It's unclear to me _why_ this is undefined here, though, which makes me less confident in the solution.

aalves08:
@mantis-toboggan-md updated PR with your fix. Thanks for taking a proper look into this PR 🙏 🙇

mantis-toboggan-md (approved):
LGTM

PR #9449, opened 2023-07-28
Filtered out non-opaque secrets in backup/restore

### LINKED ISSUE #2405: Secrets in Backup/Restore should be only opaque
Noted in Rancher 2.5.3, Secrets in Backup/Restore should be only opaque secrets, grouped by namespace, ordered alphabetically (use secretselector component, needs to support multiple namespaces for this case).

Secrets aren’t in any sensible alphabetical order. We could probably do this alphabetical by namespace, but namespaces aren’t shown. Since the key is at the bottom, so maybe by creation date? Or sortable?

eva-vashkevich:
### Summary
Fixes #2405 
Dropdown is already only showing secrets that belongs to the current namespace.

Sorted and filtered out non-opaque secrets in "Encryption config secret" drop down in "Backup" and "Restore" wizards

Go to the cluster -> "Rancher backups" -> "Backups"/"Restores" -> Create backup/restore
Select "Encrypt backups using an Encryption Config secret (Recommended)"
"Encryption config secret" dropdown should only contain Opaque secrets and they should be sorted alphabetically

### DIFF
--- a/shell/edit/resources.cattle.io.backup.vue
+++ b/shell/edit/resources.cattle.io.backup.vue
@@ -16,8 +16,7 @@ import { allHash } from '@shell/utils/promise';
 import { NAMESPACE, _VIEW } from '@shell/config/query-params';
 import { sortBy } from '@shell/utils/sort';
 import { get } from '@shell/utils/object';
-
-const OPAQUE_SECRET_TYPE = 'Opaque';
+import { formatEncryptionSecretNames } from '@shell/utils/formatter';
 
 export default {
 
@@ -111,14 +110,7 @@ export default {
     },
 
     encryptionSecretNames() {
-      const filtered = this.allSecrets.filter(
-        (secret) => (secret.data || {})['encryption-provider-config.yaml'] &&
-        secret.metadata.namespace === this.chartNamespace &&
-        !secret.metadata?.state?.error &&
-        secret._type === OPAQUE_SECRET_TYPE
-      );
-
-      return filtered.map((secret) => secret.metadata.name).sort();
+      return formatEncryptionSecretNames(this.allSecrets, this.chartNamespace);
     },
 
     storageOptions() {
--- a/shell/edit/resources.cattle.io.backup.vue
+++ b/shell/edit/resources.cattle.io.backup.vue
@@ -16,6 +16,9 @@ import { allHash } from '@shell/utils/promise';
 import { NAMESPACE, _VIEW } from '@shell/config/query-params';
 import { sortBy } from '@shell/utils/sort';
 import { get } from '@shell/utils/object';
+
+const OPAQUE_SECRET_TYPE = 'Opaque';
+
 export default {
 
   components: {
@@ -108,7 +111,14 @@ export default {
     },
 
     encryptionSecretNames() {
-      return this.allSecrets.filter((secret) => (secret.data || {})['encryption-provider-config.yaml'] && secret.metadata.namespace === this.chartNamespace && !secret.metadata?.state?.error).map((secret) => secret.metadata.name);
+      const filtered = this.allSecrets.filter(
+        (secret) => (secret.data || {})['encryption-provider-config.yaml'] &&
+        secret.metadata.namespace === this.chartNamespace &&
+        !secret.metadata?.state?.error &&
+        secret._type === OPAQUE_SECRET_TYPE
+      );
+
+      return filtered.map((secret) => secret.metadata.name).sort();
     },
 
     storageOptions() {
--- a/shell/edit/resources.cattle.io.restore.vue
+++ b/shell/edit/resources.cattle.io.restore.vue
@@ -13,6 +13,9 @@ import { SECRET, BACKUP_RESTORE, CATALOG } from '@shell/config/types';
 import { allHash } from '@shell/utils/promise';
 import { get } from '@shell/utils/object';
 import { _CREATE } from '@shell/config/query-params';
+
+const OPAQUE_SECRET_TYPE = 'Opaque';
+
 export default {
 
   components: {
@@ -93,7 +96,14 @@ export default {
     },
 
     encryptionSecretNames() {
-      return this.allSecrets.filter((secret) => !!(secret.data || {})['encryption-provider-config.yaml'] && secret.metadata.namespace === this.chartNamespace && !secret.metadata?.state?.error).map((secret) => secret.metadata.name);
+      const filtered = this.allSecrets.filter(
+        (secret) => !!(secret.data || {})['encryption-provider-config.yaml'] &&
+        secret.metadata.namespace === this.chartNamespace &&
+        !secret.metadata?.state?.error &&
+        secret._type === OPAQUE_SECRET_TYPE
+      );
+
+      return filtered.map((secret) => secret.metadata.name).sort();
     },
 
     isEncrypted() {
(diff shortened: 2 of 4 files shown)


shell/edit/resources.cattle.io.backup.vue
```diff
@@ -108,7 +111,14 @@ export default {
...
     encryptionSecretNames() {
-      return this.allSecrets.filter((secret) => (secret.data || {})['encryption-provider-config.yaml'] && secret.metadata.namespace === this.chartNamespace && !secret.metadata?.state?.error).map((secret) => secret.metadata.name);
+      const filtered = this.allSecrets.filter(
+        (secret) => (secret.data || {})['encryption-provider-config.yaml'] &&
+        secret.metadata.namespace === this.chartNamespace &&
+        !secret.metadata?.state?.error &&
+        secret._type === OPAQUE_SECRET_TYPE
+      );
+
+      return filtered.map((secret) => secret.metadata.name).sort();
```
cnotv, line 121: Mind that `metadata` is optional and name may not exist for any reason, so it may be worth to filter it further.

Also if you want to chain with more filters I would recommend something like:

```ts
this.allSecrets
      .filter((secret) => (
        secret.data || {})['encryption-provider-config.yaml'] &&
        secret.metadata.namespace === this.chartNamespace &&
        !secret.metadata?.state?.error &&
        secret._type === OPAQUE_SECRET_TYPE
      )
      .map((secret) => secret.metadata.name)
      .sort()
```

Also we should add an unit test for this.

shell/edit/resources.cattle.io.restore.vue
```diff
@@ -93,7 +96,14 @@ export default {
     },
 
     encryptionSecretNames() {
-      return this.allSecrets.filter((secret) => !!(secret.data || {})['encryption-provider-config.yaml'] && secret.metadata.namespace === this.chartNamespace && !secret.metadata?.state?.error).map((secret) => secret.metadata.name);
+      const filtered = this.allSecrets.filter(
```
cnotv, line 99: It seems like this logic is repeated and could be replaced by a function, which is easier to test.

cnotv (changes requested):

shell/utils/__tests__/formatter.test.ts
```diff
@@ -0,0 +1,143 @@
+import { formatEncryptionSecretNames } from '@shell/utils/formatter';
+
+describe('formatter', () => {
+  const secrets = [
```
cnotv, line 4: Could you please filter up the used object to just the relevant properties used for this computation?
A whole object makes very complicate to understand the aim of the test and the involved data.

shell/utils/formatter.js
```diff
@@ -0,0 +1,12 @@
+
+const OPAQUE_SECRET_TYPE = 'Opaque';
```
cnotv, line 2: Please consider using the `./shell/config` for static configuration.
On fact for this case it seems to exist already [`/shell/config/secret.js`](https://github.com/rancher/dashboard/blob/master/shell/config/secret.js)

shell/utils/__tests__/formatter.test.ts
```diff
@@ -0,0 +1,143 @@
...
+        relationships:     null,
+        resourceVersion:   '4306',
+        state:             {
+          error: false, message: 'Resource is always ready', name: 'active', transitioning: false
+        },
+        uid: 'fcc4aa44-ce0c-4a7d-bd67-e77040b889a6'
+      }
+    }];
+  const chart = 'test';
+  const testCases = [[chart, 2], ['test1', 0]];
```
cnotv, line 130: Unless there's a `describe` scope case, it is convenient to avoid unnecessary constants declaration. A direct definition of the value within the `each` seems more appropriate.

shell/utils/__tests__/formatter.test.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  it.each(testCases)( 'should show correct number of secrets', ( chartVal: string, result: number) => {
+    const res = formatEncryptionSecretNames(secrets, chartVal);
+
+    expect(res).toHaveLength(result);
+  });
+  it('should return correct results in a correct order', () => {
+    const res = formatEncryptionSecretNames(secrets, chart);
+
+    expect(res[0]).toBe('test2');
```
cnotv, line 140: Would it not be more convenient to have the following?

```ts
expect(res).toStrictlyEqual(['test2', 'test5']);
```

shell/utils/formatter.js
```diff
@@ -0,0 +1,12 @@
...
+
+export function formatEncryptionSecretNames(secrets, chartNamespace) {
```
cnotv, line 4: If you are entirely aware of the whole logic, could you add a comment to the function? If it's just a abstraction and your knowledge is _opaque_ (😆), which happens to everyone and very often in our code 😓, we may avoid to add the comment for now.

cnotv (changes requested):
Thanks for adding a test for this shared computation. I've added a couple of notes regards the test.

shell/utils/formatter.js
eva-vashkevich: We were trying to figure out where this requirement came from, but we couldn't trace it. So, we basically just know that at some point a couple years ago it was decided that we should only show opaque secrets. I doubt a comment will be helpful, unfortunately

cnotv (approved):
Thank you for the patience and review all the points! It looks better now.

LGTM

shell/utils/formatter.js
```diff
@@ -0,0 +1,12 @@
+
+import { SECRET_TYPES } from '@shell/config/secret';
+
+export function formatEncryptionSecretNames(secrets, chartNamespace) {
+  return secrets.filter(
+    (secret) => (secret.data || {})['encryption-provider-config.yaml'] &&
+        secret.metadata.namespace === chartNamespace &&
+        !secret.metadata?.state?.error &&
+        !!secret.metadata?.name &&
```
mantis-toboggan-md, line 9: Why the check for `metadata.name`? This should always be defined (though during secret creation `generateName` may be set instead). I'm sure this additional check wont break anything, just wondering why it was added

shell/utils/__tests__/formatter.test.ts
```diff
@@ -0,0 +1,76 @@
+import { formatEncryptionSecretNames } from '@shell/utils/formatter';
+
+describe('formatter', () => {
+  const secrets = [
```
mantis-toboggan-md, line 4: It would be good to include one secret with metadata.state.error === true to test the one other bit of filtering in this function.

shell/utils/formatter.js
eva-vashkevich: @cnotv brought up a point that metadata is optional, so name might not exist

shell/utils/__tests__/formatter.test.ts
```diff
@@ -0,0 +1,88 @@
...
+      data:     { hash: 'test', 'encryption-provider-config.yaml': 'MTIzNFFhYWEh' },
+      metadata: {
+        namespace: 'test',
+        state:     {
+          error: false, message: 'Resource is always ready', name: 'active', transitioning: false
+        }
+      }
+    },
+    {
+      id:       'test6',
```
cnotv, line 65: Sorry what's the purpose of the new error test?

eva-vashkevich: It is one of the conditions based on which we are filtering out secrets. so it's good to check that case as well

cnotv (dismissed):
LGTM

shell/utils/formatter.js
mantis-toboggan-md: I double-chcked the kubernetes docs, it's always gonna be defined https://kubernetes.io/docs/concepts/overview/working-with-objects/names/

mantis-toboggan-md (approved):
lgtm

PR #9451, opened 2023-07-31
Implement the directive of positive number

DaiYuzeng:

https://github.com/harvester/harvester/issues/3977

### Technical notes summary

Add `v-positive-number` directive for the input which is the number type.

@WuJun2016 Please take a look, thanks!


DaiYuzeng:
@WuJun2016 PTAL, thanks!

WuJun2016:
I think the limit here is positive integers, so may be the name of the directive should be `v-positive-int-number`  ?

mantis-toboggan-md (approved):

PR #9456, opened 2023-07-31
Add tagged release option to publish-pkgs script

### LINKED ISSUE #9419: Add workflow to build extensions in a repo on a release creation (github)
Add workflow to build extensions in a repo on a release creation (github)

Not everyone wants to have a CI/CD like workflow. Some (again Opni) would like to only build the extension when a release is created/tagged. I believe our scripts are setup in such a way that this would be relatively easy to do. Maybe we just needs docs showing how you can create builds using tags or generate different workflows.

jordojordo:
### Summary
Fixes #9419 

This adds an option to the `publish-pkgs` script to allow for workflow builds to be triggered by [Tagged Releases](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release) and added documentation explaining the process. Depending on the tag name (which needs to match either the package name or main extension name), the publish script will only build for the specified tag.

If it is a container build, it will build normally while checking for the tag name to match the extension name. If it is the chart build, the script will only build the extension package which matches the tag name.

An example of the publish script using these additions can be found in [this repository](https://github.com/jordonsuse/rancher-extension-2) - the workflows have been triggered by creating tagged releases.

Also updated the shell package version.


docusaurus/docs/extensions/publishing.md
```diff
@@ -42,6 +42,113 @@ Running any of the commands above will only build your images - in order to publ
 yarn publish-pkgs -r <REGISTRY> -o <ORGANIZATION> -p -c
 ```
 
-## Additional Information
+## Additional Release Configuration
```
aalves08, line 45: My opinion here is that we are missing what the chapter `Running the Publish Script` brought in a way: some clear cut examples on how to run the publish pkg script for both packages and containers. What do you think @jordojordo ?

aalves08 (commented):
Overall, lgtm, just left a comment regarding the documentation

docusaurus/docs/extensions/publishing.md
jordojordo: You're right, there was not any clear examples on running the publish script for the package chart builds. I have added those steps, let me know what you think.

aalves08 (approved):
lgtm

PR #9461, opened 2023-08-01
fixed bad import in NavLinks

eva-vashkevich:
Fixed compilation warning because of the wrong import in ui.cattle.io.navlink.vue


momesgin (approved):
LGTM

PR #9486, opened 2023-08-04
Support array type for cni values in yaml

### LINKED ISSUE #9378: CNI Cluster Configuration Discrepancy in Rancher-Provisioned RKE2 Clusters
Internal reference: SURE-6647
Reported in 2.7.3

Multiple-CNI configuration for rke2 standalone clusters are defined as a yaml list as referenced here - https://docs.rke2.io/install/network_options#using-multus

However, the Multiple-CNI configuration for rke2 downstream clusters requires a comma-delimited list of strings instead of a yaml list. The suspicion is because the CNI selection is a single-select element, meaning each combination is included as a single, comma-separated entry.

After some inspection, a user tracked down this line, where the cni is broken up:

`rke2.vue @ line: 815\n const parts = cni.split(',').map(x => `rke2-${ x }`);\n\n`

The Rancher documentation references the RKE2 documentation as the complete reference of options for RKE2 Kubernetes Clusters in YAML. This would imply that all options would behave identically between setting options in the Rancher UI with YAML and in the `/etc/rancher/rke2/config.yaml` file - https://docs.rke2.io/install/configuration

The user would like to know if this a UI validation edge-case that should be supported or if they are expected to define our cluster CNI as a comma-separated list?

Repro steps:

- Create a new downstream RKE2 cluster
- When setting the CNI, switch to Edit as YAML and configure the cni as stated in the RKE2 docs:
```
 cni:
- multus
- canal
```
- The error `"ERROR: e.split is not a function. (In 'e.split(",")', 'e.split' is undefined)"` appears.
- Switch the CNI configuration to read: 
```
cni: multus,canal
```
- The cluster provisions successfully.

Notes:
Backend ruled out in that the `"ERROR: e.split is not a function"` error comes from the Rancher dashboard code when it tries to split the array, as it is expecting a string value, not an array.

momesgin:
### Summary
Just a simple array and string type checks for CNI values.

Fixes #9378

### DIFF
--- a/shell/edit/provisioning.cattle.io.cluster/rke2.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/rke2.vue
@@ -792,10 +792,10 @@ export default {
       const names = [];
       const cni = this.serverConfig.cni;
 
-      if ( cni ) {
-        const parts = cni.split(',').map((x) => `rke2-${ x }`);
-
-        names.push(...parts);
+      if (typeof cni === 'string') {
+        names.push(...cni.split(',').map((x) => `rke2-${ x }`));
+      } else if (Array.isArray(cni)) {
+        names.push(...cni.map((x) => `rke2-${ x }`));
       }
 
       if (this.showCloudProvider) { // Shouldn't be removed such that changes to it will re-trigger this watch


rak-phillip (approved):
LGTM

Issue #9482, opened 2023-08-03
Show message to users who try to login during GUID <> AD migration


MKlimuszka:
We want to be able to display a unique message when a login fails because AD login is disabled due to the migration.

For Rancher 2.7.6:

    While the migration configmap object indicates that a migration is in progress, logins are rejected for ActiveDirectory.
    The migration script can provide a custom error message, and this may be updated to reflect the migration progress (as a percentage) for particularly large environments.
        The error returned from the API call will look like:
            {
            "baseType": "error",
            "code": "ClusterUnavailable",
            "message": "Login is disabled while migration is running",
            "status": 503,
            "type": "error"
            }
        If we want to show a custom message in the UI, that will require a UI change, otherwise the user will just see the standard, "An error occurred logging in. Please try again."
        If it's useful, it is possible for the migration script to create a RancherUserNotification object.  I think the UI would still need to be updated to display that notification on the login page since afaik, there is nothing doing that right now.
    Local users are not affected and may log in, primarily because we want to ensure local admins are always empowered to investigate the ongoing migration and troubleshoot if necessary

gaktive:
@nwmac may have an existing pattern to use here. He's checking for a prior PR.

gaktive:
PR: https://github.com/rancher/dashboard/pull/9487

mantis-toboggan-md:
Backend has a pr up; waiting on an image w/ that code to continue work on this

mantis-toboggan-md:
UI side is merged; testing will require https://github.com/rancher/rancher/pull/42325 to be merged

sowmyav27:
There is a typo in this error message `An error occured logging in: login is disabled while migration is running`
 . It should be `occurred` instead.

sowmyav27:
On latest 2.7.6-head, i can see the correct error message - `An error occurred logging in: login is disabled while migration is running`.

- Edit configmap from Finished status to Running status.
- Noticed this in rancher logs `2023/08/18 18:45:52 [INFO] [migrate-ad-user] ad-guid migration is currently running, refusing to run again concurrently but the admin/std users were still able to login`
- Edited the Authconfig and changed Finished status to Running status.
- Admin and Std user are blocked from logging in with this ^ error message

PR #9487, opened 2023-08-04
Auth: Preserve error code

nwmac:
WIP: Preserve the error code from the auth flow

Not sure both changes are needed for: #9482 

Probably only the change in `authenticated.js` is needed to preserve error code - will probably need something to preserve and pass along the error message.

### DIFF
--- a/shell/middleware/authenticated.js
+++ b/shell/middleware/authenticated.js
@@ -209,6 +209,13 @@ export default async function({
     if ( route.name === 'index' ) {
       return redirect(302, '/auth/login');
     } else {
+      // Don't lose the error code
+      const errorCode = route.query?.errorCode;
+
+      if (errorCode) {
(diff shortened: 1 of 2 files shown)


mantis-toboggan-md:
Not sure about the changes in `verify.vue` - does AD auth hit that page?? Per slack we need to catch this error from "The api call for activedirectory?action=login" which I think would mean passing on `err.message` here: 
https://github.com/rancher/dashboard/blob/02c7cee83316ef7e1a1103dd567322823a8b4eb0/shell/store/auth.js#L334-L343


I'll doublecheck once BE has an image we can use to test this with.

nwmac:
@mantis-toboggan-md yeah, I don't think those changes are needed.

nwmac:
Closing this out as stale

PR #9495, opened 2023-08-07
Add reusable workflows for extension builds

### LINKED ISSUE #9290: Convert extension actions to support reusable workflows

The workflow files that are generated for extensions (found [here](https://github.com/rancher/dashboard/tree/master/shell/creators/pkg/files/.github/workflows)) should be migrated and converted into reusable workflows.

There is a gap between updating an extension's `@rancher/shell` package and updating the workflows that they utilize to build and publish the extension packages.

The process of updating the workflows and scripts which they target could benefit from using [reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows).
This would allow us to update any workflow actions/scripts independently and remove the required intervention of the extension developer to make manual updates to their extension workflow files.

jordojordo:
### Summary
Fixes #9290 

This converts the original workflows for extension chart and catalog builds to use the [resuable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows) method.

Now when creating an extension, there will only be 1 workflow file created (`.github/workflows/build-extensions.yml`) when the `-w` option is passed to `yarn create @rancher/pkg` that contains jobs for both the chart and catalog builds.

These jobs will now be pointing to the workflows found in the dashboard repository at:
- `.github/workflows/build-extension-charts.yml`
- `.github/workflows/build-extension-catalog.yml`

They are versioned by the branch when calling the file by appending the `@<branch>` to the file name (defaults to `master`).

### Technical notes summary

The jobs in `build-extensions.yml` target the reusable workflows with:
```yaml
uses: rancher/dashboard/.github/workflows/build-extension-<type>.yml@master
```

Each job has `permissions` that enable the workflow to write the necessary assets into the repository, as well as a few inputs based on the job.
***build-extension-charts inputs:***
```yaml
with:
  target_branch: gh-pages
```

***build-extension-catalog inputs:***
```yaml
with:
  registry_target: ghcr.io
  registry_user: ${{ github.actor }}
secrets: 
  registry_token: ${{ secrets.GITHUB_TOKEN }}
```

These are configurable and translate to the options passed to the `yarn publish-pkgs` script.


.github/workflows/build-extension-catalog.yml
```diff
@@ -1,20 +1,34 @@
...
+      registry_token:
+        required: true
 
 jobs:
   build:
     name: Build container image
     if: github.ref == 'refs/heads/main' && github.event_name != 'pull_request'
     runs-on: ubuntu-latest
-    permissions: write-all
+    permissions:
```
aalves08, line 24: do we need all these permissions? have you checked if there are some that aren't need? a bit nitpicky, I know, but we should only use the essential ones, imo

.github/workflows/build-extension-catalog.yml
```diff
@@ -61,4 +75,10 @@ jobs:
...
       - name: Build and push UI image
         run: |
-          yarn publish-pkgs -cp -r ${{ env.REGISTRY }} -o ${{ github.repository_owner }}
+          publish="yarn publish-pkgs -cp -r ${{ inputs.registry_target }} -o ${{ github.repository_owner }}"
+          
+          if [[ -n "${{ inputs.tagged_release }}" ]]; then
+            publish="$publish -t ${{ inputs.tagged_release }}"
+          fi
+          
+          $publish
```
aalves08, line 84: I don't know much about bash scripting, but you cleared the reference to `yarn`. does this still work as expected?

aalves08 (commented):
Overall, great job mate! 💪 

Again, not the the best reviewer of bash scripting, but apart from my comments (which should be valid for both `build-extension` files) I don't see anything that stands out. I am sure that this has been quite tested. Would love to have a demo from you mate.

I have a general question:
In terms of breaking changes, which implications does this have for current users/builders of extensions?

.github/workflows/build-extension-catalog.yml
jordojordo: I've updated the permissions to only include the ones needed, previously we were requiring `write-all` which is certainly not necessary. I also updated the documentation to explain the necessary requirements.

The Github documentation around permissions can be a bit [verbose](https://docs.github.com/en/rest/overview/permissions-required-for-github-apps?apiVersion=2022-11-28) - but in short:
- `actions` requires write to [cancel a workflow](https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#cancel-a-workflow-run)
- `contents` requires read for `actions/checkout` and requires write to `put` the contents of the built extension charts [into a branch](https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#create-or-update-file-contents)
- `deployments` requires write when [deploying  `gh-pages`](https://docs.github.com/en/rest/deployments/deployments?apiVersion=2022-11-28#create-a-deployment)
- `packages` requires write when a catalog image is created to [create the package](https://docs.github.com/en/packages/managing-github-packages-using-github-actions-workflows/publishing-and-installing-a-package-with-github-actions#publishing-a-package-using-an-action)
- `pages` requires write to [request and create page builds](https://docs.github.com/en/rest/pages/pages?apiVersion=2022-11-28#request-a-github-pages-build) for the deployment

aalves08 (approved):
LGTM @jordojordo .

aalves08:
@jordojordo does this have implications in terms of breaking changes for current extension developers?

Issue #9496, opened 2023-08-08
Extension Helm charts are incorrectly filtered when using display-name annotation


jtomasek:
**Setup**
- Rancher version: v2.7.5
- Rancher UI Extensions: v2.7.5

**Describe the bug**
`catalog.cattle.io/display-name` annotation on extension Helm chart is used to define a human readable name of the extension which is then used to display it in the extensions list

[The logic which combines the helm charts list](https://github.com/rancher/dashboard/blob/master/shell/pages/c/_cluster/uiplugins/index.vue#L248) with uiplugins list is incorrect because the chart name is set to `chart.chartDisplayName` instead of `chartName`. The relevant uiplugin is then not found and this causes a plugin to be listed in both `Installed` and `Available` tabs.

**To Reproduce**
In Rancher dashboard add a repository which includes a chart which has a `catalog.cattle.io/display-name` annotation. Install the chart.

**Result**
The same plugin is listed as `Installed` and `Available`.

**Expected Result**
The plugin is only visible in `Installed` tab.

**Screenshots**
KubeVirt chart uses `catalog.cattle.io/display-name` annotation to define its human readable name. After it is installed it is visible in both `Available` and `Installed` tabs:
<img width="982" alt="Screenshot 2023-08-08 at 10 45 22" src="https://github.com/rancher/dashboard/assets/1121740/4bbf472e-158d-4eb6-956d-5a750ff6e1eb">
<img width="983" alt="Screenshot 2023-08-08 at 10 45 09" src="https://github.com/rancher/dashboard/assets/1121740/d88b332b-4593-45a2-9f1c-30bc90363518">

aalves08:
Closed in favor of https://github.com/rancher/dashboard/pull/9544

PR #9497, opened 2023-08-08
Correctly filter helm charts with display-name annotation

jtomasek:
When calculating available extension helm charts and combining them with uiplugins the chart item name must be set to chartName instead of chartDisplayName.

Fixes https://github.com/rancher/dashboard/issues/9496

### DIFF
--- a/shell/pages/c/_cluster/uiplugins/index.vue
+++ b/shell/pages/c/_cluster/uiplugins/index.vue
@@ -245,7 +245,7 @@ export default {
         // Label can be overridden by chart annotation
         const label = uiPluginAnnotation(UI_PLUGIN_CHART_ANNOTATIONS.DISPLAY_NAME) || chart.chartNameDisplay;
         const item = {
-          name:         chart.chartNameDisplay,
+          name:         chart.chartName,
           label,
           description:  chart.chartDescription,
           id:           chart.id,


aalves08 (commented):
Hello @jtomasek 👋 

I'll have to say that I don't agree with this change. Looking at the definition of a chart https://github.com/rancher/dashboard/blob/master/shell/store/catalog.js#L518-L519 you can see that `chartNameDisplay` defaults to what populates `chartName`.

I am not contesting the fact you've found a bug, but the fix itself is just like a "bandaid" over the real problem, which should be what exactly caused the bug if you decide to use the annotation `CATALOG_ANNOTATIONS.DISPLAY_NAME`.

Can you provide me the repo + branch where this extension lives so I can try it out? 

Either on [Kubewarden](https://github.com/rancher/kubewarden-ui/blob/gh-pages/index.yaml) or [Elemental](https://github.com/rancher/elemental-ui/blob/main/index.yaml) we don't use the `CATALOG_ANNOTATIONS.DISPLAY_NAME` to override the `chartNameDisplay`, so there must be something on the underlying logic of what's installed or not that doesn't find the correct match and that's where the problem lies.

FYI @richard-cox

aalves08:
Digging a bit @jtomasek , if I had to guess, them problem lies with the different `all.find((c) => c.name === p.name)`, line 303 as an example, which I believe it should reflect something like the logic around line 308.

https://github.com/rancher/dashboard/blob/master/shell/pages/c/_cluster/uiplugins/index.vue#L303-L308

This is just an educated guess, not the exact solution or exactly where the problem lies. I would start from there. 

Cheers

jtomasek:
Yes, but If a display-name annotation is supposed to be used as a human readable label I believe that that logic is correct. The name should not be overwritten by display-name annotation contents. Only label should (and it is). On the other hand, the logic which my PR is changing is I believe incorrect as it sets the name to display-name annotation contents which it shouldn't. It should set just the label which it correctly does.

aalves08:
Closed in favor of https://github.com/rancher/dashboard/pull/9544

PR #9498, opened 2023-08-08
fix growl text overflow

### LINKED ISSUE #14793: Truncate the system messages we show as "growls" / toast
Currently, every system message (received from the BE) we show as a growl / toast message gets displayed in-full. This might be ok for reasonable-long messages (3-5 lines of text), but not so much for (really) lengthy ones:

Image

Next to usability, accessibility and IxD issues, these lengthy system messages are now the root cause of a new problem: starting with Rancher 2.12, we have introduced a new notification centre - part of the initial specs was to double these system messages (when they appear) as new notifications, each with its own message type (error, warning, info, ...) to allow the user to view / digest them at their own pace / availability.

However, these lengthy system messages, in their current shape and form, don't quite fit the notification centre. Physically, they can be fitted, but that's not the point. The main issue is that the current solution breaks one of the design fundamentals: it doesn't take into account the user interest in these system messages in the first place (we have no idea if the user is actually interested in reading the full message) and it's not empathetic to the user's context (be considerate to distracting the user from what s/he's focused on). The result is a high cognitive load and a high interruption rate as well (intrusive behaviour).

**Solution**

Please **limit the body of the system message to a maximum of 3 lines of text**, followed by the suffix: `... read more` (read more = secondary text link). When the user clicks the "read more" link, the suffix is replaced by the rest of the message (we do not provide the reverted "collapse" mechanism, once you expand the text you can't collapse it back).

⚠️ Use this pattern ^ for both places where the system message is displayed: growl / toast and notification.

⚠️ When the user interacts with the "read more" link inside a growl / toast message, stop the countdown timer that automatically "kills" it (we switch to manual dismissal only) - it means the user is interested in reading it.

WuJun2016:
Fixes #14793

I've updated this PR to add the height limit and scroll to the growl message rather than the outer container - so each growl now has a restricted height and the growl scrolls, rather than the a scroll bar being added to the entire app for the growl list.

See below:

### Technical notes summary

Added a `max-height` and `overflow-y` to the growl message text to ensure it does not grow to a very largge height.

### Screenshot/Video

image

### Checklist
- [x] The PR has a Milestone

### DIFF
--- a/shell/components/GrowlManager.vue
+++ b/shell/components/GrowlManager.vue
@@ -155,7 +155,7 @@ export default {
 
   .growl-list {
     max-height: calc(100vh - 50px);
-    overflow: hidden;
+    overflow: scroll;
   }
 
   .growl {
--- a/shell/components/GrowlManager.vue
+++ b/shell/components/GrowlManager.vue
@@ -155,7 +155,7 @@ export default {
 
   .growl-list {
     max-height: calc(100vh - 50px);
-    overflow: scroll;
+    overflow: hidden;
   }
 
   .growl {
@@ -203,6 +203,10 @@ export default {
 
         > P {
           padding-top: 2px;
+
+          // Limit size of message and scroll just the message, not the entire growl, so that the title and icon are always visible
+          max-height: 200px;
+          overflow-y: scroll;
 
           &.has-title {
             margin-top: 5px;


richard-cox:
I think this requires a UX review, we shouldn't be showing so much content in a growl. cc @kwwii

richard-cox:
> > I think this requires a UX review, we shouldn't be showing so much content in a growl. cc @kwwii
> 
> There's one situation to consider: there's only one message, but it's a very long one. If we don't allow scrolling, then a portion of the message is invisible。

Think the aim of the growls are to show short, simple messages. Anything that requires scrolling kind of breaks that pattern

gaktive:
With the new notification centre, we should verify how the text overflow looks and if concatenation is possible.

nwmac:
I've updated this PR to add the height limit and scroll to the growl message rather than the outer container - so each growl now has a restricted height and the growl scrolls, rather than the a scroll bar being added to the entire app for the growl list.

See below:

image

nwmac:
@richard-cox Updated with what I think is a reasonable fix for now.

aalves08 (approved):
LGTM @nwmac

PR #9502, opened 2023-08-08
Fix usage of plugin metadata annotations

jtomasek:
Annotations for an extension/plugin should be defined in package.json at rancher.annotations object so that the annotations are correctly copied when generating a Helm chart and correctly used in the dashboard.

Fixes https://github.com/rancher/dashboard/issues/9501


aalves08:
Closed in favor of https://github.com/rancher/dashboard/pull/9544

PR #9503, opened 2023-08-08
Show ad migration login error message

### LINKED ISSUE #9482: Show message to users who try to login during GUID <> AD migration

mantis-toboggan-md:
### Summary
Fixes #9482

Currently when a login request fails we show one of two generic error messages. This PR updates the error-catching behavior to show a specific error message when one has been provided, except 401 errors, which still show `Invalid username or password. Please try again.`

Screen Shot 2023-08-08 at 4 09 52 PM

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -3008,6 +3008,7 @@ login:
     noResponse: "No response received"
   error: An error occurred logging in. Please try again.
   clientError: Invalid username or password. Please try again.
+  specificError: 'An error occured logging in: {msg}'
   useLocal: Use a local user
   loginWithProvider: Log in with {provider}
   username: Username
--- a/shell/pages/auth/login.vue
+++ b/shell/pages/auth/login.vue
@@ -149,7 +149,7 @@ export default {
         return this.t('login.error');
       }
 
-      return this.err;
+      return this.err?.length ? this.t('login.specificError', { msg: this.err }) : '';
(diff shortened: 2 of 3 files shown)


mantis-toboggan-md:
Oof, sorry about that @aalves08 - I was verifying that my changes didn't affect the custom login error banner (configurable in global settings>banners). I've disabled that setting now so only the AD migration message should show.

aalves08 (approved):
No more strange copy appearing. All good! 🏁

PR #9514, opened 2023-08-10
locationConfig hash param

### LINKED ISSUE #9511: Add ability to add cols to table inside specific tabs
Using the `addTableColumn` method from the [Extensions API](https://rancher.github.io/dashboard/extensions/api/table-columns) we should be able to target a table in a specific tab. For that we need to expand the [LocationConfig](https://rancher.github.io/dashboard/extensions/api/common#locationconfig) admissable parameters to cover this usecase.

Currently, if there's a detail view with multiple tabs + tables, if we target the combination of resource + mode (detail) we will add the same table column to all tables.

With the desired improvement, we should be able to target the table of our choice.

aalves08:
Fixes #9511 

- add hash param to `locationConfig` object in extensions to allow fine-tune of application of an extension point 
- **Sneaky fix**: change logic around extension table cols to better reflect col location based on page user is browsing and not just resource, because same resource can be in multiple pages, but extra table col may only be meaningful in a specific page **->** 
_this will introduce breaking change as target is now different. I believe it's for the best as this allows us to selectively target tables inside resource detail page, which combined with the new `hash` parameter, allows for a finer control of the application of any extension enhancement._

- create a new package `yarn create @rancher/pkg test`
- on the `index.ts` file for the newly created package, replace the code with the following:
```
import { importTypes } from '@rancher/auto-import';
import { IPlugin, TableColumnLocation } from '@shell/core/types';

// Init the package
export default function(plugin: IPlugin): void {
  // Auto-import model, detail, edit from the folders
  importTypes(plugin);

  // Provide plugin metadata from package.json
  plugin.metadata = require('./package.json');

  plugin.addTableColumn(
    TableColumnLocation.RESOURCE,
    {
      resource: ['provisioning.cattle.io.cluster'], mode: ['detail'], hash: ['node-pools']
    },
    {
      name:     'test-col',
      label:       'test col',
      getValue: (row: any) => {
         return 100;
      },
    }
  );
}
```
- Go to `cluster management`, click on `local` cluster to see it's details, make sure `test col` appears on the list of `machine pools` (before the changes to the table col mechanism, one would target this table by setting the `locationConfig` to `{ resource: ['management.cattle.io.node'] }`.
- Check other tabs to make sure `test col` doesn't appear on other table of the same details page

FYI @richard-cox @nwmac

### DIFF
--- a/docusaurus/docs/extensions/api/table-columns.md
+++ b/docusaurus/docs/extensions/api/table-columns.md
@@ -6,7 +6,7 @@ Table Columns are added to Rancher via the `addTableColumn` method.
 
 *(Rancher version v2.7.2)*
 
->**IMPORTANT NOTE:** on **Rancher version v2.8.0** we've introduced breaking changes to the behaviour of this extension enhancement (Table Columns), as it before you would target the resource name of the table you were trying to extend, which was different from the usage of the [LocationConfig object](./common#locationconfig) in any of the other extension enhancements available. With these new changes, the [LocationConfig object](./common#locationconfig) will be used to target a specific page that contains a table and add it to that particular one, therefore having a better control of the new table column appearance.
+>**IMPORTANT NOTE:** on **Rancher version v2.8.0** we've introduced breaking changes to the behaviour of this extension enhancement (Table Columns). Previously, you would target the resource name of the table you were trying to extend, which was different from the usage of the [LocationConfig object](./common#locationconfig) in any of the other extension enhancements available. With these new changes, the [LocationConfig object](./common#locationconfig) will be used to target a specific page that contains a table and add it to that particular one, therefore having a better control of the new table column appearance.
 
 This method adds a table column to a `ResourceTable` element-based table on the UI.
 
--- a/docusaurus/docs/extensions/api/table-columns.md
+++ b/docusaurus/docs/extensions/api/table-columns.md
@@ -6,6 +6,8 @@ Table Columns are added to Rancher via the `addTableColumn` method.
 
 *(Rancher version v2.7.2)*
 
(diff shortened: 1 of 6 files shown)


aalves08:
I know we are really late in the release process, but is there any chance that this can go in into the 2.7.7 release? Had to ask...  SUMA extension will make use of this mechanism. @richard-cox @nwmac

docusaurus/docs/extensions/api/table-columns.md
```diff
@@ -6,7 +6,9 @@ Table Columns are added to Rancher via the `addTableColumn` method.
 
 *(Rancher version v2.7.2)*
 
-This method adds a table column to a `SortableTable`/`ResourceList` element-based table on the UI.
+>**IMPORTANT NOTE:** on **Rancher version v2.8.0** we've introduced breaking changes to the behaviour of this extension enhancement (Table Columns), as it before you would target the resource name of the table you were trying to extend, which was different from the usage of the [LocationConfig object](./common#locationconfig) in any of the other extension enhancements available. With these new changes, the [LocationConfig object](./common#locationconfig) will be used to target a specific page that contains a table and add it to that particular one, therefore having a better control of the new table column appearance.
```
jordojordo, line 9: > as it before you would target

This should be "previously, you would target".

jordojordo (changes requested):
One small wording change, otherwise lgtm.

aalves08:
@jordojordo when you have some spare time, can you approve this? I need this to go in so that I can save the work done on the SUMA extension as it was mostly done using `yarn link` to this branch. 🙏

jordojordo (approved):

PR #9528, opened 2023-08-11
Added deprecation warning for OPA gatekeeper

### LINKED ISSUE #9483: Add deprecation notice for OPA Gatekeeper Chart
https://jira.suse.com/browse/SURE-6270

In the past, there were notices of deprecation for chart ex:- https://github.com/rancher/rancher/releases/tag/v2.5.0 `Deprecation Features` 

We want to deprecate OPA Gatekeeper and we want to send this notication in v2.8.0.

I think the warning/notices could be shown at the following pages 

**Installed Apps**

If the user has OPA Gatekeeper installed, then this warning/notice could be shown. 

**OPA Gatekeeper Chart Info Page**

When the user clicks on the Chart icon the user is redirected to the info page from where he could click on Install. So, showing a warning here would be nice.

**Additional context**

momesgin:
### Summary
Added deprecation warning for OPA gatekeeper on its install page as well as its overview page.

Fixes #9483 

### Screenshot/Video
https://github.com/rancher/dashboard/assets/135728925/a3e86f2c-63c9-4866-a2db-77c31dcf4e15

### DIFF
--- a/shell/mixins/chart.js
+++ b/shell/mixins/chart.js
@@ -8,6 +8,7 @@ import { CATALOG as CATALOG_ANNOTATIONS } from '@shell/config/labels-annotations
 import { SHOW_PRE_RELEASE, mapPref } from '@shell/store/prefs';
 import { NAME as EXPLORER } from '@shell/config/product/explorer';
 import { NAME as MANAGER } from '@shell/config/product/manager';
+import { OPA_GATE_KEEPER_ID } from '@shell/pages/c/_cluster/gatekeeper/index.vue';
 
 import { formatSi, parseSi } from '@shell/utils/units';
 import { CAPI, CATALOG } from '@shell/config/types';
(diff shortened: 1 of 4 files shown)


shell/mixins/chart.js
```diff
@@ -185,6 +186,10 @@ export default {
         }
       }
 
+      if (this.chart?.id === OPA_GATE_KEEPER_ID) {
+        warnings.unshift(this.t('gatekeeperIndex.deprecated', {}, true));
+      }
```
rak-phillip, line 191: Is this the recommended way to show a deprecation warning? I'm noticing that there's already a `showDeprecated` computed prop and I'm wondering if that's used for the same thing.

momesgin: `showDeprecated` is not used for showing deprecation messages. We filter the charts that have a `deprecated` property set to `true`, but if `showDeprecated` is also `true` then we bypass this filter.

rak-phillip (approved):
LGTM

PR #9512, opened 2023-08-10
Correctly filter helm charts with display-name annotation

aalves08:
Fixes #9496 

- Adds fix on https://github.com/rancher/dashboard/pull/9497
- Adds e2e test to cover regressions

**Note:** https://github.com/rancher/dashboard/pull/9497 should be closed once this is approved and merged

To reproduce:
- add the `ui-plugin-examples` repo to `local` -> `apps` -> `repositories`
https://github.com/rancher/ui-plugin-examples -> branch `main`

- Install `clock` example v2 in extension
- after install and reload make sure extension card for the `clock` extension appears only on `installed` tab and not in `available` at the same time. It should only be in `installed` tab

PR #9544, opened 2023-08-15
Fix extension displayName annotation conditions - add configuration docs

### LINKED ISSUE #9318: Add the ability to customize chart annotations for extensions

It would be nice to have the ability to update the annotations which are set for an Extension's chart, for instance the `catalog.cattle.io/kube-version` and `catalog.cattle.io/rancher-version` annotations.

In order to update these annotations for an Extension, the Chart.yaml needs to be updated manually, then the assets and index.yaml need to be rebuilt in order to contain the new annotations. It would be better to have an option within the `yarn publish-pkgs` script that would allow the user to determine the values for the `kube-version` and `rancher-version` annotations.

Add options to the `yarn publish-pkgs` script to allow updating the annotations. For instance:
```sh
yarn publish-pkgs --kube-version ">= 1.16.0-0 < 1.22.0-0" --rancher-version ">= 2.7.2-0 < 2.8.0-0"
```

Perhaps instead of adding options to the script, it would be possible to define a set of options within a manifest like `rancher-config.yml` at the root of the extension which could apply these annotations automatically to each package that would be built as a result of running the `publish-pkgs` script.

### LINKED ISSUE #9501: Iconsistent usage of extension annotations

jordojordo:
### Summary
Fixes #9501 Fixes #9318

The issues mentioned by @jtomasek about how we displayed the `catalog.cattle.io/display-name` for an extension was different depending on the existence of an extension chart. Also we were not searching for the correct property (`rancher.annotations`) for this annotation. Enforcing the "annotations" property within the `package.json` should be the correct approach.

There was also an issue in how we were determining if an extension was installed by comparing the `plugin.name` with the `chart.name` - However, the `chart.name` was determined by the `chartDisplayName` rather than the actual chart name. So when the `catalog.cattle.io/display-name` was declared, the `chartNameDisplay`, which did not match the name of the installed extension, was being used to filter all `installed` extensions from the "Available" tab. 

If needed, I can provide an extension repository with annotations for testing.

- Modified the `package.json` for Harvester Manager to include the `annotations` property
- Fixed how we were calling the `uiPluginAnnotation` method by supplying the necessary `chart`
- Changed the `available` extensions to use the `chart.chartName` instead of `chart.chartNameDisplay`
- Properly filtering for `rancher.annotations` on installed extensions without a chart
- Added e2e for testing that an installed extension does not appear in the "Available" tab
  - The [Clock extension](https://github.com/rancher/ui-plugin-examples) should be able to utilize the `catalog.cattle.io/display-name` correctly now
- Added documentation on how to configure the extension package annotations.

### Screenshot/Video

***Before***

***After***

### DIFF
--- a/docusaurus/docs/extensions/extensions-configuration.md
+++ b/docusaurus/docs/extensions/extensions-configuration.md
@@ -3,3 +3,43 @@
 Follow instructions [here](./extensions-getting-started.md) to scaffold your extension. This will assist you in the creation of an extension as a top-level product inside Rancher Dashboard.
 
 Once you've done so, there are some initialization steps specific to extensions. Beyond that, extensions largely work the same as the rest of the dashboard. There are a set of top-level folders that can be defined and used as they are in the dashboard: `chart`, `cloud-credential`, `content`, `detail`, `edit`, `list`, `machine-config`, `models`, `promptRemove`, `l10n`, `windowComponents`, `dialog`, and `formatters`. You can read about what each of these folders does [here](../code-base-works/directory-structure.md).
+
+## Extension Package Metadata
+
+Each extension package has the ability to customize certain aspects when it comes to compatibility with Rancher Manager/Kubernetes or displaying extension names. These are determined by the `rancher.annotations` object applied to the `package.json` of an extension package.
+
+These annotations allow you to specify compatibility with Kubernetes, Rancher Manager, the Extensions API, and the Rancher UI version by relying on [semver ranges](https://www.npmjs.com/package/semver/v/6.3.0#ranges). As well as version compatibility, you can also specify a Display Name for the Extension package as it appears on the "Extensions" page within the UI.
+
+### Configurable Annotations
+
+| Annotation | Value | Description |
+| ------ | :------: | --------------|
+| `catalog.cattle.io/kube-version` | `Range` | Determines if the Kubernetes version that Rancher Manager is utilizing is compatible with the Extension package. |
+| `catalog.cattle.io/rancher-version` | `Range` | Determines the compatibility of the installed Rancher Manager version with the Extension package. |
+| `catalog.cattle.io/ui-extensions-version` | `Range` | Determines the Extensions API version that is compatible with the Extension package. |
+| `catalog.cattle.io/ui-version` | `Range` | Determines the Rancher UI version that is compatible with the Extension package. |
+| `catalog.cattle.io/display-name` | `String` | Specifies the Display Name for an Extension package's card on the "Extensions" page. |
+
+### Example Configuration
+
+Here's an example configuration of an extensions `package.json`:
+
+___`./pkg/my-package/package.json`___
+```json
+{
+  "name": "my-package",
+  "description": "my-package plugin description",
+  "version": "0.1.0",
+  "rancher": {
+    "annotations": {
+      "catalog.cattle.io/kube-version": ">= v1.26.0-0 < v1.28.0-0",
+      "catalog.cattle.io/rancher-version": ">= 2.7.7-0 < 2.8.0-0",
+      "catalog.cattle.io/ui-extensions-version": ">= 1.1.0",
+      "catalog.cattle.io/ui-version": ">= 2.7.7-0 < 2.8.0-0",
+      "catalog.cattle.io/display-name": "My Super Great Extension"
+    }
+  },
+  ...
+}
+```
+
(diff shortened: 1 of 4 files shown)


docusaurus/docs/extensions/extensions-configuration.md
```diff
@@ -2,4 +2,44 @@
...
\ No newline at end of file
+Once you've done so, there are some initialization steps specific to extensions. Beyond that, extensions largely work the same as the rest of the dashboard. There are a set of top-level folders that can be defined and used as they are in the dashboard: `chart`, `cloud-credential`, `content`, `detail`, `edit`, `list`, `machine-config`, `models`, `promptRemove`, `l10n`, `windowComponents`, `dialog`, and `formatters`. You can read about what each of these folders does [here](../code-base-works/directory-structure.md).
+
+## Extension Package Metadata
+
+Each extension package has the ability to customize certain aspects when it comes to compatibility with Rancher Manager/Kubernetes or displaying extension names. This are determined by the `rancher.annotations` object applied to the `package.json` of an extension package.
```
codyrancher, line 9: *These are determined

docusaurus/docs/extensions/extensions-configuration.md
```diff
@@ -2,4 +2,44 @@
...
+
+### Configurable Annotations
+
+| Annotation | Value | Description |
+| ------ | :------: | --------------|
+| `catalog.cattle.io/kube-version` | `Range` | Determines if the Kubernetes version that Rancher Manager is utilizing is compatible with the Extension package. |
+| `catalog.cattle.io/rancher-version` | `Range` | Determines the compatibility of the installed Rancher Manager version with the Extension package. |
+| `catalog.cattle.io/ui-extensions-version` | `Range` | Determines the Extensions API version that is compatible with the Extension package. |
+| `catalog.cattle.io/ui-version` | `Range` | Determines the Rancher UI version that is compatible with the Extension package. |
+| `catalog.cattle.io/display-name` | `String` | Specifies the Display Name for an Extension package's card on the "Extensions" page. |
```
codyrancher, line 21: Might be nice to include a screenshot to visually convey the area that's affected.

codyrancher (dismissed):

jtomasek:
This also fixes https://github.com/rancher/dashboard/issues/9496 and IIUC it makes https://github.com/rancher/dashboard/pull/9512/files, https://github.com/rancher/dashboard/pull/9497 and https://github.com/rancher/dashboard/pull/9502 obsolete.

IMHO it would be better to contribute commits to existing PRs instead of creating new one(s) causing the open ones obsolete.

aalves08:
@jtomasek I agree with you, but in this particular case, since we have strict requirements in terms of testing and we needed the fix in order to do the testing, it's much easier for us to do it all in one place which we fully control, rather than having to wait to merge stuff in (we haven't branched yet) and doing it sequentially.

aalves08:
@jordojordo @jtomasek closed other PRs in favor of this one. 🙏

codyrancher (approved):

PR #9552, opened 2023-08-16
Disable local cluster workspace change

Shavindra:
### Summary
Fixes rancher/dashboard#7745 

- If the `cluster === isLocal` then hide the option to change the workspace
- Add E2E to for GitRepo

- Navigate to Continuous Delivery
- Click on the local cluster menu three dots on the right and look for `Change workspace` 
- `Change workspace` option should not be visible

- Changing the workspace of other clusters should not be affected.
- NB: RKE2 clusters workspace can't be changed. Only RKE 1 cluster workspaces can be changed.

### DIFF
--- a/cypress/e2e/tests/pages/fleet/dashboard.spec.ts
+++ b/cypress/e2e/tests/pages/fleet/dashboard.spec.ts
@@ -21,7 +21,7 @@ describe.only('Fleet Dashboard', { tags: '@adminUser' }, () => {
     gitRepoCreatePage.goTo();
     gitRepoCreatePage.setRepoName(repoName);
     gitRepoCreatePage.selectWorkspace('fleet-local');
-    gitRepoCreatePage.setGitRepoUrl('https://github.com/Shavindra/fleet-examples.git');
+    gitRepoCreatePage.setGitRepoUrl('https://github.com/rancher/fleet-test-data.git');
     gitRepoCreatePage.setBranchName();
     // NB - This step is here because DOM may not be ready
     gitRepoCreatePage.goToNext();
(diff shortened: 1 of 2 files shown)


shell/models/fleet.cattle.io.cluster.js
```diff
@@ -79,6 +79,14 @@ export default class FleetCluster extends SteveModel {
     return false;
   }
 
+  get canChangeWorkspace() {
+    return !this.isRke2 && !this.isLocal;
+  }
+
+  get isLocal() {
+    return this.metadata.name === 'local' || this.metadata?.labels?.[FLEET_LABELS.CLUSTER_NAME] === 'local';
```
cnotv, line 87: I thought we had somewhere the `'local'` as static configuration but we don't. Feel free to add or resolve this comment without changes.

cnotv (changes requested):
Seems fine, just a formal change.

shell/types/menu-actions.ts
```diff
@@ -0,0 +1,11 @@
+// NB: for some reason, this is not being picked up by the linter as being used.
+export enum MenuActions {
+    Pause = 'Pause', // eslint-disable-line no-unused-vars
```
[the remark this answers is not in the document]
cnotv: Sorry my mistake, this is obviously fine if used for E2E. 
We should probably move it within the E2E area to avoid confusion and keep it outside the code.

cnotv (changes requested):
I've updated the comment after clarifying the use of the enum.

shell/config/types.js
```diff
@@ -311,3 +311,5 @@ export const AUTH_TYPE = {
   _SSH:   '_ssh',
   _S3:    '_S3'
 };
+
+export const LOCAL_CLUSTER = 'local'
```
cnotv, line 315: 🏆

PR #9560, opened 2023-08-17
Feat/app bar init

scures:
### Summary
Fixes rancher/dashboard#9407

### What's new
- Introduce the `APP Bar` which is always visible & slightly modified cleaning up some of the visuals.
- Clusters now have a 3-letter identifier name when collapsed (it's automatically generated in the format: `local` -> `LCL`, `Word-Wide-Web` -> `WWW`, `33abdf` -> `3DF`)
- There's a tooltip while the mouse over and collapsed.

Note: I create a new `ClusterProviderIconMenu.vue` which handles the menu badges since the original (ClusterProviderIcon) one seems to be used in many places and for our Menu use cases was a complete refactor

EDIT: Updated the screenshots to display the right images

image

image

image

image

### DIFF
--- a/shell/components/nav/TopLevelMenu.vue
+++ b/shell/components/nav/TopLevelMenu.vue
@@ -841,6 +841,8 @@ export default {
       }
 
       .none-matching {
+        width: 100%;
+        text-align: center;
         padding: 8px
       }
 


cnotv:
There's an issue with the E2E due to the change of markup:
```
AssertionError: Timed out retrying after 60000ms: Expected to find content: ' Configuration ' within the element: [ <div.category>, 1 more... ] but never did.
    at Context.eval (webpack:///./cypress/e2e/tests/pages/global-settings/branding.spec.ts:38:56)
```

Screenshot 2023-08-17 at 14 21 40

shell/components/ClusterProviderIconMenu.vue
```diff
@@ -0,0 +1,149 @@
...
+        result = `${ word[0] }${ word[Math.floor(word.length / 2)] }${ word[word.length - 1] }`;
+        break;
+      case 2:
+        result = `${ segments[0][0] }${ segments[1][0] }${ segments[1][segments[1].length - 1] }`;
+        break;
+      default:
+        result = segments.slice(0, 3).map((segment) => segment[0]).join('');
+      }
+
+      return result.toUpperCase();
```
cnotv, line 44: Would it not better to uppercase with CSS?

shell/components/ClusterProviderIconMenu.vue
```diff
@@ -0,0 +1,149 @@
...
+    },
```
cnotv, line 45: Maybe a comment about what is about? Is it about create acronymous of the label?
Also we should make a test for this.

shell/components/ClusterProviderIconMenu.vue
```diff
@@ -0,0 +1,149 @@
...
+    </svg>
+    <img
+      v-else
+      class="cluster-os-logo"
+      :src="cluster.providerNavLogo"
+    >
+  </div>
+</template>
+
+<style lang="scss" scoped>
```
cnotv, line 113: Don't we have application wise size at all?
I think icon size should probably defined in the SCSS.

cnotv (changes requested):
I am not so sure about the guidelines for element selections by class in the E2E, maybe ask @richard-cox 

For the rest there's just a couple of points.

Functionality looks good.

https://github.com/rancher/dashboard/assets/5009481/9da11311-2713-4bd1-8c3a-5a9843c625a1

shell/components/ClusterProviderIconMenu.vue
cnotv: Yeah sounds good to me too 👍

cnotv: Ok I've seen it now in the E2E, kinda odd but at least is tested.

shell/components/ClusterProviderIconMenu.vue
cnotv: I think everything should be pretty much consistent, however if you discussed with the designers and they said that is completely different, then we can do the same code wise.

cnotv (approved):
LGTM

richard-cox (changes requested):
Haven't checked out the code yet, but until then had a click around


Functional
- while on the home page clicking on `x more` gives a fail whale error page `Product product not found`. In the cluster explorer page nothing happnes
- Long cluster names aren't handled. Before they were truncated with `...`. I think though in the designs we have a new, better way to handle this?
- Have more than 4 clusters. Set cluster list size to four. Go to home page. Expand burger menu. Everything looks good (4 cluster icons shown). Now got to preferences page and the size the 4 cluster icons are shown in is reduced to 2 and a bit
- Running the dashboard locally, navigate to the Cluster Management --> Drivers page. The Loading indicator (which i think is used in other places) covers the side menu.
- Have more than 10 clusters, scroll to the bottom of the burger menu, search for a cluster. The search box doesn't ping around the screen given the change to the cluster list (great!) but the search results appear outside of visual range (not so great). This also leaves the burger menu in an odd state when collapsed


Purely Visual
- `x` icon in the cluster search box isn't vertically centred
- the burger menu doesn't quite look horizontally centrally aligned. Might be my eyes though image
- when the burger menu expands the branding icon is shown from within the burger menu. when it collapses the branding icon in the header is shown. these two aren't horizontally the same. this was the case before, but there was a transition to slowly fade in the burger menu branding icon. Not sure if that was a workaround, but it's a bit jarring now when expanding the menu and the branding icon (notionally) wiggles


Design wise
- Subjective - The new larger text looks out of place. It also makes the problem of lots of clusters worst (bigger rows mean less shown in the same vertical space)
- Subjective - There's quite a big gap between groups. See `Global Configuration`. This compounds the issue when there are lots of clusters
- The `Global Configuration` menu item used to be in a group titled `Legacy Apps`. This is probably important info that's been lost. Was there anything design wise for this?


Some demo screenshots
image

image


image

image

image

richard-cox (changes requested):
- Didn't get very far through code again, just e2e things
- The preference `Number of clusters to show in side menu` isn't working. That dictacts the minimum size of the vertical area used to show clusters, aka 'always show x clusters'. 
  - with 9 clusters and setting a 10, only 3 and a bit are visible
  - with 9 clusters and setting to 5
    - not quite sure the `5 clusters` text is right. the number changes to whatever the preference is and the clusters are still scrollable
  - Think the above strangeness should be resolved by the minimum height of the scrollable area being set to the height of the number of clusters to show, and the `x clusters` text showing the total number of clusters
  - the padding/margin for the `5 clusters` is very large in comparison to the `See all` text at the bottom. Feels like this should be uniformly similar and small for both 
    - image
- There's a large void at the bottom of the side bar
  - image
- All group labels are `GLOBAL APPS`
- The home icon isn't sized in line with others, so it looks missaligned
  - image
- slightly subjective - The lines denoting groups when collapsed have a different top and bottom padding, feels like these should be the same. For instance the legacy group when collapsed should be in the centre 
  - image

cypress/e2e/tests/navigation/side-nav/main-side-menu.spec.ts
```diff
@@ -1,14 +1,17 @@
...
 
 Cypress.config();
 describe('Side Menu: main', () => {
+  let modal;
+
   beforeEach(() => {
     cy.login();
     HomePagePo.goTo();
     BurgerMenuPo.toggle();
+    modal = new ModalWithCardPo();
```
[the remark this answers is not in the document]
richard-cox: odd, it's only used in one test (possibly a new change). i didn't grok it at the time, but `ModalWithCardPo` should inherit from `ComponentPo` and the `card` initialised in the ctor should be created each time it's needed (which normally is bad, but cypress does weird things when selectors aren't executed on demand)

richard-cox (changes requested):
Visual / Layout
- The home icon still looks misaligned with the burger menu and cluster abbreviations
  - image
- For clusters with looong names, the cursor icon and popover flashes when moving the cursor over the provider square. The cursor icon flashes between it's correct state and the `text` cursor
- This one is still pending
  - when the burger menu expands the branding icon is shown from within the burger menu. when it collapses the branding icon in the header is shown. these two aren't horizontally the same. this was the case before, but there was a transition to slowly fade in the burger menu branding icon. Not sure if that was a workaround, but it's a bit jarring now when expanding the menu and the branding icon (notionally) wiggles
- Have more than 10 clusters (cluster search field showing), filter by random text such that there is no matching cluster, collapse the burger menu. The text is truncated `No Matchi`. Solutions could involve wrapping text (possibly with work break too, though this might look bad), different text, show nothing, show an icon?
  - image

Discussion
- With 11 clusters and a vertical height all clusters can't fit, vertical scrolling of the whole side bar happens. This also happens with 9 clusters (no search / see all). Feels like all non cluster items should consume the available space until the clusters hit a minimum height. This would solve a lot of the odd double scrolling (whole side bar AND cluster part). Is this something that's tackled in Richa's PR?

Pending
I still need to test MCM Apps (not sure these are supported anymore, but there's code in there) and the impact on the standalone modes (used by epinio and harvester)

shell/components/ClusterProviderIconMenu.vue
```diff
@@ -0,0 +1,161 @@
...
+      class="cluster-badge-logo"
+      :class="{ 'disabled': !isEnabled }"
+      :style="{borderBottom: cluster.badge?.color ? `4px solid ${cluster.badge?.color}` : ''}"
+    >
+      <span
+        class="cluster-badge-logo-text"
+      >
+        {{ smallIdentifier(cluster.label) }}
+      </span>
+      <!-- {{ cluster.badge.iconText }} -->
```
richard-cox, line 77: debug line, is this nuked in the other pr?

shell/components/ClusterProviderIconMenu.vue
```diff
@@ -0,0 +1,161 @@
...
+    </div>
+    <!-- eslint-enable -->
```
richard-cox, line 79: this would only be needed if there was an `eslint-disable`. this section is going away though right?

shell/components/nav/TopLevelMenu.vue
```diff
@@ -726,18 +755,153 @@ export default {
...
+        span {
+          display: flex;
+          align-items: center;
+        }
+      }
+
       .clusters {
         overflow-y: auto;
-        overflow-x: hidden;
+        max-height: 35vh;
```
richard-cox, line 772: This introduces a void between clusters and categories. I don't think it's needed anymore?

image

shell/components/nav/TopLevelMenu.vue
```diff
@@ -57,6 +59,10 @@ export default {
       return this.features(LEGACY);
     },
 
+    emptyCluster() {
```
richard-cox, line 62: To avoid a `computed` property this can just be added to `data`

shell/layouts/default.vue
```diff
@@ -318,6 +318,14 @@ export default {
     overflow-y: auto;
     min-height: 0px;
 
+    &:has(.side-menu) {
+    padding-left: $app-bar-collapsed-width;
```
richard-cox, line 322: comment only. in theory having this in the default layout and not other layouts should break those other layouts. that doesn't seem to be the case though, still investigating

richard-cox (changes requested):
Visual / Layout
- ~The `Home` text is misaligned with the cluster and app names~ Now continued in review comment, with suggested fix
- With legacy turned on and an extension installed with with a menu item... there's a blip at about 750px window height where no clusters are visible. With legacy turned off (but still with the extension menu item) it does not happen


Still Remaining
- ~The home icon still looks misaligned with the burger menu and cluster abbreviations~
  - image

- ~For clusters with looong names, the cursor icon and popover flashes when moving the cursor over the provider square. The cursor icon flashes between it's correct state and the text cursor~ Now continued in review comment, with suggested fix

For discussion with team
- ~when the burger menu expands the branding icon is shown from within the burger menu. when it collapses the branding icon in the header is shown. these two aren't horizontally the same. this was the case before, but there was a transition to slowly fade in the burger menu branding icon. Not sure if that was a workaround, but it's a bit jarring now when expanding the menu and the branding icon (notionally) wiggles~
  - Additional notes - This is only an issue on the home page. the image when app bar is expanded aligns with the app bar content
- ~Vertical padding of horizontal (.side-menu .body .category-title) currently 15px.~
  
For resolution in the app pinning PR
- ~Rename of shell/components/ClusterProviderIconMenu.vue and removal of the cluster provider icon~

Comment Only
- In standalone mode the loading indicators are fine. The product icon is too far over to the left and the side nav is still 300px, but I'll sort out both of those once all related PRs have merged

shell/layouts/default.vue
richard-cox: ok, i think default is really root and other layouts sit inside. apart from blank, unauthed and standalone. blank and unauthed don't contain the app bar. standalone ui's will use the normal layouts. the standalone layout is new, and not associated with the standalone uis

shell/assets/styles/global/_tooltip.scss
```diff
@@ -5,7 +5,7 @@
 
   display: block !important;
   z-index: z-index('tooltip');
-  max-width: 50vw;
+  max-width: 90vw;
```
richard-cox, line 8: This isn't needed. The issue with long names is due to vtoolip (specifically the popper.js component under the hood) getting confused with it's boundary element and moving the position from right to left. There's another flag which hides the poppover on hover... which then removes the tooltip. This is the weirdness when the cursor changes from disabled to text (text because the popover is covering the cluster icon).

Hacky fix is in getTooltipConfig 

```
      return {
        content:       this.shown ? null : item,
        placement:     'right',
        popperOptions: {
          modifiers: { preventOverflow: { enabled: false } }
        }
      };
```

Proper fix is probably to set `boundariesElement` to the jquery element for body i guess, but lets avoid any jquery mallarky

richard-cox (changes requested):
Two items remaining in https://github.com/rancher/dashboard/pull/9560#pullrequestreview-1615066589. 

~Two review comments with suggested fixes~

richard-cox (approved):
Last remaining item in https://github.com/rancher/dashboard/pull/9560#pullrequestreview-1615066589 can be done in the pin pr

Other items for afterwards / other PR
- cow provider icon replaces `LCL` acronym 
- discussion - cluster list consuming all available space, not maxed at 10
- discussion - adding more colour / branding given removal of provider icon

PR #9570, opened 2023-08-19
Fix project edit UI crash

### LINKED ISSUE #9558: Rancher UI crashes when adding incomplete Project Resource Quotas
Internal reference: SURE-6733
Reported in 2.7.5

Issue description:
Rancher UI "crashes" when adding incomplete Project Resource Quotas. 

Business impact:
No real impact because you can go back and add the "missing" part

Troubleshooting steps:
Easy to reproduce 

Repro steps:

1. create a dummy project 
2. add a namespace to the project
3. Edit Project configuration --> Resource Quotas (there should be no quota listed at that time) 
4. add Resource, i.e. "CPU Limit" and only add a value for the Project Limit (no entry for the Namespace)  and press "Save". 

Result: Rancher UI "crashes" and the message " "Cannot read properties of null (reading 'limit')"  is getting displayed

Workaround:
Add values to all fields defined

Actual behavior:
Rancher UI "crashes"

Expected behavior:
Display an error message instead, i.e. "does not have all fields defined on a resourceQuota". This error message gets already displayed when you have already one Resource Quota defined and you re-edit the project configuration by adding a second, but incomplete set of resource quotas

jonatan5524:
### Summary
Fixes #9558

When there is no meaningful changes in the ResourceQuota objec the object is null and then the v-model object gets error when trying to get limit field from null

Local tested with chrome Version 116.0.5845.96 

### Screenshot/Video

[Screencast from 2023-08-19 13-46-33.webm](https://github.com/rancher/dashboard/assets/37303618/018bd6f6-4539-4c30-9f15-a7c0ccbebb7d)

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "start:dev": "NODE_ENV=dev yarn start",
     "start:prod": "NODE_OPTIONS=--max_old_space_size=4096 DEV_PORTS=true NODE_ENV=production yarn start",
     "generate": "yarn build",
-    "dev-debug": "node --inspect ./node_modules/.bin/vue-cli-service",
+    "dev-debug": "node --inspect ./node_modules/.bin/vue-cli-service serve",
     "cy:e2e": "cypress open --e2e --browser chrome",
     "cy:open": "cypress open",
     "cy:run": "cypress run --browser chrome",
--- a/shell/components/form/ResourceQuota/ProjectRow.vue
+++ b/shell/components/form/ResourceQuota/ProjectRow.vue
@@ -57,7 +57,7 @@ export default {
       @input="updateType($event)"
     />
     <UnitInput
-      v-model="value.spec.resourceQuota.limit[type]"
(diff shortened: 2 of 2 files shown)


richard-cox (changes requested):
Unfortunately these changes cause linting errors 
```
Optional chaining cannot appear in 'v-model' directives.eslint-plugin-vue
```

It's probably best to use a computed method for both, with a get that returns the x?.limit and a set that applies it (if it exists).

richard-cox (approved):
I've pushed a change, more detail in the commit comments.

@jonatan5524 Would you be able to take a quick look? if those seem ok we can merge (once passed CI)

jonatan5524:
> v-model was split in two to handle the bad input (missing `resourceQuota` / `namespaceDefaultResourceQuota`) and also the failure to assign limits following a failed save (empty limits set to null before the resource is saved). Just guarantees both safe input and assignment.

Got it! You can merge by me! Thank you :)

Issue #9573, opened 2023-08-22
GitreRepos correctDrift self healing - Cannot read properties of undefined (reading 'enabled')


Shavindra:
**Describe the bug**
GitRepo edit view throws `Cannot read properties of undefined (reading 'enabled')`

**To Reproduce**
1. Get Dashboard instance before the commit https://github.com/rancher/dashboard/pull/9310
2. Create a GitRepo
3. Update to the latest GitRepo
4. Navigate to edit view

**Result**
Throws `Cannot read properties of undefined (reading 'enabled')`

**Expected Result**
It shouldn't throw an error

**Screenshots**

**Additional context**
This is related to https://github.com/rancher/dashboard/pull/9310 . 

`applyDefaults()` Only apply to `_CREATE`
https://github.com/rancher/dashboard/blob/7d8a223eab88b23c3a956b9b985f22e396bfafca/shell/components/ResourceDetail/index.vue#L250-L252

Therefore when creating a new repo this would work. However, not when editing existing repo.

PR #9579, opened 2023-08-22
Handle correctdrift for previously created repos.

Shavindra:
### Summary
Fixes rancher/dashboard#9573 

### Technical notes summary
If `correctDrift === undefined` set `correctDrift.enabled` to `false`

1. Create a GitRepo with Dashboard < v2.8
2. Navigate to GitRepo > Edit

- GitRepo Create / Edit

### DIFF
--- a/shell/edit/fleet.cattle.io.gitrepo.vue
+++ b/shell/edit/fleet.cattle.io.gitrepo.vue
@@ -1,4 +1,5 @@
 <script>
+import Vue from 'vue';
 import { exceptionToErrorsArray } from '@shell/utils/error';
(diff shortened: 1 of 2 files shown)


richard-cox (changes requested):
i created a git repo in 2.7.5. I've then gone to the dashboard running this branch locally. The `Enable Self-Healing` option is unclickable.

Looks like this solves it

```
      Vue.set(this.value.spec, 'correctDrift', { });
      Vue.set(this.value.spec.correctDrift, 'enabled', false);
```

richard-cox (approved):

PR #9612, opened 2023-08-29
New cluster template isn't displayed after repository refresh

### LINKED ISSUE #9510: New cluster template isn't displayed after repository refresh
**Setup**
- Rancher version: 2.7.4, 2.7.5, 2.7.6-rc2
- Browser type & version: Chrome 114.0.5735.198

**Describe the bug**
When you add a new cluster template to a git repository and then refresh the repository in the rancher UI from Cluster Management->Advanced->Repositories the UI doesn't update with the new template until you do a hard refresh.

**To Reproduce**
1. Deploy rancher (versions above)
2. Add a repository to rancher
3. update the repository with a new template
4. refresh the repository from rancher without leaving cluster management (Cluster Management->Advanced->Repositories)
5. Create cluster.

NOTE: this bug doesn't show up if you refresh the repo from local->Apps->Repositories. 

**Result**
The new template will not be displayed.

**Expected Result**
The template will be displayed after the refresh occurs without needing a hard refresh of the page or any other page hopping.

aalves08:
### Summary
Fixes #9510 

- force fetch and reset of catalog/load in order to get updated cluster templates every time we hit refresh on the cluster repo

- Fork https://github.com/susesgartner/cluster-template-examples so that you can edit it
- **NOTE:** The below sequence should be without any browser refresh of the tab that shows your Dashboard
- Go to `Cluster Management` -> `Advanced` -> `Repositories` and add your fork as a repo (branch main)
- Go `Cluster Management` -> `Clusters` and hit `Create`
- Check that on the options for cluster creation you have 2 cluster templates (if you haven't modified your forked repo)
- Go to `Cluster Management` -> `Advanced` -> `Repositories` and stay on that screen
- In a new tab, go to your fork of https://github.com/susesgartner/cluster-template-examples and edit the `index.yaml` file to only show `cluster-template1 `
- Go back to the tab where you are on the `Repositories` view of Dashboard and go to the 3 dots menu for your repo/fork and select the option `refresh`
- Go `Cluster Management` -> `Clusters` and hit `Create`
- Check that on the options for cluster creation you have 1 cluster template (you should only see `cluster-template1`, as per the edit of `index.yaml`)

### Screenshot/Video

https://github.com/rancher/dashboard/assets/97888974/81a9c32a-2aaa-42b7-bf98-57455f46eb4a

### DIFF
--- a/shell/edit/provisioning.cattle.io.cluster/index.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/index.vue
@@ -83,7 +83,9 @@ export default {
       mgmtClusters: this.$store.dispatch('management/findAll', { type: MANAGEMENT.CLUSTER }),
       provClusters: this.$store.dispatch('management/findAll', { type: CAPI.RANCHER_CLUSTER }),
 
-      catalog: this.$store.dispatch('catalog/load'),
+      // have force: true and reset: true because of cluster templates, which can be changed and updated via repo refresh
+      // https://github.com/rancher/dashboard/issues/9510
+      catalog: this.$store.dispatch('catalog/load', { force: true, reset: true }),
     };
 
     if (this.$store.getters[`management/canList`](MANAGEMENT.NODE_DRIVER)) {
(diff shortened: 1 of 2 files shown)


richard-cox (changes requested):
We have a general problem re-fetching the whole catalogue at places where we don't strictly need to. Here we'll block every edit of cluster. Is there a place it can go where we would expect the catalogue to change that's not in the normal flow of using the dashboard?

richard-cox (approved):

Issue #9009, opened 2022-11-23
When Adding SSH Public and Private keys via the GIT Repo Config page, SSH key ends up as null.


rfutterman91:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I am trying to connect my AWS Code Commit repo to Rancher Fleet GIT Repos. I enter in my public and private keys in OPENSSH format. Once I save and look back at the config it shows SSH Key: Null

This is the Error I see on the Git Repo dashboard:

Git ls-remote  ssh://git-codecommit.us-gov-west-1.amazonaws.com/v1/repos/linc-noss-fleet  refs/heads/main error: exit status 128, detail:  gitjob@git-codecommit.us-gov-west-1.amazonaws.com: Permission denied  (publickey).  fatal: Could not read from remote repository.  Please make sure you have the correct access rights and the repository exists.

the repository exists as I am copying it from AWS CodeCommit

### Expected Behavior

SSH Keys to be generated properly

### Steps To Reproduce

1. Continuous Delivery > Git repo > Add Repository
2. Use SSH keys generated from AWS Code Commit in OpenSSH format
3. Save 
4. Look at the config for the repo and notice 'SSH Key: null'

### Environment

```markdown
- Architecture:
- Fleet Version:
- Cluster:
  - Provider:
  - Options:
  - Kubernetes Version:
```

sbulage:
I am able to see the secrets created via `GitRepo` UI page, already created secrets (`http/ssh`) both are populated. Also, used them in order to deploy applications from private GitHub repository.

![Selection_103](https://github.com/rancher/dashboard/assets/12134754/34a6d1d4-0deb-40d5-8405-448b3886d95a)

PR #9624, opened 2023-08-30
Handle missing secret name.

Shavindra:
### Summary
Fixes rancher/dashboard#9009 

- Change the displayed text when the secret name is missing

- Navigate to Continuous Delivery and create a Git Repo with SSH key with following format for public key `ssh-rsa JR3lV1JzQbncvoVkWSYL= test@test.localhost` and private key `6G64vPAW5L`
- **NB:** it doesn't need to be a real working Private / Public key. You can just copy past text.
- Click Edit GitRepo
- Update the Git repo by creating a new Git Repo with  following format for public key `ssh-rsa JR3lV1JzQbncvoVkWSYL= ` and private key `6G64vPAW5L`
- Go back to edit again and it should display the credential properly without null

- Git Repo Create Helm/Git credentials dropdown

### Screenshot/Video
image

### DIFF
--- a/shell/components/form/SelectOrCreateAuthSecret.vue
+++ b/shell/components/form/SelectOrCreateAuthSecret.vue
@@ -230,7 +230,7 @@ export default {
             dataPreview, subTypeDisplay, metadata, id
           } = x;
 
-          const label = subTypeDisplay && dataPreview ? `${ metadata.name } (${ x.subTypeDisplay }: ${ dataPreview })` : `${ metadata.name } (${ x.subTypeDisplay })`;
+          const label = subTypeDisplay && dataPreview ? `${ metadata.name } (${ subTypeDisplay }: ${ dataPreview })` : `${ metadata.name } (${ subTypeDisplay })`;
 
           return {
             label,


shell/components/form/SelectOrCreateAuthSecret.vue
```diff
@@ -226,10 +226,16 @@ export default {
 
           return true;
         }).map((x) => {
+          const {
+            dataPreview, subTypeDisplay, metadata, id
+          } = x;
+
+          const label = subTypeDisplay && dataPreview ? `${ metadata.name } (${ x.subTypeDisplay }: ${ dataPreview })` : `${ metadata.name } (${ x.subTypeDisplay })`;
```
aalves08, line 233: A bit nitpicky, but it looks cleaner if you use the variable `subTypeDisplay`:

```
const label = subTypeDisplay && dataPreview ? `${ metadata.name } (${ subTypeDisplay }: ${ dataPreview })` : `${ metadata.name } (${ subTypeDisplay })`;
```

aalves08 (commented):
Only a very small comment codewise. Testing is fine! 

Overall, just the remark that this is not to be QA'd at this point, which means that we should create some testing for this. I believe that a simple unit test on this component to cover this scenario over possible regressions should suffice.

aalves08 (approved):
typo fixed. Same general comment stands, but I'll approve it

PR #9628, opened 2023-08-30
Fix correctDrift banner typo.

Shavindra:
Fixing a small typo


richard-cox (approved):
Make sure to check / update state in `master` as well

PR #9643, opened 2023-09-01
Aws In-tree support

### LINKED ISSUE #9110: Ability to deploy AWS external tree provider in RKE2
Internal Ref: SURE-4760

The external AWS cloud provider (CCM) deploys onto a cluster and relies on the hostname and/or --provider-id flag (kubelet) to describe the node and initialise into the cluster.

With v2-provisioning, the hostname is always set to a node pool naming convention (OS (see userdata example below) & with -hostname-override), and there appears to be no way of providing the values dynamically to the -provider-id flag.

### LINKED ISSUE #9204: Remove Support for AWS In-Tree Cloud Provider if Using k8s 1.27+
Internal reference: SURE-6511

With Kubernetes 1.27 and up, the AWS in-tree cloud provider no longer works so we need to ensure that users upgrading from earlier versions are aware and have a way to migrate off of it.

From Hostbusters, @kinarashah is looking at how the upgrade works from the backend. @Sahota1225 & @snasovich are aware of the UI needs. We'll need to figure out the UX around this with the appropriate advance notifications prior to upgrade and then the behaviour during upgrade: having radio buttons disabled, proper dropdowns, banners, etc.

@mantis-toboggan-md knows the basic details.

nwmac:
Fixes #9110 
Fixes #9204 

Create an AWS cluster and select the AWS Cloud Provider.

Edit the cluster and change the k8s version to 1.27 - we now change the provider to 'External' and show a banner. If you change back to 1.26, we revert to the AWS Cloud provider.

Note: Added some test code (commented out) that can be un-commented to add a fake 1.27 version for dev/testing.

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -1846,6 +1846,7 @@ cluster:
       header: Cloud Provider Config
       defaultValue:
         label: Default - RKE2 Embedded
+      unsupported: The current Cloud Provider is not supported by this version of Kubernetes. The Cloud Provider has been changed to External. Please use the Cloud Provider Config to supply an out-of-tree configuration as needed.
     security:
       header: Security
     cis:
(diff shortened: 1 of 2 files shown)


mantis-toboggan-md (approved):
lgtm - updated to remove test code and fix merge conflict

nwmac:
Thanks @mantis-toboggan-md  - I recall we said we'd add link to the docs in the banner, but I don't have that and will need to chase with @gaktive - we can add that in another PR.

PR #9654, opened 2023-09-05
Fix: Missing of spacing to min-width in pagination area

ChrisEdS:
### Summary
Fixes https://github.com/rancher/dashboard/issues/9653

In some cases the minimum width of the pagination text (200px) may not be sufficient. In case of wider text, the text will directly hit the navigation elements. Minor CSS adjustments in this PR fixes the behaviour.

Seen at Harvester Dashboard under Advanced -> PCI Devices (when enough devices are added to trigger the pagination)

Not expected (fix only applies to pagination area. The view has not changed when the pagination text is inside the min-width of 200px)

### Screenshot/Video
Before:
Screenshot 2023-09-04 at 18 12 02

After:
Screenshot 2023-09-04 at 18 11 45

### DIFF
--- a/shell/components/SortableTable/index.vue
+++ b/shell/components/SortableTable/index.vue
@@ -1933,12 +1933,13 @@ export default {
   }
 
   .paging {
-    margin-top: 10px;
+    margin-top: 20px;
     text-align: center;
 
     SPAN {
       display: inline-block;
       min-width: 200px;
+      padding: 0 20px 0 20px;
     }
   }
   </style>


cnotv:
@kwwii This change requires design approval before proceed.

I am also not able to reproduce the issue globally component wise, so it may be worth to review the specific issue itself.
Another alternative is to have the pagination component abstracted and added to Storybook as well.

For comparison:

Event log:

Screenshot 2023-09-06 at 08 16 05

Storybook:

Screenshot 2023-09-06 at 08 20 58

richard-cox:
@cnotv i thought the key is a super long resource name, and also maybe a high number of resources. also the border shown for suse / prime branding. however we do show a small gap

image

there could be some styling in the harvester side that is pushing them close together. @ChrisEdS can you confirm you see this in the latest version of Rancher?

ChrisEdS:
So since I changed the code in main branch I'd bet that it will be in the latest version as well. I will play around with Rancher deployments this week anyway. Let me confirm it afterwards. It could be that in Rancher (without extensions) there is no text or number combination which triggers this behaviour, though.

ChrisEdS:
> image

Ah, now I see what you mean, indeed - there is a gap, really small but still.

richard-cox:
> > image
> 
> Ah, now I see what you mean, indeed - there is a gap, really small but still.

Great, can we close this PR & issue?

ChrisEdS:
> > > image
> > 
> > 
> > Ah, now I see what you mean, indeed - there is a gap, really small but still.
> 
> Great, can we close this PR & issue?

Can you confirm that this view is what we get when my PR is applied? If yes, then it is worthless, if no I'd vote for merging since the view (the gap) looks also squeezed

cnotv:
Again, this has to be approved by a designer (@kwwii @edenhernandez-suse @manuelecarlini ) and it has to be verified globally.
In Storybook this does not seems to break with default text, so I imagine is because we are allowing custom text there
https://rancher.github.io/storybook/?path=/story/components-table--pagination

This is how it looks adopting the text:
Screenshot 2023-09-28 at 12 48 53

This with your fix:
Screenshot 2023-09-28 at 12 49 45

ChrisEdS:
Got it. Closed from my side

PR #9662, opened 2023-09-07
Reset network property when region is updated

### LINKED ISSUE #9636: [RFE] - request to clear out VPC/Subnet value (if configured) when changing the Region field for AWS Node driver v2prov clusters
When provisioning a downstream rke2/k3s AWS node driver cluster, if a VPC/Subnet is selected (for a particular region), then the Region field is changed, the VPC/Subnet value persists, but is now invalid, as the region has changed.

Ideally, when the Region field is changed, the VPC/Subnet value (if configured) would be cleared out, allowing the user to re-select a value from the VPC/Subnets dropdown

An alternative solution could be to add form validation to prevent the creation of downstream v2prov AWS Node driver clusters, when the selected VPC/Subnet is not found in the desired Region selected (in the Machine Pools configurations)

**Additional context**
Without this enhancement, it is easier for users to get into a situation where their cluster won't provision, as they might expect.

jordojordo:
### Summary
Fixes #9636 

The `value.subnetId`, `value.vpcId`, and `selectedNetwork` properties were not being updated/reset when the region was updated. 

Added a call to the `updateNetwork` method within the `watch` for `value.regions`, this will reset the `value.subnetId`, `value.vpcId`, and `selectedNetwork` properties to the original `null` value.

### Technical notes summary

I decided adding in the form validation as suggested in the original issue would over-complicate the machine-config when simply resetting the value would suffice.

### DIFF
--- a/shell/machine-config/amazonec2.vue
+++ b/shell/machine-config/amazonec2.vue
@@ -303,6 +303,7 @@ export default {
     },
 
     'value.region'() {
+      this.updateNetwork();
       this.$fetch();
     },
 


momesgin (approved):
LGTM

PR #9684, opened 2023-09-12
Remove v-popover manual trigger for userMenu.

Shavindra:
### Summary
Fixes rancher/dashboard#8975 

- Removed manual trigger on the user menu popover
- Added separate test on click and view of the user menu items

- Ran the E2E multiple times

- User menu navigation - This should still work as it is.


nwmac:
@richard-cox Is this needed? I think you added some robustness to the tests that makes this obsolete?

PR #9689, opened 2023-09-12
Reduce performance impact of LabeledSelect getOptionLabel

### LINKED ISSUE #9668: LabeledSelect performs poorly with >20 item lists
Seen during rancher/dashboard#8965 , where there are ~800 vm sizes available in the default region.

This appears to be a consequence of changes here https://github.com/rancher/dashboard/pull/8154/files - `getOptionLabel` runs for each option, so in the azure aks case we're iterating over 800 options 800 times...

Any fix here needs to ensure https://github.com/rancher/dashboard/issues/8149 isn't reintroduced, particularly the behavior seen in https://github.com/rancher/dashboard/pull/8133

mantis-toboggan-md:
### Summary
Fixes #9668 

Two things contributing to this:
1. `getOptionLabel` iterates over every option for each option, to address https://github.com/rancher/dashboard/issues/8149
2. `getOptionLabel` is called _a lot_ 

By adding a console.log to `getOptionLabel` I can see that the function re-runs for each option when the user mouses over options. Even providing the minimum necessary configuration to v-select I see this happening, which makes me think it is baked into the component library we are using.

This PR updates `getOptionLabel` to skip the label-checking logic when options don't have separate labels and values. This means that the LabeledSelect performance with very long lists of options is _only_ fixed if the options prop is an array of strings or numbers. I couldn't work out how else to address #9668 without re-introducing https://github.com/rancher/dashboard/issues/8149 because the matter of `getOptionLabel` being called over 9000 times for even the shortest of options arrays seems to be out of our hands.

The bug seen in https://github.com/rancher/dashboard/pull/8133 was the original motivation for the `isOutdated` check and has not been re-introduced by this change, as verified by rke2 unit tests and manual testing.

This bug was seen in code not yet merged so to test, alter any instance of LabeledSelect to have a lot of options: `:options="Array.from(Array(800).keys())"` and verify opening and scrolling the list of options isn't laggy. 

PSA version-dependent default labeling should be verified as still working too.

### Screenshot/Video
old:

https://github.com/rancher/dashboard/assets/42977925/0ee27c4d-81a9-4866-a39d-96c549f2fe90

new:

https://github.com/rancher/dashboard/assets/42977925/b06419a4-d219-49a9-bce8-597cbd4f2591

### DIFF
--- a/shell/components/form/LabeledSelect.vue
+++ b/shell/components/form/LabeledSelect.vue
@@ -155,11 +155,9 @@ export default {
         const newOption = this.getUpdatedOption(option);
 
         if (newOption) {
-          if (newOption) {
-            const label = get(newOption, this.optionLabel);
+          const label = get(newOption, this.optionLabel);
 
-            return this.localizedLabel ? this.$store.getters['i18n/t'](label) || label : label;
-          }
+          return this.localizedLabel ? this.$store.getters['i18n/t'](label) || label : label;
         }
       }
 
--- a/shell/components/form/LabeledSelect.vue
+++ b/shell/components/form/LabeledSelect.vue
@@ -180,14 +180,9 @@ export default {
     // If the option's label changed in parent but value did not, the label wont be automatically updated here
     // Ensure that the label being shown is still present in the options prop and find the new one if not
     getUpdatedOption(option) {
-      let out;
-      const isOutdated = !this.options.find((opt) => option[this.optionLabel] === opt[this.optionLabel]);
+      const isOutdated = this.options && !this.options.find((opt) => option[this.optionLabel] === opt[this.optionLabel]);
 
-      if (isOutdated && this.options) {
-        out = this.options.find((opt) => isEqual(this.reduce(option), this.reduce(opt)));
-      }
-
-      return out;
+      return isOutdated ? this.options.find((opt) => isEqual(this.reduce(option), this.reduce(opt))) : undefined;
     },
 
     positionDropdown(dropdownList, component, { width }) {
--- a/shell/components/form/LabeledSelect.vue
+++ b/shell/components/form/LabeledSelect.vue
@@ -149,16 +149,20 @@ export default {
         return;
       }
 
-      // Force to update the option label if prop has been changed
-      const isOutdated = !this.options.find((opt) => option[this.optionLabel] === opt[this.optionLabel]);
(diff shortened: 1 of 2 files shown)


cnotv:
Thanks for the videos :)

mantis-toboggan-md:
I've refactored `getOptionLabel` to test that the outdated option label checking logic is not called when options are strings. #8149 is tested by existing LabeledSelect unit tests and #8133 is tested by existing rke2 unit tests.

shell/components/form/LabeledSelect.vue
```diff
@@ -149,16 +149,17 @@ export default {
...
-      const isOutdated = !this.options.find((opt) => option[this.optionLabel] === opt[this.optionLabel]);
-
-      if (isOutdated && this.options) {
-        const newOption = this.options.find((opt) => isEqual(this.reduce(option), this.reduce(opt)));
+      // This check is only needed if its possible for an option's label to change without the option's value changing - we can skip this if options are just strings or numbers
+      // HOWEVER even if strings are passed to v-select the 'option' in the slot is normalized to {label: <option>} so we have to check the options prop here instead of the 'option' itself
+      if (typeof this.options[0] === 'object') {
+        const newOption = this.getUpdatedOption(option);
 
         if (newOption) {
```
cnotv, line 157: I think there's a typo here during the refactoring, because it's repeated twice

shell/components/form/LabeledSelect.vue
```diff
@@ -178,6 +179,19 @@ export default {
...
+    getUpdatedOption(option) {
+      let out;
+      const isOutdated = !this.options.find((opt) => option[this.optionLabel] === opt[this.optionLabel]);
+
+      if (isOutdated && this.options) {
+        out = this.options.find((opt) => isEqual(this.reduce(option), this.reduce(opt)));
+      }
+
+      return out;
+    },
```
cnotv, line 191: Since we not reassign it, would be more convenient to use a constant, otherwise it may confuse.

Also if we check `this.options`, we should do it before find out about outdated values.

```ts
      const isOutdated = this.options && !this.options.find((opt) => option[this.optionLabel] === opt[this.optionLabel]);

      return isOutdated
         ? this.options.find((opt) => isEqual(this.reduce(option), this.reduce(opt)))
         : undefined;
```

mantis-toboggan-md: Makes sense, done

cnotv (approved):
LGTM

PR #9722, opened 2023-09-15
Removed duplicted error banner when adding user to the cluster

### LINKED ISSUE #8593: Errors during cluster role template binding creation are rendered twice
The cluster role template binding edit component is using cruresource to display errors, as well as including its own error banner, leading to the same error being rendered twice:

Side note: the issue implied in the error message above, where a user who can't add cluster members is shown the option to do so anyway, is tracked in https://github.com/rancher/dashboard/issues/8594

eva-vashkevich:
### Summary
Fixes #8593
management.cattle.io.clusterroletemplatebinding had an error banner,however it is using CruResource that already handles errors. Duplicated banner was redundant since they display the same array of errors.

Removed banner from management.cattle.io.clusterroletemplatebinding component

### Technical notes summary
N/A

Test that the error is still displayed when adding user fails.

N/A

### Screenshot/Video
D0CD508D-EE2C-447D-8C97-B3E66FC85F66_1_201_a


mantis-toboggan-md (approved):
lgtm

PR #9760, opened 2023-09-22
node to management node controller remaps status into internalNodeStatus

Sean-McQ:
Code that handles this on the backend: https://github.com/rancher/rancher/blob/9129a7b0496c4c2a5895118595c37da5097c8aca/pkg/controllers/managementuser/nodesyncer/nodessyncer.go#L750

### Summary
Fixes #


cnotv:
Sorry but I see just a change from scoped variable to method `addresses()`, how is this supposed to fix the issue?

Sean-McQ:
@cnotv So I added the `addresses` getter to both `node` and `management.cattle.io.node` to make things more consistent between the two but crucially, you'll notice that the getter in `management.cattle.ui.node` now looks at `this.status?.internalNodeStatus?.addresses` if it can't find anything at `this.status?.addresses` because Steve remaps the `status` on nodes into `status.internalNodeStatus` before packaging it into the rancher specific type for nodes (`management.cattle.io.node`).

cnotv:
> addresses

Could we write a test then, something like this?

```ts
import MgmtNode from '@shell/models/management.cattle.io.node';

describe('class MgmtNode', () => {
  describe('should always return addresses', () => {
    it.each([
      { addresses: [] },
      { internalNodeStatus: { addresses: [] } },
      undefined
    ])('given %p', (status) => {
      const addresses = [];
      const mgmtNode = new MgmtNode({});

      mgmtNode.status = status;

      expect(mgmtNode.addresses).toStrictEqual(addresses);
    });
  });
});
```

Sean-McQ:
"always" return is a bit of a strong word, especially considering we're getting values back from an API so testing thoroughly makes a ton of sense. I went ahead and wrote a few quick and simple tests for the addresses and internal/external ip getters on the models. Covered both the new cases I added and also handled the older ones including the negative cases.

shell/models/__tests__/management.cattle.io.node.ts
```diff
@@ -0,0 +1,82 @@
+import MgmtNode from '@shell/models/management.cattle.io.node';
+
+describe('class MgmtNode', () => {
+  const t = jest.fn(() => 'baz');
+  const ctx = { rootGetters: { 'i18n/t': t } };
+
+  const resetMocks = () => {
+    // Clear all mock function calls:
+    jest.clearAllMocks();
```
cnotv, line 9: Jest has an hook for before/after where you can reset the mocks. This is definitely not the way to do it.

Also may I ask you what is the reason to reset them? It should be sufficient what is defined globally in the Jest setup.

shell/models/__tests__/management.cattle.io.node.ts
```diff
@@ -0,0 +1,82 @@
...
+  };
+
+  it('should not return addresses if they are not present in the resource status, the internalNodeStatus, or the rkeNode key in status', () => {
+    const mgmtNode = new MgmtNode({ status: {} }, ctx);
+
+    expect(mgmtNode.addresses).toStrictEqual([]);
+    resetMocks();
+  });
+  it('should return addresses if they are present directly on the resource status', () => {
```
cnotv, line 18: You have 2 tests where you repeat `should return addresses if they`. It should be contained in the same description having this part of text and remove it from the test. It will helps it also to be conciser.

shell/models/__tests__/management.cattle.io.node.ts
```diff
@@ -0,0 +1,82 @@
...
+    expect(mgmtNode.addresses).toStrictEqual(['foo']);
+    resetMocks();
+  });
+  it('should return addresses if they are not present directly on the resource status but are on "status.internalNodeStatus"', () => {
+    const mgmtNode = new MgmtNode({ status: { internalNodeStatus: { addresses: ['bar'] } } }, ctx);
+
+    expect(mgmtNode.addresses).toStrictEqual(['bar']);
+    resetMocks();
+  });
+  it('should return an internalIp if addresses includes an object with an appropriate type and address', () => {
```
cnotv, line 30: You have 4 tests where you repeat `should return an internalIp if`. It should be contained in the same description having this part of text and remove it from the test. It will helps it also to be conciser.

shell/models/__tests__/management.cattle.io.node.ts
```diff
@@ -0,0 +1,82 @@
...
+    const mgmtNode = new MgmtNode({ status: { addresses: [{ type: 'InternalIP', address: 'foo' }] } }, ctx);
+
+    expect(mgmtNode.internalIp).toStrictEqual('foo');
+    resetMocks();
+  });
+  it('should return an externalIp if addresses includes an object with an appropriate type and address', () => {
```
cnotv, line 36: You have 4 tests where you repeat `should return an externalIp if`. It should be contained in the same description having this part of text and remove it from the test. It will helps it also to be conciser.

shell/models/__tests__/management.cattle.io.node.ts
```diff
@@ -0,0 +1,82 @@
...
+  };
+
+  it('should not return addresses if they are not present in the resource status, the internalNodeStatus, or the rkeNode key in status', () => {
+    const mgmtNode = new MgmtNode({ status: {} }, ctx);
+
+    expect(mgmtNode.addresses).toStrictEqual([]);
+    resetMocks();
+  });
+  it('should return addresses if they are present directly on the resource status', () => {
+    const mgmtNode = new MgmtNode({ status: { addresses: ['foo'] } }, ctx);
```
cnotv, line 19: Repeated value `['foo']` should be a constant to increase readability. Please consider all the other cases too, so I don't have to go one by one.

shell/models/__tests__/management.cattle.io.node.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    expect(mgmtNode.externalIp).toStrictEqual('bar');
+    resetMocks();
+  });
+  it('should return an appropriate message if there is no internalIp to display', () => {
+    const mgmtNode = new MgmtNode({ status: {} }, ctx);
+
+    expect(mgmtNode.internalIp).toStrictEqual('baz');
+    expect(t).toHaveBeenCalledTimes(1);
+    expect(t).toHaveBeenCalledWith('generic.none');
```
cnotv, line 71: Where is `generic.none` coming from?
It also should be a separated test, so it would explain what is that for and what you are trying to test. Currently is not clear.

shell/models/__tests__/management.cattle.io.node.ts
```diff
@@ -0,0 +1,82 @@
...
+    expect(mgmtNode.externalIp).toStrictEqual('bar');
+    resetMocks();
+  });
+  it('should return an internalIp if internalNodeStatus.addresses includes an object with an appropriate type and address', () => {
+    const mgmtNode = new MgmtNode({ status: { internalNodeStatus: { addresses: [{ type: 'InternalIP', address: 'foo' }] } } }, ctx);
+
+    expect(mgmtNode.internalIp).toStrictEqual('foo');
+    resetMocks();
+  });
+  it('should return an externalIp if internalNodeStatus.addresses includes an object with an appropriate type and address', () => {
```
cnotv, line 48: This test can be parametrized.

shell/models/__tests__/management.cattle.io.node.ts
```diff
@@ -0,0 +1,82 @@
...
+  });
+  it('should return an appropriate message if there is no internalIp to display', () => {
+    const mgmtNode = new MgmtNode({ status: {} }, ctx);
+
+    expect(mgmtNode.internalIp).toStrictEqual('baz');
+    expect(t).toHaveBeenCalledTimes(1);
+    expect(t).toHaveBeenCalledWith('generic.none');
+    resetMocks();
+  });
+  it('should return an appropriate message if there is no externalIp to display', () => {
```
cnotv, line 74: This test can be parametrized.

shell/models/__tests__/node.ts
```diff
@@ -0,0 +1,52 @@
+import Node from '@shell/models/management.cattle.io.node';
+
+describe('class MgmtNode', () => {
```
cnotv, line 3: As far as I can see, there's a lot of duplicated logic. Not entirely sure how this is handled in our OOP architecture, but I guess we may have a subclass to be extended for both the models and have 1 test suite there?

cnotv (changes requested):
There's some remarks about the tests.

cnotv:
> "always" return is a bit of a strong word, especially considering we're getting values back from an API so testing thoroughly makes a ton of sense. I went ahead and wrote a few quick and simple tests for the addresses and internal/external ip getters on the models. Covered both the new cases I added and also handled the older ones including the negative cases.

"Always" in the pure sense of type signature of the function returning always an `Array`. I agree to define API cases though, it makes more clear what we have no control of.

shell/models/__tests__/management.cattle.io.node.ts
```diff
@@ -0,0 +1,82 @@
...
+  const t = jest.fn(() => 'baz');
+  const ctx = { rootGetters: { 'i18n/t': t } };
+
+  const resetMocks = () => {
+    // Clear all mock function calls:
+    jest.clearAllMocks();
+  };
+
+  it('should not return addresses if they are not present in the resource status, the internalNodeStatus, or the rkeNode key in status', () => {
+    const mgmtNode = new MgmtNode({ status: {} }, ctx);
```
cnotv, line 13: This should be `undefined` based on the code.

shell/models/__tests__/management.cattle.io.node.ts
Sean-McQ: You're talking about the beforeEach and afterEach hooks for Jest. I try not to use those unless I have to if for no other reason than because it breaks a linting rule, if using it is a best practice for us then we should probably remove that rule.

This method is pretty much the way described in the eslint-plugin-jest docs concerning this rule and also mirrors another implementation in the code found in `shell/components/nav/WindowManager/__tests__/ContainerShell.test.ts` where it's used for the same purpose.

In my case, I'm resetting the mocks between tests both to keep each test isolated but also because I'm asserting that the `t` getter is only called once when needed. Strictly speaking, I don't need to pass in CTX and/or reset the mock for most of these tests but it was part of my boilerplate for creating new tests.

cnotv:
Since I've seen a lot of my comments are usually part of linters, do we want to integrate rules for them at least in the tests?
[This is for the string](https://github.com/SonarSource/eslint-plugin-sonarjs/blob/master/docs/rules/no-duplicate-string.md).
If we want to integrate code smell, I used [rules from SonarQube](https://github.com/SonarSource/eslint-plugin-sonarjs) which detects also function duplication.

shell/models/__tests__/management.cattle.io.node.ts
cnotv: I personally have found that every time I have to place it in a `beforeEach` I am doing something not correct. We should review this if we find more valid cases.

Specifically in your case you have the CTX called multiple times, because it should be instanciated instead of using the same value.

cnotv: I have not reviewed `ContainerShell`, so I cannot talk about it here.

shell/models/__tests__/management.cattle.io.node.ts
Sean-McQ: Done

shell/models/__tests__/management.cattle.io.node.ts
Sean-McQ: Done

shell/models/__tests__/management.cattle.io.node.ts
Sean-McQ: Done

shell/models/__tests__/management.cattle.io.node.ts
Sean-McQ: I feel like this is perhaps gilding the lily but it's a simple change so your wish is my command.

For reference, you never have to go one by one on these, if you mention that something is a problem and I go about fixing it, I'll change it everywhere I think it's relevant.

shell/models/__tests__/management.cattle.io.node.ts
Sean-McQ: `generic.none` is coming from the model code itself, it's the translation string that gets fed to `t` if the model getter for internalIp or externalIp don't return an address. This code is part of this model so it belongs in this test suite. This test doesn't touch the external method `t` but ensures that the code in the model is calling it with the expected params given the right circumstances which is exactly what these unit tests are for.

shell/models/__tests__/management.cattle.io.node.ts
Sean-McQ: It could be and if there were enough permutations to it, I would parameterize it and run several variables across eachother but in this case, there's not too many use cases, only a single variable, and most of the tests are incredibly simple so the added overhead of parameterizing the test and looping through test conditions seems like it would do more harm than good from a readability perspective and towards expanding this test suite for the remaining model.

shell/models/__tests__/management.cattle.io.node.ts
Sean-McQ: It could be and if there were enough permutations to it, I would parameterize it and run several variables across eachother but in this case, there's not too many use cases, only a single variable, and most of the tests are incredibly simple so the added overhead of parameterizing the test and looping through test conditions seems like it would do more harm than good from a readability perspective and towards expanding this test suite for the remaining model.

shell/models/__tests__/node.ts
Sean-McQ: Duplicated logic isn't a major issue here in this unit test land where we want our tests to be isolated and deterministic. Centralizing logic in tests can save you a lot of work when you've got confounding variables in your test suite but doing so needlessly can lead to mistakes such as an error in the central logic causing all tests to fail without indicating why in an obvious way.

shell/models/__tests__/management.cattle.io.node.ts
Sean-McQ: actually, it should return an empty array which is what it's doing and what I'm testing for: 
image

Sean-McQ:
My general thought is that we tend to get a bit too pedantic about code hygiene, in particular in tests which aren't actually production code. Obviously readability is important but the line between "this isn't up to standard" and "this isn't how I would write it" gets confused sometimes. I'm good with keeping the linters the way they are, enforcing their rules as-is, and relying on them to define the standard of code for the codebase.

cnotv:
> My general thought is that we tend to get a bit too pedantic about code hygiene, in particular in tests which aren't actually production code. Obviously readability is important but the line between "this isn't up to standard" and "this isn't how I would write it" gets confused sometimes. I'm good with keeping the linters the way they are, enforcing their rules as-is, and relying on them to define the standard of code for the codebase.

As we have no linting rules, no Typescript, no test coverage, I dunno how can this be pedantic, speaking frankly. I also have never enforced any pattern or standard, so not so sure what are we talking about.
I never said "this isn't how I would write it", rather give a hint and you are free to agree or not, as it happened in the past.
Just don't be surprised if also other people don't get it.

If you prefer me to just catch possible errors and that's it, we could do that, because it's more important to don't impair the work.

cnotv (approved):
LGTM

PR #9761, opened 2023-09-22
checks to see if membership editor is project and adds a project filter

### LINKED ISSUE #8574: Cannot load 'Members' page of Project in Rancher UI

Sean-McQ:
### Summary
Fixes #8574 
Progresses #9341

This looks to be primarily a performance issue wherein loading a project edit page with a 'members' list (actually the `projectRoleTemplateBindings` resource) can either simply fail or get stuck in a repeating pattern at scale due to the way that `projectRoleTemplateBindings` scales across projects and clusters. Adding a project filter to the `projectRoleTemplateBindings` should cut down on the cost of this API call significantly.

The membership editor will now check if it's type is set to Norman's `projectRoleTemplateBindings` and will append a `projectId` field (Norman specific, Steve's `management.cattle.io.projectRoleTemplateBindings` uses `projectName` but is not used on this page.

While not strictly required by this PR, changes have been added to better support filtering and Sorting for Steve.

Changed out `uriEncodeComponent` for `uriEncode` since `uriEncodeComponent` was encoding characters that it shouldn't such as ":". Tested thoroughly and all existing tests still pass. New tests added to cover Steve specific url options.

### Technical notes summary
Please refer to the ReadMe on https://github.com/rancher/steve to verify URL options correctness for Steve. Specific Correctness of the "projectId" field for Norman's `projectRoleTemplateBindings` can be found on the collection itself under the `filters` key and the API call can also always be called directly or inspected in the network tab of the browser's developer tools to validate correctness. The actual spec used for Norman's query options can be found [here](https://github.com/rancher/api-spec/blob/master/specification.md#filtering).

This is primarily a performance enhancement, if you'd like to actually see the specific performance impact, you'll want to create a rancher instance with a downstream cluster using ~400 projects and ~150 users on a single-node rancher instance but for the most part what is needed is just simple functional validation. The members list does load, it shows the correct members, no broken API calls, etc...

No Steve API calls were using filters and/or sorting in them so nothing should break there, all existing unit tests on urlOptions still pass so I'm not assuming any regressions there either.

### DIFF
--- a/shell/components/form/Members/MembershipEditor.vue
+++ b/shell/components/form/Members/MembershipEditor.vue
@@ -56,8 +56,13 @@ export default {
   },
 
   async fetch() {
+    const roleBindingRequestParams = { type: this.type, opt: { force: true } };
+
+    if (this.type === NORMAN.PROJECT_ROLE_TEMPLATE_BINDING && this.parentId) {
+      Object.assign(roleBindingRequestParams, { opt: { filter: { projectId: this.parentId.split('/').join(':') } } });
+    }
     const userHydration = [
-      this.schema ? this.$store.dispatch(`rancher/findAll`, { type: this.type, opt: { force: true } }) : [],
+      this.schema ? this.$store.dispatch(`rancher/findAll`, roleBindingRequestParams) : [],
       this.$store.dispatch('rancher/findAll', { type: NORMAN.PRINCIPAL }),
       this.$store.dispatch(`management/findAll`, { type: MANAGEMENT.ROLE_TEMPLATE }),
       this.$store.dispatch(`management/findAll`, { type: MANAGEMENT.USER })
--- a/shell/plugins/steve/__tests__/getters.spec.ts
+++ b/shell/plugins/steve/__tests__/getters.spec.ts
@@ -68,8 +68,14 @@ describe('steve: getters', () => {
     it('returns a string with a single filter statement applied if a single filter statement is applied', () => {
       expect(urlOptionsGetter('foo', { filter: { bar: 'baz' } })).toBe('foo?bar=baz');
     });
+    it('returns a string with a single filter statement applied and formatted for steve if a single filter statement is applied and the url starts with "/v1"', () => {
+      expect(urlOptionsGetter('/v1/foo', { filter: { bar: 'baz' } })).toBe('/v1/foo?filter=bar=baz&exclude=metadata.managedFields');
+    });
     it('returns a string with a multiple filter statements applied if a single filter statement is applied', () => {
       expect(urlOptionsGetter('foo', { filter: { bar: 'baz', far: 'faz' } })).toBe('foo?bar=baz&far=faz');
+    });
+    it('returns a string with a multiple filter statements applied and formatted for steve if a single filter statement is applied and the url starts with "/v1"', () => {
+      expect(urlOptionsGetter('/v1/foo', { filter: { bar: 'baz', far: 'faz' } })).toBe('/v1/foo?filter=bar=baz&far=faz&exclude=metadata.managedFields');
     });
     it('returns a string with an exclude statement for "bar" and "metadata.managedFields" if excludeFields is a single element array with the string "bar" and the url starts with "/v1/"', () => {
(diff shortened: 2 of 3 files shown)


aalves08 (commented):
@Sean-McQ changes seem to be "trivial" but I don't have any info on how to reproduce this and also I don't know about this `filter` option on the `findAll` 😛 . I would say that if @mantis-toboggan-md is okay with this change, I'm also okay with it.

Sean-McQ:
Changed out uriEncodeComponent for uriEncode since uriEncodeComponent was encoding characters that it shouldn't such as ":". Tested thoroughly and all existing tests still pass. While I was in there I also addressed the fact that filters and sorting criteria for Steve requests wouldn't be formatted correctly for that API so I added in some code to support that. New tests added to cover Steve specific url options. There's still quite a bit of functionality we're not using in Steve's URL options but that's covered under #9341

shell/components/form/Members/MembershipEditor.vue
```diff
@@ -56,8 +56,13 @@ export default {
   },
 
   async fetch() {
+    const roleBindingRequestParams = { type: this.type, opt: { force: true } };
+
+    if (this.type === NORMAN.PROJECT_ROLE_TEMPLATE_BINDING) {
+      Object.assign(roleBindingRequestParams, { opt: { filter: { projectId: this.parentId.split('/').join(':') } } });
```
aalves08, line 62: Is there any chance of `this.parentId` might be `undefined`? If so then `this.parentId.split` would throw an error. (optional chaining)

shell/plugins/steve/__tests__/getters.spec.ts
```diff
@@ -86,8 +92,17 @@ describe('steve: getters', () => {
...
+    });
     it('returns a string with a sorting criteria if the sort option is provided and an order if sortOrder is provided', () => {
       expect(urlOptionsGetter('foo', { sortBy: 'bar', sortOrder: 'baz' })).toBe('foo?sort=bar&order=baz');
     });
+    it('returns a string with a sorting criteria formatted for steve if the sort option is provided and an order if sortOrder is provided and the url starts with "/v1"', () => {
+      expect(urlOptionsGetter('/v1/foo', { sortBy: 'bar', sortOrder: 'baz' })).toBe('/v1/foo?exclude=metadata.managedFields&sort=bar');
+    });
+    it('returns a string with a sorting criteria formatted for steve if the sort option is provided and an order if sortOrder is "desc" and the url starts with "/v1"', () => {
+      expect(urlOptionsGetter('/v1/foo', { sortBy: 'bar', sortOrder: 'desc' })).toBe('/v1/foo?exclude=metadata.managedFields&sort=-bar');
+    });
```
aalves08, line 106: Very interesting all of these options! never knew about these. Nice! 🤜

aalves08 (commented):
just a small comment, but otherwise looks good. Still need to test it. Once you have that system ready, let me know 🙏

shell/components/form/Members/MembershipEditor.vue
Sean-McQ: It shouldn't be possible for it to be undefined or null while the type is `NORMAN.PROJECT_ROLE_TEMPLATE_BINDING` since the route requires it but it's trivial to add code to ensure that that's so prior to running so I'll do that.

shell/plugins/steve/__tests__/getters.spec.ts
Sean-McQ: Thank @cmurphy, she gave us all sorts of goodies in Steve!

aalves08:
All good @Sean-McQ ! 🚢 it

aalves08 (approved):
LGTM!

PR #9762, opened 2023-09-22
Safe ClusterRoleTemplateBinding with a non-existent user

### LINKED ISSUE #9729: Creating a ClusterRoleTemplateBinding with a non-existent user breaks the Cluster Membership page
**Setup**
- Rancher version: dev-2.8 (also on 2.7.7)
- Rancher UI Extensions: None
- Browser type & version: Firefox

**Describe the bug**
If you create a CRTB that does not have a user that exists, when you open the Cluster Membership page the list of members will not load. Doing the same for PRTBs does not cause an issue in the Project Membership page. Also, if you use a field other than userName, such as userPrincipalName it doesn't cause the UI to break.

**To Reproduce**

With public API in 2.8.0:

1. Create a Cluster Role Template in the Users & Authentication tab
2. Open a kubectl shell for your local cluster
3. Apply the following CRTB via kubectl apply -f crtb.yaml

```
kind: ClusterRoleTemplateBinding
metadata:
  name: test-crtb
  namespace: local
clusterName: local
roleTemplateName: <clusterRoleTemplate.Name>
userName: fake-user
```

Alternatively, this can be done through the Norman API in 2.7.7:
1. Create a Cluster Role Template in the Users & Authentication tab
2. Use the following curl command:
```
curl -u "${CATTLE_ACCESS_KEY}:${CATTLE_SECRET_KEY}" \
-X POST \
-H 'Accept: application/json' \
-H 'Content-Type: application/json' \
"${RANCHER_URL}/v3/clusterroletemplatebindings" -d \
'{  
    "clusterId": "local",  
    "groupId": "",  
    "groupPrincipalId": "",  
    "name": "",  
    "namespaceId": "",  
    "roleTemplateId": "${ROLE_TEMPLATE}",  
    "userId": "fake-user",  
    "userPrincipalId": ""  
}'

```

**Result**

Go to the "Cluster and Project Members" tab and see that the list of Cluster Memberships won't populate any members, not even those previously visible.

**Expected Result**
All the members should be listed, and if the user doesn't exist use the same image from PRTBs:
image

**Screenshots**
What the broken UI looks like:
image

**Additional context**
Issue in rancher/rancher with some additional context and more screenshots:
https://github.com/rancher/rancher/issues/42822

torchiaf:
### Summary
Fixes #9729

### DIFF
--- a/shell/components/form/Members/ClusterPermissionsEditor.vue
+++ b/shell/components/form/Members/ClusterPermissionsEditor.vue
@@ -198,7 +198,7 @@ export default {
     async principalProperty() {
       const principal = await this.principal;
 
-      return principal.principalType === 'group' ? 'groupPrincipalId' : 'userPrincipalId';
+      return principal?.principalType === 'group' ? 'groupPrincipalId' : 'userPrincipalId';
     },
 
     onAdd(principalId) {
--- a/shell/models/management.cattle.io.clusterroletemplatebinding.js
+++ b/shell/models/management.cattle.io.clusterroletemplatebinding.js
@@ -43,7 +43,7 @@ export default class CRTB extends HybridModel {
 
   get principalId() {
     // We've either set it ourselves or it's comes from native properties
-    return this.principalName || this.userPrincipalName || this.groupPrincipalName;
+    return this.principalName || this.userPrincipalName || this.groupPrincipalName || '';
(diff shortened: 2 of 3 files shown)


aalves08 (approved):
LGTM

PR #9768, opened 2023-09-25
fix: Adds margins to the app-bar when the global banner is enabled

### LINKED ISSUE #9752: Header and Footer banners should be full width
With the app bar changes, the header and footer banners don't go full width - they should:

image

Note, they are still full width on the login view, so this would also give consistency:

image

scures:
### Summary
Fixes #9752 

### Test
Go to https://localhost:8005/c/local/settings/banners and enable the Header Banner, it should now take the whole width on the top of the page. Do the same for the Footer or both, same behavior is expected. 

### Screenshot/Video

image

### DIFF
--- a/shell/assets/styles/base/_variables.scss
+++ b/shell/assets/styles/base/_variables.scss
@@ -3,6 +3,7 @@ $body-font: 'Lato', arial, helvetica, sans-serif;
 $mono-font: 'Roboto Mono', monospace;
 
 $app-bar-collapsed-width: 70px;
+$app-bar-height-with-global-banner: 2em;
 $max-width: 1440px !default;
 $min-width: 75% !default;
 $input-height: 61px;
--- a/shell/assets/styles/base/_variables.scss
+++ b/shell/assets/styles/base/_variables.scss
@@ -3,7 +3,6 @@ $body-font: 'Lato', arial, helvetica, sans-serif;
 $mono-font: 'Roboto Mono', monospace;
 
 $app-bar-collapsed-width: 70px;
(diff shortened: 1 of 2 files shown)


nwmac:
@richard-cox Can you review this one please

richard-cox (changes requested):
There's a lot of over complex code here which can be simplified to (with settings fix of x === 'true' and reactivity fix for byId)

- bannerConfiguration can just be a computed property
  - ```
    bannerConfiguration() {
      const settings = this.$store.getters['management/all'](MANAGEMENT.SETTING);
      const bannerSettings = settings.find((s) => s.id === SETTING.BANNERS);

      if (bannerSettings) { // if this isn't reactive it can go back in to a `bannerSettings` computed property)
        const parsed = JSON.parse(bannerSettings.value);
        const { showHeader, showFooter } = parsed;

        return {
          showHeader: showHeader === 'true',
          showFooter: showFooter === 'true',
        };
      }

      return undefined;
    },
    ```
- styles applied via css selector rather than in code
  - `:class="{'bannerHeader' : bannerConfiguration.showHeader, 'bannerFooter' : bannerConfiguration.showFooter}"`
  - `.side-menu { &.bannerHeader { margin-top: 28px; } &.bannerFooter { margin-bottom: 28px; } }`
  
This means
- no additional data properties, watch or methods required
- as the styles are in scss the height can then match the FixedBanner .banner { line-height: 2em} instead of hardcoding a pixel size to match. so that 2em becomes a scss variable somewhere

richard-cox (approved):

PR #9784, opened 2023-09-26
Update bulk action condition to check for empty resource actions

### LINKED ISSUE #9779: Rancher UI crash when mouse hovers over action buttons
Downstream Cluster : Cluster --> Projects/Namespaces 

If there is no namespace listed under "Not in a Project" then the UI crashes if mouse hovers over one of the three the action buttons ( <MOVE> // <Download YAML> // <Delete>).

By default, it seems, there is always a namespace "local" listed in the subsection "Not in a Project". But if someone moves this namespace to a project or deletes it, and there is no other namespace, then the UI crashes as described. 

Even after browser refresh the issue happens again and again until you create a simple namespace which is not assigned to an existing project – the name of the new namespace does not have to be local, you can name it whatever you want. The only requirement for the new namespace is that it does not belong to a project. 

Troubleshooting steps:

Make sure that there is no namespace available which does not belong to a project
Hover mouse over one of the action buttons and the UI crashes with
e.availableActions is undefined console.js:31:30

  e console.js:31

  e client.js:111

  Babel 7

  errorHandler client.js:113

  VueJS 16

  setBulkActionOfInterest index.vue:799

  mouseover index.vue:1

  VueJS 33

jordojordo:
### Summary
Fixes #9779 

When removing a namespace from the "Not in a Project" group, this group would still be ran within a method looking for `availableActions`. With the project not being a real resource (or "Project"), it would not have any available actions and would cause the computed method `canRunBulkActionOfInterest` to fail.

Added to the first condition to also check for the existence of a resource's `availableActions`.

### Screenshot/Video

https://github.com/rancher/dashboard/assets/40806497/c3799716-c52a-418c-b8d1-02b55a297e4e

### DIFF
--- a/shell/components/SortableTable/index.vue
+++ b/shell/components/SortableTable/index.vue
@@ -1,6 +1,7 @@
 <script>
 import { mapGetters } from 'vuex';
 import day from 'dayjs';
+import isEmpty from 'lodash/isEmpty';
 import { dasherize, ucFirst } from '@shell/utils/string';
 import { get, clone } from '@shell/utils/object';
 import { removeObject } from '@shell/utils/array';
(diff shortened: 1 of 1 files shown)


aalves08 (approved):
LGTM

PR #9789, opened 2023-09-27
Projects ns extension point

jordojordo:
### Summary
This adds the `ExtensionPanel` component to the Projects/Namespaces resource list. As this list was a custom component, it does not use the standard `ResourceList` component which already contains the `ExtensionPanel`

Also bumped the shell package version to `0.3.25` to allow extensions to utilize this feature.

### DIFF
--- a/shell/components/ExplorerProjectsNamespaces.vue
+++ b/shell/components/ExplorerProjectsNamespaces.vue
@@ -5,6 +5,8 @@ import { STATE, AGE, NAME } from '@shell/config/table-headers';
 import { uniq } from '@shell/utils/array';
 import { MANAGEMENT, NAMESPACE, VIRTUAL_TYPES } from '@shell/config/types';
 import { PROJECT_ID, FLAT_VIEW } from '@shell/config/query-params';
+import { PanelLocation, ExtensionPoint } from '@shell/core/types';
+import ExtensionPanel from '@shell/components/ExtensionPanel';
 import Masthead from '@shell/components/ResourceList/Masthead';
 import { mapPref, GROUP_RESOURCES, ALL_NAMESPACES } from '@shell/store/prefs';
 import MoveModal from '@shell/components/MoveModal';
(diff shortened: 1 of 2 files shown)


aalves08 (approved):
LGTM! now you got me wondering if should add the `PanelLocation.RESOURCE_LIST` stuff to the top of the `ResourceTable` instead of the `ResourceList` component. We would capture much more usecases and without any breaking changes to the user (I believe). What's your opinion? 

It would just be moving the extension panel that is on `ResourceList` (and ofc the variables) to `ResourceTable`.

jordojordo:
I thought about adding it there, but I believe this would cause more confusion when a resource is [using a custom list component](https://github.com/rancher/dashboard/blob/70ad28eb3de9d6df515adb5a08777d204314acbe/shell/components/ResourceList/index.vue#L208-L215). 

The Projects/Namespaces seems to be the only example of this case where listing the resources for Projects and Namespaces isn't using this `hasCustomList` since it is a unique case of combing two resources together.

aalves08:
> I thought about adding it there, but I believe this would cause more confusion when a resource is [using a custom list component](https://github.com/rancher/dashboard/blob/70ad28eb3de9d6df515adb5a08777d204314acbe/shell/components/ResourceList/index.vue#L208-L215).
> 
> The Projects/Namespaces seems to be the only example of this case where listing the resources for Projects and Namespaces isn't using the `hasCustomList` feature since it is a unique case of combing two resources together.

The thing is there are tables without being in a `ResourceList` (eg: tables in tabs, like details page of clusters -> machine pools) that cannot be targeted by this panel. I found a quite interesting usecase on SUMA extension because of that. Adding it there would not only capture `ResourceList` coming from custom components (at least they should all use `ResourceTable` and not `SortableTable`, but would have to check) and also incorporate those extra cases that a table may contain.

Either that or we just create an additional extension point. 🤔

jordojordo:
>Either that or we just create an additional extension point.

Perhaps we could add another panel location rather than another extension point? Either way, I think for the scope of this PR is fine for a last minute addition to `v2.8.0`.

aalves08:
> > Either that or we just create an additional extension point.
> 
> Perhaps we could add another panel location rather than another extension point? Either way, I think the scope of this PR is fine for a last minute addition to `v2.8.0`.

Correct for both parts of your comment my friend 😄

PR #9857, opened 2023-10-04
Updated docs urls

### LINKED ISSUE #9854: Modify docs URLs to use new domain ranchermanager.docs.rancher.com
Docs are working on updating there versioned links (up to /v2.7 right now, but will add /v2.8 this week).

They would like to see if the UI can use the current domain (https://ranchermanager.docs.rancher.com/) instead of the old one (https://rancher.com/docs/rancher/) so they don't need to add redirects

NEW: https://ranchermanager.docs.rancher.com/ replaces 
OLD: https://rancher.com/docs/rancher/

momesgin:
### Summary

Replaced the old docs urls with the new ones. In some cases the english language code(`en`) needed to be removed from the url otherwise you would visit the "Page not found" page.

Fixes #9854

### DIFF
--- a/docusaurus/docs/code-base-works/auth-providers.md
+++ b/docusaurus/docs/code-base-works/auth-providers.md
@@ -37,7 +37,7 @@ Use the steps below to set up a Keycloak instance for dev environments and confi
 
    > Double check the client has the correct checkboxes set, specifically the Mappers `group` entry.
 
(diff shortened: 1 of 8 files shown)


mantis-toboggan-md (approved):
lgtm, verified all links are working

momesgin:
> lgtm, verified all links are working

Thanks for double-checking them 🙏

PR #9865, opened 2023-10-06
now properly escaping key names prior to diffing objects

### LINKED ISSUE #9249: Helm Chart values not parsed correctly (PodSpec.nodeSelector: got "map", expected "string")
**Rancher Server Setup**
- Rancher version: **v2.6.2**
- Installation option (Docker install/Helm Chart): **Helm Chart**
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): **RKE1, Kubernetes v1.19.10**
- Proxy/Cert Details: **Nginx Ingress Controller (came with RKE1 cluster), Rancher-generated TLS certificate (cert-manager installed)**

- Kubernetes version: **v1.21.5**	
- Cluster Type (Local/Downstream): **Downstream**
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): **VMware vSphere provider**

**Describe the bug**
Unable to deploy Nginx Ingress Controller on downstream cluster (VMware vSphere provider) from Rancher UI (Apps & Marketplace) with `controller.nodeSelector` set to `node-role.kubernetes.io/worker: 'true'`.  
Deployment with the same values is successful when deploying with helm from CLI:
```
cat > nginx-controller-load-balancer-service.yaml <<EOF
controller:
  nodeSelector:
    node-role.kubernetes.io/worker: 'true'
EOF

helm install ingress-nginx ingress-nginx/ingress-nginx --namespace ingress-nginx -f nginx-controller-load-balancer-service.yaml --version=4.0.8
```
Since it works from CLI, this seems to be Rancher issue.

**To Reproduce**
- Add `ingress-nginx` repository to Rancher (Apps & Marketplace) using this URL (from their docs): `https://kubernetes.github.io/ingress-nginx`
- Try to deploy Nginx Ingress Controller from Rancher UI (Helm Chart version 4.0.8) changing `controller.nodeSelector` to in values yaml (single change in line 149):
```
  nodeSelector:
    node-role.kubernetes.io/worker: 'true'
```

**Result**
Helm install error:
> helm install --namespace=ingress-nginx --timeout=10m0s --values=/home/shell/helm/values-ingress-nginx-4.0.8.yaml --version=4.0.8 --wait=true ingress-nginx /home/shell/helm/ingress-nginx-4.0.8.tgz
> Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(Deployment.spec.template.spec.nodeSelector.kubernetes): invalid type for io.k8s.api.core.v1.PodSpec.nodeSelector: got "map", expected "string"

**Expected Result**
Successful installation.

**Additional context**
Default selector only works:
```
  nodeSelector:
    kubernetes.io/os: linux
```
Keeping default selector works:
```
  nodeSelector:
    kubernetes.io/os: linux
    node-role.kubernetes.io/worker: 'true'
```
Does not work:
```
  nodeSelector:
    node-role.kubernetes.io/worker: true
```
```
  nodeSelector:
    node-role.kubernetes.io/worker: 'true'
```
```
  nodeSelector:
    node-role.kubernetes.io/worker: "true"
```

Sean-McQ:
### Summary
Fixes #9249 
we were incorrectly creating a nested object where we should have had a primitive value when an object's key included a period. The root cause was that we weren't properly escaping the key names by surrounding them with double-quotes so that the `set` function knows not to split the string on those particular keys. This caused the "set" function to create additional levels of nesting and to create false keys that the API was not expecting.

The `diff` utility function now wraps individual keys in quotation marks so it doesn't confuse the `set` utility function when a key in a diff includes a dot.

### Technical notes summary
It's possible that the full repro steps in the parent issue might still produce an error given the steps and depending on your environment, this mostly has to do with those steps being insufficient for creating an additional nginx ingress in some environments and isn't related to the issue fixed here.

For testing purposes, we just want to verify that the payload we're hitting the API with doesn't have an incorrectly nested value.

For testing purposes, we just want to verify that the payload we're hitting the API with doesn't have an incorrectly nested value.

`/shell/components/ResourceDetail/index.vue` and `/shell/edit/provisioning.cattle.io.cluster/rke2.vue` both use the diff function which shouldn't have changed but it's be good to load a ResourceDetail page and edit an rke2 cluster just to make sure nothing untowards is happening outside of my test environment.

### DIFF
--- a/shell/utils/__tests__/object.test.ts
+++ b/shell/utils/__tests__/object.test.ts
@@ -1,5 +1,5 @@
 import {
-  clone, get, getter, isEmpty, toDictionary, remove
+  clone, get, getter, isEmpty, toDictionary, remove, diff, definedKeys
 } from '@shell/utils/object';
 
 describe('fx: get', () => {
@@ -157,6 +157,72 @@ describe('fx: remove', () => {
     [{ level1: { level2: true }, other: true }, 'level1', { other: true }],
   ])('should evaluate %p as %p', (obj, path, expected) => {
     const result = remove(obj, path);
+
+    expect(result).toStrictEqual(expected);
+  });
+});
+
+describe('fx: diff', () => {
+  it('should return an object including only the differences between two objects', () => {
+    const from = {
+      foo: 'bar',
+      baz: 'bang',
+    };
+    const to = {
+      foo:  'bar',
+      bang: 'baz'
+    };
+
+    const result = diff(from, to);
+    const expected = {
+      baz:  null,
+      bang: 'baz'
+    };
+
+    expect(result).toStrictEqual(expected);
+  });
+  it('should return an object and dot characters in object should still be respected', () => {
+    const from = {};
+    const to = { foo: { 'bar.baz': 'bang' } };
+
+    const result = diff(from, to);
+    const expected = { foo: { 'bar.baz': 'bang' } };
+
+    expect(result).toStrictEqual(expected);
+  });
+});
+
+describe('fx: definedKeys', () => {
+  it('should return an array of keys within an array', () => {
+    const obj = {
+      foo: 'bar',
+      baz: 'bang',
+    };
+
+    const result = definedKeys(obj);
+    const expected = ['"foo"', '"baz"'];
+
+    expect(result).toStrictEqual(expected);
+  });
+  it('should return an array of keys with primitive values and their full nested path', () => {
+    const obj = {
+      foo: 'bar',
+      baz: { bang: 'bop' },
+    };
+
+    const result = definedKeys(obj);
+    const expected = ['"foo"', '"baz"."bang"'];
+
+    expect(result).toStrictEqual(expected);
+  });
+  it('should return an array of keys with primitive values and their full nested path with quotation marks to escape keys with dots in them', () => {
+    const obj = {
+      foo: 'bar',
+      baz: { 'bang.bop': 'beep' },
+    };
+
+    const result = definedKeys(obj);
+    const expected = ['"foo"', '"baz"."bang.bop"'];
 
     expect(result).toStrictEqual(expected);
   });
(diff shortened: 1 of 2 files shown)


shell/utils/__tests__/object.test.ts
```diff
@@ -161,3 +161,69 @@ describe('fx: remove', () => {
...
+
+describe('fx: definedKeys', () => {
+  it('should return an array of keys within an array', () => {
+    const obj = {
+      foo: 'bar',
+      baz: 'bang',
+    };
+
+    const result = definedKeys(obj);
+    const expected = ['"foo"', '"baz"'];
```
cnotv, line 203: Why both single and double quotes? I am not so sure to understand why do we wrap strings in quotes 🤔

cnotv:
Are you sure that the use of periods was not intended? This is usually how you would use [`get()` in Lodash](https://lodash.com/docs/4.17.15#get), while keys with a period are more the exception for the K8S world, no?

shell/utils/__tests__/object.test.ts
Sean-McQ: In this case, I'm testing for the presence of the double-quotes in the string. If I just wrapped them in double quotes here they would be wrapping the string but not part of the string.

The double-quotes here serve the purpose of escaping any potential `.` characters that might confuse the `set` utility function.

Sean-McQ:
Yes, I'm sure the use of periods in the path was not intended as these keys are defined in yaml and typed which is the source of the error described in the ticket.

There are three new tests for definedKeys and two additional ones for the `diff` utility function. The `diff` utility function is the only place in the code that uses the `definedKeys` utility function if you don't count its recursive call to itself so testing that the `diff` utility function has the same input/output after my changes (save for the fixed behavior) ensures that this amount of test coverage is adequate.

mantis-toboggan-md (approved):

PR #5348, opened 2022-03-10
Fix gitrepo model bundle getter filter - #5326

Shavindra:
## Summary

Issue #5326 - Continuous Delivery Dashboard | Incorrect Bundles count

## Technical notes

- If 2 git repos have similar names filtering works incorrectly due to `` bundle.name.includes(`${ this.name }-`) `` accounting for substring. For example, if repo 1 has name `test` then another repo has `test1`, in the filtering `test1` will include `test1`
- It also breaks for instances where 2 different workspaces have same repo name

## Steps to test

1. Create additional workspace
    1. Navigate to Continuous Delivery > Advanced > Workspaces > create
2. Add a git repo to newly created workspace
    1. Navigate to Continuous Delivery > Git Repos > Add repository
    Name: test
    Repository URL: https://github.com/rancher/fleet-examples.git
    Path: /simple
    Branch: master

3. Add a git repo to Local workspace
    1. Navigate to Continuous Delivery > Git Repos > Add repository
    Name: test
    Repository URL: https://github.com/rancher/fleet-examples.git
    Path: /simple
    Branch: master

3. Add another git repo to Local workspace
    1. Navigate to Continuous Delivery > Git Repos > Add repository
    Name: testanother
    Repository URL: https://github.com/rancher/fleet-examples.git
    Path: /simple
    Branch: master

4. Navigate back to dashboard and it should show correct number of bundles

image

PR #9867, opened 2023-10-06
Fix gitrepo detail view bundle count

### LINKED ISSUE #9866: GitRepo detail view displays incorrect number of bundles
To repo:
1. Create a non-empty gitrepo named `test`
2. Create another non-empty gitrepo `test-abc`
3. Go to the detail view for the first repo - # of bundles is wrong

Found while investigating SURE-6692

mantis-toboggan-md:
### Summary
Fixes #9866 

This PR updates the bundle-filtering logic in FleetSummary to use the `repoName` getter, which references an annotation of the associated repo's name, instead of a regex. This is imiliar to a previous fix: https://github.com/rancher/dashboard/pull/5348

### Technical notes summary
Also cleaned up misleading unnecessary filtering in FleetBundles -- the bundles there are already filtered by repo.

1. Gitrepo detail view when repos have similar names

### Screenshot/Video
before and after:
Screen Shot 2023-10-06 at 8 35 00 AM
Screen Shot 2023-10-06 at 8 49 27 AM

### DIFF
--- a/shell/components/fleet/FleetBundles.vue
+++ b/shell/components/fleet/FleetBundles.vue
@@ -26,7 +26,9 @@ export default {
   },
 
   async fetch() {
-    this.allFleet = await this.$store.getters['management/all'](FLEET.CLUSTER);
+    if (this.$store.getters['management/schemaFor']( FLEET.CLUSTER )) {
+      this.allFleet = await this.$store.getters['management/all'](FLEET.CLUSTER);
+    }
   },
 
   data() {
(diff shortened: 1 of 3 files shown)


nwmac:
@Shavindra I'm a bit confused - I thought we'd already fixed the issue where bundle's weren't being matched correctly, because we were assuming  `name-` prefix rather than using the field that linked them to a GitRepo ?

nwmac:
@Shavindra Can you review this, please

PR #9868, opened 2023-10-06
Fixed RKE cluster being shown as Imported

### LINKED ISSUE #6836: Provider field on Imported clusters states "Imported" instead of cluster type
**Setup**

- Rancher version: v2.6.7 & v2.6-head Commit ID: f900172
- Browser type & version: Brave - v1.43.89 (Chromium v105.0.5195.102)

**Describe the bug**

Clusters that are imported into Rancher are showing their Provisioner as "Imported" instead of the cluster type (e.g. RKE, K3S, etc.)

**To Reproduce**

1. Start a Rancher server, and login as admin
2. Navigate to the Home page
- Note that the Provider column for the local cluster states "Imported Imported"
4. Create a cluster using RKE and then import the cluster into Rancher
- Note that the Provider column for the local cluster states "Imported Imported"
5. Navigate to the Home page
- Note that the Provider column states "Imported Imported"

**Result**

The Provider column for the local cluster states "Imported Imported"

**Expected Result**

The Provider column contains the correct information

**Screenshots**

v2.6.7:
image
image

v2.6-head:
image
image

**Additional Information**

This is reproducible as admin or as a standard user

eva-vashkevich:
### Summary
Fixes #6836

We were not identifying RKE1 clusters correctly.

Changed the way we identify cluster provisioner type to recognize RKE1 clusters correctly and moved the logic to determine if cluster is imported from the Cluster Provider component to the model for better consistency.

Cluster management and Home pages should be showing correct "provider" field for each cluster.
Cluster edit should be showing correct fields.

Tested using  Chrome 117.0.5938.132
Tested localy:
RKE1 is showing correct info as local and as imported clusters.
K3s is showing correct info as local, created through UI and imported clusters.
RKE2 is showing correct info as local and imported clusters.
AKS is showing correct info as created and imported cluster.

Cluster edit might be affected.

### Screenshot/Video

Screenshot 2023-10-05 at 3 24 06 PM


momesgin:
@eva-vashkevich Just curious what happens to imported k3s clusters, as of my understanding `provisioner()` would return `‘k3s’`, then `isImported()` would return `false`, and consequently `isImportedK3s()` would also return `false`, is this intentional?

eva-vashkevich:
@momesgin I've checked and it does identify imported k3 correctly, though your comment made me re-check rke2 case and I was missing identifying DO correctly, so thank you.

momesgin:
> @momesgin I've checked and it does identify imported k3 correctly, though your comment made me re-check rke2 case and I was missing identifying DO correctly, so thank you.

happy that it helped somehow 😆

mantis-toboggan-md (approved):

PR #9878, opened 2023-10-09
Automation: RadioGroup: Expose isDisabled slot prop

### LINKED ISSUE #8977: [@rancher/components] RadioGroup.vue: Expose disabled slot prop for indexed slots
Exposing a slot prop to communicate when the radio group is disabled will be helpful for tracking the overall disabled state when composing components.

An example of this usage can be seen in rancher-sandbox/rancher-desktop#4758

cnotv:
### Summary
Fixes #8977

### Technical notes summary
Added test using an example wrapper to integrate slot case.

### DIFF
--- a/pkg/rancher-components/src/components/Form/Radio/RadioGroup.test.ts
+++ b/pkg/rancher-components/src/components/Form/Radio/RadioGroup.test.ts
@@ -1,0 +1,37 @@
+import { mount } from '@vue/test-utils';
+import { RadioGroup } from './index';
+
+describe('component: RadioGroup', () => {
+  describe('when disabled', () => {
+    /**
+     * NOTE: Current setup does not match guidelines for the test of scoped slots
+     * https://test-utils.vuejs.org/guide/advanced/slots.html#Scoped-Slots
+     * This seems not covered till the latest version of the library
+     * https://github.com/vuejs/vue-test-utils
+     */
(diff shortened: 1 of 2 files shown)


pkg/rancher-components/src/components/Form/Radio/RadioGroup.test.ts
```diff
@@ -0,0 +1,37 @@
+import { mount } from '@vue/test-utils';
+import { RadioGroup } from './index';
+
+describe('component: RadioGroup', () => {
+  describe('when disabled', () => {
+    /**
+     * NOTE: Current setup does not match guidelines for the test of scoped slots
```
rak-phillip, line 7: This test looks very similar to the recommendation for v1 of Vue Test Utils, is there something that I'm missing in this note?

https://v1.test-utils.vuejs.org/api/options.html#scopedslots

rak-phillip (commented):
I think this change looks great, I just have a question about the note before the test.

pkg/rancher-components/src/components/Form/Radio/RadioGroup.test.ts
cnotv: That's a very good point, I did not notice that they changed also this.

rak-phillip (approved):
LGTM

PR #9912, opened 2023-10-13
add deprecation notice for restricted-admin role

aalves08:

- add deprecation notice for restricted-admin role in `user creation/edit` + `Role template list view` + `edit/detail view for global role of restricted-admin`

### Screenshot/Video

Screenshot 2023-10-13 at 11 35 22
Screenshot 2023-10-13 at 14 22 34
Screenshot 2023-10-13 at 14 54 44
Screenshot 2023-10-13 at 14 54 51
Screenshot 2023-10-13 at 14 54 55

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -4613,6 +4613,7 @@ rbac:
       restricted-admin:
         label: Restricted Administrator
         description: Restricted Admins have full control over all resources in all downstream clusters but no access to the local cluster.
+        deprecation: 'Warning: The Restricted Administrator role has been deprecated as of Rancher 2.8.0 and will be removed in a future release - Check out the <a href="{releaseNotesUrl}" target="_blank" rel="noopener noreferrer nofollow">Release Notes</a>'
       user:
         label: Standard User
         description: Standard Users can create new clusters and manage clusters and projects they have been granted access to.
--- a/shell/components/GlobalRoleBindings.vue
+++ b/shell/components/GlobalRoleBindings.vue
@@ -373,7 +373,6 @@ export default {
 
   .deprecation-notice {
     margin: 8px 0 8px 20px;
-    color: var(--error);
   }
   .role-group {
     .type-title {
--- a/shell/components/GlobalRoleBindings.vue
+++ b/shell/components/GlobalRoleBindings.vue
@@ -94,6 +94,7 @@ export default {
     };
   },
   computed: {
+    ...mapGetters(['releaseNotesUrl']),
     ...mapGetters({ t: 'i18n/t' }),
 
     isCreate() {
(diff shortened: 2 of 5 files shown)


shell/assets/translations/en-us.yaml
```diff
@@ -4613,6 +4613,7 @@ rbac:
       restricted-admin:
         label: Restricted Administrator
         description: Restricted Admins have full control over all resources in all downstream clusters but no access to the local cluster.
+        deprecation: 'Warning: The Restricted Administrator role has been deprecated as of Rancher 2.8.0 and will be removed in a future release - Check out the <a href="{releaseNotesUrl}" target="_blank" rel="noopener noreferrer nofollow">Release Notes</a>'
```
nwmac, line 4616: The target blank doesn't seem to work - it opens in the existing tab - not sure what is wrong.

shell/components/GlobalRoleBindings.vue
```diff
@@ -364,6 +370,11 @@ export default {
 </style>
 <style lang='scss' scoped>
   $detailSize: 11px;
+
+  .deprecation-notice {
+    margin: 8px 0 8px 20px;
+    color: var(--error);
```
nwmac, line 376: I think we should remove the red - it stands out too much - just having the text is sufficient, I think.

nwmac (changes requested):
If its easy, wopuld also be good to add the banner here:

image

aalves08:
> If its easy, wopuld also be good to add the banner here:
> 
> image

Done

shell/assets/translations/en-us.yaml
aalves08: It was the `v-clean-html` directive that was sanitizing it. Check if you agree with my change on the PR (added a hook to put the `target` back on links if `rel` property is well populated.

shell/plugins/clean-html-directive.js
```diff
@@ -17,6 +17,13 @@ const ALLOWED_TAGS = [
   'strong',
 ];
 
+DOMPurify.addHook('afterSanitizeAttributes', (node) => {
```
nwmac, line 20: @aalves08 yeah - Not so keen on this solution.

I assume all of our links that open in a new tab or broken? If so, I'd suggest removing this form this PR and filing an issue for that bug with the clean-html directive and that we fix it there and get this PR in.

Feels like we need a way to allow the target to be allowed by the clean directive, rather than adding it when we see the other attributes.

nwmac (changes requested):

shell/plugins/clean-html-directive.js
nwmac: Ref might be useful: https://github.com/cure53/DOMPurify/issues/317

aalves08: Changes reverted. Will create another issue and PR.

nwmac (approved):

PR #9913, opened 2023-10-13
Fix RBAC issue with workload create/edit screen

### LINKED ISSUE #9863: [BUG] User-base users cannot use secrets in Rancher 2.7.5+
**Rancher Server Setup**
- Rancher version: 2.7.5 or newer
   - Helm chart, RKE2; Both brand new latest installs and upgraded older versions.

- Kubernetes version: 1.24.x
   - Doesn't matter

**User Information**
  - Custom users created with "user-base" global permissions. Standard users are not effected.

**Describe the bug**
In Rancher 2.7.5, users that are created with user-base global permissions (standard users can still use secrets) are no longer able to use secrets. Users like this can use secrets fine in 2.6.x.

**To Reproduce**
- Create user with "user-base" global permissions; Nothing else customized. Starting with standard user level seems to create users that can still use secrets.
- Explore any cluster (any will work, local or downstream), go to Cluster => Projects/Namespaces
- Create a project
- Add the new user, any project permissions that "should" work for this (Owner, Member, or Custom)
- Create a registry secret
- Log out and back in as the new user
   - Create another registry secret if you want to
- Go to the cluster in question and try to create something using the registry secret, such as a deployment or pod
- No secrets will be available under Pull Secrets
- Secrets WILL be available under Storage => Secrets

It doesn't matter who creates the secret, which is why I suggested making them under both admin and the created user in question.

This seems like a bug since users can create and manage secrets fine under Storage > Secrets, but are unable to actually use them, and did work fine in older versions of rancher as expected.

**Result**
Limited users given permission to use secrets cannot use them.

**Expected Result**
Limited users given permission to use secrets should be able to use them.

**Additional context**
This bug was already confirmed by SUSE, and this was opened to track it.

nwmac:
Fixes #9863 

The workload create/edit screen assumes access to two resources that the user might not have access to - in that case, the fetch errors out and we do not fetch secrets etc, so the form does not operate correctly.

Hard to automate test as this requires RBAC setup.

To test:

- As admin, create a user with user-base
- Create a project and add a namespace
- Create an image pull secret (registry secret) int this new namespace
- Give the new user member access to the project
- Login as the new user
- Go to create a new workload in the namespace in the project and in general, check the 'Pull Secrets' dropdown in the Image section - with this bug, this will be empty. With this fix, the list should show the secret that was created.

image

### DIFF
--- a/shell/edit/workload/mixins/workload.js
+++ b/shell/edit/workload/mixins/workload.js
@@ -146,10 +146,20 @@ export default {
 
   async fetch() {
     // TODO Should remove these lines
-    await allHash({
-      rancherClusters:  this.$store.dispatch('management/findAll', { type: CAPI.RANCHER_CLUSTER }),
-      harvesterConfigs: this.$store.dispatch('management/findAll', { type: HCI.HARVESTER_CONFIG }),
-    });
+    // ? The results aren't stored, so don't know why we fetch?
+
+    // User might not have access to these resources - so check before trying to fetch
+    const fetches = {};
+
+    if (this.$store.getters[`management/canList`](CAPI.RANCHER_CLUSTER)) {
+      fetches.rancherClusters = this.$store.dispatch('management/findAll', { type: CAPI.RANCHER_CLUSTER });
+    }
+
+    if (this.$store.getters[`management/canList`](HCI.HARVESTER_CONFIG)) {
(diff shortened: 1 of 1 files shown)


aalves08 (approved):
testing and codewise, LGTM

PR #9914, opened 2023-10-13
add additional configuration banner for amazon cloud provider

### LINKED ISSUE #9110: Ability to deploy AWS external tree provider in RKE2

aalves08:
### Summary
Fixes #9110 

- Add warning banner for amazon cloud provider when kubernetes version is `gte` to `v1.27`

- Go to `Cluster Management` -> and create a `RKE2` `Amazon EC2` cluster
- The `basics` Tab, select the combo of `kube version` at least `v1.27` + cloud provider `Amazon`
- A new banner should appear just like the screenshot

### Screenshot/Video

Screenshot 2023-10-13 at 12 48 07

### DIFF
--- a/shell/edit/provisioning.cattle.io.cluster/Basics.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/Basics.vue
@@ -377,6 +377,13 @@ export default {
       const canNotEdit = this.clusterIsAlreadyCreated && !this.unsupportedCloudProvider;
 
       return canNotEdit;
+    },
+
+    /**
+     * Display warning about additional configuration needed for cloud provider Amazon if kube >= 1.27
+     */
+    showCloudProviderAmazonAdditionalConfigWarning() {
+      return !!semver.gte(this.value.spec.kubernetesVersion, 'v1.27.0') && this.agentConfig['cloud-provider-name'] === 'aws';
     }
   },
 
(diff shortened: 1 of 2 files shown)


shell/edit/provisioning.cattle.io.cluster/Basics.vue
```diff
@@ -413,6 +420,12 @@ export default {
         v-clean-html="t('cluster.harvester.warning.cloudProvider.incompatible', null, true)"
       />
     </Banner>
+    <Banner
+      v-if="showCloudProviderAmazonAdditionalConfigWarning"
+      color="warning"
+    >
+      <span v-clean-html="t('cluster.banner.cloudProviderAddConfig', {}, true)" />
```
Sean-McQ, line 427: It doesn't look like the translation string uses any arguments so I'm not sure you need the second and third parameter on the `t` function here.

Sean-McQ (approved):
One small nit but it does what it says on the label.

shell/edit/provisioning.cattle.io.cluster/Basics.vue
aalves08: It needs the last arg, so that it renders the HTML (anchor) that the translation contains. Usually the second arg is `null` but an empty object also works.

aalves08:
> One small nit but it does what it says on the label.

Thanks for the review mate!

PR #9921, opened 2023-10-16
add hook to re-add `target _blank` to external links cleaned by v-clean-directive

### LINKED ISSUE #9920: target="_blank" doesn't work in the UI with v-clean-html
**Setup**
- Rancher version: Any
- Rancher UI Extensions: 2.8-head (master)
- Browser type & version: Any

**Describe the bug**

A link using `v-clean-html` and with a `target="_blank"` doesn't open in a new tab because of the `v-clean-html` directive that removes the `target` attribute.

**To Reproduce**

Add a HTML element to a Vue view like (just as an example... find another translation string with an anchor `<a>` inside:
```
<p
    v-if="role.id === 'restricted-admin'"
    v-clean-html="t('rbac.globalRoles.role.restricted-admin.deprecation', { releaseNotesUrl }, true)"
    class="deprecation-notice"
  />
```
click on it via the UI

**Result**
Doesn't open the link in a new tab

**Expected Result**

Should open the link in a new tab

**Screenshots**

**Additional context**

aalves08:
add hook to re-add `target _blank` to external links cleaned by v-clean-directive

### Summary
Fixes #9920 

- add hook to re-add `target _blank` to external links cleaned by v-clean-directive

Add a HTML element to a Vue view like (just as an example... find another translation string with an anchor `<a>` inside:
```
<p
    v-if="role.id === 'restricted-admin'"
    v-clean-html="t('rbac.globalRoles.role.restricted-admin.deprecation', { releaseNotesUrl }, true)"
    class="deprecation-notice"
  />
```
click on it via the UI
Make sure the link is opened in a new tab

### DIFF
--- a/shell/plugins/clean-html-directive.js
+++ b/shell/plugins/clean-html-directive.js
@@ -17,6 +17,14 @@ const ALLOWED_TAGS = [
   'strong',
 ];
 
+DOMPurify.addHook('afterSanitizeAttributes', (node) => {
+  // set all elements owning target to target=_blank if rel property includes the correct values, making it safe
+  // Solution based on https://github.com/cure53/DOMPurify/issues/317
+  if ('rel' in node && node.rel.includes('noopener') && node.rel.includes('noreferrer') && node.rel.includes('nofollow')) {
+    node.setAttribute('target', '_blank');
+  }
+});
+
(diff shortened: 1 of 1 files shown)


nwmac (changes requested):
I don't like the fact we're adding target blank based on the other attributes - there may be cases where you want to set those, but use a different target.

Feels like we should preserve target blank somehow.

PR #9927, opened 2023-10-16
fix(GroupRules): fix multiple alert rules modification errors

### LINKED ISSUE #9926: Add multiple Alerting Rules in PrometheusRule to always display the same value
**Setup**
- Rancher version: v2.7.8
- Rancher UI Extensions:
- Browser type & version:

**Describe the bug**

When adding a second or more `Alerting Rules` to a `PrometheusRule`, the newly added `Alerting Rules` are the same as the first one

**To Reproduce**

1. Enable rancher monitoring, navigate to `PrometheusRule` and add multiple `Alerting Rules`
**Result**
All `Alerting Rules` are the same, and if you modify one of the `Alerting Rules`, the other forms will also take effect.

**Screenshots**

**Additional context**

prometheus

ly5156:
### Summary
Fixes #9926

1. Before adding a default rule, should make a copy of the default rule


torchiaf (approved):
LGTM

PR #9928, opened 2023-10-16
Support "target=_blank" in A links via clean-html directive

### LINKED ISSUE #9920: target="_blank" doesn't work in the UI with v-clean-html

nwmac:
Fixes #9920

To test:

Go to User Mgmt and the Restricted Admin role page.

Click on 'Release Notes' in the top banner and verify that this now opens in a new tab.

### DIFF
--- a/shell/initialize/index.js
+++ b/shell/initialize/index.js
@@ -13,7 +13,7 @@ import { setContext, getLocation, getRouteData, normalizeError } from '../utils/
 import { createStore } from '../config/store.js';
 
 /* Plugins */
-
+import { loadDirectives } from '@shell/plugins';
 import '../plugins/portal-vue.js';
 import cookieUniversalNuxt from '../utils/cookie-universal-nuxt.js';
 import axios from '../utils/axios.js';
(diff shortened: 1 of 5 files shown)


aalves08 (approved):
LGTM! Tested and it works. I thought you weren't into the hooks thing but it was more about the sanitisation logic.

richard-cox (changes requested):
This works fine on the role detail page.

This didn't work for me on the role list page. Looks like something is removing the `target` attribute... but adding console output to hooks (and directive inserted, etc functions) in `clean-helm-directive` provides no output ... as if the directive code was not running. 

Also checked the create user page and that also removes the target attribute.

I figure something's off locally but can't work out what

Edit: Apart from above, code looks good

aalves08 (approved):
I tested these latest changes with Elemental installed and we don't have Elemental overriding this, so I would say LGTM. This will be needed for links to open in a new tab on markdown files https://github.com/rancher/dashboard/pull/9956

nwmac:
@richard-cox I can't reproduce the issues you see

richard-cox (approved):
Could not reproduce the issue i saw in https://github.com/rancher/dashboard/pull/9928#pullrequestreview-1680586192

PR #9952, opened 2023-10-19
[INT] Removed vue-meta library

### LINKED ISSUE #9538: Remove vue-meta
# Description

Replace `vue-meta`, discontinued since 3 years but [has replacement for Vue3](https://github.com/nuxt/vue-meta/tree/next), still discontinued since 1 year, there's also no envisioned use case, as we don't update metadata.

eva-vashkevich:
### Summary
Fixes rancher/dashboard#9538

Removed vue-meta library

N/A

### Technical notes summary

Removed Meta from the project
 
This change is internal and shouldn't require additional testing.

Sending requests can be affected.

### DIFF
--- a/shell/mixins/brand.js
+++ b/shell/mixins/brand.js
@@ -4,6 +4,7 @@ import { SETTING } from '@shell/config/settings';
 import { findBy } from '@shell/utils/array';
 import { createCssVars } from '@shell/utils/color';
 import { _ALL_IF_AUTHED } from '@shell/plugins/dashboard-store/actions';
+import { setTitle } from '@shell/config/private-label';
 
 const cspAdaptorApp = ['rancher-csp-adapter', 'rancher-csp-billing-adapter'];
 
@@ -150,6 +151,7 @@ export default {
   },
   mounted() {
     this.setBodyClass();
+    setTitle();
   },
   methods: {
     setCustomColor(color, name = 'primary') {
--- a/shell/mixins/brand.js
+++ b/shell/mixins/brand.js
@@ -144,6 +144,44 @@ export default {
       }
     }
   },
+  mounted() {
+    const cssClass = `overflow-hidden dashboard-body`;
+    const body = document.getElementsByTagName('body')[0];
+    let title = getVendor();
+    let bodyClass, brandMeta;
+
+    if (getVendor() === 'Harvester') {
+      const ico = require(`~shell/assets/images/pl/harvester.png`);
+
+      title = 'Harvester';
+      const link = document.createElement('link');
+
+      link.hid = 'icon';
+      link.rel = 'icon';
+      link.type = 'image/x-icon';
+      link.hrefv = ico;
+      const head = document.getElementsByTagName('head')[0];
+
+      head.appendChild(link);
+    }
+
+    if ( this.brand ) {
+      try {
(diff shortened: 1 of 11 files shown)


nwmac (approved):

nwmac (changes requested):
The e2e test is failing because there is a genuine error here.

The vue-meta library wires in to some lifecycle hooks to set the title and other attributes.

Our branding mixin adds the `head()` method which vue-meta will call. Since we have removed vue-meta, this is not being called, so the title (and other attributes) are not being set.

See: https://v2.nuxt.com/docs/components-glossary/head/

shell/mixins/brand.js
```diff
@@ -144,6 +144,44 @@ export default {
...
+          bodyClass = `theme-${ this.theme } overflow-hidden dashboard-body`;
+          this.$store.dispatch('prefs/setBrandStyle', this.theme === 'dark');
+        }
+      } catch {
+        bodyClass = `theme-${ this.theme } ${ cssClass }`;
+      }
+    }
+
+    document.title = title;
+    body.className = bodyClass;
```
mantis-toboggan-md, line 183: This will set `body.className` to `'undefined'` when `this.brand` is falsey - the default value for the brand setting is '' so by default the body class is set to`'undefined'`. The dashboard defaults to a light theme, so you'll have to switch to dark theme to observe that the theme isn't being applied.

mantis-toboggan-md (changes requested):
I hit a couple issues with this PR.I noticed that when the private label is changed, `document.title` isn't updated until the page is refreshed. I think you can fix that by moving the title and body class computations into a computed prop, then using a watcher on that computed prop to re-set the document title/body class when they differ from the computed prop.  eg
```
watch: {
    ...
    
    someKindaTitleBodyClassComputedProp({ title }) {
      if (title !== document.title) {
        // re-set document.title and body class here
      }
    }
 }
  ```

I also noticed that global settings have not been loaded when the `mounted` function runs: I think you could move the call to load those out of `fetch` and into `mounted`  to resolve that.

File: shell/mixins/brand.js [hunk 2]
```diff
@@ -159,47 +166,26 @@ export default {
...
-    if ( this.brand ) {
-      try {
-        brandMeta = require(`~shell/assets/brand/${ this.brand }/metadata.json`);
-      } catch {
-        return out;
+    },
+    setBodyClass() {
+      const body = document.getElementsByTagName('body')[0];
+      const cssClass = `overflow-hidden dashboard-body`;
+      let bodyClass = `theme-${ this.theme } ${ cssClass }`;
```
mantis-toboggan-md, line 173: I think you still need to set `document.title` to `getVendor()` as part of this process: it ensures that the title is "Rancher Prime" when the ui-brand global setting indicates that this is a paid version of the dash and a custom private label hasn't been set. Set ui-brand to 'suse' to check this behaviour.

mantis-toboggan-md (changes requested):

mantis-toboggan-md (approved):
lgtm

nwmac:
@eva-vashkevich Can you mark the conversations as resolved for the ones you have addressed

eva-vashkevich:
> @eva-vashkevich Can you mark the conversations as resolved for the ones you have addressed
Done. I think it still wants you to explicitly approve it.

nwmac (approved):

PR #9956, opened 2023-10-20
Wrong rendering of Markdown files

### LINKED ISSUE #9951: [BUG] Wrong rendering of Markdown files
### Rancher Server Setup

- Rancher version: 2.7.6

### Information about the Cluster

N/A

### User Information

N/A

### Describe the bug

Markdown files are rendered in a wrong way in the Apps section.

### To Reproduce

- Go to Apps, click on Charts
- Select an app

### Result

A `README` file is rendered but some elements (especially tables) are wrong. The information is unreadable.

### Expected Result

- The output could be improved in terms of margins, sizes (headers, paragraphs), etc...
- Tables are rendered correctly.

### Screenshots

screenshot1

### Additional context

A good render can improve a lot the readability and user experience.

screenshot2

aalves08:
### Summary
Fixes #9951 

- remove `DOMpurify` from markdown component (`v-clean-html` does exactly this, so need to duplicate it)
- add missing `html` tags to allowed tags in `v-clean-html` so that markdown content is rendered correctly

**NOTE:** links in markdown won't open in a new tab unless https://github.com/rancher/dashboard/pull/9928 is merged as well.

- Go to `https://<!--- RANCHER BASE URL --->/c/local/apps/charts/chart?repo-type=cluster&repo=rancher-partner-charts&chart=tomcat&version=10.10.10` and check that markdown is rendered correctly

### Screenshot/Video

Screenshot 2023-10-20 at 11 07 18
Screenshot 2023-10-20 at 11 07 22

### DIFF
--- a/shell/components/Markdown.vue
+++ b/shell/components/Markdown.vue
@@ -20,9 +20,14 @@ export default {
 
   computed: {
     html() {
-      return this.marked.parse(this.value, {
+      console.log('HTML', this.marked.parse(this.value, {
         renderer: this.markedRenderer,
         breaks:   true
+      }));
+
+      return this.marked.parse(this.value, {
+        renderer: this.markedRenderer,
+        // breaks:   true
       });
     },
   },
@@ -87,8 +92,20 @@ export default {
 }
 
 .markdown {
+    blockquote {
+      color: rgb(101, 109, 118);
+      border-left: 0.25em solid rgb(208, 215, 222);
+      padding: 0 1em;
+      margin-bottom: 16px;
+    }
+
+    table {
+      border-collapse: collapse;
+    }
+
     TH {
       text-align: left;
+      border: 1px solid #e3e7eb;
     }
 
     table tr th {
@@ -109,6 +126,7 @@ export default {
       text-align: left;
       margin: 0;
       padding: 6px 13px;
+      border: 1px solid #e3e7eb;
     }
 
     table tr th :first-child, table tr td :first-child {
--- a/shell/components/Markdown.vue
+++ b/shell/components/Markdown.vue
@@ -20,14 +20,9 @@ export default {
 
   computed: {
     html() {
-      console.log('HTML', this.marked.parse(this.value, {
-        renderer: this.markedRenderer,
-        breaks:   true
-      }));
-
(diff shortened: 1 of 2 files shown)


recena:
@aalves08 Thank you. It seems tables do not have styles (look & feel).

shell/plugins/clean-html-directive.js
```diff
@@ -15,6 +15,18 @@ const ALLOWED_TAGS = [
...
+  'h3',
+  'h4',
+  'h5',
+  'h6',
+  'table',
+  'thead',
+  'tr',
+  'th',
+  'tbody',
+  'td',
```
recena, line 29: Why not `blockquote` for example?

richard-cox (commented):
@rak-phillip can you take a look at this one? I think it's mac specific, i couldn't reproduce at all.

Changes generally look ok. The custom instance of dompurity used in markdown component is removed and it now uses the global one from the clean directive.

aalves08:
@rak-phillip take into account that changes from https://github.com/rancher/dashboard/pull/9928 need to go in and this PR will need to be adjusted accordingly

shell/plugins/clean-html-directive.js
aalves08: will add it... This PR needs to be revisited as it's dependent of #9928 , which will lead to changes here.

rak-phillip:
@aalves08 All in all, I think the change looks good.

However, I'm testing against 2.8-head and I can't reproduce the problem as described in the issue. For example, I see the output rendered as expected while reviewing the Apache Tomcat chart

image

Is there an environment where I can see this issue in action so that we can validate that the change fixes the problem?

aalves08:
@rak-phillip I was able to repro this on a `2.7.6` system like the original issue. I'm running a `2.8.0-rc3` right now and I cannot repro it, which is strange. I don't know what changed 🤔

aalves08:
@rak-phillip can I get an approval here? I've just updated the PR after the merge of https://github.com/rancher/dashboard/pull/9928 by Neil

rak-phillip (approved):
I think this change looks good. I cannot reproduce the original issue. Even still, I think this change makes sense.

PR #9969, opened 2023-10-23
Update port-issue workflow permissions

mantis-toboggan-md:
I hit two issues with the backport workflow that didn't come up in my own testing repo initially:
1. The workflow needs permission to write issues, see https://github.com/rancher/dashboard/blob/master/.github/workflows/add-to-triage-label.yaml#L10
2. My membership in the Rancher org was set to private. I updated my membership setting (`https://github.com/orgs/rancher/people?query=<GITHUB LOGIN>`), not sure if there's an alternative workkaround that could be added to this workflow.


mantis-toboggan-md:
just double-checked and merely updating my rancher org memerbship visibility was enough for the backport bot to work, so i'm clsing this.

PR #9992, opened 2023-10-27
Fixed chart yaml scroll issue when there is an error at the bottom

### LINKED ISSUE #9680: Errors when editing chart yaml not fully visible
Setup
Rancher version: 2.7-head

Describe the bug
If there are a lot of issues with the chart yaml when editing, the errors at the bottom of the screen are cut off and the user can't see them all as the page won't scroll. The user could still fix some errors and save again to shorten the list and see more errors, but that's not the greatest UX.

To Reproduce
1. Deploy a Rancher 2.7-head cluster
2. On local or downstream, go to install a chart
3. Click the Edit YAML button
4. Make a bunch of errors in the yaml
5. Click Save

Result
Error list is shown, but cut off at the bottom if there are more than a few errors

Expected Result
Page to be scrollable with full error list shown

bisht-richa:
### Summary
Fixes #9680

Added fixed from 2.8 to 2.7 head to fix chart yaml scroll issue when there is an error at the bottom.

### Screenshot/Video

https://github.com/rancher/dashboard/assets/18264463/5046a97d-7a7a-4b58-9944-618be680002e

### DIFF
--- a/shell/components/Wizard.vue
+++ b/shell/components/Wizard.vue
@@ -397,6 +397,7 @@ export default {
             color="error"
             :label="err"
             :closable="true"
+            class="footer-error"
             @close="errors.splice(idx, 1)"
           />
         </div>
@@ -468,7 +469,6 @@ $spacer: 10px;
   flex-direction: column;
   flex: 1;
   padding: 0;
-  height: 0;
   justify-content: flex-start;
 }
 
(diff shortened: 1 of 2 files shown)


bisht-richa:
> This looks like the same issue as #9917. Need to understand the difference between this solution and it's (https://github.com/rancher/dashboard/pull/9918/files)

Yes, it is the same bug but in 2.7 head.

bisht-richa:
This bug has been resolved in 2.8.

PR #10019, opened 2023-11-07
Fixed scale up button not showing up

### LINKED ISSUE #9909: RKE2: Cannot Scale up node pool after scaling nodepool to 0 nodes
**Setup**
- Rancher version: v2.8-head
- Browser type & version: Chrome 117.0.5938.132

**Describe the bug**
When scaling a node pool to 0 nodes in a nodepool the UI option for scaling the nodepool up goes away and you can no longer add nodes to the cluster in that nodepool

**To Reproduce**
1. Create a AWS RKE2 cluster using k8s 1.27 and the following nodepool setup
pool1: 3 node all roles
pool2: 2 cp nodes
2. Scale down pool2 to 0 nodes
3. Attempt to scale pool2 back up.

**Result**
The option to scale up in the UI disappears 

**Expected Result**
Node pool should be able to be scaled back up.

**Screenshots**
image

Additional context: 
I was able to edit the cluster and add 2 nodes to the nodepool and they provisioned successfully. This makes it seem that this is only a UI issue as the backend can provision properly.
This behavior seems to be specific of RKE2, I tried to reproduce this on RKE1 but was unable to.

momesgin:
### Summary
Fixes #9909 

### Technical notes summary
Separating the condition for rendering the progress bar from the condition for the scaling buttons.

### Screenshot/Video

image


aalves08 (approved):
LGTM

PR #10033, opened 2023-11-13
Use full text label "Refresh" on refresh button instead of just icon

### LINKED ISSUE #8645: Use full text label "Refresh" on refresh button instead of just icon
For improved accessibility, always use a text label in buttons

aalves08:
### Summary
Fixes #8645 

- create new `manual-refresh` action to not confuse with other refresh actions 
- update translations to match new `AsyncBtn` action
- remove tooltip from manual refresh button (we already have the text)

Go to `Settings` -> `Performance` ->  `Manual Refresh` -> and enable manual refresh with a low threshold (ex: 5)
Go to `local` cluster -> `events` -> and check that the button appears with the correct text label

### Screenshot/Video

Screenshot 2023-11-13 at 12 16 56

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -691,6 +691,15 @@ asyncButton:
     successIcon: checkmark
     waiting: ''
     waitingIcon: refresh
+  manual-refresh:
+    action: ''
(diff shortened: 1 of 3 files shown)


momesgin (approved):
lgtm

PR #10056, opened 2023-11-20
Disable rke1 UI

### LINKED ISSUE #9823: Disable RKE1 completely through UI
Internal reference: SURE-6568

Request description:
As Rancher is a UI that various users are using to start clusters on top of the Harvester, they would would like to have an option/button to disable RKE1 features completely to avoid accidental deployment.

Actual behavior:
RKE1 features are always available

Expected behavior:
It should be possible to disable RKE1 features in the configuration 

There will be a new feature flag added

When this feature flag is NOT set, we should hide RKE1 features from the UI:

- The toggle switch in provisioning should be hidden - the user will only see RKE2
- Hide the 'RKE1 Configuration' section in Cluster Management

image

We should write automated tests for this feature.

See linked backend issue - suggest feature flag is, name: `rke1-ui`, description: `Enable RKE1 provisioning in the Rancher UI` - default is true.

Ahead of backend adding the required feature flag, you can add it yourself for development by importing/applying the following yaml:

```
kind: Feature
metadata:
  name: rke1-ui
spec:
  value: null
status:
  default: true
  description: Enable RKE1 provisioning in the Rancher UI
  dynamic: true
  lockedValue: null
```

bisht-richa:
### Summary
Fixes #9823

- Added  RKE1 features flag

When this feature flag is NOT set, we should hide RKE1 features from the UI:

- The toggle switch in provisioning should be hidden - the user will only see RKE2
- Hide the 'RKE1 Configuration' section in Cluster Management

### To Test

 importing/applying the following yaml:
```js
kind: Feature
metadata:
  name: rke1-ui
spec:
  value: null
status:
  default: true
  description: Enable RKE1 provisioning in the Rancher UI
  dynamic: true
  lockedValue: null
```

### Screenshot/Video

Screenshot 2023-11-22 at 14 06 08


shell/store/features.js
```diff
@@ -27,6 +27,7 @@ export const mapFeature = function(name) {
 export const MULTI_CLUSTER = create('multi-cluster-management', true);
 export const LEGACY = create('legacy', false);
 export const RKE2 = create('rke2', true);
+export const RKE1_UI = create('legacy', true);
```
nwmac, line 30: Wrong name

nwmac (changes requested):

shell/store/features.js
bisht-richa: This is to test, I will change it once I finish creating a test for this change.

shell/edit/provisioning.cattle.io.cluster/index.vue
```diff
@@ -266,8 +271,7 @@ export default {
     subTypes() {
       const getters = this.$store.getters;
       const isImport = this.isImport;
-      const isElementalActive = !!this.activeProducts.find((item) => item.name === ELEMENTAL_PRODUCT_NAME);
-
+      const isElementalActive = !!getters['type-map/activeProducts'].find((item) => item.name === ELEMENTAL_PRODUCT_NAME);
```
nwmac, line 274: @aalves08 Not sure why this change - any ideas? I think this should probably be removed - doesn't seem relevant to the overall change.

aalves08: This doesn't make sense. I ran this branch, but with the original `const isElementalActive = !!this.activeProducts.find((item) => item.name === ELEMENTAL_PRODUCT_NAME);` and all was fine with cluster creation and Elemental.

I've reverted the change and pushed the code @nwmac

aalves08: Ok, this change is because of the unit test created... 💦 it's because Richa must have hit a roadblock when mocking a namespaced getter `mapGetters('type-map', ['activeProducts'])`.

Let me see if there's an alternative.

aalves08: Unit test fixed. Pushed changes

aalves08:
Gave this a whirl and I think it's good to go @nwmac . Waiting your approval (or review dismiss) 🙏

aalves08 (approved):
LGTM

PR #10074, opened 2023-11-24
APP Bar quick fixes

scures:
### Summary
Fixes https://github.com/rancher/dashboard/issues/9780

Takes care of: 
- [x]  Blocks borders
- [x]  Burger icon section bottom border
- [x]  Pin icon size
- [x]  Deployed bar shadow
- [x]  Reduced transitions time
- [x]  Filter input field

Check the original issue for details 

### Screenshot/Video

image

It also fixes the visual representation of the custom color:
Old vs New
image

### DIFF
--- a/shell/components/ClusterIconMenu.vue
+++ b/shell/components/ClusterIconMenu.vue
@@ -7,11 +7,6 @@ export default {
       type:     Object,
       required: true,
     },
-    isSelected: {
-      type:     Boolean,
-      required: false,
-      default:  false,
-    },
   },
   computed: {
     isEnabled() {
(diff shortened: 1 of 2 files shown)


richard-cox (changes requested):
- Not sure if this was introduced in this PR, if not feel free to either address here or create a separate issue. The `See All` is handing in the middle of the section after filtering

image

image
- Regarding https://github.com/rancher/dashboard/pull/9954#discussion_r1380199253, i did a quick comparison between this PR and current head and it's a bit of a regression. In head there's only a few places, in PR it affects common things like `Cluster`, `Workloads`, sections within `Workloads`, etc). In the short term can we apply the css workaround from the comment and inline reference #10010 which will fix it later on?

richard-cox (approved):

PR #10084, opened 2023-11-27
Layout removal

### LINKED ISSUE #10059: Extensions: Replace Nuxt layouts with VueRouter nested routes

codyrancher:
### Summary
Removes the concept of nuxt layouts and replaces the concept with nested routes. 

Fixes #10059

There was some existing work to support adding nested routes which upon inspection never worked. This change fixes the issues and we can now support nested routes.

A page renders under each product

Extensions will need to be changed to support the new nested routes.

An example diff of the elemental required changes:
image

TODO:
We should remove the layout information from our dev/extension docs and we should add docs around nested routes.


shell/core/plugin-routes.ts
```diff
@@ -34,70 +22,25 @@ export class PluginRoutes {
...
   }
 
-  // Ensure we put back any routes that the plugin that is being uninstalled added
-  public uninstall(plugin: any) {
-    // List of routes we need to restore
-    const restore: RouteInfo[] = [];
-
-    Object.keys(this.replacedRoutes).forEach((routeName) => {
-      const info = this.replacedRoutes[routeName];
-
```
codyrancher, line 44: Uninstall really shouldn't be necessary. We require the user refresh the page on uninstall.

shell/core/plugins.js
```diff
@@ -222,9 +222,6 @@ export default function({
         promises.push(...this.removeTypeFromStore(store, 'management', Object.keys(plugin.types.models)));
       }
 
-      // Uninstall routes
```
codyrancher, line 225: No longer needed. See above.

shell/initialize/App.js
```diff
@@ -7,19 +7,15 @@ import NuxtError from '../layouts/error.vue';
...
 export default {
   render(h) {
     const loadingEl = h('NuxtLoading', { ref: 'loading' });
 
-    const layoutEl = h(this.layout || 'nuxt');
     const templateEl = h('div', {
       domProps: { id: '__layout' },
       key:      this.layoutName
-    }, [layoutEl]);
+    }, [h('nuxt')]);
```
codyrancher, line 18: We still require the nuxt element which is a thin wrapper around <router-view>. I ran into some issues which I no longer recall when switching to router-view. I'm guessing <nuxt> does some component registration for fetch etc.

shell/components/layouts/Cluster.vue
codyrancher, line 1: Files under `components/layouts` are used in the vue router nested routes.
Reduces the amount of changes necessary here and improves re-usability slightly.

shell/core/plugin-routes.ts
```diff
@@ -107,87 +50,86 @@ export class PluginRoutes {
...
     });
 
     const newRouter: Router = new Router({
       mode:   'history',
       routes: [...orderedPluginRoutes, ...allRoutes]
     });
 
-    // handle plugin routes with parent
-    pluginRoutesWithParents.forEach((r: any) => {
-      newRouter.addRoute(r.parent, r.route);
```
codyrancher, line 187: This didn't work. The routes wouldn't render even though I believe the pages were found.

shell/core/plugin-routes.ts
```diff
@@ -6,20 +6,8 @@ interface RouteInfo {
...
-
-type RouteInstallHistory = {
-  [route: string]: RouteInstallInfo[]
-}
-
 export class PluginRoutes {
   router: Router;
-  pluginRoutes: RouteConfig[] = [];
-
-  replacedRoutes: RouteInstallHistory = {};
```
codyrancher, line 22: We don't need to track this information any longer since we don't need to handle uninstall.

jordojordo (commented):
I'm seeing a lot of strange behavior on both Firefox and Chrome. I tested every product page I had available, the only page which wouldn't work was the extensions page: `/c/local/uiplugins` - this would just load a blank screen with navigation. I could not see the `/c/cluster/uiplugins` route within the vue dev tools either.

uiplugin-page

Other general strangeness:
- There is no vertical scroll anywhere, any resource view, list, or even the side-nav

- Logging in requires a refresh, an error in the console mentions duplicate routes:
```console
[vue-router] Named Route 'index' has a default child route. When navigating to this named route (:to="{name: 'index'}"), the default child route will not be rendered. Remove the name from this route and use the name of the default child route for named links instead. vue-router.esm.js:16
[vue-router] Duplicate named routes definition: { name: "index", path: "" }
```

https://github.com/rancher/dashboard/assets/40806497/9d43b08a-fe54-4afd-b0dd-f19b27028764

- The cluster "Explore" button within the Cluster Management view leads to an empty view

https://github.com/rancher/dashboard/assets/40806497/4f47679a-c9b6-429f-ad0d-57309ccd3e3f

- The Loading overlay for resource views overlaps the sidenav

https://github.com/rancher/dashboard/assets/40806497/2921b48b-8fa3-4e3f-935c-f49e56ec673d

shell/config/router.js
```diff
@@ -18,363 +18,359 @@ export const routerOptions = {
...
+    },
+    {
+      path:      '/',
+      component: () => interopDefault(import('../components/layouts/Navigation.vue')),
+      name:      'index',
+      children:  [
+        {
+          path:      '',
+          component: () => interopDefault(import('../pages/index.vue')),
+          name:      'index'
```
aalves08, line 40: Duplicate with line 40. `index`

aalves08 (commented):
Overall the same comments as @jordojordo . No scrolls, login needs refresh, cannot access `uiplugins`, etc.

I've found the duplicate name on the router definition as well.

I don't know if it's on purpose or not, but looking at the dev tools you've got two possible paths for `c/:cluster`... funny enough the `/apps` works and `uiplugins` doesn't... (is it because of the redirect in `shell/pages/c/_cluster/apps/index.vue?`)
 
Screenshot 2023-12-06 at 14 40 01

I could find the `uiplugins` route on the dev tools @jordojordo 
Screenshot 2023-12-06 at 14 39 25

With the changes in `plugin-routes` I was a bit afraid you had messed up the route appending system for extensions (they need to be prepended so that they won't hit wildcarded routes) but that seems fine according to the dev tools

Screenshot 2023-12-06 at 14 49 15

I couldn't install any other extensions (no access to `/uiplugins`) but Elemental was already on my instance and it doesn't show any side navs

Screenshot 2023-12-06 at 14 36 20

Screenshot 2023-12-06 at 14 52 32

codyrancher:
I ended up simplifying the approach. Instead of doing the layout removal along with the nested layouts/templates I decided to just go with only the layout removal.

I also added layout deprecation messages and some reasonable defaults for extensions that don't specify a parent and still use layouts.

shell/config/router.js
```diff
@@ -6,375 +6,396 @@ import scrollBehavior from '../utils/router.scrollBehavior.js';
...
-  }, {
-    path:      '/auth/setup',
-    component: () => interopDefault(import('../pages/auth/setup.vue' /* webpackChunkName: "pages/auth/setup" */)),
-    name:      'auth-setup'
-  }, {
-    path:      '/auth/verify',
-    component: () => interopDefault(import('../pages/auth/verify.vue' /* webpackChunkName: "pages/auth/verify" */)),
-    name:      'auth-verify'
-  }, {
-    path:      '/rio/mesh',
```
aalves08, line 78: I've seen that you've deleted the `rio/mesh` page as well. What was this used for?

aalves08 (changes requested):
There are still several issues happening on Dashboard:

**1) login still requires a page refresh**

https://github.com/rancher/dashboard/assets/97888974/b8504a6e-c5d5-485f-8739-e50ec198a4e9

**2) kubewarden routes are broken**


https://github.com/rancher/dashboard/assets/97888974/45e02382-a19a-4b3b-85fc-dfef149b85c3

This almost surely be related to the way Kubewarden routes have been added to the routes tree. On the current code in `master` we "prepend" the extension routes to the routes array. With this we are mitigating the fact that we could hit a wildcarded route like `c/:cluster/:product/:resource` before the matcher finds Kubewarden's unique route `c/:cluster/kubewarden/<-- whatever-kubewarden-resource -->`

Also, we cannot add `elemental` routes under `c/:cluster` as it's a "global" product not a cluster one. Ex: `elemental/c/_/<-- whatever-elemental-resource -->`. Elemental routes also need to be "prepended" (added to the top) of the router array on the correct section.

routes-tree

You can check where to add the extension route by looking at it's `name` property. Either it starts with `c-cluster-xxxx` or it's a `xxxx-c-cluster-`.

**3) no scroll on diagnostics page**

scroll-diagnostics-page


**4) bad link on `Users & Authentication` -> `Role templates` (check lower left corner of screenshot - that's why the link it's highlighted)**


bad-link

**5) Overall problem with navigation in different products, such as `fleet`, `elemental`, etc , on a "sub-menu" navigation level**


https://github.com/rancher/dashboard/assets/97888974/381a02db-8488-414f-a314-682e2cdc4c55


https://github.com/rancher/dashboard/assets/97888974/a4479763-b526-49bf-a510-37788e994a1f


https://github.com/rancher/dashboard/assets/97888974/3bbc948c-27d0-4681-8cf8-7117e3fa16b9

**6) There's something wrong with the `feature flag` part under `settings`**


https://github.com/rancher/dashboard/assets/97888974/0f3ebf56-5841-4052-bc5f-1bc441053614

shell/config/router.js
codyrancher: It was really the basis for fleet but is no longer used https://github.com/rancher/rio

codyrancher:
@aalves08 I'll just comment on these separately so we can have separate threads.

> 1) login still requires a page refresh

Are you doing anything special here? It seems to completely work for me.

https://github.com/rancher/dashboard/assets/55104481/7dba48aa-958b-41e2-b8cc-1281e1af1f6e

codyrancher:
> 3) no scroll on diagnostics page

Fixed with https://github.com/rancher/dashboard/pull/10084/commits/7374ddc9f1f93546f06a7aa7eaf0edb34f13a747

Turns out the default layout wasn't scoped and affected all other layouts. It also looks like the unscoped styles are now lazy loaded instead of all loaded and app initialization. I moved the default styles to the global layout scss.

codyrancher:
> 4) bad link on Users & Authentication -> Role templates (check lower left corner of screenshot - that's why the link it's highlighted)

Fixed with https://github.com/rancher/dashboard/pull/10084/commits/098d9c380b32cbe2c2e34503ea7c3fea4659114c

codyrancher:
> 6) There's something wrong with the feature flag part under settings

This was actually true of navigating to any page that shared the same component. Changing parameters doesn't force components to be rerendered. To recreate the expected behavior I went ahead and force a rerender using :key. 

Since I was touching <nuxt> I also gave <router-view> a try and it looks like it works now so I don't think we'll need to open a separate issue.

codyrancher:
> 2) kubewarden routes are broken
It looks like you were correct, pushing the routes to the front appear to resolve this issue.

image

aalves08:
> @aalves08 I'll just comment on these separately so we can have separate threads.
> 
> > 1. login still requires a page refresh
> 
> Are you doing anything special here? It seems to completely work for me.
> 
>  logging-in.mp4

Nothing different, mate. Just testing this on `Chrome Version 120.0.6099.62 ` running your branch locally 🤷 If you want to get together to show you and help debug this, ping me

shell/core/plugin-routes.ts
aalves08: I looked at the code and couldn't quite trace it back correctly, but my main concern here is what happens on a extension `upgrade`? Does it do an uninstall of the previous version of the extension? If so, then if we aren't removing the previous extension routes, do they duplicate?

aalves08:
Overall, it's just the problem with the login and a question around the routes uninstall removal, afaik.

Also, we should give this a run with Harvester before we merge it (navigate throught it + access old a new Harvester clusters).

Not for this PR, since it's quite extensive, but we should look into the possibility of removing `shell/components/nuxt/nuxt.js` and `shell/components/nuxt/nuxt-child.js` in a near future. 

Great work @codyrancher 👏

codyrancher:
> > @aalves08 I'll just comment on these separately so we can have separate threads.
> > > 1. login still requires a page refresh
> > 
> > 
> > Are you doing anything special here? It seems to completely work for me.
> > logging-in.mp4
> 
> Nothing different, mate. Just testing this on `Chrome Version 120.0.6099.62 ` running your branch locally 🤷 If you want to get together to show you and help debug this, ping me

I finally figured out the issue with this one and why you were reproducing and I wasn't. The login issue was only occurring when an extension was installed.

codyrancher:
> Overall, it's just the problem with the login and a question around the routes uninstall removal, afaik.
> 
> Also, we should give this a run with Harvester before we merge it (navigate throught it + access old a new Harvester clusters) and make this pass the e2e gates.
> 
> Not for this PR, since it's quite extensive, but we should look into the possibility of removing `shell/components/nuxt/nuxt.js` and `shell/components/nuxt/nuxt-child.js` in a near future.
> 
> Great work @codyrancher 👏

I haven't used harvester at all so if you have any experience let me know what's required. I agree we should to a run through however.

shell/core/plugin-routes.ts
codyrancher: I tested the change when updating and we still prompt the user to reload the page so I think it's fine.

aalves08 (changes requested):
I've given this another round of testing and all of the previous flagged issues seem resolved, including the login problem. Kudos on finding the cause of that one @codyrancher  👏 

Today I got access to a system with a Harvester cluster imported and I've found that this branch has problems when refreshing a Harvester cluster route.

How it's supposed to behave


https://github.com/rancher/dashboard/assets/97888974/69d3c3c4-59b9-43b3-997e-1cb4ab8a826a


How it's behaving on this branch


https://github.com/rancher/dashboard/assets/97888974/6759cf7f-71b5-4971-ba88-7b33d7eb3dae

I can give you access to the system (just DM me).

A good test to do is also to try a dummy harvester cluster route, ex:

`https://localhost:8005/harvester/c/some-dummy-cluster-id/harvesterhci.io.dashboard#cluster-metrics`

and try to see if the redirect also works well.

codyrancher:
> I've given this another round of testing and all of the previous flagged issues seem resolved, including the login problem. Kudos on finding the cause of that one @codyrancher 👏
> 
> Today I got access to a system with a Harvester cluster imported and I've found that this branch has problems when refreshing a Harvester cluster route.
> 
> How it's supposed to behave
> Screen.Recording.2024-01-03.at.16.17.54.mov
> 
> How it's behaving on this branch
> Screen.Recording.2024-01-03.at.16.21.25.mov
> 
> I can give you access to the system (just DM me).
> 
> A good test to do is also to try a dummy harvester cluster route, ex:
> 
> `https://localhost:8005/harvester/c/some-dummy-cluster-id/harvesterhci.io.dashboard#cluster-metrics`
> 
> and try to see if the redirect also works well.

This is now resolved primarily with https://github.com/rancher/dashboard/pull/10084/files#diff-3cae65949742ad5fbccf99ac31dfdf9a51ab2c08f434b6ed67ec386c608bb632R281-R318

I really think our authentication and page loading accidentally work.... There's so many side effects being relied on to make things work.

shell/initialize/client.js
```diff
@@ -215,13 +204,14 @@ function callMiddleware(Components, context, layout) {
     if (layout.options.middleware) {
       midd = midd.concat(layout.options.middleware);
     }
-    Components.forEach((Component) => {
-      if (Component.options.middleware) {
-        midd = midd.concat(Component.options.middleware);
-      }
-    });
   }
```
jordojordo, line 207: Will `layout` ever be defined here?

shell/initialize/client.js
```diff
@@ -667,30 +620,6 @@ function addHotReload($component, depth) {
...
-        promise.then(() => {
-          this.setLayout(layout);
-          Vue.nextTick(() => hotReloadAPI(this));
-        });
-
-        return promise;
-      })
-
       .then(() => {
         return callMiddleware.call(this, Components, context, this.layout);
```
jordojordo, line 624: Is `this.layout` still necessary here?

jordojordo (changes requested):
A few general questions.

Also, should we remove the ability to add `layouts` from Extensions? I think you have already removed the loading of layouts from plugins but we are still [importing any layout directories here](https://github.com/rancher/dashboard/blob/7fce213bd89653a2c4249eba382e92b3ec607137/shell/pkg/auto-import.js#L3), not sure if this will cause issues.

shell/initialize/client.js
codyrancher: It won't. I removed this.layout entirely. Removing this as well as the layout argument from middleware.

shell/initialize/client.js
codyrancher: No, removed.

shell/core/plugin-routes.ts
```diff
@@ -107,87 +50,86 @@ export class PluginRoutes {
...
         if (foundParentRoute) {
-          pluginRoutesWithParents.push(r);
+          foundParentRoute.children = foundParentRoute?.children || [];
+          foundParentRoute.children.unshift(r.route);
         }
       }
 
       if (!foundParentRoute) {
-        orderedPluginRoutes.push(r.route);
+        orderedPluginRoutes.unshift(r.route);
```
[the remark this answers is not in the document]
codyrancher: So routing order matters in vue-router. We need to push the more specific routes for plugins to be before the default or the default will match first. I'll take a look at this elemental issue.

shell/components/templates/home.vue
```diff
@@ -109,6 +112,14 @@ export default {
...
       grid-area: header;
     }
+
+    &.dashboard-padding-left {
+      padding-left: $app-bar-collapsed-width;
+
+      .overlay-content-mode {
+        left: calc(var(--nav-width) + $app-bar-collapsed-width);
+      }
+    }
```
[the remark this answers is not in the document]
codyrancher: It's not. Removed.

codyrancher:
> A few general questions.
> 
> Also, should we remove the ability to add `layouts` from Extensions? I think you have already removed the loading of layouts from plugins but we are still [importing any layout directories here](https://github.com/rancher/dashboard/blob/7fce213bd89653a2c4249eba382e92b3ec607137/shell/pkg/auto-import.js#L3), not sure if this will cause issues.
We should

> A few general questions.
> 
> Also, should we remove the ability to add `layouts` from Extensions? I think you have already removed the loading of layouts from plugins but we are still [importing any layout directories here](https://github.com/rancher/dashboard/blob/7fce213bd89653a2c4249eba382e92b3ec607137/shell/pkg/auto-import.js#L3), not sure if this will cause issues.

We should remove it and I'll go ahead and remove the docs as well. I'll wait to do the new docs until we're close to merging in case anything changes.

shell/core/plugin-routes.ts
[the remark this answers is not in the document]
codyrancher: @jordojordo you were right about the cause. We still want to push the new routes to the front but because we were pushing them one at a time we changed the order which caused the issue you illustrated here.

We wanted:
`[...[new routes], ...[old routes]]`

We got:
`[...[setuor wen], ...[old routes]]`

jordojordo (approved):

jordojordo (approved):

PR #10103, opened 2023-11-30
Add automated workflow to publish @rancher/shell to npm

### LINKED ISSUE #9796: Add automated workflow to publish @rancher/shell to npm
Currently in order to publish a new version of the `@rancher/shell` package to npm, a user with access needs to manually run the `./scripts/publish-shell` script.

It would be nice to have a workflow that would run automatically when detecting a change to the `./shell/package.json` version to publish the new version of the the shell to npm.

aalves08:
### Summary
Fixes #9796 

- Add automated workflow to publish @rancher/shell to npm

### DIFF
--- a/.github/workflows/release-shell-pkg.yaml
+++ b/.github/workflows/release-shell-pkg.yaml
@@ -1,9 +1,10 @@
 name: Publish Shell Package
 
-on:
-  push:
-    tags:
-        - 'shell-pkg-v*'
+# on:
+#   push:
+#     tags:
+#         - 'shell-pkg-v*'
+on: workflow_dispatch
 
 jobs:
   build:
--- a/.github/workflows/release-shell-pkg.yaml
+++ b/.github/workflows/release-shell-pkg.yaml
@@ -1,0 +1,33 @@
+name: Publish Shell Package
+
+on:
+  push:
+    tags:
+        - 'shell-pkg-v*'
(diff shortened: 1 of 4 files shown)


.github/workflows/release-shell-pkg.yaml
```diff
@@ -5,7 +5,7 @@ name: Publish Shell Package
 #     tags:
 #         - 'shell-pkg-v*'
 on:
-  pull_request:
+  pull_request_target:
```
rak-phillip, line 8: Ah, so the `pull_request_target` trigger provides access to secrets because it runs in the context of the target repo. That definitely makes sense.

Now that I'm thinking about this holistically, I'm wondering if pull requests are where we would want to publish shell releases. I would expect a release or a tag to trigger the publish to npm. Additionally, it appears that there are some security considerations to keep in mind when using this trigger

> Warning: For workflows that are triggered by the pull_request_target event, the GITHUB_TOKEN is granted read/write repository permission unless the permissions key is specified and the workflow can access secrets, even when it is triggered from a fork. Although the workflow runs in the context of the base of the pull request, you should make sure that you do not check out, build, or run untrusted code from the pull request with this event. [^1]

[^1]: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

PR #10107, opened 2023-12-01
Fix select does not select default storage class

### LINKED ISSUE #10090: Rancher UI error reading 'storageclass.kubernetes.io/is-default-class' if a storageclass doesn't have an annotations block
Internal Reference: SURE-7199

When installing the Rancher Backups chart via the Rancher UI on the local cluster, you get an error when trying to select "Use an existing storage class" when you have a storage class without annotations.

Steps to reproduce - on the local cluster:

1. import the following YAML to create a new Storage Class, using the 'Import YAML' feature:

```
allowVolumeExpansion: false
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: test-annotations
parameters:
  encrypted: 'true'
  iopsPerGB: '0'
  type: gp2
provisioner: kubernetes.io/aws-ebs
reclaimPolicy: Delete
volumeBindingMode: Immediate
```
3. Go to Charts, Rancher Backups and click Install
4. Click Next to go to 'Edit Options'
5. Click on 'Use an existing storage class' under 'Default Storage Location'

Browser console logs shows an error:

```
TypeError: Cannot read properties of undefined (reading 'storageclass.kubernetes.io/is-default-class')
    at index.vue:68:1
    at Array.filter (<anonymous>)
    at a.defaultStorageClass (index.vue:68:1)
...
```

nwmac:
Fixes #10090 

This is a simple bug fix - the UI assumed that a storage class has annotations - if it does not, the default storage class is not auto-selected on the Rancher Backups chart install page.

This used to cause a full-screen error page due to another bug that has been fixed.

This PR fixes the remaining auto-selection bug.

Added an automated e2e test to cover regression testing.

### DIFF
--- a/cypress/e2e/blueprints/charts/rancher-backup-chart.ts
+++ b/cypress/e2e/blueprints/charts/rancher-backup-chart.ts
@@ -1,0 +1,35 @@
+
+// No annotatiosn - test case that broke auto-selection of the default storage class
+export const exampleStorageClass = {
+  allowVolumeExpansion: false,
+  apiVersion:           'storage.k8s.io/v1',
+  kind:                 'StorageClass',
+  metadata:             { name: 'test-no-annotations' },
+  parameters:           {
+    encrypted: 'true',
+    iopsPerGB: '0',
+    type:      'gp2'
+  },
+  provisioner:       'kubernetes.io/aws-ebs',
+  reclaimPolicy:     'Delete',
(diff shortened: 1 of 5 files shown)


nwmac:
@aalves08 Rebased and bumped - all passing gates now - can you review again. Thanks.

aalves08 (approved):
LGTM

Issue #9978, opened 2023-10-24
[BUG] Cannot read properties of undefined (reading 'cloudprovider.harvesterhci.io/ipam')


skaven81:
**Rancher Server Setup**
- Rancher version: 2.7.6
- Installation option (Docker install/Helm Chart): Helm chart on RKE1 1.23.16
- Proxy/Cert Details: Rancher manages TLS termination via a TCP load balancer

**Information about the Cluster**
- Kubernetes version: 1.23.16
- Cluster Type (Local/Downstream): downstream RKE1 Rancher "custom"

**User Information**
- What is the role of the user logged in? Admin

**Describe the bug**
Both admins and unprivileged users trying to access the configuration panel for a specific workload (which was created using the Rancher UI) get a Vue error and a long traceback in the browser console.

**To Reproduce**
1. Log into Rancher
2.  Access the cluster
3. Filter to the workload's Project
4. Visit the "Workloads" page
5. Click the name of the workload (a Deployment)
6. Click the Config button at the top right

**Result**
Vue error appears, "Cannot read properties of undefined (reading 'cloudprovider.harvesterhci.io/ipam')"

**Expected Result**
The config page appears.

**Screenshots**
![image](https://github.com/rancher/rancher/assets/5861261/842c1317-f9e6-431c-9216-f32594f9b678)

**Additional context**
![image](https://github.com/rancher/rancher/assets/5861261/5cf39b47-3fc2-4eb9-b905-130b79571b9c)

![image](https://github.com/rancher/rancher/assets/5861261/8b429612-df30-4c16-85fd-24c29f628052)

```
console.js:31  Cannot read properties of undefined (reading 'cloudprovider.harvesterhci.io/ipam')
e.forEach.console.<computed> @ console.js:31
(anonymous) @ client.js:111
p @ regeneratorRuntime.js:44
(anonymous) @ regeneratorRuntime.js:125
(anonymous) @ regeneratorRuntime.js:69
r @ asyncToGenerator.js:3
s @ asyncToGenerator.js:22
(anonymous) @ asyncToGenerator.js:27
(anonymous) @ asyncToGenerator.js:19
(anonymous) @ client.js:113
$n @ vue.runtime.esm.js:3031
Bn @ vue.runtime.esm.js:3008
Hn @ vue.runtime.esm.js:3024
Hr @ vue.runtime.esm.js:4031
Qi.t._init @ vue.runtime.esm.js:5693
a @ vue.runtime.esm.js:5826
[...]

PR #10110, opened 2023-12-04
Fix err when save deployment and service annotations is null

hhstu:
### Summary
Fixes https://github.com/rancher/dashboard/issues/9978 fix err when save deployment and  service annotations is null


torchiaf (approved):
LGTM

nwmac:
@torchiaf You can go ahead and merge as the author doesn't have permission to do so.

PR #10111, opened 2023-12-04
Hide provider labels for Harvester clusters

### LINKED ISSUE #9892: Visual bug on the node pool view "Unknown - no location"
Internal reference: SURE-7049
Reported in 2.7.6

Beneath the title of the pool (Cluster Management -> Click on one of the clusters), there is a grey subtitle that contains the following: Unknown – No Location / No Size ($RESOURCE_POOL_NAME). 

This is confusing for users about the relevance, which looks to be tied to broken metrics in that the correct location isn't being shown.

Screenshot available upon request.

torchiaf:
### Summary
Fixes #9892

### Technical notes summary

`isHarvester` could be also calculated by checking the provider name of the cluster.
Other solution to show the label: find required information for Harvester machine pools = `this.machines.pool`

Clusters details page

### DIFF
--- a/shell/detail/provisioning.cattle.io.cluster.vue
+++ b/shell/detail/provisioning.cattle.io.cluster.vue
@@ -775,10 +775,10 @@ export default {
                   v-clean-html="t('resourceTable.groupLabel.notInANodePool')"
                 />
                 <div
-                  v-if="group.ref && group.ref.template"
+                  v-if="group.ref && group.ref.providerSummary"
                   class="description text-muted text-small"
                 >
-                  {{ group.ref.providerDisplay }} &ndash;  {{ group.ref.providerLocation }} / {{ group.ref.providerSize }} ({{ group.ref.providerName }})
+                  {{ group.ref.providerSummary }}
                 </div>
               </div>
               <div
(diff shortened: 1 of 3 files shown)


shell/detail/provisioning.cattle.io.cluster.vue
```diff
@@ -775,7 +775,7 @@ export default {
                   v-clean-html="t('resourceTable.groupLabel.notInANodePool')"
                 />
                 <div
-                  v-if="group.ref && group.ref.template"
+                  v-if="group.ref && group.ref.template && !group.ref.isHarvester"
```
nwmac, line 778: I wonder if we could call this `providerSummary` (or something) and have that as a getter that returns the formatted summary.

The default on the machinedeployment model would be to return the HTML as currently formatted unless its Harvester in which case it could return undefined - this if could then be conditional on `providerSummary` being truthy

torchiaf: Updated by adding the new `providerSummary` getter, thanks!

mantis-toboggan-md (approved):
lgtm

PR #10114, opened 2023-12-05
Move extensions gitlab workflow

jordojordo:
### Summary
This moves the directories `.gitlab/workflows` into the `shell/scripts` directory and updates the extensions gitlab-ci remote url to the new location.

Also bumps the `@rancher/shell` package version to `0.4.1`.


nwmac (commented):
Shouldn't the app creator copy the GitLab workflow when you create a new app and then reference the workfload in the repo, rather than referencing the one from the dashboard repo ?

jordojordo:
We used this approach for versioning the workflow, this is mentioned in the [dev docs](https://rancher.github.io/dashboard/extensions/publishing#gitlab-integration) - but if someone wanted to use it in that way they could copy the workflow into their own ci file rather than referencing the dashboard repo version.

nwmac (approved):

PR #10119, opened 2023-12-07
remove "process." mentions from Dashboard and adjust code

### LINKED ISSUE #10120: Remove all "process." mentions in the code

Remove all of the `process.client`, `process.server` and `process.static` mentions from Dashboard code as one of the measures to remove Nuxt related code from Rancher Dashboard codebase.

aalves08:
### Summary
Fixes #10120 

- Remove all of the `process.client`, `process.server` and `process.static` mentions from Dashboard and adjust code.

Notes: 
- `process.client` was being set to `true` by default: https://github.com/rancher/dashboard/blob/master/shell/vue.config.js#L390
- `process.server` was exclusively related to SSR
- `process.static` seemed to be tied to nuxt generate, which is related to pre-rendering and was only `true` when the command `nuxt generate` ran, which we don't have on the codebase anymore

### DIFF
--- a/shell/plugins/steve/subscribe.js
+++ b/shell/plugins/steve/subscribe.js
@@ -597,9 +597,9 @@ const defaultActions = {
   },
 
   rehydrateSubscribe({ state, dispatch }) {
-    // if ( state.wantSocket && !state.socket ) {
-    //   dispatch('subscribe');
-    // }
+    if ( state.wantSocket && !state.socket ) {
+      dispatch('subscribe');
+    }
   },
 
   reconnectWatches({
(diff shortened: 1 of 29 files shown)


shell/plugins/steve/subscribe.js
```diff
@@ -601,9 +597,9 @@ const defaultActions = {
   },
 
   rehydrateSubscribe({ state, dispatch }) {
-    if ( process.client && state.wantSocket && !state.socket ) {
-      dispatch('subscribe');
-    }
+    // if ( state.wantSocket && !state.socket ) {
```
codyrancher, line 600: Looks like this needs to come back

aalves08: you are correct @codyrancher . Thanks for spotting this 🙏

codyrancher (approved):

PR #10126, opened 2023-12-11
remove page transitions from Dashboard

### LINKED ISSUE #10124: remove transitions (page transitions)

Remove mentions/logic related to page transitions

aalves08:
### Summary
Fixes #10124 

- remove page transitions from Dashboard

### Technical notes summary

- seems like the package `vue2-transitions` wasn't needed anymore, so I proceeded to remove it
-

### DIFF
--- a/shell/components/nuxt/nuxt-child.js
+++ b/shell/components/nuxt/nuxt-child.js
@@ -53,7 +53,7 @@ export default {
       routerView = h('keep-alive', { props: props.keepAliveProps }, [routerView])
     }
 
-    return h('transition', {
+    return h('div', {
       on: listeners
     }, [routerView])
   }
--- a/shell/components/nuxt/nuxt-child.js
+++ b/shell/components/nuxt/nuxt-child.js
@@ -29,11 +29,6 @@ export default {
     data.nuxtChildDepth = depth
 
     const listeners = {}
-    listenersKeys.forEach((key) => {
-      if (typeof transition[key] === 'function') {
-        listeners[key] = transition[key].bind(_parent)
-      }
-    })
     
     // Add triggerScroll event on beforeEnter (fix #1376)
     const beforeEnter = listeners.beforeEnter
--- a/shell/components/nuxt/nuxt-child.js
+++ b/shell/components/nuxt/nuxt-child.js
@@ -48,7 +48,10 @@ export default {
       routerView = h('keep-alive', { props: props.keepAliveProps }, [routerView])
     }
 
-    return h('div', {
+    // this needs to be a "transition" or another non-rendering component, 
+    // otherwise we will break pages like the charts wizard or the extensions main screen (DOM would render an additional element and break CSS)
+    // we can deal with this later once we remove this component and <nuxt /> component
+    return h('transition', {
       on: listeners
     }, [routerView])
   }
--- a/shell/components/nuxt/nuxt-child.js
+++ b/shell/components/nuxt/nuxt-child.js
@@ -18,10 +18,8 @@ export default {
 
     data.nuxtChild = true
     const _parent = parent
-    const transitions = parent.$nuxt.nuxt.transitions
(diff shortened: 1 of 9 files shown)


shell/components/nuxt/nuxt-child.js
```diff
@@ -57,66 +43,14 @@ export default {
...
-      }
-    }
-
     let routerView = h('routerView', data)
 
     if (props.keepAlive) {
       routerView = h('keep-alive', { props: props.keepAliveProps }, [routerView])
     }
 
     return h('transition', {
```
codyrancher, line 52: Is this still needed?

shell/initialize/App.js
```diff
@@ -21,25 +21,10 @@ export default {
...
-    const transitionEl = h('transition', {
-      props: {
-        name: 'layout',
-        mode: 'out-in'
-      },
-      on: {
-        beforeEnter(el) {
-          // Ensure to trigger scroll event after calling scrollBehavior
-          window.$nuxt.$nextTick(() => {
-            window.$nuxt.$emit('triggerScroll');
```
codyrancher, line 33: I think you'll still need to emit this event on page transition.

It looks like there's some transitions which won't scroll to the top of the page when switching pages.


https://github.com/rancher/dashboard/assets/55104481/3c034cb9-f914-46f8-bdff-5f209eb52af7

shell/components/nuxt/nuxt-child.js
aalves08: Not really... Can be just a plain `div` afaik.

shell/initialize/App.js
aalves08: Well, this one was driving me crazy but I just realized that it never worked! I am running a 2.8.0-rc3 instance and the UI that is built with it fails on the move from Pods to Workloads (or vice versa... one of these two still has the problem).
I haven't been able to fix it but I think it's because the Pods list view is essentially the same as Workloads, although interestingly it does trigger a route change and everything seems to be correct through my investigation in https://github.com/rancher/dashboard/blob/master/shell/utils/router.scrollBehavior.js

aalves08: Fixed it! 💦  it wasn't registered as subtype of workloads... man, this one was tough

codyrancher: Nice, find. Well done.

codyrancher (dismissed):

shell/config/product/explorer.js
```diff
@@ -103,6 +103,9 @@ export function init(store) {
     componentForType(WORKLOAD_TYPES[key], WORKLOAD);
   }
 
+  // Deal with POD (it wasn't registered as belonging to the workloads, so it wasn't using the list view of workloads)
+  componentForType(POD, WORKLOAD);
```
richard-cox, line 107: Pods both are and aren't workloads. Why is this needed?

Importantly though they have their own detail and edit components which are now bypassed with this line. interestingly edit works, but detail is broken.

richard-cox (changes requested):
- The extension page is broken, could be an issue with various layouts? 
  image

aalves08:
> * The extension page is broken, could be an issue with various layouts?
>   image

Unfortunately I found this today, along with the Charts installation wizard (our e2e tests got it). I now understand why and will add a fix for it. It reverted all of broken screens to the correct state.

shell/config/product/explorer.js
aalves08: It was to try and fix https://github.com/rancher/dashboard/pull/10126#discussion_r1423204878 . This one was a "bug" that already existed. I prefer to revert this change and have the "same old" bug rather than breaking a complete page.

The problem is that the PODS list view currently uses a default list view and I've just noticed that if you switch between resource list INSIDE the same PRODUCT to another that uses the default list view it doesn't re-adjust the scrolling.

Yesterday I thought I understood why, but apparently I still don't 😅 

I'll revert this change.

richard-cox (approved):

PR #10144, opened 2023-12-12
Initial rancher-api implementation

### LINKED ISSUE #10097: Create Rancher API for extensions to access/manage resources
It would be nice to have an API to easily access and manage resources rather than calling store functions. This would make it nicer for extensions as they would not need to dive into the dashboard store code to know what functions are available.

The API should be declarative and discoverable.

jordojordo:
### Summary
Fixes #10097 

This is the initial proposal for the rancher-api spec with a few methods on the cluster and rancher classes to interact with cluster based resources.

Initialized by creating a Vue prototype for each endpoint:
- `$cluster`
  - fetching/managing cluster level resources
- `$rancher`
  - when interacting with the management/local or Rancher cluster
- `$shell`
  - to interact with `@rancher/shell` components/methods
- `$extension`
  - for extension product related methods
  
This allows for Intellisense on all methods within each class for proper definitions and arguments of each method. Also adds some basic types.


shell/initialize/index.js
```diff
@@ -121,6 +123,12 @@ async function createApp(ssrContext, config = {}) {
   // Add this.$router into store actions/mutations
   store.$router = router;
 
+  // Add racher-api classes to Vue prototype
+  Vue.use(rancherApiPlugin, { store });
+  Vue.use(clusterApiPlugin, { store });
+  Vue.use(shellApiPlugin, { store });
+  Vue.use(extensionApiPlugin, { store });
```
codyrancher, line 130: Will these make the API available in the bootstraping methods of the extensions? If so, isn't there a possibility that some of the stores won't be loaded yet? We might have to make that clear or error out gracefully if that's the case.

shell/plugins/rancher-api/base-cluster-class.ts
```diff
@@ -0,0 +1,135 @@
...
+   *                The format is 'key=value', where 'key' is the label name and 'value' is the label value.
+   *                For example, 'app=kubewarden-policy-server-default' would match resources labeled with
+   *                'app' as 'kubewarden-policy-server-default'.
+   *  - `namespace`: Specify the namespace of the resource (if applicable).
+   *  - `force`: Force a new dispatch request, bypassing any cached data.
+   *
+   * @returns `Promise<any>` A promise that resolves with the fetched resource(s).
+   */
+  async get(type: string, options?: ResourceFetchOptions): Promise<any> {
+    if ( this.getSchema(type) ) {
```
codyrancher, line 112: We probably want to inform the consumer if the schema doesn't exist instead of failing silently. I notice this pattern in other parts of this pr as well.

shell/plugins/rancher-api/base-cluster-class.ts
```diff
@@ -0,0 +1,135 @@
...
+   *  - `selector`: A string used to match resources based on their labels.
+   *                The format is 'key=value', where 'key' is the label name and 'value' is the label value.
+   *                For example, 'app=kubewarden-policy-server-default' would match resources labeled with
+   *                'app' as 'kubewarden-policy-server-default'.
+   *  - `namespace`: Specify the namespace of the resource (if applicable).
+   *  - `force`: Force a new dispatch request, bypassing any cached data.
+   *
+   * @returns `Promise<any>` A promise that resolves with the fetched resource(s).
+   */
+  async get(type: string, options?: ResourceFetchOptions): Promise<any> {
```
codyrancher, line 111: It's not obvious to me by looking at this signature how I should retrieve all items of a resource (from looking at the code it's obvious) which I think is problematic. Maybe changing options to filter instead would make that clear but I suspect we'll want to leave it as options to support pagination in the future.

I think I'd be in favor of splitting the functionality of findAll and findMatching/find into two different methods. If there's a better idea for making the usage obvious I'd be fine with something else.

shell/plugins/rancher-api/cluster-api-class.ts
```diff
@@ -0,0 +1,90 @@
...
+ * ClusterApi extends the BaseClusterApi to provide specific functionalities for interacting with
+ * resources within the current cluster context. It includes methods to create, edit, and delete
+ * resources. This class is specifically designed for the $cluster context and should not be
+ * confused with CAPI.
+ */
+
+import { SteveResource, ResourceManageOptions } from '@shell/types/rancher-api/cluster';
+import BaseClusterApi from './base-cluster-class';
+
+export default class ClusterApi extends BaseClusterApi {
```
codyrancher, line 11: I have a suspicion we'll want to expose more of the methods from shell/store.index.js.

A few examples:
getCluster
getClusterId
isRancher

codyrancher (commented):
I like the groups you split the api into.

It looks like all of these will be available in components. Do you intend to expose any of these to the bootstraping methods of extensions?

shell/initialize/index.js
jordojordo: This doesn't make the API available to extensions after some initial testing, I don't think the types are being properly referenced in an extension.

Perhaps creating a helper function to check for the existence of the store before running any commands wouldn't be a bad idea, but I believe the plugins (extensions) are initialized on line 324 which is after both the store and prototypes are created:
```js
  if (process.client && typeof plugin === 'function') {
    await plugin(app.context, inject);
  }
```

shell/plugins/rancher-api/cluster-api-class.ts
jordojordo: I decided not to dive too into the weeds with implementation before we agreed on a proper plan of this API.

shell/plugins/rancher-api/base-cluster-class.ts
jordojordo: I was trying to emulate how running a `kubectl get <resource>` would return all of a resource (sans the namespace), maybe this isn't the best approach. I figured having Intellisense that explains the design of the function and options in detail would provide clear instructions on how to use the method.

I'm open to splitting the functionality as well, this may help limit confusion with further implementation.

nwmac:
@jordojordo As discussed, I think we need to:

- Split this into a few PRs - one for the base changes to support adding different APIs and then PRs for the APIs
- Review each API - i think for some like shell, we should namespace the functionality, e.g. `$shell.dialog.<api>` etc - so we can richer APIs with a partitioning via the 'namespace'
- Include an API doc generator, so we get API docs for these APIs in the extensions docs - should look at `TypeDoc` and `https://typedoc-plugin-markdown.org/plugins/docusaurus`

jordojordo:
Closing in favor of #13839

PR #10155, opened 2023-12-14
Add automated workflow to publish @rancher/shell to npm

aalves08:
### Summary
Fixes #10103

- Add workflow to publish `shell` package to NPM with a release tag

### DIFF
--- a/.github/workflows/release-shell-pkg.yaml
+++ b/.github/workflows/release-shell-pkg.yaml
@@ -1,13 +1,9 @@
 name: Publish Shell Package
 
-#
-# on:
-#   push:
-#     tags:
-#         - 'shell-pkg-v*'
 on:
-  pull_request:
-    # types: [converted_to_draft, ready_for_review]
+  push:
+    tags:
+        - 'shell-pkg-v*'
 
 jobs:
   build:
@@ -21,19 +17,19 @@ jobs:
         fetch-depth: 0
         persist-credentials: false
 
-    # - uses: actions/setup-node@v3
-    #   with:
-    #     node-version: '16.x'
+    - uses: actions/setup-node@v3
+      with:
+        node-version: '16.x'
 
-    # - name: Validate Plugin build system
-    #   run: ./shell/scripts/test-plugins-build.sh
-    #   shell: bash
+    - name: Validate Plugin build system
+      run: ./shell/scripts/test-plugins-build.sh
+      shell: bash
 
-    # - name: Reset repository (file system)
-    #   run: | 
-    #     git reset --hard HEAD
-    #     echo $YARN_REGISTRY
-    #     cat ./shell/package.json
+    - name: Reset repository (file system)
+      run: | 
+        git reset --hard HEAD
+        echo $YARN_REGISTRY
+        cat ./shell/package.json
 
     - uses: actions/setup-node@v3
       with:
--- a/.github/workflows/release-shell-pkg.yaml
+++ b/.github/workflows/release-shell-pkg.yaml
@@ -25,6 +25,10 @@ jobs:
       run: ./shell/scripts/test-plugins-build.sh
       shell: bash
 
+    # Reset the local (ci) repository state because
+    # The previous step (Validate Plugin build system) changes
+    # the version number on package.json for Shell in order
+    # to release it in verdaccio for it's tests
     - name: Reset repository (file system)
       run: | 
         git reset --hard HEAD
--- a/.github/workflows/release-shell-pkg.yaml
+++ b/.github/workflows/release-shell-pkg.yaml
@@ -7,6 +7,7 @@ name: Publish Shell Package
 #         - 'shell-pkg-v*'
 on:
   pull_request_target:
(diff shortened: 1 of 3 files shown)


.github/workflows/release-shell-pkg.yaml
```diff
@@ -0,0 +1,48 @@
+name: Publish Shell Package
+
+# on:
+#   push:
+#     tags:
+#         - 'shell-pkg-v*'
+on:
+  pull_request_target:
```
rak-phillip, line 8: Ah, so the `pull_request_target` trigger provides access to secrets because it runs in the context of the target repo. That definitely makes sense.

Now that I'm thinking about this holistically, I'm wondering if pull requests are where we would want to publish shell releases. I would expect a release or a tag to trigger the publish to npm. Additionally, it appears that there are some security considerations to keep in mind when using this trigger

> Warning: For workflows that are triggered by the pull_request_target event, the GITHUB_TOKEN is granted read/write repository permission unless the permissions key is specified and the workflow can access secrets, even when it is triggered from a fork. Although the workflow runs in the context of the base of the pull request, you should make sure that you do not check out, build, or run untrusted code from the pull request with this event. [^1]

[^1]: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

I originally left this comment in https://github.com/rancher/dashboard/pull/10103#discussion_r1426787558

aalves08: Yeah, the trigger is a tag like Rancher Components. I just wanna test it in a simpler way (on a PR). If it works, then I can finish (uncomment) the code on the workflow and just finalise the PR

aalves08: Kudos to Neil for discovering this 👏 👏 👏

aalves08:
@rak-phillip it worked! https://github.com/rancher/dashboard/actions/runs/7210456092/job/19643640369

Going to adjust this PR to reflect the correct trigger and steps and ask you for a review 🙏 🙇  Thank you for the time spent on it 🙇

aalves08:
FYI @nwmac @jordojordo @codyrancher @richard-cox

package.json
```diff
@@ -17,6 +17,7 @@
     "publish-pkg": "./shell/scripts/publish-pkg.sh",
     "serve-pkgs": "./shell/scripts/serve-pkgs",
     "publish-shell": "./shell/scripts/publish-shell.sh",
```
rak-phillip, line 19: Do we intend to remove `publish-shell` in the future?

aalves08: Good point. We can remove the script mention, but we will leave the shell script in

.github/workflows/release-shell-pkg.yaml
```diff
@@ -0,0 +1,46 @@
...
+        cat ./shell/package.json
+
+    - uses: actions/setup-node@v3
+      with:
+        node-version: '16.x'
+        registry-url: 'https://registry.npmjs.org'
+        scope: '@rancher'
+    
+    - name: Install packages
+      run: yarn install
```
rak-phillip, line 41: I think that we should be using `yarn install --frozen-lockfile` in ci. 

> If you need reproducible dependencies, which is usually the case with the continuous integration systems, you should pass --frozen-lockfile flag.

.github/workflows/release-shell-pkg.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+    - name: Validate Plugin build system
+      run: ./shell/scripts/test-plugins-build.sh
+      shell: bash
+
+    - name: Reset repository (file system)
+      run: | 
+        git reset --hard HEAD
+        echo $YARN_REGISTRY
+        cat ./shell/package.json
```
rak-phillip, line 32: I'm a little confused by this step, why is is required to run `git reset --hard HEAD` before we publish?

rak-phillip (commented):
I left two more comments. I think this is looking good 🙂

.github/workflows/release-shell-pkg.yaml
aalves08: Because the step `Validate Plugin build system` messes with the file system in the pipeline (especially the version number on `package.json` so that it can be published to verdaccio to do it's checks).

We don't want to mess with that, so resetting the repo locally seems to be a good candidate.

Another alternative would be to do another checkout step (I think... I don't believe I tested it).

This is a keeper

.github/workflows/release-shell-pkg.yaml
aalves08: updated. Good one

.github/workflows/release-shell-pkg.yaml
rak-phillip: Shall we leave a comment that describes this behavior for posterity? As it stands, this is the only thing in this action that is potentially confusing.

aalves08: Added :)

rak-phillip (approved):
Looks great 🚀

PR #10161, opened 2023-12-14
Cluster management

yonasberhe23:
### Summary
This pr covers the functionality within Cluster Management (Cloud Credentials, Drivers, Advanced options, etc.)
View Test Plan [here](https://github.com/rancher/qa-tasks/issues/735)

### Technical notes summary
- `Cloud Credentials` and `Node Templates` tests will be filtered out of Git Actions test runs using `@jenkins` tag. Those specs rely on secrets that are stored in Jenkins therefore they will only be run in the Jenkins pipeline.
- This PR relies on a code change in the corral-packages repo which gives our tests access to aws cloud credentials https://github.com/rancherlabs/corral-packages/pull/33 (I will not merged this PR until the associated pr is approved/merged)


cypress/e2e/po/components/sortable-table.po.ts
```diff
@@ -56,6 +56,14 @@ export default class SortableTablePo extends ComponentPo {
     return cy.getId('sortable-table-promptRemove');
   }
 
+  downloadYamlButton() {
+    return cy.getId('sortable-table-download');
+  }
+
+  downloadKubeConfigButton() {
```
richard-cox, line 63: this is specific to the prov cluster table and should be in that po.

slightly longer, but could also use rowActionMenuOpen(row name).getMenuItem('Download KubeConfig')

cypress/e2e/po/edit/cloud-credentials.po.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  name(): LabeledInputPo {
+    return LabeledInputPo.byLabel(this.self(), 'Name');
+  }
+
+  description(): LabeledInputPo {
+    return LabeledInputPo.byLabel(this.self(), 'Description');
+  }
+
+  accessKey(): LabeledInputPo {
```
richard-cox, line 34: accessKey, secretKey and defaultRegion are specific to the type of cloud cred created. would suggest a parent class for that type with these three in that extending this base po

cypress/e2e/po/edit/machine-deployments.po.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  static goTo(path: string): Cypress.Chainable<Cypress.AUTWindow> {
+    throw new Error('invalid');
+  }
+
+  constructor(clusterId = 'local', nsName?: string, machineSetName?: string) {
+    super(MachineDeploymentsCreateEditPo.createPath(clusterId, nsName, machineSetName));
+  }
+
+  saveCreateForm(): AsyncButtonPo {
```
richard-cox, line 19: In term terms of component hierarchy it goes ResourceDetail --> ResourceYaml --> Footer (which contains the button).

So would suggest either creating a Footer PO (containing the button) and adding it to ResourceYamlPo, or just adding this function to ResourceYamlPo. In both cases you would access them via  `ResourceDetailPo` `.resourceYaml`

This pattern also applies to the other machine resources.

cypress/e2e/po/edit/cloud-credentials.po.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  secretKey(): LabeledInputPo {
+    return LabeledInputPo.byLabel(this.self(), 'Secret Key');
+  }
+
+  defaultRegion(): LabeledSelectPo {
+    return new LabeledSelectPo('.vs__dropdown-toggle');
+  }
+
+  saveCreateForm(): AsyncButtonPo {
```
richard-cox, line 46: Comment out of order, see saveCreateForm in machine-x po.

This pattern has crept in here too.

These go ResourceDetail --> CruResource. So the save button should be in CruResourcePo and accessed via ResourceDetailPo.

This pattern applies in other places `data-testid="form-save` is used

cypress/e2e/po/edit/cloud-credentials.po.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  constructor(clusterId = 'local', id?: string) {
+    super(CloudCredentialsCreateEditPo.createPath(clusterId, id));
+  }
+
+  cloudServiceOptions(): CruResourcePo {
+    return new CruResourcePo('[data-testid="cru-form"]');
+  }
+
+  name(): LabeledInputPo {
```
richard-cox, line 26: name and description (and in other places namespace) all come from a `NameNsDescription` component, so should be accessed via the `NameNsDescription` PO

cypress/e2e/po/edit/pod-security-admissions.po.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  static goTo(path: string): Cypress.Chainable<Cypress.AUTWindow> {
+    throw new Error('invalid');
+  }
+
+  constructor(clusterId = 'local', id?: string) {
+    super(PodSecurityAdmissionsCreateEditPo.createPath(clusterId, id));
+  }
+
+  name(): LabeledInputPo {
```
richard-cox, line 20: Access via NameNsDescription PO

cypress/e2e/po/edit/pod-security-admissions.po.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  description(): LabeledInputPo {
+    return LabeledInputPo.byLabel(this.self(), 'Description');
+  }
+
+  saveCreateForm(): AsyncButtonPo {
+    return new AsyncButtonPo('[data-testid="form-save"]', this.self());
+  }
+
+  saveAndWaitForRequests(method: string, url: string) {
```
richard-cox, line 32: i think these kind of props can also go in the same PO as the save button

cypress/e2e/po/edit/repositories.po.ts
```diff
@@ -0,0 +1,56 @@
+import PagePo from '@/cypress/e2e/po/pages/page.po';
+import LabeledInputPo from '@/cypress/e2e/po/components/labeled-input.po';
+import AsyncButtonPo from '@/cypress/e2e/po/components/async-button.po';
+import LabeledSelectPo from '@/cypress/e2e/po/components/labeled-select.po';
+import RadioGroupInputPo from '@/cypress/e2e/po/components/radio-group-input.po';
+
+export default class RepositoriesCreateEditPo extends PagePo {
```
richard-cox, line 7: There's a couple of different `Repositories` concepts, would suggest this one is referred to as `ChartRepositories` (as it's specific to helm chart repos)

cypress/e2e/po/lists/repositories-list.po.ts
```diff
@@ -0,0 +1,7 @@
+import BaseResourceList from '@/cypress/e2e/po/lists/base-resource-list.po';
+
+export default class RepositoriesListPo extends BaseResourceList {
```
richard-cox, line 3: same naming convention for this one, suggestion was ChartRepositories

cypress/e2e/po/pages/cluster-manager/machine-deployments.po.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  static navTo() {
+    const sideNav = new ProductNavPo();
+
+    BurgerMenuPo.burgerMenuNavToMenubyLabel('Cluster Management');
+    sideNav.groups().contains('Advanced').click();
+    sideNav.navToSideMenuEntryByLabel('MachineDeployments');
+  }
+
+  create() {
```
richard-cox, line 29: the create is part of `ResourceListMastheadPo`. added there it can be accessed via `this.list().masthead()`. applies to other machine-x pages

cypress/e2e/po/pages/cluster-manager/node-templates.po.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  list(): EmberSortableTablePo {
+    return new EmberSortableTablePo('.sortable-table');
+  }
+
+  addTemplate() {
+    return cy.iFrame().contains('.right-buttons .btn', 'Add Template');
+  }
+
+  bulkActions(label: string) {
```
richard-cox, line 42: this and actionMenu should be part of the ember list

cypress/e2e/po/pages/cluster-manager/node-templates.po.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  actionMenu(): EmberSelectPo {
+    return new EmberSelectPo('.ember-basic-dropdown-content');
+  }
+
+  form(): EmberFormRkeTemplatesPo {
+    return new EmberFormRkeTemplatesPo('form.horizontal-form ');
+  }
+
+  formActions(): EmberFormRkeTemplatesPo {
```
richard-cox, line 54: what's the different between form and formActions? they both return EmberFormRkeTemplatesPo

cypress/e2e/po/pages/cluster-manager/pod-security-admissions.po.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  static navTo() {
+    const sideNav = new ProductNavPo();
+
+    BurgerMenuPo.burgerMenuNavToMenubyLabel('Cluster Management');
+    sideNav.groups().contains('Advanced').click();
+    sideNav.navToSideMenuEntryByLabel('Pod Security Admissions');
+  }
+
+  create() {
```
richard-cox, line 27: similar to `masthead-create-yaml"` this should be accessed via ResourceListMastheadPo

cypress/e2e/po/pages/cluster-manager/rke-templates.po.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  mainRow(): EmberSortableTablePo {
+    return new EmberSortableTablePo('table.sortable-table .main-row');
+  }
+
+  actionMenu(): EmberSelectPo {
+    return new EmberSelectPo('.ember-basic-dropdown-content');
+  }
+
+  form(): EmberFormRkeTemplatesPo {
```
richard-cox, line 45: form and formActions return the same object

cypress/e2e/po/pages/extensions.po.ts
```diff
@@ -4,7 +4,7 @@ import LabeledSelectPo from '@/cypress/e2e/po/components/labeled-select.po';
 import TabbedPo from '@/cypress/e2e/po/components/tabbed.po';
 import ActionMenuPo from '@/cypress/e2e/po/components/action-menu.po';
 import NameNsDescriptionPo from '@/cypress/e2e/po/components/name-ns-description.po';
-import ReposListPagePo from '@/cypress/e2e/po/pages/repositories.po';
+import RepositoriesPagePo from '@/cypress/e2e/po/pages/cluster-manager/repositories.po';
```
richard-cox, line 7: ok, i see the weirdness now.

`repositories` should be `chart-repositories` and remain in the generic place. that will cover the helm chart repo list in both the cluster manager and explorer products

richard-cox (changes requested):
Think i've reached the half way mark and reviewed all non-spec files. Will return and do those soon

cypress/e2e/tests/pages/manager/machine-deployments.spec.ts
```diff
@@ -0,0 +1,157 @@
+import PromptRemove from '@/cypress/e2e/po/prompts/promptRemove.po';
+import MachineDeploymentsPagePo from '@/cypress/e2e/po/pages/cluster-manager/machine-deployments.po';
+import * as path from 'path';
+import * as jsyaml from 'js-yaml';
+
+describe('MachineDeployments', { testIsolation: 'off', tags: ['@manager', '@adminUser'] }, () => {
+  const machineDeploymentsPage = new MachineDeploymentsPagePo('_');
+  const runTimestamp = +new Date();
```
richard-cox, line 8: i think now is the time to resolve the issue of a common approach to making names for resources created by e2e tests.

We need a common function that will create it, suggest something like
```
const runTimestamp = +new Date();
function createE2EResourceName(context: string) {
   return `e2e-test-` + runTimestamp + `-` + context
}
```

We can start using it here and then gradually move the other usages over

cypress/e2e/tests/pages/manager/machine-deployments.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+  const machineDeploymentsName = `e2e-machinedeployment-name-${ runTimestamp }`;
+  const machineDeploymentsNameClone = `e2e-machinedeployment-name-${ runTimestamp }-clone`;
```
richard-cox, line 10: this can just be 
```
= `{machineDeploymentsName}-clone`
```

cypress/e2e/tests/pages/manager/node-templates.spec.ts
```diff
@@ -0,0 +1,147 @@
+import ClusterManagerListPagePo from '@/cypress/e2e/po/pages/cluster-manager/cluster-manager-list.po';
+import NodeTemplatesPagePo from '@/cypress/e2e/po/pages/cluster-manager/node-templates.po';
+import ClusterManagerCreateRke1Amazonec2PagePo from '@/cypress/e2e/po/edit/provisioning.cattle.io.cluster/create/cluster-create-rke1-amazonec2.po';
+import EmberPromptRemove from '@/cypress/e2e/po/components/ember/ember-prompt-remove.po';
+
+// will only run this in jenkins pipeline where cloud credentials are stored
+describe('Node Templates', { testIsolation: 'off', tags: ['@manager', '@jenkins', '@adminUser'] }, () => {
```
richard-cox, line 7: comment only, would you be able to share (privately) where jenkins results are published?

cypress/e2e/tests/pages/manager/pod-security-admissions.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+    cy.viewport(1380, 720);
+  });
+
+  it('can create a policy security admission', () => {
+    PodSecurityAdmissionsPagePo.navTo();
+    podSecurityAdmissionsPage.waitForPage();
+    podSecurityAdmissionsPage.create();
+    podSecurityAdmissionsPage.createPodSecurityAdmissionForm().waitForPage();
+    podSecurityAdmissionsPage.createPodSecurityAdmissionForm().name().set(policyAdmissionName);
+    podSecurityAdmissionsPage.createPodSecurityAdmissionForm().description().set(policyAdmissionDescription);
```
richard-cox, line 27: would we be able to set some admission control and exemptions fields and confirm they get sent correctly? same goes for edit

cypress/e2e/tests/pages/manager/repositories.spec.ts
```diff
@@ -0,0 +1,131 @@
+import PromptRemove from '@/cypress/e2e/po/prompts/promptRemove.po';
+import RepositoriesPagePo from '@/cypress/e2e/po/pages/cluster-manager/repositories.po';
+import * as path from 'path';
+import * as jsyaml from 'js-yaml';
+
+describe('Repositories', { testIsolation: 'off', tags: ['@manager', '@adminUser'] }, () => {
```
richard-cox, line 6: Similar naming comment to the PO's, this one is for the Cluster Management Helm Repo's

cypress/e2e/tests/pages/manager/cluster-manager.spec.ts
```diff
@@ -320,6 +321,62 @@ describe('Cluster Manager', { testIsolation: 'off', tags: ['@manager', '@adminUs
     clusterDashboard.waitForPage(undefined, 'cluster-events');
   });
 
+  it('can navigate to Cluster Management Page', () => {
```
richard-cox, line 324: as this is the root navigation to the area could this `it` come first? that way if it fails it's clear all other failures will be due to a navigation issue

richard-cox (changes requested):
Review finished. Just another set of small tweaks required

cypress/e2e/po/components/ember/ember-prompt-deactivate.po.ts
```diff
@@ -0,0 +1,11 @@
+import EmberComponentPo from '@/cypress/e2e/po/components/ember/ember-component.po';
+
+export default class EmberPromptDeactivate extends EmberComponentPo {
+  constructor() {
+    super('.modal-container');
```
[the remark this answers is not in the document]
yonasberhe23: hmm i guess `.model-open` works for all of these cases. I'll consolidate and just use EmberModalPo

cypress/e2e/po/edit/machine-deployments.po.ts
yonasberhe23: makes sense now, thanks for clarifying that hierarchy!

cypress/e2e/po/pages/extensions.po.ts
yonasberhe23: ah I see, I forgot two areas contain repositories. thanks for the callout.

moved back to `cypress/e2e/po/pages/repositories.po.ts` and updated file name to `chart-repositories`

cypress/e2e/po/lists/repositories-list.po.ts
yonasberhe23: @richard-cox looks like there was an existing po for this `cypress/e2e/po/lists/catalog.cattle.io.clusterrepo.po.ts` so I'll use that one instead of the one I created and rename it. `RepoListPo` to `ChartRepositoriesListPo`

yonasberhe23: should i also update the fiile name from `catalog.cattle.io.clusterrepo.po` to `chart-repositories.po`?

cypress/e2e/tests/pages/manager/machine-deployments.spec.ts
yonasberhe23: Good call! I've created a custom command called `createE2EResourceName`. We can use it in the tests via [aliasing](https://docs.cypress.io/guides/core-concepts/variables-and-aliases#Sharing-Context). So far I've update `machine-deployments.spec` and `repositories.spec`. I will wait for your feedback before updating the rest of the tests in this PR. Please see updated code below:

custom command: https://github.com/rancher/dashboard/blob/ca300951b879e5acb27ed8b908cd238225ae8129/cypress/support/commands/commands.ts#L66

test usage: https://github.com/rancher/dashboard/blob/ca300951b879e5acb27ed8b908cd238225ae8129/cypress/e2e/tests/pages/manager/machine-deployments.spec.ts#L18

cypress/e2e/tests/pages/manager/pod-security-admissions.spec.ts
yonasberhe23: this is in progress

cypress/e2e/tests/pages/manager/node-templates.spec.ts
yonasberhe23: yes i can share that with you.

cypress/e2e/po/components/ember/ember-input.po.ts
```diff
@@ -6,12 +6,12 @@ export default class EmberInputPo extends EmberComponentPo {
    * @param value Value to be typed
    * @returns
    */
-  set(value: string): Cypress.Chainable {
-    this.input().should('be.visible');
-    this.input().focus();
-    this.input().clear();
+  set(value: string, index = 0): Cypress.Chainable {
```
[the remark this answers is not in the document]
richard-cox: The `ctor` approach would work by something like below, as input just looks at self

```
new EmberInputPo(<selector to input field>)
```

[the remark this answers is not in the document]
yonasberhe23: @richard-cox I don't think the `ctor` approach works in this case since the input fields do not have unique identifiers (other than an ember id (e.g. id="ember175") which is inconsistent between forms so that can't be used here). For this approach to work, we would need to add a `data-testid` to these fields however I’m unfamiliar with ember code. could you help me with this?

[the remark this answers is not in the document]
richard-cox: In this context selectors are css selectors and don't need to contain `data-testid`. For instance it could be `[data-testid="<something parent like?"] input:nth-of-type(1)`

cypress/e2e/po/edit/pod-security-admissions.po.ts
```diff
@@ -0,0 +1,49 @@
+import PagePo from '@/cypress/e2e/po/pages/page.po';
+import NameNsDescription from '@/cypress/e2e/po/components/name-ns-description.po';
+import LabeledSelectPo from '@/cypress/e2e/po/components/labeled-select.po';
+import InputPo from '~/cypress/e2e/po/components/input.po';
```
richard-cox, line 4: `'~/` and below should be `@/`

cypress/support/commands/commands.ts
```diff
@@ -59,3 +59,12 @@ Cypress.Commands.add('iFrame', () => {
     .should('not.be.empty')
     .then((body) => cy.wrap(body));
 });
+
+/**
+ * Create resource name
+ */
+Cypress.Commands.add('createE2EResourceName', (context) => {
+  const runTimestamp = +new Date();
```
richard-cox, line 67: This needs to live outside of the function, otherwise it will be new every time it's called

cypress/e2e/tests/pages/manager/machine-deployments.spec.ts
richard-cox: Looks good, one minor comment inline. I'll keep this thread open to cover the other places it needs to be called.

And TIL `.as` will assign the response to `this`.

richard-cox (changes requested):
I haven't gone through the code, just comments. Just let me know when it's ready to review again

cypress/e2e/tests/pages/manager/pod-security-admissions.spec.ts
yonasberhe23: Oh I forgot to mention that this is done. See changes here: https://github.com/rancher/dashboard/pull/10161/commits/110beb8aff8bcf600dffa9071241005a1d0e1569

cypress/e2e/po/components/ember/ember-input.po.ts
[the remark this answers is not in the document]
yonasberhe23: Unfortunately, using `form input:nth-of-type(1)` targets all 4 inputs 

Screenshot 2024-01-19 at 12 22 38 PM

cypress/e2e/po/components/ember/ember-prompt-deactivate.po.ts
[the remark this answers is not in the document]
yonasberhe23: done in https://github.com/rancher/dashboard/pull/10161/commits/7abd6ad67c0c357ebfce662a15afcd01f4982f5a

cypress/e2e/po/components/ember/ember-input.po.ts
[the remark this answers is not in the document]
richard-cox: That was just an example. You would need to tweak the selector to be more specific. I'm guessing there are other forms on the page with input elements. So maybe something like `form.model-container`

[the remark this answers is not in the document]
yonasberhe23: I've tried all sorts of selector combinations. :nth-of-type(1) still targets all 4 inputs on the modal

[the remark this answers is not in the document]
richard-cox: `.modal-open .inline-form:nth-of-type(2) input` will do the trick

cypress/e2e/po/components/sortable-table.po.ts
```diff
@@ -56,6 +56,10 @@ export default class SortableTablePo extends ComponentPo {
     return cy.getId('sortable-table-promptRemove');
   }
 
+  downloadYamlButton() {
```
richard-cox, line 59: missed this. yaml download will be something specific to a generic resource list rather than the core sortable table

richard-cox (changes requested):
Looking good! One comment regarding selector, and one that just needs a function to move.

That's a lot of great value in this PR, looking forward to merging.

cypress/e2e/po/components/ember/ember-input.po.ts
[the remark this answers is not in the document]
yonasberhe23: that did do the trick, thank you @richard-cox! updated in https://github.com/rancher/dashboard/pull/10161/commits/e093c5c62425893677980aaa3cb60e30b0946f49

cypress/e2e/po/components/sortable-table.po.ts
yonasberhe23: @richard-cox so should it live in `/po/components/resource-table.po.ts`?

richard-cox: I thought so, and checked, there's already a downloadYamlButton in there that can be used

yonasberhe23: oh okay, I moved it from `resource-table.po` to `sortable-table.po` in this commit https://github.com/rancher/dashboard/pull/10161/commits/805d37f7aba0835eb5200b3ee0c3b42a198ea300 but I can move it back though

richard-cox (approved):

PR #10163, opened 2023-12-15
[Automation] - Fixing Qase Suites dupes by increasing results limit and using string matching

izaac:
### Summary
This PR is fixing the Jenkins pipeline Qase reporting. The report script was creating duplicated Test Suites on Qase on certain conditions.

This fix increases the results limit to the maximum the Qase QQL API allow which is 100 but to prevent hitting this I've added a string lookup that will get unique results using the Qase QQL.

Sometimes the logic that checks if a Suite exist got limited results returned by the Qase API as the default limit is 10. Due to this as there wasn't a match then a new Suite was getting created.

Jenkins automation pipeline reporting


yonasberhe23 (approved):
LGTM

PR #10164, opened 2023-12-15
Remove coverage for E2E

### LINKED ISSUE #10150: Remove coverage for E2E
# Description

Remove coverage for E2E tests. I causes issues and it's not convenient for computing code coverage.

> Measuring code coverage for E2E tests is difficult and challenging due to the intricate nature of these tests. Instead of using code coverage, requirement coverage might be the better way to go. This is because the focus of E2E tests is to cover the requirements of your test, not to focus on the source code.

https://web.dev/articles/ta-code-coverage#:~:text=Measuring%20code%20coverage%20for%20E2E,focus%20on%20the%20source%20code.

cnotv:
### Summary
Fixes #10150

Allow to re-run just failed jobs and cut down the risks of flaky tests on the way.

### Technical notes summary

Removed coverage for E2E and flags which will speed up process. Workflow is kept the same even if just unit tests for clearance and open to changes.


torchiaf (approved):
LGTM

Issue #9232, opened 2023-06-28
E2E Tests for Kubewarden


nwmac:
Build out tests for the Kubewarden extension that checks that it works correctly.

TBD - install KW, check views work, create a policy etc

nwmac:
Closing as won't fix

PR #10180, opened 2023-12-18
App upgrade notification shown for incorrect app (corrected filters and added further tests)

### LINKED ISSUE #9613: App upgrade notification shown for incorrect app

cnotv:
### Summary
Fixes #9613

Changes are the same of https://github.com/rancher/dashboard/pull/9806, except that `repoType`, `repoName`, `chartName` are not mandatory anymore, as they will just return no charts if none of the criteria is provided.
If parameters are provided, only charts with all of them will be returned.

### Technical notes summary
Added tests for:
- Chart match logic
  - Check `given only all matching parameters if provided` for Epinio case.
  - Check `given any matching parameter` for returning always charts if anything matches (reason this is reopened).
- Chart install view version selection

#### Kubewarden installation
I have a 2 breaking errors when I try locally with master to install Kubewarden to reproduce the mentioned case:
- Copy to clipboard says function does not exists
- `addChart` says the `repo` is actually `null`

After setting conditional variables it works, although I keep it for another time, since my PR covered already more than needed and the error is on master as well (also just when running locally). 

### Screenshot/Video
Screenshot 2023-12-18 at 19 59 31

### DIFF
--- a/shell/models/__tests__/catalog.cattle.io.app.test.ts
+++ b/shell/models/__tests__/catalog.cattle.io.app.test.ts
@@ -1,0 +1,80 @@
+import CatalogApp from '@shell/models/catalog.cattle.io.app';
+
+describe('class CatalogApp', () => {
+  describe('should return upgradeAvailable for this chart', () => {
+    describe('as false to be excluded from some logic if', () => {
+      const expectation = false;
+
+      it('is fleet', () => {
+        const catalogApp = new CatalogApp({ spec: { chart: { metadata: { annotations: { 'fleet.cattle.io/bundle-id': 'anything' } } } } });
+
+        expect(catalogApp.upgradeAvailable).toStrictEqual(expectation);
+      });
+
+      it('is managed,', () => {
+        const catalogApp = new CatalogApp({ spec: { chart: { metadata: { annotations: { 'catalog.cattle.io/managed': 'anything' } } } } });
+
+        expect(catalogApp.upgradeAvailable).toStrictEqual(expectation);
+      });
+    });
+
+    describe('as latest version', () => {
+      it('excluding pre-releases', () => {
+        const version = '1';
+        const catalogApp = new CatalogApp({});
+
+        catalogApp.spec = { chart: { metadata: { version: '0' } } };
+
+        jest.spyOn(catalogApp, '$rootGetters', 'get').mockReturnValue({
+          'catalog/chart': () => ({ versions: [{ version }] }),
+          currentCluster:  jest.fn(),
+          'prefs/get':     () => false
+        });
+
+        expect(catalogApp.upgradeAvailable).toStrictEqual(version);
+      });
+    });
+
+    describe('as null', () => {
+      const expectation = null;
+
+      it('if no chart', () => {
+        const catalogApp = new CatalogApp({});
+
+        jest.spyOn(catalogApp, '$rootGetters', 'get').mockReturnValue({
+          'catalog/chart': () => [],
+          currentCluster:  jest.fn(),
+          'prefs/get':     () => false
+        });
+
+        expect(catalogApp.upgradeAvailable).toStrictEqual(expectation);
+      });
+
+      it('if no version is available', () => {
+        const catalogApp = new CatalogApp({});
+
+        jest.spyOn(catalogApp, '$rootGetters', 'get').mockReturnValue({
+          'catalog/chart': () => [],
+          currentCluster:  jest.fn(),
+          'prefs/get':     () => false
+        });
+
+        expect(catalogApp.upgradeAvailable).toStrictEqual(expectation);
+      });
+
+      it('if current version is the latest', () => {
+        const catalogApp = new CatalogApp({});
+
+        catalogApp.spec = { chart: { metadata: { version: '1' } } };
+
+        jest.spyOn(catalogApp, '$rootGetters', 'get').mockReturnValue({
+          'catalog/chart': () => ({ versions: [{ version: '0' }] }),
+          currentCluster:  jest.fn(),
+          'prefs/get':     () => false
+        });
+
+        expect(catalogApp.upgradeAvailable).toStrictEqual(expectation);
+      });
+    });
+  });
+});
--- a/shell/models/catalog.cattle.io.app.js
+++ b/shell/models/catalog.cattle.io.app.js
@@ -76,7 +76,7 @@ export default class CatalogApp extends SteveModel {
     const showPreRelease = this.$rootGetters['prefs/get'](SHOW_PRE_RELEASE);
     const currentChartVersion = this.spec?.chart?.metadata?.version;
     const versions = !showPreRelease
-      ? matchingChart.versions.filter((v) => !isPrerelease(v.version))
+      ? matchingChart.versions?.filter((v) => !isPrerelease(v.version))
       : compatibleVersionsFor(matchingChart, workerOSs, showPreRelease);
     const newChartVersion = versions?.[0]?.version;
     const isOlder = compare(currentChartVersion, newChartVersion) < 0;
--- a/shell/models/catalog.cattle.io.app.js
+++ b/shell/models/catalog.cattle.io.app.js
@@ -41,23 +41,10 @@ export default class CatalogApp extends SteveModel {
   }
 
   matchingChart(includeHidden) {
-    const chart = this.spec?.chart;
-
-    if ( !chart ) {
-      return;
-    }
-
-    const chartName = chart.metadata?.name;
-    const preferRepoType = chart.metadata?.annotations?.[CATALOG_ANNOTATIONS.SOURCE_REPO_TYPE];
-    const preferRepoName = chart.metadata?.annotations?.[CATALOG_ANNOTATIONS.SOURCE_REPO_NAME];
-    const match = this.$rootGetters['catalog/chart']({
-      chartName,
-      preferRepoType,
-      preferRepoName,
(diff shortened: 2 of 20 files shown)


cnotv:
Also I noticed that we had already a talk about tests for Kubewarden here:
https://github.com/rancher/dashboard/issues/9232

shell/models/catalog.cattle.io.app.js
```diff
@@ -40,69 +41,48 @@ export default class CatalogApp extends SteveModel {
...
-    if (!showPreRelease) {
-      versions = chart.versions.filter((v) => !isPrerelease(v.version));
-    }
-
-    versions = compatibleVersionsFor(chart, workerOSs, showPreRelease);
-
-    const newestChart = versions?.[0];
-    const newestVersion = newestChart?.version;
-
-    if ( !thisVersion || !newestVersion ) {
```
richard-cox, line 100: this check, for truthy currentChartVersion and newestVersion, has been lost. i noticed that `compare` now has 

```
if (in1 === undefined && in2 === undefined) {
    return null;
}
```
if that was meant for a replacement, would vote to keep `compare` untouched and to bring it back to this context (as compare is used in other places). So it would be

```
if (!currentChartVersion || !newestVersion) {
    return null;
}
```

richard-cox, line 100: Apart from the check for undefined above, the other changes in upgradeAvailable are just tidying up? just want to be sure i haven't missed something

shell/models/catalog.cattle.io.app.js
```diff
@@ -1,15 +1,16 @@
+/* eslint-disable multiline-ternary */
```
richard-cox, line 1: can this be applied directly to the line / block

shell/models/catalog.cattle.io.app.js
```diff
@@ -40,69 +41,48 @@ export default class CatalogApp extends SteveModel {
   }
 
   matchingChart(includeHidden) {
-    const chart = this.spec?.chart;
-
-    if ( !chart ) {
```
richard-cox, line 45: this early exit has been lost

shell/models/__tests__/catalog.cattle.io.app.test.ts
```diff
@@ -0,0 +1,80 @@
...
+        const catalogApp = new CatalogApp({ spec: { chart: { metadata: { annotations: { 'catalog.cattle.io/managed': 'anything' } } } } });
+
+        expect(catalogApp.upgradeAvailable).toStrictEqual(expectation);
+      });
+    });
+
+    describe('as latest version', () => {
+      it('excluding pre-releases', () => {
+        const version = '1';
+        const catalogApp = new CatalogApp({});
```
richard-cox, line 24: comment only. the first arg to models is the json for the object (normally received by http request). to use less lines the spec can be passed there. to prefer clarity would vote leaving as is

shell/models/__tests__/catalog.cattle.io.app.test.ts
```diff
@@ -0,0 +1,80 @@
...
+
+      it('is managed,', () => {
+        const catalogApp = new CatalogApp({ spec: { chart: { metadata: { annotations: { 'catalog.cattle.io/managed': 'anything' } } } } });
+
+        expect(catalogApp.upgradeAvailable).toStrictEqual(expectation);
+      });
+    });
+
+    describe('as latest version', () => {
+      it('excluding pre-releases', () => {
```
richard-cox, line 22: the title of this test is wrong, it returns the success case where there is an upgrade available

shell/models/__tests__/catalog.cattle.io.app.test.ts
```diff
@@ -0,0 +1,80 @@
...
+        jest.spyOn(catalogApp, '$rootGetters', 'get').mockReturnValue({
+          'catalog/chart': () => [],
+          currentCluster:  jest.fn(),
+          'prefs/get':     () => false
+        });
+
+        expect(catalogApp.upgradeAvailable).toStrictEqual(expectation);
+      });
+
+      it('if no version is available', () => {
```
richard-cox, line 53: this is the same as above (should it include `catalogApp.spec = { chart: { metadata: {  } } };`?)

shell/store/__tests__/catalog.test.ts
```diff
@@ -0,0 +1,169 @@
...
+      });
+
+      it.each([
+        {
+          chartName: 'rancher-alerting-drivers',
+          repoType:  'cluster',
+          repoName:  'rancher-charts',
+        },
+        { chartName: 'rancher-alerting-drivers' },
+        { repoType: 'cluster' },
```
richard-cox, line 71: given only repo type or repo name i don't think we should match any chart in that repo. feels like the minimum would be the chart name

richard-cox (changes requested):
@jordojordo Can you validate kube warden side is fine? In the description Giuseppe mentions some kubewarden strageness

cnotv:
> @jordojordo Can you validate kube warden side is fine? In the description Giuseppe mentions some kubewarden strageness

It seems to crash just in case of lack of charts, so I omitted it.

shell/types/catalog.ts
```diff
@@ -0,0 +1,30 @@
+export interface IChartVersions {
+  annotations?: Record<string, any>,
+  version: string,
+}
+
+export interface IChart {
```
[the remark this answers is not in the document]
cnotv: I am aware of the JS file which I'm not going to touch 😄 
What are you suggesting about the interface? I did not get it.

shell/models/catalog.cattle.io.app.js
cnotv: It fails if I keep as it is, I'll try to check what I can do.
Ideally would be convenient to have real data to understand better the suggestion.

shell/models/__tests__/catalog.cattle.io.app.test.ts
cnotv: No it's not necessary for computation.

jordojordo:
The sections where I previously had trouble with the `catalog/chart` getter have been resolved as I've added the `repoName` and `repoType` to each instance of those calls.

>addChart says the repo is actually null

I didn't notice any issues here.

>Copy to clipboard says function does not exists

I did run into the issue with the `$copyText` method not existing where I have used the `CopyCode` component [here](https://github.com/rancher/kubewarden-ui/blob/cab1ec7cf0108da976f6d7486c2d04e377ee96db/pkg/kubewarden/components/Dashboard/InstallView.vue#L296-L298).


https://github.com/rancher/dashboard/assets/40806497/c7b86cbb-8fcf-4df6-922c-a826472454da

shell/types/catalog.ts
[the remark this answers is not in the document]
richard-cox: This would just be to link the two together. So if someone updates the object they also know to update the interface, or want to know where the object is instantiated. 

```
/* 
 * ...
 * Created in `addChart` shell/store/catalog.js
 */
export interface IChart
```

```
// Instance of `IChart`
obj = classify(ctx, {
```

shell/models/__tests__/catalog.cattle.io.app.test.ts
richard-cox: to check explicitly for a missing version it would need to get passed the check for a missing chart and spec. so by providing the object above the test should hit the code that it's aiming to check. 

as it stands the test is the same as for `if no chart` and doesn't exercise anything different

shell/types/catalog.ts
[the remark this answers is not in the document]
cnotv: Oh but I have absolutely no idea what do we have there, so I have just put what I know and then later someone will have to update just as I did, otherwise nobody would ever write anything, no?

shell/models/__tests__/catalog.cattle.io.app.test.ts
cnotv: but the test is passing, otherwise we are missing one and I did not realize it.

shell/types/catalog.ts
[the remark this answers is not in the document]
richard-cox: That's fine, but can you still add the suggested comments?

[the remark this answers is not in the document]
cnotv: yeah absolutely, just taking time :)

shell/models/__tests__/catalog.cattle.io.app.test.ts
richard-cox: the two tests `if no chart` and `if no version is available` are exactly the same. both tests should provide and env such that only the condition referenced in the test title is exercised

cnotv:
Closing since is outdated and replaced by a new issue: https://github.com/rancher/dashboard/issues/11465

PR #10184, opened 2023-12-19
[Automation] - Refactor Jenkins Qase reporter to not use QQL endpoints and improve multi branch report creation

izaac:
### Summary

I've noticed that when running the Jenkins pipeline using the recurring Multi branch the report generated in Qase sometimes had some type of collisions and the report generated extra test cases on different branches. 

Debugging the issue I found that the QQL searches comparisons were unstable as sometimes it matched the proper string in the returned json object or not at all. 

There were some other edge cases like when either a test case or test suite returned multiple matches (duplicates) the logic wasn't smart enough to handle the differences.

To fix this I removed the QQL queries replacing them with direct calls to the `cases` and `suites` Qase API endpoints to then use the `id` and `title` to have more control over the data.

Multi branch running of the Automation UI tests. `v2.6`, `v2.7` and `v2.8`

### Test done
Execute the multi branch pipeline and analyze the generated reports.
Validate the oldest test case or the test case with the lowest id value is used when duplicates of a test case title is detected in the results.
As a search string is used already to check for existing suites. This ensure when the count is zero means the current suite title is new.
Validate only suites with test cases get selected when looking to update a test cases. This takes care of ignoring duplicates

### DIFF
--- a/cypress/jenkins/junit_to_qase.py
+++ b/cypress/jenkins/junit_to_qase.py
@@ -135,12 +135,15 @@ skipped_cases_ids = []
 failed_cases_ids_stack = {}
 
 
-def qase_cases(project, case_id=None):
+def qase_cases(project, case_id=None, suite_id=None, limit=100):
     if case_id is not None:
         return requests.get("{0}/{1}/{2}".format(cases_url, project, case_id),
                             headers=qase_headers)
+    elif suite_id is not None:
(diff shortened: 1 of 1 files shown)


izaac:
The more recent update just renamed a variable and rebased.

yonasberhe23 (approved):
LGTM

izaac:
Thanks for taking a look. Merging

PR #10195, opened 2023-12-21
[Automation] - Fixing logic for running Jenkins tests on existing Rancher

izaac:
### Summary
Fixes:

An issue when configuring the Jenkins tests to run on an `existing` Rancher. The test was doing the `local` type instead.
`recurring` is unaffected.

Jenkins jobs execution using an existing Rancher, running the job on demand.


yonasberhe23 (approved):
lgtm

PR #10197, opened 2023-12-22
fix(ContainerLogs): fix display error if log is truncated

### LINKED ISSUE #10196: Logs display errors if there are multibyte utf8 characters truncated
**Setup**
- Rancher version: v2.8.0
- Browser type & version: Chrome v120.0.6099.109 arm64

**Describe the bug**

If a single log entry is too long, the log message pushed by the backend may be truncated, and if the log entry contains multibyte utf8 characters, the complete utf8 characters may be truncated, resulting in the ui being unable to display the truncated characters.

**To Reproduce**

The following mirror will replicate the problem：

arm64 image：liuyan/test:v1
amd64 image: liuyan/test:v2

**Result**
Truncated characters are displayed as `�`

**Expected Result**

The truncated utf8 characters need to be merged in order to be displayed correctly.

**Screenshots**

image

**Additional context**

ly5156:
### Summary
Fixes #10196

### DIFF
--- a/shell/components/nav/WindowManager/ContainerLogs.vue
+++ b/shell/components/nav/WindowManager/ContainerLogs.vue
@@ -2,7 +2,7 @@
 import { saveAs } from 'file-saver';
 import AnsiUp from 'ansi_up';
 import { addParams } from '@shell/utils/url';
-import { base64Decode } from '@shell/utils/crypto';
+import { base64DecodeToBuffer } from '@shell/utils/crypto';
 import { LOGS_RANGE, LOGS_TIME, LOGS_WRAP } from '@shell/store/prefs';
 import LabeledSelect from '@shell/components/form/LabeledSelect';
 import { Checkbox } from '@components/Form/Checkbox';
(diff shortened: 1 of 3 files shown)


shell/components/nav/WindowManager/ContainerLogs.vue
```diff
@@ -273,11 +315,47 @@ export default {
...
+            truncatedLog = data;
+
+            return;
+          }
+          data = lines.slice(0, -1).join('\n');
+          truncatedLog = lines.slice(-1);
+        }
         // Websocket message may contain multiple lines - loop through each line, one by one
-        data.split('\n').forEach((line) => {
+        data.trim().split('\n').forEach((line) => {
```
aalves08, line 358: understand why you do the `trim` but `data.trim().split('\n')` I feel it's again a sequence of operations that might throw an error if something doesn't start off with the correct type or expected content

aalves08 (changes requested):
Left some comments. Overall a bit lost reviewing this as I don't have any experience with these kind of operations done by your functions.

We do need to add some unit tests to this PR in order to verify the integrity of the basic usecases that this component covers + the functionality you are adding here.

ly5156:
@aalves08 
Thanks for reviewing the code, I've added some comments.

aalves08 (approved):
LGTM

PR #10203, opened 2023-12-28
Fix chart installation question groups are "None" if there is only 1 group defined

### LINKED ISSUE #9959: Chart installation question groups are "None" if there is only 1 group defined
On kubewarden-defaults we display `(None)` tab instead of `Settings` (that is defined in questions.yaml).

Screenshot from 2023-10-20 09-31-54

aalves08:
### Summary
Fixes #9959 

- fix issue where if `tabbed=multiple` and only 1 group would show a tab  called `None`
- as per comment, added flag to hide tabs section if there's only 1 group (cannot make it a generalised usecase since machine config is cluster provisioning needs tabs to be shown at all time when creating machine pools)

- Go to `<!---YOUR-RANCHER-BASE-URL-->/c/local/apps/charts/install?repo-type=cluster&repo=rancher-charts&chart=rancher-vsphere-cpi&version=102.1.0%2Bup1.5.1` and check that no tabs appear (only 1 group - Configuration)
- Go to `<!---YOUR-RANCHER-BASE-URL-->/c/local/apps/charts/install?repo-type=cluster&repo=rancher-charts&chart=epinio&version=103.0.2%2Bup1.10.0` and check that tabs appear since there's more than 1 group

### Screenshot/Video

_Before_
Screenshot 2023-12-28 at 12 51 07

_After_
Screenshot 2024-01-02 at 15 24 31
Screenshot 2024-01-02 at 15 25 30


nwmac:
@aalves08 Could we add automated tests for this?

Also, in this case, can we hide the side tabs altogether when there is only one option? Perhaps we can add an option to the tab component to support this and then in this usage, turn this on?

shell/components/Tabbed/index.vue
```diff
@@ -86,6 +91,11 @@ export default {
     sortedTabs() {
       return sortBy(this.tabs, ['weight:desc', 'labelDisplay', 'name']);
     },
+
+    // hide tabs based on tab count IF flag is active
+    hideTabs() {
+      return !this.hideSingleTab || (this.hideSingleTab && this.sortedTabs.length > 1);
```
nwmac, line 97: Is this logic correct?

In the usage below, we're really using it as showTabs, not hideTabs (in the v-if).

Suggest we leave this as `hideTabs`, but change the v-if below to `!hideTabs`.

I think the logic here can be simply:

`return this.hideSingleTab && this.sortedTabs.length === 1;`

I think that's easier to read?

nwmac (changes requested):

shell/components/Tabbed/index.vue
aalves08: I did the opposite 🤦 . Totally correct @nwmac . Going to change it. Thanks

cypress/e2e/tests/pages/charts/question-tabs.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+        const epinioVersion = '103.0.2%2Bup1.10.0';
+        const chartsEpinioPage = `${ chartsPageUrl }&chart=epinio&version=${ epinioVersion }`;
+        const chartsPage: ChartsPage = new ChartsPage(chartsEpinioPage);
+
+        chartsPage.goTo();
+        chartsPage.goToInstall().nextPage();
+
+        chartsPage.tabsCountOnInstallQuestions().should('have.length.above', 3);
+      });
+    });
```
jordojordo, line 35: Considering the current project status, should we use a different chart other than Epinio?

aalves08: True... True... 🤦

shell/components/Questions/index.vue
```diff
@@ -258,7 +258,7 @@ export default {
       }
 
       if ( this.tabbed === 'multiple' ) {
-        return this.groups.length > 1;
+        return this.groups.length;
```
nwmac, line 261: If I was being picked I'd add `!!` to return a boolean rather than rely on a truthy/falsey

aalves08: you are correct @nwmac . Fixed

jordojordo (approved):
lgtm

Issue #10104, opened 2023-11-30
Replace `Vue.extend` with `defineComponent`


rak-phillip:
Vue 2.7 introduced a backported Vue 3 feature known as `defineComponent` [^1]. Migrating from `Vue.extend` to `defineComponent` and resolving any potential issues that arise can help ease the overall transition to Vue 3.

## Acceptance Criteria

- [x] Instances of `Vue.extend` are replaced with the equivalent usage of `defineComponent`
- [x] Specific issues that might arise as a result of this migration are resolved
   The Vue 3 migration guide[^2] can provide guidance on how to resolve these issues 

## Identified usages of Vue.extend

- [x] ./pkg/rancher-components/src/components/BadgeState/BadgeState.vue
- [x] ./pkg/rancher-components/src/components/Banner/Banner.vue
- [x] ./pkg/rancher-components/src/components/Card/Card.vue
- [x] ./pkg/rancher-components/src/components/Form/Checkbox/Checkbox.vue
- [x] ./pkg/rancher-components/src/components/Form/Radio/RadioButton.vue
- [x] ./pkg/rancher-components/src/components/Form/Radio/RadioGroup.vue
- [x] ./pkg/rancher-components/src/components/Form/TextArea/TextAreaAutoGrow.vue
- [x] ./pkg/rancher-components/src/components/Form/ToggleSwitch/ToggleSwitch.vue
- [x] ./pkg/rancher-components/src/components/LabeledTooltip/LabeledTooltip.vue
- [x] ./pkg/rancher-components/src/components/StringList/StringList.vue
- [x] ./pkg/harvester/pages/c/_cluster/projectsnamespaces.vue
- [x] ./shell/components/form/Error.vue
- [x] ./shell/components/form/Footer.vue
- [x] ./shell/components/form/GitPicker.vue
- [x] ./shell/components/DraggableZone.vue
- [x] ./shell/components/PodSecurityAdmission.vue
- [x] ./shell/components/AsyncButton.vue
- [x] ./shell/mixins/create-edit-view/index.js

### Not updating

The following mixins should be replaced with their composable counterparts:

- ./shell/mixins/compact-input.ts
- ./shell/mixins/labeled-form-element.ts

The following files are related to our init process and will need to be refactored along with larger nuxt removal/vue 3 migration efforts:

- ./shell/utils/nuxt.js
- ./shell/initialize/client.js

[...]

PR #10220, opened 2024-01-03
Bump typescript to 4.5.5

rak-phillip:
### Summary

I came across the following tsc warnings while working on #10104:

```
Excessive stack depth comparing types 'Vue<Record<string, any>, Record<string, any>, never, never, (event: string, ...args: any[]) => Vue<Record<string, any>, Record<string, any>, never, never, ...>> | HTMLElement' and 'ComponentPublicInstance<{}, {}, {}, {}, {}, {}, {}, {}, false, ComponentOptionsBase<any, any, any, any, any, any, any, any, any, any>> | Element | Vue<...>'.ts(2321)
Excessive stack depth comparing types 'ComponentPublicInstance<{}, {}, {}, {}, {}, {}, {}, {}, false, ComponentOptionsBase<any, any, any, any, any, any, any, any, any, any>> | Element | Vue<...>' and 'Vue<Record<string, any>, Record<string, any>, never, never, (event: string, ...args: any[]) => Vue<Record<string, any>, Record<string, any>, never, never, ...>> | HTMLElement'.ts(2321)
const ref: (ComponentPublicInstance<{}, {}, {}, {}, {}, {}, {}, {}, false, ComponentOptionsBase<any, any, any, any, any, any, any, any, any, any>> | Element | Vue<...>)[]
```

It appears that we are running up against the TypeScript type depth limiter while trying to infer the correct type for Vue refs. I've noted references to this changing in TypeScript from around a time that coincides with the current version that we use (4.1.6) [^1][^2].

Bumping TypeScript to 4.5.5 easily resolves this problem. I'm targeting version 4.5.5 because this is the targeted version for new Vue 2.x projects.

supports #10221

[^1]: https://github.com/microsoft/TypeScript/pull/45025
[^2]: https://github.com/microsoft/TypeScript/pull/45711

### DIFF
--- a/package.json
+++ b/package.json
@@ -185,7 +185,7 @@
     "start-server-and-test": "1.13.1",
     "style-loader": "3.3.2",
     "ts-jest": "27.1.4",
-    "typescript": "4.1.6",
+    "typescript": "4.5.5",
     "vue": "2.7.14",
     "vue-template-compiler": "2.7.14",
     "webpack-bundle-analyzer": "4.5.0",
--- a/pkg/rancher-components/package.json
+++ b/pkg/rancher-components/package.json
@@ -43,7 +43,7 @@
     "jsonpath-plus": "6.0.1",
     "sass": "1.55.0",
     "sass-loader": "10.2.1",
-    "typescript": "4.1.6",
+    "typescript": "4.5.5",
     "vue": "2.7.14",
     "vue-template-compiler": "2.7.14"
   },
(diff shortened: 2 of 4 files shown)


codyrancher (approved):
Locally I verified that global, cluster and extension pages still function with this version of typescript still installed and the tests all pass.

Because it's low effort to revert this change if any issues arise I don't think there's a good reason to hold this change up.

I like the reasoning for choosing this version and I also ran into the same issue which prompted this change last week.

richard-cox (approved):

PR #10221, opened 2024-01-03
Replace `vue.extend` with `defineComponent` in Rancher Components

### LINKED ISSUE #10104: Replace `Vue.extend` with `defineComponent`

rak-phillip:
### Summary
closes #10104

This replaces the usage of `Vue.extend` with `defineComponent` throughout multiple components located under `pkg/rancher-components`. `defineComponent` provides better type inference, improved TypeScript support, and will be the recommended way to define components when Dashboard migrates to Vue 3.

- Replaced instances of `Vue.extend` with `defineComponent` for components located under `pkg/rancher-components`

- BadgeState.vue
- Banner.vue
- Card.vue
- Checkbox.vue
- RadioButton.vue
- RadioGroup.vue
- TextAreaAutoGrow.vue
- ToggleSwitch.vue
- LabeledTooltip.vue
- StringList.vue


rak-phillip:
@codyrancher I'm requesting a review from you because you previously reviewed #10226. I opted to make use of your suggested change[^1] from our previous conversation. After doing some investigation, I found that this behavior is reproducible in new Vue 2/TypeScript applications using Vue test utils. It also looks like new Vue 2 projects using Vite exclude tests from the tsconfig in the same way. So, it seems that maybe this is indeed a problem that we can revisit after the Vue  3 migration 🙂

[^1]: https://github.com/rancher/dashboard/pull/10226#discussion_r1443359130

pkg/rancher-components/src/components/Form/Radio/RadioButton.vue
```diff
@@ -134,7 +144,7 @@ export default Vue.extend({
     @click.stop="clicked($event)"
   >
     <input
-      :id="_uid+'-radio'"
+      :id="randomString"
```
codyrancher, line 147: I think the intention is to invoke the randomString method here. Right now the function is getting set as the ID. 

I think it would be better to have a data property used here so that the id stays consistent if the component is re-rendered because of a :key change or something like that.

pkg/rancher-components/src/components/Form/Radio/RadioButton.vue
```diff
@@ -115,12 +116,21 @@ export default Vue.extend({
     /**
      * Emits the input event.
      */
-    clicked({ target }: { target?: HTMLElement }) {
-      if (this.isDisabled || target?.tagName === 'A') {
+    clicked(event: MouseEvent | KeyboardEvent) {
+      const target = event.target;
+
+      if (this.isDisabled || (target instanceof HTMLElement && target.tagName === 'A')) {
```
codyrancher, line 122: Just for my own information, in what scenarios will target be something other than HTMLElement?

pkg/rancher-components/src/components/Form/ToggleSwitch/ToggleSwitch.vue
```diff
@@ -41,7 +41,7 @@ export default Vue.extend({
   },
 
   methods: {
-    toggle(neu: boolean | string | number) {
+    toggle(neu: boolean | 'true' | 'false' | undefined | null) {
```
codyrancher, line 44: If this type could be the same as the state type declared above in data it might make sense to declare a union type and use it in both locations.

pkg/rancher-components/src/components/LabeledTooltip/LabeledTooltip.vue
```diff
@@ -29,9 +29,14 @@ export default Vue.extend({
...
   computed: {
-    iconClass() {
+    iconClass(): string {
       return this.status === 'error' ? 'icon-warning' : 'icon-info';
     }
+  },
+  methods: {
+    isObject(value: string | Record<string, unknown>): value is Record<string, unknown> {
+      return typeof value === 'object' && value !== null && !!value.content;
+    }
```
codyrancher, line 39: Could lodash be used here? I'm not sure what the dependencies look like for rancher-components

rak-phillip: We don't import the entirety of lodash inside of `rancher-components`, so we can add an import for `lodash.isobject`[^1] to support this. 

[^1]: https://www.npmjs.com/package/lodash.isobject

pkg/rancher-components/src/components/Form/Radio/RadioButton.vue
rak-phillip: Good catch, that's not quite intended at all 🙂

> I think it would be better to have a data property used here so that the id stays consistent if the component is re-rendered because of a :key change or something like that.

I think handling it this way will be an improvement overall, thanks for the suggestion!

pkg/rancher-components/src/components/Form/Radio/RadioButton.vue
rak-phillip: We're really using `instanceof` to provide a more specific type check so that we can access `Event.target` without complaints from TypeScript.

Event.target is typed as `EventTarget | null`; the original code would produce compiler warnings

```
Property 'tagName' does not exist on type 'EventTarget'.ts
```

pkg/rancher-components/src/components/BadgeState/BadgeState.vue
```diff
@@ -47,7 +47,7 @@ export default Vue.extend({
   },
 
   computed: {
-    bg(): string | null {
```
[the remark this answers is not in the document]
rak-phillip: Hmm, you're right, we default it to `null` but that doesn't appear to be picked up..

On line 62, we have the following usage of bg

https://github.com/rancher/dashboard/blob/e72694892b8982c7a9fe677c994fd66d0cb866c4/pkg/rancher-components/src/components/BadgeState/BadgeState.vue#L62

which will trigger the tsc warning when we allow null

```
A computed property name must be of type 'string', 'number', 'symbol', or 'any'. ts(2464)
```

I think that the array syntax might be more correct in this case. It should evaluate the same way as it did today without this collision.

codyrancher (approved):

PR #10224, opened 2024-01-04
Comment out Chrome 116 pinning as no longer needed

nwmac:
Commented out a workaround that was added to the e2e CI workflow that pinned the Chrome version.

There was a bug in Chrome > 116 and < 119 that broke our CI.


aalves08 (approved):
LGTM

PR #10225, opened 2024-01-04
fix issue with duplicate labels on editing a cluster cloud  credential

### LINKED ISSUE #10020: Cluster credentials: duplicate names possible but not working
Internal reference: SURE-6920
Reported in 2.7.5

Issue description:
It is possible in the Rancher GUI to create a cluster credential with an identical name to an already existing one. Then it is listed twice in the selection menu. It can even be selected but it seems never to be used.

Use case:
The user wanted to replace a cluster credential for a Harvester cluster which was not working anymore (reason unclear and not subject to this case).

They created a second one with the same name and selected it to be the new credential, but cluster access was still not working. Obviously, the old one was still used.

The only way to replace the old cloud credential was to create a new one with a different name and assign it to the existing cluster.

This should either not be possible to create or it should work with both credentials. 

Business impact:
There is a UI bug where someone can create a cluster credential with the name of an already existing credential. It seems the older one is then used anyway. There should be a validation that prevents re-use of the same name or both credentials should work when selected. 

Troubleshooting steps:
On a cluster with nonworking cloud credentials, a new one has been created with the same name. However, the new one was never used. It looks like the name is used for finding the object and the first one found (in this case older) is always used.

Repro steps:

In Rancher UI create a cloud credential with the same name as the existing one
Select a new credential when starting the cluster

Workaround:
None

Actual behavior:
It is possible to create a cloud credential with the name of the existing one. But then once selected during cluster creation it is not used

Expected behavior:
It should either not be possible to create a cloud credential with the name of an existing credential or it should be possible the one that was selected.

aalves08:
### Summary
Fixes #10020 

- fix issue with duplicate labels on editing a cluster cloud credential
- update label on `SelectCredential` to display combination of name and id (optional?)
- fix label and placeholder on `cloudcredential` edit view 
- fix translation for credential name placeholder 

**Note:** Cannot create e2e tests for creating cloud credentials, since we need valid tokens for any of the options available on the UI.

### Technical notes summary

I’ve compared the code for `SelectCredential`  for `2.7.5` and `master` and nothing relevant changed (template and `options`  variable logic), which to me means that we’ve always had this problem.

https://github.com/rancher/dashboard/blob/release-2.7.5/shell/edit/provisioning.cattle.io.cluster/SelectCredential.vue vs https://github.com/rancher/dashboard/blob/master/shell/edit/provisioning.cattle.io.cluster/SelectCredential.vue

The fix: by default `LabeledSelect`  uses as `option-key`  the `label` property of the `options` variable. In this case the `label`  isn’t unique, we just add `option-key="value"`  to get it’s “unique” key from the `value` property.

The other thing about creating a cluster (being RKE1 or RKE2) with duplicate creds works flawlessly. Why?
That’s because the `v-model`  isn’t populated, so the component doesn’t default to the first option with the same `label` always. We are able to pick one and it works.

- Create two valid credentials with the same provider, using the same name. Ex: `digital-ocean-cred`
- revert the code change that makes the label unique on `SelectCredential` (so that you can check that the fix would work for duplicate labels)
- Create a RKE2 cluster for that provider and select the first credential on the list
- Go to `view YAML` and note the cloud credential id that was "appended" to the cluster
- Edit that same cluster and select the second credential on the list 
- Go to `view YAML` again and check that the cloud credential id that was "appended" to the cluster is now the one from the edit operation

### Screenshot/Video

_Previous behaviour (while editing cluster)_

https://github.com/rancher/dashboard/assets/97888974/877a75ac-285a-4bef-9124-0a6bee80cefb

_Correct behaviour_

https://github.com/rancher/dashboard/assets/97888974/a6d6ea68-2ed7-48eb-9d96-af2d462ead38

_Fixed label and placeholder on cloud creation screen_
Screenshot 2024-01-04 at 16 11 28

### DIFF
--- a/cypress/e2e/po/pages/cluster-manager/cluster-manager-list.po.ts
+++ b/cypress/e2e/po/pages/cluster-manager/cluster-manager-list.po.ts
@@ -18,7 +18,7 @@ export default class ClusterManagerListPagePo extends PagePo {
     super(ClusterManagerListPagePo.createPath(clusterId));
   }
 
-  static navTo() {
+  navTo() {
     BurgerMenuPo.toggle();
     BurgerMenuPo.burgerMenuNavToMenubyLabel('Cluster Management');
   }
--- a/cypress/e2e/po/pages/cluster-manager/cluster-manager-list.po.ts
+++ b/cypress/e2e/po/pages/cluster-manager/cluster-manager-list.po.ts
@@ -43,4 +43,8 @@ export default class ClusterManagerListPagePo extends PagePo {
     return this.list().masthead().actions().eq(1)
       .click();
   }
+
+  editCluster(name: string) {
+    this.sortableTable().rowActionMenuOpen(name).getMenuItem('Edit Config').click();
+  }
 }
--- a/cypress/e2e/po/pages/cluster-manager/cluster-manager-list.po.ts
+++ b/cypress/e2e/po/pages/cluster-manager/cluster-manager-list.po.ts
@@ -18,7 +18,7 @@ export default class ClusterManagerListPagePo extends PagePo {
     super(ClusterManagerListPagePo.createPath(clusterId));
   }
 
-  navTo() {
+  static navTo() {
     BurgerMenuPo.toggle();
     BurgerMenuPo.burgerMenuNavToMenubyLabel('Cluster Management');
   }
--- a/cypress/e2e/tests/pages/manager/cloud-credential.spec.ts
+++ b/cypress/e2e/tests/pages/manager/cloud-credential.spec.ts
@@ -82,7 +82,7 @@ describe('Cloud Credential', () => {
           const editClusterPage = new ClusterManagerEditGenericPagePo(clusterName);
 
           editClusterPage.selectOptionForCloudCredentialWithLabel(`${ credsName } (${ createdCloudCredsIds[1] })`);
-          editClusterPage.saveEditCluster();
+          editClusterPage.save();
 
           cy.wait('@dummyProvClusterSave').then(({ request }) => {
             expect(request.body.spec.cloudCredentialSecretName).to.equal(createdCloudCredsIds[1]);
--- a/cypress/e2e/tests/pages/manager/cloud-credential.spec.ts
+++ b/cypress/e2e/tests/pages/manager/cloud-credential.spec.ts
@@ -58,7 +58,7 @@ describe('Cloud Credential', () => {
     const createdCloudCredsIds = [];
 
     for (let i = 0; i < cloudCredsToCreate.length; i++) {
-      cy.createRancherResource('v3', 'cloudcredentials', cloudCredentialCreatePayload(cloudCredsToCreate[i].name, cloudCredsToCreate[i].token)).then((resp: Cypress.Response<any>) => {
+      cy.createRancherResource('v3', 'cloudcredentials', JSON.stringify(cloudCredentialCreatePayload(cloudCredsToCreate[i].name, cloudCredsToCreate[i].token))).then((resp: Cypress.Response<any>) => {
         createdCloudCredsIds.push(resp.body.id);
 
         if (i === 0) {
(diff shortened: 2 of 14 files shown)


shell/edit/provisioning.cattle.io.cluster/SelectCredential.vue
```diff
@@ -104,7 +104,7 @@ export default {
     options() {
       const out = this.filteredCredentials.map((obj) => {
         return {
-          label: obj.nameDisplay,
+          label: `${ obj.nameDisplay }/${ obj.id }`,
```
nwmac, line 107: I don't think we should add the IDs like this - and formatting with a `/` is a bit confusing,

If I was being really clever here, I'd look through the list of credential names first and make a note of the ones that have duplicates.

Then for the label, if the name is one with a duplicate I'd add the id in parenthesis afterwards, otherwise I'd format the ones without duplicates as they currently are (name only).

nwmac (changes requested):
Can we add an automated test?

shell/edit/provisioning.cattle.io.cluster/SelectCredential.vue
aalves08: Yeah, was unsure about this, but I thought the best procedure was to add some kind of a distinction between those duplicates on the `LabeledSelect`

richard-cox (changes requested):

cypress/e2e/tests/pages/manager/cloud-credential.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+              body:       clusterProvDigitalOceanSingleResponse(clusterName, resp.body.id, machinePoolId),
+            });
+          }).as('dummyClusterListLoad');
+        }
+
+        // code to edit cluster and final checks needs to be after all "then's" otherwise the cloud cred id's aren't stored yet on it's variable...
+        if (i === cloudCredsToCreate.length - 1) {
+          const clusterList = new ClusterManagerListPagePo('local');
+
+          clusterList.goTo();
```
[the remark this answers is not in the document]
aalves08: This can't be a `navTo` because we aren't on any page to begin with in this `spec`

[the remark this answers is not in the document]
richard-cox: gotcha. the goTo can move to `beforeEach`, which can change to `before`. `clusterList.goTo` can then be `navTo`. that should chop out a 2 page loads and only add some much quicker navs

shell/edit/provisioning.cattle.io.cluster/SelectCredential.vue
```diff
@@ -103,8 +103,11 @@ export default {
 
     options() {
       const out = this.filteredCredentials.map((obj) => {
+        const duplicateFound = this.filteredCredentials.find((filtered) => filtered.nameDisplay === obj.nameDisplay);
```
richard-cox, line 106: ok, my bad, i got this wrong. map must alter in place because this is now broken (for me it always returns true).

adding ` && filtered.id !== obj.id` will fix it. 

i did wonder about performance. both map and find will scale (cpu) badly with a lot of credentials.  i'm not sure how realistic that is, but would suggest something like this instead...

```
const duplicates = {};

this.filteredCredentials.forEach((cred) => {
  duplicates[cred.nameDisplay] = duplicates[cred.nameDisplay] === null ? true : null;
});

const out = this.filteredCredentials.map((obj) => ({
  // if credential name is duplicated we add the id to the label
  label: duplicates[obj.nameDisplay] ? `${ obj.nameDisplay } (${ obj.id })` : obj.nameDisplay,
  value: obj.id,
}));
```

richard-cox (changes requested):
sorry, i made a wrong suggestion, details in comment

cypress/e2e/po/pages/cluster-manager/cluster-manager-list.po.ts
```diff
@@ -18,7 +18,7 @@ export default class ClusterManagerListPagePo extends PagePo {
     super(ClusterManagerListPagePo.createPath(clusterId));
   }
 
-  static navTo() {
+  navTo() {
```
richard-cox, line 21: This should stay as static and can be accessed via `ClusterManagerListPagePo.navTo`. I think this change will be the cause of some of the e2e failures

richard-cox (changes requested):

richard-cox (approved):

PR #10226, opened 2024-01-04
Replace `Vue.extend` with `defineComponent` in `LabeledInput.vue`

rak-phillip:
### Summary
contributes to #10104 

Rewrote `compact-input.ts` and `labeled-form-element.ts` mixins as composables to address TypeScript issues when using mixins with `defineComponent`. The new composables are written to adhere to the conventions and best practices [^1] as best as possible, taking into consideration what's available to us in Vue 2.7.14. 

Moving to composables positions us for a smoother Vue 3 migration.

[^1]: https://vuejs.org/guide/reusability/composables.html#conventions-and-best-practices

Any forms that utilize `LabeledInput`

### DIFF
--- a/pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.test.ts
+++ b/pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.test.ts
@@ -1,4 +1,3 @@
-
 import { mount } from '@vue/test-utils';
 import { LabeledInput } from './index';
 
@@ -6,7 +5,7 @@ describe('component: LabeledInput', () => {
   it('should emit input only once', () => {
     const value = '2';
     const delay = 1;
-    const wrapper = mount(LabeledInput, {
+    const wrapper = mount(LabeledInput as any, {
       propsData: { delay },
       mocks:     { $store: { getters: { 'i18n/t': jest.fn() } } }
     });
--- a/pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.vue
+++ b/pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.vue
@@ -101,6 +101,7 @@ export default defineComponent({
 
   setup(props, { emit }) {
     const {
+      focused,
       onFocusLabeled,
       onBlurLabeled,
       isDisabled,
@@ -110,6 +111,7 @@ export default defineComponent({
     const { isCompact } = useCompactInput(props);
 
     return {
+      focused,
       onFocusLabeled,
       onBlurLabeled,
       isDisabled,
@@ -123,7 +125,6 @@ export default defineComponent({
     return {
       updated:          false,
       validationErrors: '',
-      focused:          false,
     };
   },
 
--- a/pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.vue
+++ b/pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.vue
@@ -6,9 +6,8 @@ import { escapeHtml } from '@shell/utils/string';
 import cronstrue from 'cronstrue';
 import { isValidCron } from 'cron-validator';
 import { debounce } from 'lodash';
-import { useLabeledFormElement } from '@shell/composables/useLabeledFormElement';
+import { useLabeledFormElement, labeledFormElementProps } from '@shell/composables/useLabeledFormElement';
 import { useCompactInput } from '@shell/composables/useCompactInput';
-import { _EDIT } from '@shell/config/query-params';
 
 declare module 'vue/types/vue' {
   interface Vue {
@@ -22,6 +21,7 @@ export default defineComponent({
   inheritAttrs: false,
 
   props: {
+    ...labeledFormElementProps,
     /**
      * The type of the Labeled Input.
      * @values text, cron, multiline, multiline-password
@@ -54,11 +54,6 @@ export default defineComponent({
     tooltip: {
       default: null,
       type:    [String, Object]
-    },
-
-    tooltipKey: {
-      type:    String,
-      default: null
     },
 
     /**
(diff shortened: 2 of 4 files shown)


pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.test.ts
```diff
@@ -6,7 +6,7 @@ describe('component: LabeledInput', () => {
   it('should emit input only once', () => {
     const value = '2';
     const delay = 1;
-    const wrapper = mount(LabeledInput, {
+    const wrapper = mount(LabeledInput as any, {
```
rak-phillip, line 9: I'm not a fan of using `any` here, but I'm failing to find a solution that will satisfy vue test utils expectations here.

codyrancher: It doesn't appear to cause any issues for me if you remove the type cast

image

pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.vue
```diff
@@ -90,24 +101,85 @@ export default (
...
+    required: {
+      type:    Boolean,
+      default: false,
+    },
+
+    disabled: {
+      type:    Boolean,
+      default: false,
+    },
+  },
```
codyrancher, line 152: I think you mentioned this during our conversation. I don't think we want to have to define these props here or anywhere the composables are used. At the very least I think we should define the props in the composable files and allow the consumer to spread them in the object here. If there's a better solution we should do that instead.

pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.vue
```diff
@@ -144,7 +216,7 @@ export default (
       if (this.type !== 'cron' || !this.value) {
         return;
       }
-      if (!isValidCron(this.value)) {
+      if (typeof this.value === 'string' && !isValidCron(this.value)) {
```
codyrancher, line 219: This seems like it should be pushed into isValidCron. The function should enforce it's own type expectations.

pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.vue
```diff
@@ -173,15 +245,22 @@ export default (
...
      * The max length for the Labeled Input.
      */
-    _maxlength(): number | null {
+    _maxlength(): number | string {
       if (this.type === 'text' && this.maxlength) {
         return this.maxlength;
       }
 
-      return null;
+      return '';
```
codyrancher, line 253: Why this change?

shell/composables/useCompactInput.ts
```diff
@@ -0,0 +1,21 @@
+import { computed } from 'vue';
+import type { ComputedRef } from 'vue';
```
codyrancher, line 2: Is there a reason to keep these separate?

rak-phillip: It's note required, I tend to prefer to try and keep it clear when an import is only used for type annotations. I'm not opposed to merging the two imports.

pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.vue
rak-phillip: `InputHTMLAttributes.maxlength` on input has a type of `Numberish | undefined` (`Numberish` is just `number | string`).

I found that an empty string indicates an unlimited `maxlength` in my testing and felt that it was appropriate, but I suppose that we could opt for `undefined` instead of `null` and get the same outcome.

pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.vue
rak-phillip: I would agree, but `isValidCron` is provided by the `cron-validator` package.

pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.vue
rak-phillip: Yeah, I think it should be possible to do this. I was on the fence about implementing this right now, as I think that the PR is getting quite large. 

I don't think that taking this on will add too much complexity on top of what's represented here, so I'll give it a try.

rak-phillip: This was quite straightforward to perform and we still get prop types exposed in our templates. Let me know what you think!

pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.test.ts
rak-phillip: It looks like the errors I'm receiving are reported by tsc. Can you let me know what the output of `yarn build:lib` is for you without the type cast?

pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.vue
```diff
@@ -90,24 +96,38 @@ export default (
...
+      isCompact,
+    };
   },
 
   data() {
     return {
       updated:          false,
-      validationErrors: ''
+      validationErrors: '',
+      focused:          false,
```
rak-phillip, line 126: I think that `focused` needs to be exposed through the composable.

pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.test.ts
codyrancher: Yeah, running yarn build:lib fails. I've spent a heck of alot of time trying to make this work in a better way. I managed to get a variation of https://github.com/vuejs/vue-test-utils/issues/2026#issuecomment-1460894573 to partially work.

Unfortunately I had to pull in the entire test-utils definition in to fix other issues. I think using `any` is fine until we migrate to vue3 and a new version of test-utils.

codyrancher (approved):

pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.test.ts
codyrancher: After posting this I did have one other thought... Why is the test getting built when running `yarn build:lib`? If it doesn't need to be bundled into the library I found this diff works fine

image

mantis-toboggan-md (approved):

pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.test.ts
rak-phillip: Wouldn't that exclude all of our tests from the compilation process, which begs the question, what's the benefit of writing our tests in TypeScript in the first place?

I might opt for `any` for now and follow up in a separate PR to address the vue-test-utils issues. I'm seeing multiple github issues that appear to be related to the one we're experiencing here, https://github.com/vuejs/vue-test-utils/issues/2026 for example

codyrancher: I think the tests are already being excluded from the compilation process since yarn test:ci functions as already noted. I personally don't see the point of typescript until composition or vue 3 is done.

I'm fine with you making the change later anyway. I just wanted to bring up what I found.

PR #10229, opened 2024-01-04
Nuxt component removal

codyrancher:
Specifically removing 
- `<Nuxt />`
- `<NuxtChild />`
- `<NuxtBuildIndicator />`
- `<NuxtError />`
- `<NuxtLoading />`


shell/initialize/client.js
```diff
@@ -543,108 +538,6 @@ function nuxtReady(_app) {
...
-
-function hotReloadAPI(_app) {
-  if (!module.hot) {
-    return;
-  }
-
-  const $components = getNuxtChildComponents(_app.$nuxt, []);
-
-  $components.forEach(addHotReload.bind(_app));
-}
```
codyrancher, line 629: I got to this when removing `getNuxtChildComponents`. I thought this might break hot reloading with asyncData components but it looks like hot reloading is already broken without these changes. We should just move the existing asyncData components to fetch.

PR #10239, opened 2024-01-09
Replace `Vue.extend` with `defineComponent` in shell

rak-phillip:
### Summary
contributes to #10104

This replaces the usage of `Vue.extend` with `defineComponent` throughout multiple components located under `shell/components/`. `defineComponent` provides better type inference, improved TypeScript support, and will be the recommended way to define components when Dashboard migrates to Vue 3.

- Replaced instances of `Vue.extend` with `defineComponent` for components located under `shell/components/`

### Technical notes summary

Omitting updates to `Vue.extend` in `shell/mixins/compact-input.ts` & `shell/mixins/labeled-form-element.ts`. Updating these mixins to use `defineComponent` caused compilation errors that I don't think will be worth assessing, given that `shell/composables/useCompactInput.ts` & `shell/composables/useLabeledFormElement.ts` exist now. I think it will be a better effort to replace these mixins with their compsable counterparts.

- AsyncButton.vue
- Certificates.vue
- DraggableZone.vue
- Error.vue
- Footer.vue
- GitPicker.vue * note: I can't find any references to this in Dashboard
- PodSecurityAdmission.vue
- management.cattle.io.podsecurityadmissionconfigurationtemplate.vue
- create-edit-view mixin

### DIFF
--- a/shell/components/AsyncButton.vue
+++ b/shell/components/AsyncButton.vue
@@ -1,5 +1,5 @@
 <script lang="ts">
-import Vue from 'vue';
+import { defineComponent, PropType, inject } from 'vue';
 import typeHelper from '@shell/utils/type-helpers';
 
 export const ASYNC_BUTTON_STATES = {
@@ -14,7 +14,13 @@ const TOOLTIP = 'tooltip';
 
 export type AsyncButtonCallback = (success: boolean) => void;
 
-export default Vue.extend<{ phase: string}, any, any, any>({
+interface NonReactiveProps {
+  timer: NodeJS.Timeout | undefined;
+}
+
+const provideProps: NonReactiveProps = { timer: undefined };
+
+export default defineComponent({
   props: {
     /**
      * Mode maps to keys in asyncButton.* translations
@@ -37,7 +43,7 @@ export default Vue.extend<{ phase: string}, any, any, any>({
       default: false,
     },
     type: {
-      type:    String,
+      type:    String as PropType<'button' | 'submit' | 'reset' | undefined>,
       default: 'button'
     },
     tabIndex: {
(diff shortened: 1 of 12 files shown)


rak-phillip:
@torchiaf I'm requesting a review from you because this change makes considerable updates to `shell/components/form/GitPicker.vue`. I don't see where this component is utilized. Would you be able to test it to make sure that there are no regressions introduced here?

cnotv (approved):
PSA parts are fully tested and there's no actual need to QA.
Other elements are all in Storybook and can be tested directly there, as any UI library should.
In any case a correct compilation should assure no breaking changes and be enough.

Can I ask you why the use of the term `NonReactiveProps`? It's good to know, but in the end it's just props, so I was wondering.

Finally I see there's a lot of issues related typing in GitPicker but I'm not going to consider it for this PR. I'll probably create an issue for listing it in Storybook and address all the TS mistakes there.
https://github.com/rancher/dashboard/issues/10240

PS: I think GitPicker has been used in Epinio.

LGTM

rak-phillip:
> Can I ask you why the use of the term NonReactiveProps? It's good to know, but in the end it's just props, so I was wondering.

These are properties that we want to expose to the Vue instance, but they will not take advantage of Vue's reactivity model. I think that `debounceTime` is a good example of this; for the most part, it's a constant that we want to expose so that we can easily use it inside of our component, but there's no need for change detection. 

> Finally I see there's a lot of issues related typing in GitPicker but I'm not going to consider it for this PR.

Interesting. I thought that I had those resolved yesterday, but I'm seeing them today after I rebooted my workstation. I wonder if something silently failed in VSCode while I was developing 🤔 

I'll investigate to see if I can't find a quick fix. If it ends up taking too long, I will follow up with an issue so we can properly address at a later date.

cnotv:
> I also added an additional Vue augment in `shell/types/vue-shim.d.ts` that should help resolve the outstanding TypeScript warning. Can you check it out to see if you're still seeing problems?

I apologize but I phrased my remark in an unclear way 😓 
There's nothing from your side to fix, as it did not get caught and I approved. There's way to define types which are inconvenient although not breaking any rule. The detail have been defined in the issue I created for Storybook, as chance to put hands on it.

cnotv (approved):

PR #10246, opened 2024-01-10
[Automation] - Jenkinsfile formatting and failure flags

izaac:
### Summary

Formatting mainly indentation for better readability and mark the current build as failed after certain error is caught.
Not updating the result on those cases makes the build as success. 

Jenkins Pipeline file updated

### Areas tested
A run was triggered on Jenkins to check that the pipeline is working after before the second stage. Usually if there's syntax errors in the script an exception is thrown immediately by the executor.  

None in the UI just the automation pipeline.

### DIFF
--- a/cypress/jenkins/Jenkinsfile
+++ b/cypress/jenkins/Jenkinsfile
@@ -50,56 +50,60 @@ node {
             }
             }
             try {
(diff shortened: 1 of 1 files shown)


yonasberhe23 (approved):
LGTM

PR #10248, opened 2024-01-10
Replace `vue.extend` with `defineComponent` in Harvester

rak-phillip:
### Summary
contributes to #10104

This replaces the usage of `Vue.extend` with `defineComponent` in Harvester. `defineComponent` provides better type inference, improved TypeScript support, and will be the recommended way to define components when Dashboard migrates to Vue 3.

- Replaced instances of `Vue.extend` with `defineComponent` for components located under `pkg/harvester`

- projectsnamespaces.vue


torchiaf (changes requested):
Change looks good, it requires a rebase.

PR #10253, opened 2024-01-11
fix(service): reset clusterIP(s) field  when clone a service

### LINKED ISSUE #10252: clone service failed
**Setup**
- Rancher version: v2.8.0
- Rancher UI Extensions: v2.8.0

**Describe the bug**

1. When cloning an existing service, you cannot delete or reset the clusterIPs field on the form page because it does not have a form that sets the clusterIPs field.

**To Reproduce**

1. clone a cluster ip service
2. set cluster ip to empty
3. save

**Result**
save failed

**Expected Result**

1. remove clusterIPs field when clone a service
ref: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.29/#service-v1-core

**Screenshots**

image

**Additional context**

ly5156:
### Summary
Fixes #10252

1. Set clusterIP to an empty string, if it exists and the value is not None when clone a service
2. Remove clusterIPs if it exists when clone a service

### DIFF
--- a/shell/edit/__tests__/service.test.ts
+++ b/shell/edit/__tests__/service.test.ts
@@ -1,0 +1,89 @@
+import { shallowMount, Wrapper } from '@vue/test-utils';
+
+import { _CLONE } from '@shell/config/query-params';
+import { cleanHtmlDirective } from '@shell/plugins/clean-html-directive';
+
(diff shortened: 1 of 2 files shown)


jordojordo (approved):
lgtm - I added a few unit tests.

Issue #10045, opened 2023-11-16
kubelet args do not show up in the Advanced tab


gaktive:
Internal reference: SURE-7211
Reported in 2.7.4

Issue description:

Kubelet args do not show up in the advanced tab after they are applied using Terraform

Business impact:

Don't have a way to know what is already configured for the kubelete args

Troubleshooting steps:

    Checked that the TF ran successfully without any issues;
    Checked that it was applied;
    checked that edit as yaml (cluster yaml) for that cluster has the right args that were configured

Repro steps:

    TF code to modify kubelet-arg
    Apply the code
    Confirm that the config has been applied
    Going in the rancher UI, under "Advanced", kubelet args are not showing up

![1699287965_1.png](https://zube.io/files/rancher/343209722d77e9434fd6f14a7c19c897-1699287965_1.png)

Workaround:
Check the config as yaml

Actual behavior:
The kubelet args do appear the UI

Expected behavior:
The kubelet args are not showing from the UI

gaktive:
Possible connection with #9988

nwmac:
Terraform puts the global args in a different place to where the UI expects them - it needs to be updated to be aware of the other location.

torchiaf:
@izaac @momesgin should we close this and open a new issue to track the missing E2E tests ?

momesgin:
@torchiaf I'm working on manually testing this, I believe that would be sufficient enough to close, @yonasberhe23 ?

yonasberhe23:
@momesgin correct, you can close this once the fix is verified

momesgin:
My QA process to test this:

First I created a DO cluster using Terraform without providing `machine_selector_config` or `machine_global_config`,

then I added `machine_selector_config`:
```
    machine_selector_config {
      machine_label_selector{
        match_labels = {
          "mo-from-terraform" = "test-machine-selector-config"
        }
      }
    }
```
and the result:
![image](https://github.com/rancher/dashboard/assets/135728925/34efa61e-9b90-4d41-955c-a51127ae5e71)

in the next step I added `machine_global_config`:
```
    machine_global_config = <<EOF
     kubelet-arg:
       - pippo
    EOF
```
and the result:
![image](https://github.com/rancher/dashboard/assets/135728925/1d50b551-e65c-4585-b2b6-7c270a4d463e)

and finally I updated both:
```
    machine_selector_config {
      machine_label_selector{
        match_labels = {
[...]

PR #10255, opened 2024-01-11
Handle `kubelet-arg` when defined in cluster's global configurations

torchiaf:
### Summary
Fixes https://github.com/rancher/dashboard/issues/10045

### Technical notes summary

Rancher UI expects to find the `kubelet-arg` configuration in `spec.rkeConfig.machineSelectorConfig[0].config` where it can also be defined in `spec.rkeConfig.machineGlobalConfig`, see docs https://ranchermanager.docs.rancher.com/reference-guides/cluster-configuration/rancher-server-configuration/rke2-cluster-configuration#machineselectorconfig.

We are assuming the UI should take the kubelet-arg value from machineGlobalConfig when defined and display it in the same panel where it shows the kubelet-arg from machineSelectorConfig.

Use cases scenarios:

- Create:
    -  only `machineSelectorConfig` is used
- Edit / View:
    - in the case that ONLY machineSelectorConfig in the Yaml, UI can only add/modify kubelet-arg in machineSelectorConfig.
    - in the case that ONLY machineGlobalConfig is defined in the Yaml, UI can only add/modify kubelet-arg in machineGlobalConfig.
    - in the case that BOTH machineSelectorConfig and machineGlobalConfig are defined, the UI handles both and existing args will be updated from where they came from.

- Save the `kubelet-arg` in machineGlobalConfig by manually edit and save cluster's YAML .
- Put the `kubelet-arg` in machineGlobalConfig by Terraform configs, sample file is attached below:

<details><summary>Terraform block</summary>
<p>

```

terraform {
 required_providers {
    rancher2 = {
    source = "rancher/rancher2"
    version = "3.2.0"
    }
  }
}

# Provider bootstrap config
provider "rancher2" {
 api_url = # your rancher url
 insecure = true
 access_key = # your access_key token
 secret_key = # your secret_key token
}

resource "rancher2_machine_config_v2" "rancher2_machine_config_v2" {
 generate_name   = "cfg-terraform"
 digitalocean_config {
  image  = "ubuntu-20-04-x64"
  region  = "nyc1"
  size   = "s-4vcpu-8gb"
 }
}

# Create a new rancher2 RKE2 Cluster
resource "rancher2_cluster_v2" "cfg-terraform" {
 name = "pr-10255-terraform"
 kubernetes_version = "v1.26.10+rke2r2"
 enable_network_policy = false
 default_cluster_role_for_project_members = "admin"
 rke_config {
  # machine_selector_config {
  #   machine_label_selector{
  #     match_labels = {
  #       "key" = "value"
  #     }
  #   }
  # }
  # machine_global_config = <<EOF
  #   kubelet-arg:
  #     - pippo
  # EOF
  machine_pools {
   name = "pool1"
   # this is important
   cloud_credential_secret_name = # DO cluster credentials
   control_plane_role = true
   etcd_role = true
   worker_role = true
   quantity = 1
   machine_config {
    kind = rancher2_machine_config_v2.rancher2_machine_config_v2.kind
    name = rancher2_machine_config_v2.rancher2_machine_config_v2.name
   }
  }
 }
}

```

</p>
</details> 

### Screenshot/Video

Yaml configs

image

UI

image

### DIFF
--- a/shell/edit/provisioning.cattle.io.cluster/tabs/Advanced.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/tabs/Advanced.vue
@@ -35,7 +35,6 @@ export default {
       type:     Object,
       required: true
     },
-
   },
 
   computed: {
@@ -44,9 +43,6 @@ export default {
     },
     agentArgs() {
       return this.selectedVersion?.agentArgs || {};
-    },
-    canRemoveKubeletRow(row, idx) {
-      return idx !== 0;
     },
     serverArgs() {
       return this.selectedVersion?.serverArgs || {};
@@ -68,6 +64,12 @@ export default {
   },
 
   methods: {
+    canRemoveKubeletRow(row, i) {
+      return i !== 0 || !this.agentConfig;
+    },
+    showEmptyKubeletArg(config) {
+      return !this.serverArg?.['kubelet-arg']?.length && !config?.['kubelet-arg']?.length;
+    },
     onInputProtectKernelDefaults(value) {
       Vue.set(this.agentConfig || this.serverConfig, 'protect-kernel-defaults', value);
     }
@@ -83,11 +85,12 @@ export default {
         v-if="agentArgs['kubelet-arg']"
         v-model="rkeConfig.machineSelectorConfig"
         class="mb-20"
+        :mode="mode"
         :add-label="t('cluster.advanced.argInfo.machineSelector.label')"
         :can-remove="canRemoveKubeletRow"
         :default-add-value="{machineLabelSelector: { matchExpressions: [], matchLabels: {} }, config: {'kubelet-arg': []}}"
       >
-        <template #default="{row}">
+        <template #default="{row, i}">
           <template v-if="row.value.machineLabelSelector">
             <h3>{{ t('cluster.advanced.argInfo.machineSelector.title') }}</h3>
             <MatchExpressions
@@ -104,11 +107,40 @@ export default {
           </h3>
 
           <ArrayList
+            v-if="i === 0 && serverConfig['kubelet-arg']"
+            v-model="serverConfig['kubelet-arg']"
+            class="mb-10"
+            data-testid="global-kubelet-arg"
+            :mode="mode"
+            :add-label="t('cluster.advanced.argInfo.machineGlobal.listLabel')"
+          >
+            <template #empty>
+              {{ '' }}
+            </template>
+          </ArrayList>
+
+          <ArrayList
+            v-if="row.value.config"
             v-model="row.value.config['kubelet-arg']"
+            data-testid="selector-kubelet-arg"
             :mode="mode"
             :add-label="t('cluster.advanced.argInfo.machineSelector.listLabel')"
             :initial-empty-row="!!row.value.machineLabelSelector"
-          />
+          >
+            <template
+              v-if="i === 0"
+              #empty
+            >
+              {{ '' }}
+            </template>
+          </ArrayList>
+
+          <div
+            v-if="i === 0 && mode==='view' && showEmptyKubeletArg(row.value.config)"
+            class="text-muted"
+          >
+            &mdash;
+          </div>
         </template>
       </ArrayListGrouped>
       <Banner
--- a/shell/edit/provisioning.cattle.io.cluster/tabs/Advanced.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/tabs/Advanced.vue
@@ -64,6 +64,13 @@ export default {
     },
     protectKernelDefaults() {
       return (this.agentConfig || this.serverConfig)['protect-kernel-defaults'];
+    },
+    kubeletArgTooltip() {
+      if (this.serverConfig?.['kubelet-arg']) {
+        return this.t(`cluster.advanced.argInfo.tooltip.${ this.agentConfig?.['kubelet-arg'] ? 'mixed-args' : 'global-args' }`, null, { raw: true });
+      }
+
+      return null;
     }
   },
 
@@ -108,6 +115,11 @@ export default {
           </template>
           <h3 v-else>
             {{ advancedTitleAlt }}
+            <i
+              v-if="kubeletArgTooltip"
+              v-clean-tooltip="kubeletArgTooltip"
+              class="icon icon-info"
+            />
           </h3>
 
           <ArrayList
--- a/shell/edit/provisioning.cattle.io.cluster/tabs/Advanced.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/tabs/Advanced.vue
@@ -1,5 +1,6 @@
 <script>
 import Vue from 'vue';
+import { _VIEW } from '@shell/config/query-params';
 import { Banner } from '@components/Banner';
 import ArrayListGrouped from '@shell/components/form/ArrayListGrouped';
 import MatchExpressions from '@shell/components/form/MatchExpressions';
@@ -38,6 +39,9 @@ export default {
   },
 
   computed: {
+    isView() {
+      return this.mode === _VIEW;
+    },
     rkeConfig() {
       return this.value.spec.rkeConfig;
     },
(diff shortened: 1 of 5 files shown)


richard-cox (changes requested):
There's a few confusing parts
- How would someone add global level kubelet-args to machineGlobalConfig?
- when viewing yaml in the create rke2 cluster page `machineSelectorConfig` is added at the top level as well as `rkeConfig`. that might be an existing issue
- it looks like we assume the global agent config is the first machineSelectorConfig without a  machineLabelSelector (see shell/models/provisioning.cattle.io.cluster.js `get agentConfig`. should that prefer `spec.rkeConfig.machineGlobalConfig` instead? If so there's a lot of other places the user will report issues
  ```
  get agentConfig() {
    // The one we want is the first one with no selector.
    // If there are multiple with no selector, that will fall under the unsupported message below.
    return this.spec.rkeConfig.machineSelectorConfig.find((x) => !x.machineLabelSelector)?.config;
  }
  ```

I think we might need @nwmac or someone RKE2 side to work through the questions above (and one inline)

torchiaf:
Thanks for the review @richard-cox , answers to your points:

1- `kubelet-arg` can be modified at global level using the 'Add global argument' button (see 2nd snapshot) when its value is already defined in the yaml; the UI allows to save the value only in spec.rkeConfig.machineGlobalConfig[0].config, which is how it works on master. Global kubelet-arg can be created only via Yaml, then via terraform. The change was intended to support 'edge' cases , so when cluster settings are defined by Terraform.
We can change this logic though.

2- Could you provide more details ? I went to creation page, added some kubelet-arg, then went to view Yaml and everything looks good:

    ```machineSelectorConfig
      rkeConfig:
        machineSelectorConfig:
          - config:
              kubelet-arg:
                - adwawd
                - awdwd
              protect-kernel-defaults: false
    ```
    
3- Reading the docs, settings have same effects if defined in both `GlobalMachineConfig` and first element in `machineSelectorConfig` array where machineLabelsSelector is empty, https://ranchermanager.docs.rancher.com/reference-guides/cluster-configuration/rancher-server-configuration/rke2-cluster-configuration#machineselectorconfig , so I think we are supporting both cases and preferring settings defined in `agentConfig`.
In general I agree with checking all other places where `agentConfig` is used and put a fallback to `globalMachineConfigs`; I would open a new issue for that.

shell/edit/provisioning.cattle.io.cluster/tabs/Advanced.vue
```diff
@@ -104,11 +107,40 @@ export default {
           </h3>
 
           <ArrayList
+            v-if="i === 0 && serverConfig['kubelet-arg']"
```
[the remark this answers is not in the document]
torchiaf: I used customer's terraform file, but here is the settings definition https://github.com/rancher/terraform-provider-rancher2/blob/376656c62c6ee854a9f70dd8e2b743b29ccbe100/docs/resources/cluster_v2.md?plain=1#L364 and it seems the relation is spec.rkeConfig.machineGlobalConfig <=> spec.rkeConfig.machineSelectorConfig.config

richard-cox:
Ok, this is starting to make sense
- When there's NO spec.rkeConfig.machineGlobalConfig['kubelet-args'] they see below. 
  - image
  - The `For any machines, use the Kubelet args` section will add an entry to spec.rkeConfig.machineSelectorConfig which contains no selector (aka they apply globally)
  - The `For machines with labels matching` section will add an entry to spec.rkeConfig.machineSelectorConfig which contains a selector
- When there IS spec.rkeConfig.machineGlobalConfig['kubelet-args'] they see the UI in your screenshot which allows users to `Add Global Argument` to update `spec.rkeConfig.machineGlobalConfig['kubelet-args'] `

I'm wondering if there's a use case where both global worlds exists (spec.rkeConfig.machineGlobalConfig['kubelet-args'] AND a spec.rkeConfig.machineSelectorConfig entry with no selector). My concern is that the UX is a bit odd, having two lists for `For any machines, use the Kubelet args` and `Add Global Argument`, when to the user they do the same thing. Can we run through this with Ken and Neil after the sync tomorrow?

richard-cox:
> 2- Could you provide more details ? I went to creation page, added some kubelet-arg, then went to view Yaml and everything looks good:

This is what i see when going straight to Edit Yaml

image

I think this might be a previous bug, but want to validate that it won't interfere with these changes

torchiaf:
> > 2- Could you provide more details ? I went to creation page, added some kubelet-arg, then went to view Yaml and everything looks good:
> 
> This is what i see when going straight to Edit Yaml
> 
> image
> 
> I think this might be a previous bug, but want to validate that it won't interfere with these changes



> > 2- Could you provide more details ? I went to creation page, added some kubelet-arg, then went to view Yaml and everything looks good:
> 
> This is what i see when going straight to Edit Yaml
> 
 Weird, let me double check.

torchiaf:
> I'm wondering if there's a use case where both global worlds exists (spec.rkeConfig.machineGlobalConfig['kubelet-args'] AND a spec.rkeConfig.machineSelectorConfig entry with no selector). My concern is that the UX is a bit odd, having two lists for `For any machines, use the Kubelet args` and `Add Global Argument`, when to the user they do the same thing. Can we run through this with Ken and Neil after the sync tomorrow?

Sure, this is a UX design matter. In my opinion, the cleanest and safer solution would be to add a new section to handle global configs , so there would be a 1:1 match with the YAML, but let's have an overview together.

torchiaf:
@richard-cox duplicated `machineSelectorConfig` is also present in `v2.8.0`. I think the one at top level is just ignored by rancher, but needs to be fixed.

shell/edit/provisioning.cattle.io.cluster/tabs/Advanced.vue
```diff
@@ -101,14 +115,48 @@ export default {
...
+            :mode="mode"
+            :add-label="t('cluster.advanced.argInfo.machineGlobal.listLabel')"
+          >
+            <template #empty>
+              {{ '' }}
+            </template>
+          </ArrayList>
+
+          <ArrayList
+            v-if="row.value.config && (row.value.config['kubelet-arg'] || !serverConfig['kubelet-arg'])"
```
richard-cox, line 139: i need to play around with this more, but i think it should be below  i.e. show if there IS a machineSelectorConfig without a selector and with 'kubelet-arg' or there is no machineGlobalConfig a 'kubelet-arg'

```
v-if="row.value.config && ((agentConfig || {})['kubelet-arg'] || !serverConfig['kubelet-arg'])"
```

to test i've been tweaking 

```
  created() {
    this.value.spec.rkeConfig.machineGlobalConfig['kubelet-arg'] = ['imfromglobal'];
    this.value.spec.rkeConfig.machineSelectorConfig[0]['kubelet-arg'] = ['imfromselector'];
  },
```

richard-cox (commented):
Still reviewing, just updating. 

There's also some strangeness going on with value.spec.rkeConfig.machineSelectorConfig[0]. looks like its populated via the kub3 version's agentArgs which is then overwritten. that's existing functionality, not a regression here

shell/edit/provisioning.cattle.io.cluster/tabs/Advanced.vue
torchiaf: I changed `row.value.config['kubelet-arg']` -> `(agentConfig || {})['kubelet-arg']`

richard-cox: That change has been reverted, how come?

richard-cox (changes requested):
Just one comment to go https://github.com/rancher/dashboard/pull/10255#discussion_r1483028473

shell/edit/provisioning.cattle.io.cluster/tabs/Advanced.vue
torchiaf: @richard-cox 

- I had a quick look to the `agentConfig` value and it's the same of row.value.config in this case; it's tested in the unit test though
-  The unit tests fail when using `agentConfig`, since it's a computed value which triggers a DOM update not cacthed in time by unit test.

richard-cox (approved):
I made a small improvement, not related to changes in this PR but in the area.

More detail in the commit message, but tweaked a translation to avoid confusing `For all machines..` --> `For any machines..` text when adding the first selector below

PR #10258, opened 2024-01-11
System namespaces curation

### LINKED ISSUE #10214: Namespaces ending with -system are automatically seen as System Namespaces
Internal reference: SURE-6914

When you create a namespace that contains `-system` it will always be handled as a system project and cannot be seen by users of the project.

We should have a curated list of Rancher's system namespaces and only use those, rather than assuming all namespaces ending in `-system` are system namespaces.

aalves08:
### Summary
Fixes #10214 

- update logic to decide if a namespace is a system one
- improve curated list of system namespaces
- update unit tests

### Technical notes summary

Agreed with @nwmac the curated list of system namespaces + the change in the `isSystem` logic

### DIFF
--- a/shell/config/system-namespaces.js
+++ b/shell/config/system-namespaces.js
@@ -12,4 +12,7 @@ export default [
   'linkerd',
   'security-scan',
   'tekton-pipelines',
+  'cis-operator-system',
+  'istio-system',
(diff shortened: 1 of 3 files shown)


nwmac (approved):
LGTM

PR #10263, opened 2024-01-12
Restructure longhorn UI proxy link to contain dynamic namespace

### LINKED ISSUE #10259: Longhorn section in Rancher UI support for Longhorn installations in Namespaces other than longhorn-system
Internal Reference: SURE-5743

longhorn-system namespace is hard-coded in the Rancher UI link
https://github.com/rancher/dashboard/blob/v2.6.9/shell/pages/c/_cluster/longhorn/index.vue#L31

We should detect the longhorn UI app service and use its namespace, so that the link works if you install Longhorn in a namespace other than `longhorn-systsem`

jordojordo:
### Summary
Fixes #10259 

The Longhorn frontend UI would fail to load as the `namespace` was hard-coded to `longhorn-system`. Although setting this namespace is the [recommended approach](https://longhorn.io/docs/1.5.3/deploy/install/install-with-helm/#installing-longhorn) when installing Longhorn with Helm, users may still determine their own namespace.

This adds a computed property to find the Longhorn UI based on the label `app=longhorn-ui` which is defined in the [Longhorn Helm chart](https://github.com/rancher/charts/blob/38aed06f77002690b4857d84e79252bed828dc82/charts/longhorn/103.2.1%2Bup1.5.3/templates/deployment-ui.yaml#L16). 
If the service is found, the `externalLinks` will contain the correct namespace of this service in the `link` property.

### Technical notes summary

I've also added a banner to display an error when no external links have been populated.

### Screenshot/Video

https://github.com/rancher/dashboard/assets/40806497/929e5993-c8c2-40d8-9792-1b96624590d0

### DIFF
--- a/shell/pages/c/_cluster/longhorn/index.vue
+++ b/shell/pages/c/_cluster/longhorn/index.vue
@@ -19,12 +19,18 @@ export default {
 
   async fetch() {
     if ( this.$store.getters['cluster/schemaFor'](SERVICE) ) {
-      await this.$store.dispatch('cluster/findAll', { type: SERVICE });
+      this.uiServices = await this.$store.dispatch('cluster/findMatching', {
+        type:     SERVICE,
+        selector: 'app=longhorn-ui'
+      });
     }
   },
 
   data() {
-    return { longhornImgSrc: require('~shell/assets/images/vendor/longhorn.svg') };
+    return {
+      longhornImgSrc: require('~shell/assets/images/vendor/longhorn.svg'),
+      uiServices:     null
+    };
   },
 
   computed: {
@@ -44,10 +50,6 @@ export default {
       }
 
       return [];
-    },
-
-    uiServices() {
-      return this.$store.getters['cluster/matching'](SERVICE, 'app=longhorn-ui');
     }
   }
 };
--- a/shell/pages/c/_cluster/longhorn/index.vue
+++ b/shell/pages/c/_cluster/longhorn/index.vue
@@ -3,42 +3,59 @@ import { mapGetters } from 'vuex';
 
 import InstallRedirect from '@shell/utils/install-redirect';
 
+import { SERVICE } from '@shell/config/types';
 import { NAME, CHART_NAME } from '@shell/config/product/longhorn';
 
+import IconMessage from '@shell/components/IconMessage';
 import LazyImage from '@shell/components/LazyImage';
+import Loading from '@shell/components/Loading';
 
 export default {
-  components: { LazyImage },
+  components: {
+    IconMessage, LazyImage, Loading
+  },
 
   middleware: InstallRedirect(NAME, CHART_NAME),
 
+  async fetch() {
+    if ( this.$store.getters['cluster/schemaFor'](SERVICE) ) {
+      await this.$store.dispatch('cluster/findAll', { type: SERVICE });
+    }
+  },
+
   data() {
-    return {
-      externalLinks:  [],
-      longhornImgSrc: require('~shell/assets/images/vendor/longhorn.svg'),
-    };
+    return { longhornImgSrc: require('~shell/assets/images/vendor/longhorn.svg') };
   },
 
-  computed: { ...mapGetters(['currentCluster']) },
+  computed: {
+    ...mapGetters(['currentCluster']),
 
-  mounted() {
-    this.externalLinks = [
-      {
-        enabled:     true,
-        iconSrc:     this.longhornImgSrc,
-        label:       'longhorn.overview.linkedList.longhorn.label',
-        description: 'longhorn.overview.linkedList.longhorn.description',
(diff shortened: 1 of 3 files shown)


shell/pages/c/_cluster/longhorn/index.vue
```diff
@@ -19,18 +22,30 @@ export default {
...
-      },
-    ];
+    uiService() {
+      return this.$store.getters['cluster/matching'](SERVICE, 'app=longhorn-ui')[0];
+    }
+  },
+
+  async mounted() {
+    if ( this.$store.getters['cluster/schemaFor'](SERVICE) ) {
+      await this.$store.dispatch('cluster/findAll', { type: SERVICE });
```
nwmac, line 35: There might be a large number of services, so for performance, we can use the label selector to look just for the longhorn one.

nwmac (changes requested):

shell/pages/c/_cluster/longhorn/index.vue
jordojordo: That was my original approach with:
```js
  async mounted() {
    if ( this.$store.getters['cluster/schemaFor'](SERVICE) ) {
      const uiService = await this.$store.dispatch('cluster/findMatching', {
        type:     SERVICE,
        selector: 'app=longhorn-ui'
      });

      if ( !isEmpty(uiService) && uiService.metadata?.namespace ) {
        this.externalLinks = [
          {
            enabled:     true,
            iconSrc:     this.longhornImgSrc,
            label:       'longhorn.overview.linkedList.longhorn.label',
            description: 'longhorn.overview.linkedList.longhorn.description',
            link:        `/k8s/clusters/${ this.currentCluster.id }/api/v1/namespaces/${ uiService.metadata.namespace }/services/http:longhorn-frontend:80/proxy/`
          },
        ];
      }
    }
  },
```

The service is never found even when it exists. If I fetch all of the services with `findAll`, then run `findMatching` this will work but defeats the purpose. Am I using `findMatching` incorrectly?


https://github.com/rancher/dashboard/assets/40806497/1d605948-ab4a-4235-a423-c6e216ebe16f

nwmac: You can supply `force: true` in the options to bypass that behaviour.

nwmac:
@jordojordo I had a look and the issue is that findMatching returns an array of matching services, not a single service which your code was looking for.

This worked for me:
```
  async mounted() {
    if ( this.$store.getters['cluster/schemaFor'](SERVICE) ) {
      await this.$store.dispatch('cluster/findAll', { type: SERVICE });

      const uiServices = await this.$store.dispatch('cluster/findMatching', { type: SERVICE, selector: 'app=longhorn-ui' });

      console.error(uiServices);

      if (uiServices && uiServices.length ===1 && uiServices[0].metadata?.namespace) {
        this.externalLinks = [
          {
            enabled:     true,
            iconSrc:     this.longhornImgSrc,
            label:       'longhorn.overview.linkedList.longhorn.label',
            description: 'longhorn.overview.linkedList.longhorn.description',
            link:        `/k8s/clusters/${ this.currentCluster.id }/api/v1/namespaces/${ uiServices[0].metadata.namespace }/services/http:longhorn-frontend:80/proxy/`
          },
        ];
      }
    }
  },
```

You might want to use fetch rather than mounted as with this, the red banner flicks up briefly before the service is found.

I would also suggest we use a vertical `IconMessage` rather than the error banner, eg:

image

Lastly, if we can avoid using lodash that would be great - generally I think we want to try and get that out of the code base, espcially for things like `isEmpty` (code above does not use that)

nwmac (changes requested):
See comment

jordojordo:
I updated this to use the `fetch` method rather than the `mounted` when fetching the services, converted the `externalLinks` into a computed property, and switched the Banner to an IconMessage component with a more specific message. Also updated the tests to fit this new structure.

shell/pages/c/_cluster/longhorn/index.vue
```diff
@@ -3,42 +3,59 @@ import { mapGetters } from 'vuex';
...
-  components: { LazyImage },
+  components: {
+    IconMessage, LazyImage, Loading
+  },
 
   middleware: InstallRedirect(NAME, CHART_NAME),
 
+  async fetch() {
+    if ( this.$store.getters['cluster/schemaFor'](SERVICE) ) {
+      await this.$store.dispatch('cluster/findAll', { type: SERVICE });
```
nwmac, line 22: We can use `findMatching`, but it was broken - this PR fixes it  https://github.com/rancher/dashboard/pull/10413

nwmac (changes requested):

jordojordo:
Restructured this to use the `findMatching` action.


https://github.com/rancher/dashboard/assets/40806497/73c6ec8d-0e50-408d-a047-064531850ef3

nwmac (approved):

enneitex:
Hey,

I know this is a bit late as this has been merged for a long time but this change means that users need mors RBAC than before to access longhorn UI. They need to be able to list services in every namespaces, this was not required before

PR #10272, opened 2024-01-15
Fix issue LabeledInput with type "multiline"

### LINKED ISSUE #10249: Global Settings: Banners: Text is removed from input field after user focuses on different page element
**Setup**
- Rancher version: v2.8.0
- Dashboard version:
- Browser type & version: Any

**Describe the bug**
Text is removed from the input field after the user focuses on a different page element. This issue is specific to localhost setup and for the Header Banner, Footer Banner, and Login Screen Banner only.

**To Reproduce**

1. run Rancher instance locally
2. login to Rancher UI as an admin
3. go to Global Settings -> Banners (https://localhost:8005/c/local/settings/banners)
4. go to 'Header Banner' section 
5. select the checkbox for 'Show Banner in Header'
6. Type banner description in input field ('test') 
7. click anywhere on the page
8. notice text is removed from the input field
9. go to remote setup (https://yonasb.qa.rancher.space/dashboard/c/local/settings/banners)
10. repeat steps 4-7
11. notice text is not removed from the input field

**Result**
Text is removed from the input field after the user focuses on a different page element.

**Expected Result**
Text remains in the input field after the user focuses on a different page element.

**Screenshots**

https://github.com/rancher/dashboard/assets/127343932/57bde139-a1c8-49c3-b1db-75951e4b496f

aalves08:
### Summary
Fixes #10249 

- fix issue where `LabeledInput` input event was not working for type `multiline` (the component used is `TextAreaAutoGrow` which emits the `value` instead of an `event`)

### DIFF
--- a/pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.vue
+++ b/pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.vue
@@ -244,9 +244,12 @@ export default defineComponent({
     /**
      * Emit on input with delay. Note: Arrow function is avoided due context
      * binding.
+     *
+     * NOTE: In multiline, TextAreaAutoGrow emits a string with the value
+     * https://github.com/rancher/dashboard/issues/10249
      */
-    delayInput(event: Event): void {
-      const value = (event?.target as HTMLInputElement)?.value;
+    delayInput(val: string | Event): void {
+      const value = typeof val === 'string' ? val : (val?.target as HTMLInputElement)?.value;
 
       this.$emit('input', value);
     },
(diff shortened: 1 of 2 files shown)


pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.vue
```diff
@@ -244,9 +244,12 @@ export default defineComponent({
     /**
      * Emit on input with delay. Note: Arrow function is avoided due context
      * binding.
+     *
+     * NOTE: In multiline, TextAreaAutoGrow emits a string with the value
+     * https://github.com/rancher/dashboard/issues/10249
      */
-    delayInput(event: Event): void {
-      const value = (event?.target as HTMLInputElement)?.value;
+    delayInput(val: string | Event): void {
```
rak-phillip, line 251: I can see that this regressed as a part of some TypeScript refactoring[^1] as a result of to move to `defineComponent`. Where `delyInput` used to always accept a string, but `$event.target.value` isn't valid inside of templates, as `$event` or `target` can be undefined. 

[^1] https://github.com/rancher/dashboard/pull/10226/files#diff-7740b11d8a0c151dff21cdb333af5b5312b35719c169e1938663019db42502e8R248-R250

PR #10278, opened 2024-01-17
Fix parse-tag-name script to use correct GITHUB_WORKFLOW_TYPE

jtomasek:
build-extension-charts.yml workflow passes "charts" as GITHUB_WORKFLOW_TYPE build-extension-catalog.yaml workflow passes "catalog" as GITHUB_WORKFLOW_TYPE

This change utilizes correct workflow types in the parse-tag-name script. For charts it checks against package.json in pkg and for catalog it checks against root package.json

Examples of failing workflows:
https://github.com/suse-edge/dashboard-extensions/actions/runs/7554197375
https://github.com/suse-edge/dashboard-extensions/actions/runs/7554197372

### DIFF
--- a/.github/workflows/build-extension-catalog.yml
+++ b/.github/workflows/build-extension-catalog.yml
@@ -35,14 +35,6 @@ jobs:
           git config user.name 'github-actions[bot]'
           git config user.email 'github-actions[bot]@users.noreply.github.com'
 
-      - name: Parse Extension Name
-        if: github.ref_type == 'tag'
-        id: parsed-name
-        env:
-          GH_TOKEN: ${{ github.token }}
-        run: |
-          yarn parse-tag-name ${{ inputs.tagged_release }} ${{ github.run_id }} "catalog"
-
       - name: Login to GitHub Container Registry
         uses: docker/login-action@v2
         with:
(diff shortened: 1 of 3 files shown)


jordojordo (commented):
This PR looks good to me and makes sense to fix the condition for the workflow type, but it will not fix the issue you're experiencing. 

The step within the reusable workflow that [runs this script](https://github.com/rancher/dashboard/blob/ebd2834bd65fd1856b81ce7197412eb880de5048/.github/workflows/build-extension-charts.yml#L40-L46) needs to be below the step that [installs the node modules](https://github.com/rancher/dashboard/blob/ebd2834bd65fd1856b81ce7197412eb880de5048/.github/workflows/build-extension-charts.yml#L72-L73). This would explain why the script is[ not found](https://github.com/suse-edge/dashboard-extensions/actions/runs/7554197375/job/20566558427#step:4:11).

I believe I originally had this step before installing the node modules to relieve the workflow from doing too much work if it was going to be unusable, but I forgot that the modules needed to be installed for this script to be accessible. I'm sorry for this oversight.

jtomasek:
@jordojordo thanks for the pointers, I've add a commit which moves the step.

jordojordo (approved):
LGTM

PR #10300, opened 2024-01-18
[2.7] AWS external cloud provider support

mantis-toboggan-md:
Fixes https://github.com/rancher/dashboard/issues/10299
Fixes https://github.com/rancher/dashboard/issues/10266

This is a backport of https://github.com/rancher/dashboard/pull/9643 and https://github.com/rancher/dashboard/pull/9914

### Summary
UX for these issues changed from what was originally outlined in the above issues. The initial plan was to remove the aws cloud provider option for k8s versions >=1.27 (rke2 cloud provider options come from an api call, so no need to explicitly remove the option in the UI). Generalized UI code was added to handle the possibility that a user is upgrading a cluster with a cloud provider that is no longer supported. 

The plan for aws external cloud provider support changed, and new UX involves selecting the 'aws' cloud provider option then adding some extra configuration. #9914 adds a banner warning the user that they need to do this.

### Technical notes summary
Per offline conversations w/ Neil and Gary we're keeping the code from #9643 in our backports - though no longer relevant, it doesn't cause bugs.

When upgrading a cluster using the aws cloud provider to >=1.27, a banner should appear warning the user that they will need to supply some extra configration for the external aws cloud provider. The same banner should appear when creating a new >=1.27 cluster with the aws cloud provider selected.

2.-7head doesn't have a k8s 1.27 option available yet. You fake a 1.27 version option for testing purposes by adding code [here](https://github.com/rancher/dashboard/blob/1c09fbd83dbc18de148487ef74c28175d5bd3455/shell/edit/provisioning.cattle.io.cluster/rke2.vue#L1215)

```
const add = JSON.parse(JSON.stringify(this.rke2Versions[this.rke2Versions.length -1]));
add.id = 'v1.27.0+rke2r1';
add.version = add.id;
this.rke2Versions.push(add);
 ```

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -1846,6 +1846,7 @@ cluster:
       header: Cloud Provider Config
       defaultValue:
         label: Default - RKE2 Embedded
+      unsupported: The current Cloud Provider is not supported by this version of Kubernetes. The Cloud Provider has been changed to External. Please use the Cloud Provider Config to supply an out-of-tree configuration as needed.
     security:
       header: Security
     cis:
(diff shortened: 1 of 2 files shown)


momesgin (approved):
lgtm

Issue #10047, opened 2023-11-17
`Enable IPv6 support` does not work with RKE2 + Cilium


masap:
Internal reference: SURE-7260

**Setup**
- Rancher version: 2.7.9
- Rancher UI Extensions:
- Browser type & version: Firefox 119.0.1 (64 bit)

**Describe the bug**
Enable IPv6 with Rancher UI but `enable-ipv6` is `false`.

```
$ kubectl -n kube-system get cm cilium-config -o json | jq '.data."enable-ipv6"'
"false"
```

**To Reproduce**
Enable IPv6 with Rancher UI and check the value with trailing command.

```
$ kubectl -n kube-system get cm cilium-config -o json | jq '.data."enable-ipv6"'
```

**Result**
```
"false"
```

**Expected Result**
```
"true"
```

**Screenshots**
![rke2-cilium-ipv6](https://github.com/rancher/dashboard/assets/1505016/f7c4ef7a-67aa-421b-af1a-fdf5cb755d08)

**Additional context**

gaktive:
Internal reference: SURE-7260
As noted, there is a PR already for us to review.

izaac:
The automated test is enough to validate the spec is sent correctly. This would need manual validation on an actual cluster.

momesgin:
I tested it with RKE2 Digital Ocean and I was able to get `"true"` after running the command in the description.

But I found a problem, if you enable ipv6 during the creation process your cluster get stuck in updating status: `waiting for cluster agent to connect`. You can still enable ipv6 later on without causing an issue. This is not happening in 2.8.

I compared the payloads of 2.8 and 2.9 and Cilium's related configs were the same, so UI probably hasn't caused a regression.

2.8:

```
{
    "type": "provisioning.cattle.io.cluster",
    "metadata": {
        "namespace": "fleet-default",
        "name": "mo-test-2-8"
    },
    "spec": {
        "rkeConfig": {
            "chartValues": {
                "rke2-cilium": {
                    "cilium": {
                        "ipv6": {
                            "enabled": true
                        }
                    }
                }
            },
            "upgradeStrategy": {
                "controlPlaneConcurrency": "1",
                "controlPlaneDrainOptions": {
                    "deleteEmptyDirData": true,
                    "disableEviction": false,
                    "enabled": false,
                    "force": false,
                    "gracePeriod": -1,
                    "ignoreDaemonSets": true,
                    "skipWaitForDeleteTimeoutSeconds": 0,
[...]

PR #10304, opened 2024-01-19
Fix incorrect spec structure for cilium ipv6 configuration

nwmac:
Fixes https://github.com/rancher/dashboard/issues/10047

Supersedes https://github.com/rancher/dashboard/pull/10048

This PR fixes the IPv6 support for RKE2 + Cilium.

It ensures migration of the setting in the case that a 1.23 cluster is upgraded and it adds automated tests.

Notes:
- Changed the `Basics.vue` Tab component to take care of the structure and emit the change for the parent `rke2.vue` to update the model with.
- Renames two e2e test to `.test.st` from `.tests.ts` so that the typescript imports work correctly in vscode
- Updated checkbox and labeled select POs to make them more reliable
- Added a single e2e test that validates the changes made to `Basics.vue` that setting ipv6 is correctly wired into the request sent to create a cluster.
- Added unit tests to cover the various cases - focused on `Basics.vue` - ensuring that the correct structrue is used depending on if the k8s version > 1.23.5. Also checks that the settings are migrated back and forward when you change the k8s version

### DIFF
--- a/cypress/e2e/po/components/checkbox-input.po.ts
+++ b/cypress/e2e/po/components/checkbox-input.po.ts
@@ -16,7 +16,7 @@ export default class CheckboxInputPo extends ComponentPo {
    * @returns
    */
   set(): Cypress.Chainable {
-    return this.input().click({ force: true });
+    return this.self().find('.checkbox-custom').click();
   }
 
   /**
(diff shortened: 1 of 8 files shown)


aalves08 (approved):
LGTM! good work Neil! Very thorough with the testing, covering both the legacy a new structures 👍 🚢 it

izaac:
The automated test is enough to validate the spec is sent correctly. This would need manual validation on an actual cluster.

PR #10341, opened 2024-01-26
Fixing an issue where a field can sometimes be undefined

codyrancher:
It was reported in slack that the bundle page would error out when visiting the bundle pages due to an undefined reference. This resolves that issue.

Found the problem after replaying the user generated HAR file which yielded this in the console:
image


aalves08 (approved):
LGTM, but still in draft

PR #10364, opened 2024-01-31
Fix parse tag name condition to check for proper workflow types

### LINKED ISSUE #10363: parse-tag-name script targeting incorrect workflow type
**Describe the bug**

When running a an Extension release workflow the [`Parse Extension Name` step](https://github.com/rancher/dashboard/blob/b08e9288473b7c19c16e8825ed67d05f931d06cf/.github/workflows/build-extension-charts.yml#L73) the `parse-tag-name` script which has a condition checking for the [`charts` workflow type](https://github.com/rancher/dashboard/blob/b08e9288473b7c19c16e8825ed67d05f931d06cf/shell/scripts/extension/parse-tag-name#L8).

This causes the workflow to be automatically canceled as it was originally intended to check for naming collisions when releasing an Extension Catalog Image. Since this is now checking a "charts" release, the release tag name should absolutely equal a `pkg/<pkg-name>` causing the workflow to be canceled.

**To Reproduce**

See [this Elemental UI workflow action](https://github.com/rancher/elemental-ui/actions/runs/7724489669/job/21056821534#step:11:36) which was canceled.

**Result**
The workflow was canceled as `GITHUB_RELEASE_TAG` (`elemental-1.3.0-rc2`) matched `PKG_NAME` (`elemental-1.3.0-rc2`).

**Expected Result**

The workflow should have passed this step.

**Additional context**

The condition which checks for the workflow type should be checking for `catalog` instead of `charts`.

jordojordo:
### Summary
Fixes #10363 

This changes the workflow type in `parse-tag-name` to check for `catalog` instead of `charts`. The intent of this condition is to determine if there are any naming collisions with the release tag name and any package names when releasing an Extension Catalog Image.

The `else` condition will be checking for naming collisions with the release tag name and the root package name.

### DIFF
--- a/shell/scripts/extension/parse-tag-name
+++ b/shell/scripts/extension/parse-tag-name
@@ -4,8 +4,8 @@ GITHUB_RELEASE_TAG=$1
 GITHUB_RUN_ID=$2
 GITHUB_WORKFLOW_TYPE=$3
 
-# Check packages for released tag name
-if [[ "${GITHUB_WORKFLOW_TYPE}" == "charts" ]]; then
+# Ensure "catalog" workflow release tag name does not match a pkg/<pkg-name>
+if [[ "${GITHUB_WORKFLOW_TYPE}" == "catalog" ]]; then
   for d in pkg/*/ ; do
     pkg=$(basename $d)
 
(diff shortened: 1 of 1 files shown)


aalves08 (approved):
LGTM

PR #10366, opened 2024-01-31
Update logic of 'hasError' in 'provisioning.cattle.io.cluster' model

### LINKED ISSUE #10043: One EKS cluster not clickable in the UI
Internal reference: SURE-7035
Reported in 2.7.6

Issue description:
The cluster EKS eks-entsvcs-pci-dev (cluster ID: c-nqdv4) is not clickable from the explorer after Rancher upgrade from 2.6.12 to 2.7.6

Business impact:
The user can't click on that cluster from the explorer

Troubleshooting steps:
    reviewed the rancher pods and cluster agent logs
    rolling restart of rancher pods
    re-deployed the cluster agents

Repro steps:
Support was not able to reproduce it, but the user wants to know what is causing this behavior 

Workaround:
None

Using the cluster management or directly accessing the link for that cluster with something like this https://hostname/dashboard/c/c-xxxxx/explorer

Actual behavior:
A cluster is not clickable from the explorer

Expected behavior:
A cluster should be and clickable when it is Active in Rancher

Addenda:
- Logs & HAR files available via the SURE ID.

aalves08:
### Summary
Fixes #10043 

- Update logic of `hasError` in `provisioning.cattle.io.cluster` model
- Add unit test for `cluster.status.conditions` which affect the outcome of `hasError` model property

- Check SURE details in #10043 and grab the HAR files
- Use https://github.com/rancher/dashboard/blob/master/docs/developer/har-files.md#yarn-dev to run this branch
- Verify that the cluster is available to "explore"

**NOTE:** I never tested it this way, so I am not 100% this will work. All of the investigation was based on the HAR files output and code checks in Dashboard which I've tried to sum up in the comments of #10043

### DIFF
--- a/shell/models/provisioning.cattle.io.cluster.js
+++ b/shell/models/provisioning.cattle.io.cluster.js
@@ -903,14 +903,18 @@ export default class ProvCluster extends SteveModel {
           return true;
         }
 
-        const errorConditions = this.status?.condition.filter((c) => c.error === true);
-        const readyConditions = this.status?.condition.filter((c) => c.type === 'Ready');
+        const errorConditions = this.status?.conditions.filter((c) => c.error === true);
+        const readyConditions = this.status?.conditions.filter((c) => c.type === 'Ready');
 
         const lastErrorCondition = errorConditions.reduce((a, b) => ((a.lastUpdateTime > b.lastUpdateTime) ? a : b));
         const lastReadyCondition = readyConditions.reduce((a, b) => ((a.lastUpdateTime > b.lastUpdateTime) ? a : b));
 
         const lastErrorConditionDate = new Date(lastErrorCondition.lastUpdateTime) || null;
         const lastReadyConditionDate = new Date(lastReadyCondition.lastUpdateTime) || null;
+
+        console.log('this.status?.conditions', lastReadyConditionDate.getTime());
+        console.log('lastReadyConditionDate.getTime()', lastReadyConditionDate.getTime());
+        console.log('lastErrorConditionDate.getTime()', lastErrorConditionDate.getTime());
 
         // fallback is false because something might be wrong with the condition 'lastUpdateTime' prop
         return lastErrorConditionDate && lastReadyConditionDate ? lastReadyConditionDate.getTime() > lastErrorConditionDate.getTime() : false;
--- a/shell/models/provisioning.cattle.io.cluster.js
+++ b/shell/models/provisioning.cattle.io.cluster.js
@@ -897,30 +897,20 @@ export default class ProvCluster extends SteveModel {
     // We now check if there's a ready condition after an error, which helps dictate the readiness of a cluster
     // Based on the findings in https://github.com/rancher/dashboard/issues/10043
     if (this.status?.conditions && this.status?.conditions.length) {
+      // if there are errors, we compare with how recent the "Ready" condition is compared to that error, otherwise we just move on
       if (this.status?.conditions.some((c) => c.error === true)) {
         // there's no ready condition and has an error, mark it
         if (!this.status?.conditions.some((c) => c.type === 'Ready')) {
           return true;
         }
 
-        const errorConditions = this.status?.conditions.filter((c) => c.error === true);
-        const readyConditions = this.status?.conditions.filter((c) => c.type === 'Ready');
+        const filteredConditions = this.status?.conditions.filter((c) => c.error === true || c.type === 'Ready');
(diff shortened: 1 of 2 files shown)


aalves08:
@richard-cox me and @nwmac chose this course of action in a sync we had earlier in the day. @nwmac I think that I've captured the correct logic of what we discussed in the code change

shell/models/provisioning.cattle.io.cluster.js
```diff
@@ -891,7 +891,37 @@ export default class ProvCluster extends SteveModel {
...
+        const errorConditions = this.status?.conditions.filter((c) => c.error === true);
+        const readyConditions = this.status?.conditions.filter((c) => c.type === 'Ready');
+
+        const lastErrorCondition = errorConditions.reduce((a, b) => ((a.lastUpdateTime > b.lastUpdateTime) ? a : b));
+        const lastReadyCondition = readyConditions.reduce((a, b) => ((a.lastUpdateTime > b.lastUpdateTime) ? a : b));
+
+        const lastErrorConditionDate = new Date(lastErrorCondition.lastUpdateTime) || null;
+        const lastReadyConditionDate = new Date(lastReadyCondition.lastUpdateTime) || null;
+
+        // fallback is false because something might be wrong with the condition 'lastUpdateTime' prop
```
nwmac, line 915: What can be wrong with the lastUpdateTime? Why do we need to convert to dates for the comparison, where as above we used the lastUpdateTime directly for comparison?

shell/models/provisioning.cattle.io.cluster.js
```diff
@@ -891,7 +891,37 @@ export default class ProvCluster extends SteveModel {
...
 
   get hasError() {
-    return this.status?.conditions?.some((condition) => condition.error === true);
+    // Before we were just checking for this.status?.conditions?.some((condition) => condition.error === true)
+    // but this is wrong as an error might exist but it might not be meaningful in the context of readiness of a cluster
+    // which is what this 'hasError' is used for.
+    // We now check if there's a ready condition after an error, which helps dictate the readiness of a cluster
+    // Based on the findings in https://github.com/rancher/dashboard/issues/10043
+    if (this.status?.conditions && this.status?.conditions.length) {
+      if (this.status?.conditions.some((c) => c.error === true)) {
```
nwmac, line 900: Can we add a comment to say - if any conditions are in error, then we use additional logic, otherwise return quickly as we know there are no errors

shell/models/provisioning.cattle.io.cluster.js
```diff
@@ -891,7 +891,37 @@ export default class ProvCluster extends SteveModel {
...
+
+        const lastErrorConditionDate = new Date(lastErrorCondition.lastUpdateTime) || null;
+        const lastReadyConditionDate = new Date(lastReadyCondition.lastUpdateTime) || null;
+
+        // fallback is false because something might be wrong with the condition 'lastUpdateTime' prop
+        return lastErrorConditionDate && lastReadyConditionDate ? lastErrorConditionDate.getTime() > lastReadyConditionDate.getTime() : false;
+      }
+
+      // here we just fallback to the previous original logic
+      return false;
```
nwmac, line 920: Don't need this as there is a fallback below

nwmac (changes requested):

shell/models/provisioning.cattle.io.cluster.js
```diff
@@ -891,7 +891,37 @@ export default class ProvCluster extends SteveModel {
...
+    // Based on the findings in https://github.com/rancher/dashboard/issues/10043
+    if (this.status?.conditions && this.status?.conditions.length) {
+      if (this.status?.conditions.some((c) => c.error === true)) {
+        // there's no ready condition and has an error, mark it
+        if (!this.status?.conditions.some((c) => c.type === 'Ready')) {
+          return true;
+        }
+
+        const errorConditions = this.status?.conditions.filter((c) => c.error === true);
+        const readyConditions = this.status?.conditions.filter((c) => c.type === 'Ready');
```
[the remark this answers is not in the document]
aalves08: I thought that the `conditions` array was like a record where we could have more entries of the same type. That simplifies the logic a lot. Thanks for the input

shell/models/provisioning.cattle.io.cluster.js
aalves08: Just wanted to make sure if the BE sent a malformed ISO date Dashboard wouldn't "crash". With the simplified logic based on the previous comment, I've updated this part

shell/models/provisioning.cattle.io.cluster.js
aalves08: true. Fixed

nwmac (approved):

PR #10380, opened 2024-02-01
update logic of `WorkloadDetailEndpoints` to consider hostname as a valid link if there is an `ingress` associated

### LINKED ISSUE #8087: Endpoints point to IP and not hostname, which complicates Ingress
Internal reference: SURE-5591
Reported in 2.6.8

The endpoints listed under Deployment Details point to IP and not hostname and are thus not useable. Suggestion is to remove Endpoints line from Deployment Details. 

Request description:
Remove or move the "Endpoints" line in the Description of a deployment.  The new tabs like Ingresses and Services are super useful, and something like this should probably be found under one of these tabs.  The admin reports that the HTTPS link created when you deploy and ingress is particularly not useful to users since the https endpoint doesn't actually map to their ingress, but to the node IP that their users don't have access to. They said that it would be better to just have the different links in the tabs associated with the appropriate ingress or service.

Right now the Services tab has links to the service endpoint in the Target column but the Ingress has a link to the service dashboard instead of the endpoints created by the ingress. 

Here is their initial report verbatim (the request evolved to moving the endpoints to tabs during the call but this provides some context):

Steps to reproduce:
1. Create a workload running a webserver
2. Add an Ingress to the workload
3. Observe in the Deployment details that there is a "443/HTTPS" link in the "Endpoints" part of the deployment details.
4. Click the "443/HTTPS" link

Expected results:
The browser opens a new tab to `https://ingress-hostname/`

Actual results:
The browser opens a new tab to `https://kubernetes-node-ip/`

Support suspects that the "Endpoints" is supposed to refer to the Service(s) associated with the Deployment. But the URL constructed for the "443/HTTPS" link doesn't make any sense in that context either, as the only Service in this case is a ClusterIP referenced in the Ingress, which would have a cluster-internal IP, not the IP of one of the cluster nodes.

Theory: there is broken code where the "443/HTTPS" link is being constructed, where it's mistakenly using the Ingress controller's IP (which is a DaemonSet using hostPort, so any of the cluster's nodes would show up as IPs). But you can't access an application via Ingress using an IP address, you have to provide a hostname.

In all, the current behavior is very confusing for users, who see a clickable "443/HTTPS" link in their Deployment and think that this is where they can go to access their app, then they get frustrated when they get a 404 from the ingress controller (because they didn't specify a hostname).

It's notable that in the "Ingresses" tab in the Deployment detail view, the "Target" URL is correctly clickable to the application's hostname. The "443/HTTPS" link under "Endpoints" should use that URL, not `https://ip-address`, unless there's no Ingress associated.

Actual behavior:
Endpoints list exists in the Description of a deployment

Expected behavior:
Endpoints should be listed in a tab appropriate to what they link to

aalves08:
### Summary
Fixes #8087 

- update logic of `WorkloadDetailEndpoints` to consider hostname as a valid link if there is an `ingress` associated
- formatter code cleanup (unused computed property)
- add unit tests

- Install repository as `git repo` -> `https://github.com/bashofmann/rodeo` - branch `master` in `apps` - `repositories` on `local` cluster
- in `apps` - `chart` on `local` cluster, find a card for `tetris` app that comes from the repo added on the step before and install it
- on `local` cluster, `workloads` - `deployments`, find the deployment associated with the deployment of `tetris`
- On the details in the top of the page, check that the `endpoints` link points to the hostname `tetris.kube-public.172.18.0.3` as per the changes on this PR 

### Screenshot/Video

**Before**
 
before

**After**

after

### DIFF
--- a/shell/components/formatter/WorkloadDetailEndpoints.vue
+++ b/shell/components/formatter/WorkloadDetailEndpoints.vue
@@ -35,7 +35,7 @@ export default {
               protocol = 'https';
             }
 
-            const linkDefaultDisplay = endpoint.port ? `${ endpoint.port }${ endpoint.protocol }` : endpoint.protocol;
+            const linkDefaultDisplay = endpoint.port ? `${ endpoint.port }/${ endpoint.protocol }` : endpoint.protocol;
 
             // If there's an ingress and it has a hostname, we use the hostname address instead
             // https://github.com/rancher/dashboard/issues/8087
--- a/shell/components/formatter/WorkloadDetailEndpoints.vue
+++ b/shell/components/formatter/WorkloadDetailEndpoints.vue
@@ -27,6 +27,7 @@ export default {
 
         try {
           out = JSON.parse(this.value);
+
           out.forEach((endpoint) => {
             let protocol = 'http';
 
@@ -34,7 +35,11 @@ export default {
               protocol = 'https';
             }
 
-            if (endpoint.addresses) {
+            // If there's an ingress and it has a hostname, we use the hostname address instead
+            // https://github.com/rancher/dashboard/issues/8087
+            if (endpoint.ingressName && endpoint.hostname) {
+              endpoint.link = `${ protocol }://${ endpoint.hostname }${ endpoint.path }`;
+            } else if (endpoint.addresses && endpoint.addresses.length) {
               endpoint.link = `${ protocol }://${ endpoint.addresses[0] }:${ endpoint.port }`;
             } else if (externalIp) {
               endpoint.link = `${ protocol }://${ externalIp }:${ endpoint.port }`;
@@ -46,26 +51,6 @@ export default {
           return out;
         } catch (err) {
           return this.value[0];
-        }
-      }
-
-      return null;
-    },
-
-    protocol() {
-      const { parsed } = this;
-
-      if ( parsed) {
-        if (this.parsed[0].protocol) {
-          return this.parsed[0].protocol;
-        }
-
-        const match = parsed.match(/^([^:]+):\/\//);
-
-        if ( match ) {
-          return match[1];
-        } else {
-          return 'link';
         }
       }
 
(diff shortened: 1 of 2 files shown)


shell/components/formatter/WorkloadDetailEndpoints.vue
```diff
@@ -90,7 +75,7 @@ export default {
         :href="endpoint.link"
         target="_blank"
         rel="nofollow noopener noreferrer"
-      ><span v-if="endpoint.port">{{ endpoint.port }}/</span>{{ endpoint.protocol }}</a>
+      >{{ endpoint.link }}</a>
```
nwmac, line 78: This changes the display for the service case as well - do we want to do this?

I'd suggest leaving it as is - can we add a property to the endpoint that is the what should be displayed for the link and set that in the loop - the services can use the original from the template.

nwmac (changes requested):

shell/components/formatter/WorkloadDetailEndpoints.vue
aalves08: I think I interpreted this correctly @nwmac 🤞 https://github.com/rancher/dashboard/pull/10380/commits/9defbd34b1f1863fa746dc12f81079d2d07321ac

PR is ready for re-review

shell/components/formatter/WorkloadDetailEndpoints.vue
```diff
@@ -27,17 +27,27 @@ export default {
...
+
           out.forEach((endpoint) => {
             let protocol = 'http';
 
             if (endpoint.port === 443) {
               protocol = 'https';
             }
 
-            if (endpoint.addresses) {
+            const linkDefaultDisplay = endpoint.port ? `${ endpoint.port }${ endpoint.protocol }` : endpoint.protocol;
```
nwmac, line 38: Missing a slash between the port and protocol

nwmac (changes requested):
Minor change

shell/components/formatter/WorkloadDetailEndpoints.vue
aalves08: Thanks for catching that 🙏

PR #10385, opened 2024-02-02
add option of 'none' for Container Network select in Basics tab of RKE2 provisioning

### LINKED ISSUE #10338: RKE2 Cluster Provisioning doesn't allow 'none' for container network
Internal Reference: SURE-7482

RKE2 allows users to deploy custom cni to replace the default canal that comes with it.

This can be set in the RKE2 config file with "cni: none".

While provisioning RKE2 downstream cluster via Rancher UI, there is no option to select none in the container network field.

You have to click edit yaml and change it manually.

Need to add the option 'none' to the list for RKE2.

aalves08:
### Summary
Fixes #10338 

- add option of 'none' for Container Network select in Basics tab of RKE2 provisioning
- add e2e test to check that option is available and can be selected

- Under `Cluster Management`, create a `RKE2 custom` cluster
- On the `Basics` tab, `Container Network` select, select the option `none`
- Check that on the network request payload has been sent with `none` and check YAML after creation to confirm as well

`spec.rkeConfig.machineGlobalConfig.cni`

### Screenshot/Video

**Confirmation of data after cluster creation**
Screenshot 2024-02-02 at 13 30 52

Screenshot 2024-02-14 at 17 48 16

### DIFF
--- a/shell/edit/provisioning.cattle.io.cluster/rke2.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/rke2.vue
@@ -330,6 +330,26 @@ export default {
 
       const out = findBy(this.versionOptions, 'value', str);
 
+      // Adding the option 'none' to Container Network select (used it Basics component)
+      // https://github.com/rancher/dashboard/issues/10338
+      if (!out.serverArgs) {
+        out.serverArgs = {};
+      }
+
+      if (!out.serverArgs.cni) {
+        out.serverArgs.cni = {};
+      }
+
+      if (!out.serverArgs.cni?.options) {
+        out.serverArgs.cni.options = [];
+      }
+
+      // there's an update loop on refresh that might include 'none'
+      // multiple times... Prevent that
+      if (!out.serverArgs.cni.options.includes('none')) {
+        out.serverArgs.cni.options.push('none');
+      }
+
       return out;
     },
 
@@ -1405,7 +1425,9 @@ export default {
       for ( const chartName of this.addonNames ) {
         const entry = this.chartVersions[chartName];
 
-        if ( this.versionInfo[chartName] ) {
+        // prevent fetching of addon config for 'none' CNI option
+        // https://github.com/rancher/dashboard/issues/10338
+        if ( this.versionInfo[chartName] || chartName.includes('none')) {
           continue;
         }
 
--- a/shell/edit/provisioning.cattle.io.cluster/rke2.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/rke2.vue
@@ -334,7 +334,7 @@ export default {
       // https://github.com/rancher/dashboard/issues/10338
       // there's an update loop on refresh that might include 'none'
       // multiple times... Prevent that
-      if (out.serverArgs.cni?.options && !out.serverArgs.cni.options.includes('none')) {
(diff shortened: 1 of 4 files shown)


eva-vashkevich:
lgtm

eva-vashkevich:
The cluster I created never became available, but so did cluster with regular configurtion, so I don't think it's because of this change

aalves08:
@eva-vashkevich I've changed this PR with the correct logic to add the option `none`. Turns out the failing unit test exposed a problem that I didn't foresee.
The logic couldn't be applied to the `Basics` component directly because it's a prop that is passed from the `rke2` component down to the `Basics` component.

Now everything is correct, I believe. Could you re-review this please? 🙏

shell/edit/provisioning.cattle.io.cluster/rke2.vue
```diff
@@ -331,6 +331,26 @@ export default {
...
+      if (!out.serverArgs) {
+        out.serverArgs = {};
+      }
+
+      if (!out.serverArgs.cni) {
+        out.serverArgs.cni = {};
+      }
+
+      if (!out.serverArgs.cni?.options) {
+        out.serverArgs.cni.options = [];
```
eva-vashkevich, line 345: I think this will make showCni on L527 to be always true? I think we should either only set none if cni options already exist or remove showCni and its dependencies.

shell/edit/provisioning.cattle.io.cluster/rke2.vue
```diff
@@ -331,6 +331,26 @@ export default {
 
       const out = findBy(this.versionOptions, 'value', str);
 
+      // Adding the option 'none' to Container Network select (used it Basics component)
```
eva-vashkevich, line 334: typo :)

eva-vashkevich (changes requested):
Sorry, I should have caught the problem you had in my last review. I found one issue with this change now.

shell/edit/provisioning.cattle.io.cluster/rke2.vue
aalves08: Yep, I think you are right. Thanks for catching this!

shell/edit/provisioning.cattle.io.cluster/rke2.vue
```diff
@@ -331,6 +331,14 @@ export default {
 
       const out = findBy(this.versionOptions, 'value', str);
 
+      // Adding the option 'none' to Container Network select (used in Basics component)
+      // https://github.com/rancher/dashboard/issues/10338
+      // there's an update loop on refresh that might include 'none'
+      // multiple times... Prevent that
+      if (out.serverArgs.cni?.options && !out.serverArgs.cni.options.includes('none')) {
```
eva-vashkevich, line 338: I'd probably do 
if (out?.serverArgs?.cni?.options && !out.serverArgs.cni.options.includes('none')) {
just in case

eva-vashkevich (commented):
Did creating an instance with 'none' setting work for you? It seems stuck in the updating state for me

aalves08:
> Did creating an instance with 'none' setting work for you? It seems stuck in the updating state for me

It's not supposed to provision because it needs additional configuration. The aim of the `none` option is to allow for custom network config.

eva-vashkevich (approved):
lgtm

Issue #7023, opened 2022-09-28
CloudCredential is not created when using a custom Docker Machine Driver.


jclab-joseph:
**Setup**
- Rancher version:  2.6.8
- Browser type & version: Chrome

**Describe the bug**
CloudCredential is not created when using a custom Docker Machine Driver.

**To Reproduce**

1. Install ProxmoxVE Machine Driver : https://github.com/lnxbil/docker-machine-driver-proxmox-ve
2. Create Cloud Credential
3. But not created.

**Result**

**Expected Result**

**Screenshots**
![image](https://user-images.githubusercontent.com/88625531/192664778-0af7830e-c75b-48c1-acc0-4256c33ac99b.png)
![image](https://user-images.githubusercontent.com/88625531/192664857-c501b668-5ec8-4a6d-a238-8abb5e5e92a9.png)
![image](https://user-images.githubusercontent.com/88625531/192664811-7a680308-2b05-4c96-9e88-f08c36540659.png)
![image](https://user-images.githubusercontent.com/88625531/192664874-2a56d491-a5fb-448e-abb0-d85b8866b111.png)

It returns 201 as response to POST, but Response Body is `null`.

**Additional context**

- There are no logs related to the rancher pod.

gaktive:
Pushing to Q4 but we need to understand how much of this is a Rancher issue. Research required.

nwmac:
The issue is that the proxmox node driver does not support a cloud credential (the `spec.addCloudCredential` property on the node driver is false).

We use this to prevent a user from creating a cloud credential for that driver type, but in the cluster creation process, we do not - we assume all cluster providers need credentials, so we show the UI for creating a credential as part of the flow. the backend API does return 201 (this is a bug and it should return an error status code) - but the credential does not get created.

We need to update the RKE2 cluster creation flow to skip credential creation based on the same criteria it uses to determine that you can/can't create one via the Cloud Credentials page.

izaac:
I reproduced the issue on Rancher `v2.7.9`
Creating a new cluster using the machine driver the UI shows the Create Credential form.


On Rancher `v2.9.0` this isn't reproducible, the cluster create workflow goes straight to filling the cluster resources details. Without errors on the Network logs.

PR #10403, opened 2024-02-06
Make rke2 clusters respect the machingConfig `addCloudCredential` property set to false

codyrancher:
### Summary
This ports RKE1 behavior over to RKE2 where we bypass the credential prompt when using a node driver with the `addCloudCredential` property set to false

This is achieved by having the cluster creation page pass the nodeDriver config object down to the rke2 component so the rke2 component can inspect the the property.

https://github.com/rancher/dashboard/issues/7023

### Technical notes summary
I tested this using the `Nutanix` driver instead of the `Proxmoxve` since would get stuck in a `Downloading` state for me.

Drivers with and without the `addCloudCredential` should be tested. 

- I know Nutanix is false while Digitalocean is true.


momesgin (approved):
LGTM

PR #10415, opened 2024-02-06
[Automation] - Use the grepTags as environment variable

izaac:
### Summary
Fixes: An small fix to use an environment variable instead of an `--env` flag for `grepTags`. Using `--env` was unstable when using tags separated by commas.

Jenkins pipeline

N/A


yonasberhe23 (approved):
lgtm

PR #10433, opened 2024-02-09
Fix SelectOrCreateAuthSecret not re-rendering properly after removing

### LINKED ISSUE #10412: RKE2 Registries - When multiple registry authentication hosts exist and one is removed, an incorrect secret name is rendered
**Setup**
- Rancher version: 2.9-head

**Describe the bug**

When you have multiple registry authentication hosts if you remove one of them the other one doesn't show its original secret name anymore. This is just a rendering issue since after save the payload still has the original secret name.

**Screenshots**

https://github.com/rancher/dashboard/assets/135728925/2eee1e79-98ff-45ce-90b4-c92ca87735e7

momesgin:
### Summary
Fixes #10412 

### Technical notes summary

Vue was not able to re-render items properly after removing an item because of not providing a unique key.

### Screenshot/Video

https://github.com/rancher/dashboard/assets/135728925/32767e9a-0f19-420a-b1de-f40d07e833c4

### DIFF
--- a/shell/edit/provisioning.cattle.io.cluster/tabs/registries/RegistryConfigs.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/tabs/registries/RegistryConfigs.vue
@@ -149,7 +149,7 @@ export default {
             />
 
             <SelectOrCreateAuthSecret
-              :key="row.value.hostname"
+              :key="`${row.value.hostname}-${row.value.authConfigSecretName}`"
               v-model="row.value.authConfigSecretName"
               :register-before-hook="wrapRegisterBeforeHook"
               :append-unique-id-to-hook="true"
--- a/shell/edit/provisioning.cattle.io.cluster/tabs/registries/RegistryConfigs.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/tabs/registries/RegistryConfigs.vue
@@ -148,6 +148,7 @@ export default {
             />
 
             <SelectOrCreateAuthSecret
+              :key="row.value.hostname"
               v-model="row.value.authConfigSecretName"
               :register-before-hook="wrapRegisterBeforeHook"
               :append-unique-id-to-hook="true"


shell/edit/provisioning.cattle.io.cluster/tabs/registries/RegistryConfigs.vue
```diff
@@ -148,6 +148,7 @@ export default {
             />
 
             <SelectOrCreateAuthSecret
+              :key="row.value.hostname"
```
rak-phillip, line 151: Is `hostname` really unique? I seem to be able to reproduce the behavior described in the issue if I have a form with two `a.registry.com` entries.

momesgin: Ideally UI should run some validation against having unique registry host names, currently UI allows to have multiple registries with the same host name when you're editing, but once you save it, it merges them into one.

momesgin: I made it more unique by combining the host name with secret name, but I'll create a separate issue later for adding validation after discussing it with the team

rak-phillip: Thanks! I think this is better and the situation should improve with better validation down the line.

rak-phillip (approved):
This change looks good. 

### Not to be addressed in this PR

I think there's a deeper design issue with ArrayList and ArrayListGrouped that's causing this behavior. I would expect a `key` to be more closely associated with the `v-for` that it's describing. I do see this in the code, but it's not represented in the rendered DOM nodes

https://github.com/rancher/dashboard/blob/1dfd3c0c2c7a72631c34f662e0031f4546e7a805/shell/components/form/ArrayList.vue#L238-L243

I think that we need a minimal reproduction of this behavior so that we can write up a clear issue to describe the problem.

PR #10445, opened 2024-02-13
Fix race condition in loadCluster

### LINKED ISSUE #10444: Race condition with `loadCluster`
There is a race condition in the authenticated middleware related to calling `loadManagement` and `loadCluster` in parallel.

At present this does not cause an error in the UI, but with performance improvements to be made, it will.

The main issue is that in `loadCluster`, the function `waitForSchema` assumes that schemas have been loaded before it then checks for the specific schema being looked for. If it is not, we get a full page error that schemas have not been loaded:

image

This is not currently a problem because the code path on loadCluster takes long enough that schemas have been loaded. If this path is optimised, the error becomes apparent.

nwmac:
Fixes #10444

This PR fixes the above issue by:

- Changing  `waitForSchema` so that it does not error out if the schemas are not loaded, but continue to wait

In terms of test, I've indicated QA/None - PR reviewer should check that the UI load still operates correctly and that refreshing on a page within cluster explorer works correctly.

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -208,8 +208,6 @@ nav:
     resourceIdNotFound: Resource { resource } with id { fqid } not found, unable to display resource details
     reload: Reload
     separator: or
-  errors:
-    noMgmtClusterSchema: Management Cluster Schema is not available
 
 product:
   apps: Apps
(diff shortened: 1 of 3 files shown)


richard-cox (dismissed):

nwmac:
@richard-cox I decided to revert the change to add the new check and error - so now this is a small change that ensures that `waitForSchema` does not fail if the schemas have not been loaded (they are loaded asynchronously) - the existing code path will not fail if the cluster scheme can not be found - it will continue on. With my original change it would now fail with an error - I decided to revert this to reduce risk and keep the original behaviour.

richard-cox (approved):

PR #10494, opened 2024-02-22
Relay an error message when the backport fails

rak-phillip:
This is a follow-up to #10478

The goal of this PR is to relay an error message when the Port PR action fails to apply a patch. The current behavior isn't very helpful and requires digging through logs for actions when a failure occurs. For example

image
|


jordojordo (approved):
lgtm - Great idea with printing the log into markdown!

PR #10520, opened 2024-02-27
Remove `event` data from store when leaving Events tab

torchiaf:
### Summary
Fixes https://github.com/rancher/dashboard/issues/10265

The Events tab inside Cluster Dashboard page - `/c/local/explorer#cluster-events` - fetches and saves the events as a new `fetchedType` into the store.

We are removing the event type from the store before the Events tab is unmounted, so the events can be fetched again when navigating to Cluster/Events page.

### Technical notes summary

- The same result can be achieved by using a custom list component for the Cluster/Events page; it seems that, using a new component and calling the `$fetchType` after the component is rendered, it fetches again the event resource. This is likely caused by an async behavior inside `resource-fetch.js` mixin and needs further investigations.
- it is not possible to call `cluster/forgetType` directly in the `beforeDestroy` because it is async and it would be called after the new Events page is rendered.

- Cluster/Events page.

### Checklist
- [x] The PR has a Milestone 


aalves08 (approved):
LGTM

Issue #10510, opened 2024-02-26
EKS: Validate node group min/max/desired values


nwmac:
Internal Reference: SURE-5957

When creating an EKS cluster, it's possible to define an invalid amount of min/desired, this creates the cluster, but fails due with validation from the AWS API.{}

This will be done as part of the port of EKS to Vue.

cpinjani:
Validation for nodegroup ASG size is missing on build `v2.9-81337b95660cba868629d6fe769bd7e8242b5aee-head, Dashboard - master 3da1205`
Hence cluster provisioning is allowed with invalid values and it eventually fails:

![image](https://github.com/user-attachments/assets/0ca5d8e6-b69c-4500-81f2-20c33656d5c0)

thehejik:
**Test report**

Validated on rancher `v2.9-637346f76ea2bb71a82ea6e00a2779f61610a2a4-head`

The problem seems to be fixed for EKS provisioning.

**Before** the change on dashboard `v2.9.0-alpha7` there is no import field validation involved:
![pasted_image008](https://github.com/user-attachments/assets/9d103833-94de-43e2-aff3-09f19ba71d0e)

**After** the change on dashboard `master ddd2dee` the validation seems to be in place and it won't let user create a cluster with wrong values.

![pasted_image007](https://github.com/user-attachments/assets/1e8ed6c2-5aea-464f-a6b2-4f667b74bfa9)

![pasted_image006](https://github.com/user-attachments/assets/7189b4ce-6959-430e-b775-7992c1d20b29)

Issue #6840, opened 2022-09-08
Subnets selection is missing once a user clicks on any other operation on the page during EKS cluster creation


dasarinaidu:
**Setup**
- Rancher version: 2.6.8
- Browser type & version: Chrome/MacOS

**Describe the bug**
EKS->Create Cluster-> Networking->Subnet->Custom: Subnets selection is missing once user click on any other operation on the configuration page. 

**To Reproduce**
Steps:

1.	Create AWS Instance
2.	Provision Rancher
3.	Create a cluster from Rancher – On Network Configuration page 
4.	Select Custom and select a Subnet under a VPC
5.	Perform any other operation on the page (except click on NEXT)
Ex: Select/Deselect Private
6.	Check for the Subnets selection

**Result**
Actual:
1.	The selected subnets information is removed, user has to reselect the Subnets
2.	The Next button is enabled

**Expected Result**
1. Selected Subnets should have the focus till the user choose to change the options

**Screenshots**
On this screen if user click on any other option - the Subnet selection is loosing the selection

<img width="1486" alt="image" src="https://user-images.githubusercontent.com/22034884/189243862-ac6c2f7f-689e-4b50-ba53-74e980edb5b1.png">

gaktive:
This appears to be an ember issue that may have been around for a while. Need to dig in.

nwmac:
WIll check when we port EKS to Vue

nwmac:
@mantis-toboggan-md One to check when the port is done - moving to your next up

PR #10581, opened 2024-03-08
EKS Provisioning

### LINKED ISSUE #8966: Move EKS from Ember to Vue
In support of moving all public cloud provisioning off Ember (https://github.com/rancher/dashboard/issues/8083 and in a way, tied to work for finding all pages off Ember in https://github.com/rancher/dashboard/issues/8718), this ticket is to move the EKS pages to being in Vue. 

Currently, to provision EKS, it's iFramed from Ember into Vue.

Original embedding work: https://github.com/rancher/dashboard/issues/2749

### LINKED ISSUE #10416: AKS cluster page while editing cluster showing old UI
**Rancher version:**
`v2.9-b127983e2509c63090538cd1a92ce7ccf8fdabfb-head`

**Describe the bug**
AKS pages are moved to Vue in https://github.com/rancher/dashboard/issues/8965 and Import AKS option is addressed in https://github.com/rancher/dashboard/issues/10302
However the AKS page after editing the cluster, for provisioned/imported clusters still shows old UI (Ember)

**Steps**
- Go to Cluster management
- Click Import existing, Import and AKS cluster
- Edit the imported/provisioned AKS cluster

**Screenshots**
image

mantis-toboggan-md:
### Summary
Fixes #8966 
#6840 
#10510 
#7295 

Also fixes #10416 

I hit problems with KeyValue and LabeledSelect in this PR, which I'm now looking into: 
https://github.com/rancher/dashboard/issues/9668 
https://github.com/rancher/dashboard/issues/10587 
The LabeledSelect bug in particular is blocking this PR.

All EKS provisioning functionality from Ember should be replicated here. The only changes are some additional validation, and fewer hardcoded lists: version, region, and instance type are dynamically loaded in this form and hardcoded in the old EKS form.

### Technical notes summary
* it is expected that some users wont have permission to load k8s versions nor kms keys - the `describeAddonVersions` and `listKeys` aws api calls can fail silently
* regions can't be loaded until we have authetnicated with aws...using a region...so the credential dropdown still uses a hardcoded list of regions, but the eks form's dropdown doesn't
* the description field in Ember doesn't work
* `nodegroupName` is not auto-generated in the Ember UI - these ec2 resources have other tags identifying them in aws so I think it's ok for us to default to something relatively meaningless like `group1`

1. Creating new EKS Clusters
2. Editing existing EKS clusters
3. Editing EKS clusters before they've finished provisioning (confirm which fields are editable at which stage)
4. EKS Launch templates https://github.com/rancher/rancher/issues/30613#issue-773898047

Ensure the correct page still loads when creating/editing/importing other cluster types (aks, gke, rke1, rke2)

### Screenshot/Video
Screenshot 2024-03-08 at 8 44 28 AM
Screenshot 2024-03-08 at 8 44 51 AM
Screenshot 2024-03-08 at 8 45 18 AM

### Checklist
- [x] The PR has a Milestone 


pkg/eks/components/AccountAccess.vue
```diff
@@ -0,0 +1,149 @@
...
+    region: {
+      type:    String,
+      default: ''
+    }
+  },
+
+  async fetch() {
+    this.defaultRegions = await this.$store.dispatch('aws/defaultRegions');
+    if (this.defaultRegions.length && !this.region) {
+      (this.$emit('update-region', DEFAULT_REGION));
```
richard-cox, line 37: un-needed brackets

pkg/eks/components/AccountAccess.vue
```diff
@@ -0,0 +1,149 @@
...
+          }
+        }
+      },
+      immediate: true
+    }
+  },
+
+  methods: {
+    async fetchRegions() {
+      const { region, credential } = this as any;
```
richard-cox, line 66: to avoid as any, this can be `const { region, credential}: { region: string, credential: string} = this`

pkg/eks/components/CruEKS.vue
```diff
@@ -0,0 +1,766 @@
...
+    CREATE(): string {
+      return _CREATE;
+    },
+
+    VIEW(): string {
+      return _VIEW;
+    },
+
+    groupedInstanceTypes() {
+      const out = {} as any;
```
richard-cox, line 373: const out: { [group: string]: any } = {};

there's a few places where this same typing can apply (see references to `as any;`)

pkg/eks/components/CruEKS.vue
```diff
@@ -0,0 +1,766 @@
...
+        } else {
+          out[type.groupLabel] = [type];
+        }
+      });
+
+      return out;
+    },
+
+    instanceTypeOptions() {
+      const out = [] as any[];
```
richard-cox, line 387: const out: {label: string, value?: string, group?: string, kind?: string}[] = [];

pkg/eks/components/CruEKS.vue
```diff
@@ -0,0 +1,766 @@
...
+
+        out.push(groupOption);
+        out.push(...instanceTypeOptions);
+      });
+
+      return out;
+    },
+
+    spotInstanceTypeOptions() {
+      const out = [] as any[];
```
richard-cox, line 408: same as above, so probably should be an official type . guessing they relate to a component as well, so should live near there. 

const out: {label: string, value?: string, group?: string, kind?: string}[] = [];

pkg/eks/components/Networking.vue
```diff
@@ -0,0 +1,240 @@
...
+        this.$emit('update:subnets', []);
+      }
+    }
+
+  },
+
+  data() {
+    return {
+      loadingVpcs:  false,
+      vpcInfo:      {} as any,
```
richard-cox, line 85: There's quite a lot of these around (search for `as any,`). Some could be defined as 
```
      vpcInfo:     {} as {
        // [key: string]: any,
        Vpcs: [],
      },
```
The commented out part could be brought it, if it needs to be more generic, but ideally not

pkg/eks/components/CruEKS.vue
```diff
@@ -0,0 +1,766 @@
...
+      iamInfo:         {} as any,
+      ec2Roles:        [],
+      eksRoles:        [],
+      instanceTypes:   [],
+      launchTemplates: []
+    };
+  },
+
+  created() {
+    const registerAfterHook = this.registerAfterHook as Function;
```
richard-cox, line 204: const registerAfterHook: Function = this.registerAfterHook;

The `as` will disregards the type check of `this.registerAfterHook`. the function is in a mixin, but ts should be ok finding it.

pkg/eks/components/CruEKS.vue
```diff
@@ -0,0 +1,766 @@
...
+    />
+
+    <div
+      v-if="hasCredential"
+      data-testid="crueks-form"
+    >
+      <Tabbed
+        class="mb-20"
+        :side-tabs="true"
+        :show-tabs-add-remove="mode !== 'view'"
```
richard-cox, line 615: _VIEW

pkg/eks/components/NodeGroup.vue
```diff
@@ -0,0 +1,675 @@
...
+
+// map between fields in rancher eksConfig and amazon launch templates
+const launchTemplateFieldMapping = {
+  imageId:      'ImageId',
+  userData:     'UserData',
+  instanceType: 'InstanceType',
+  ec2SshKey:    '',
+  resourceTags: 'TagSpecifications',
+  diskSize:     'BlockDeviceMappings'
+} as {[key: string]: string};
```
richard-cox, line 26: Mentioned above, about other places as well, but can avoid the `as` here by just typing the const

const launchTemplateFieldMapping: {[key: string]: string}

pkg/eks/components/NodeGroup.vue
```diff
@@ -0,0 +1,675 @@
...
+
+--==MYBOUNDARY==
+Content-Type: text/x-shellscript; charset="us-ascii"
+
+#!/bin/bash
+echo "Running custom user data script"
+
+--==MYBOUNDARY==--\\`;
+
+export default Vue.extend({
```
richard-cox, line 40: Needs to be one of them new fangled `defineComponent`

richard-cox (changes requested):
I've given everything a brief view. As like the AKS PR i didn't go in to detail around config options, validating how they're actually set in aws, etc. That one's best left for someone QA side with lots of AWS knowledge.

Clusters come up as before and can be edited fine.

I spotted the following generally things

- ~With no creds (or multiple?) the create creds page shows the cluster name, description and region inputs at top~
- Create Amazon EKS - Main page
  - ~Top of page - There's a difference between AKS and EKS~
    - ~New EKS has row 1 - name, description. row 2 - region. row 3 - cloud credential~
    - ~New ALS has row 1 - cloud crendential. row 2 - name, description, region, kube version~
    - ~Could these align somehow?~
      - ~Would think that EKS looks best, but maybe move cloud credential on the same row as region (order tbc). atm the values look real lost given the large width of the controls~
      - ~AKS would become row - 1 name, description. row 2 - region, cloud credential, kube version~
      - If the above sounds good, i can create an issue for the aks change
  - Node Pools
    - ~Missing `Node Pools` header (this is in new AKS, though is pretty small there, should it be bigger?)~
    - ~Node Template Details~
      - ~Can the width of the `Node Template Details` input be expanded to display the default option without expanding the drop down?~
      - ~Applying a launch template populates some of the fields, but switching back to `Default` doesn't reset those back to their original 'default' values~
      - ~To confirm, in old world i see no `Launch Template` options. It's a bug fix we now show them in the new world?~
      - ~There's some nice info text below the old `AMI ID` option. If that's still relevant can it come across?~
      - ~User Data~
        - ~Useful text in old world lost in new~
        - ~We've lost the `Read from a file` option for User Data. Super low priority though.~
  - Networking
    - ~Public / Private Networking options were grouping in an `API Server Endpoint Access` section with a nice blurb. If that's still relevant can we keep? Maybe even docs link optional~
    - ~The endpoints also have a small blurb that's missing~
  - ~Labels and Annotations~
    - ~This is now an empty drop down (no way to add either type)~
- ~Edit Amazon EKS - Main page~
  - ~Changes from previous disabled/enabled state (are these fixes?)~
    - ~Credentials~
      - ~Can change the cloud credential~
    - ~Node Pools~
      - ~Unable to change the desired/min/max of an existing pool~
      - ~Can change the gpu enabled instance and request spot instances~
    - ~Labels and Annotations~
      - ~Some of these (both types) are now editable. From memory i think there's a hardcoded list that defines them~
- Generally
  - ~I created a cluster and assigned a standard user as a cluster member~
  - ~The user can then see the cluster. They can't edit... but going to the cluster detail's `Config` page gives a number of warning notifications (below). I _think_ these shouldn't happen (user can make requests via meta/proxy with cloud credential)? It might be a general problem that applies to other cluster providers as well, if so this can be a different issue to agree a pattern?~
    - ~Request is missing Authentication Token~
    - ~Error fetching regions: MissingParameter: The request must contain the parameter AWSAccessKeyId~
    - ~The request must contain the parameter AWSAccessKeyId~
  - After creating a new cluster it failed to come up with the error 
    - `Waiting for API to be available:error getting or creating launch template: error creating launch template: InvalidLaunchTemplateName.AlreadyExistsException: Launch template name already in use. status code: 400, request id: 35b6b40b-45c5-4436-a0fc-0771aab78ba5`
    - the launch template is set to `Default` 
    - i probably re-used the cluster name of `rc-rke` and node pool name of `rc-rke`
    - do we pass through an id for the launch template, is it based on the pool name? if so do we need to make these context unique?

mantis-toboggan-md:
>       * To confirm, in old world i see no `Launch Template` options. It's a bug fix we now show them in the new world?

What options do you see in the new world that you dont in the old world? Is this a new cluster or editing an existing cluster? Same region and credential selected in both UIs? I haven't seen discrepancies. 


> * With no creds (or multiple?) the create creds page shows the cluster name, description and region inputs at top
> * Create Amazon EKS - Main page
>   
>   * Top of page - There's a difference between AKS and EKS
>     
>     * New EKS has row 1 - name, description. row 2 - region. row 3 - cloud credential
>     * New ALS has row 1 - cloud crendential. row 2 - name, description, region, kube version
>     * Could these align somehow?
>       
>       * Would think that EKS looks best, but maybe move cloud credential on the same row as region (order tbc). atm the values look real lost given the large width of the controls
>       * AKS would become row - 1 name, description. row 2 - region, cloud credential, kube version
>       * If the above sounds good, i can create an issue for the aks change

I've updated the top area to show the region and credential dropdown on the same row, then give the credential component the whole width of the form when creating a new credential.



>   * Node Pools
>     
>     * Missing `Node Pools` header (this is in new AKS, though is pretty small there, should it be bigger?)
>     * Node Template Details
>       
>       * Can the width of the `Node Template Details` input be expanded to display the default option without expanding the drop down?

I've added the header here and increased the one in AKS from h4 to h3. Played around with input widths and added various info boxes/warnings

>       * Applying a launch template populates some of the fields, but switching back to `Default` doesn't reset those back to their original 'default' values

I updated the form so that if you clear your launch template selection or select a new version without a value for a particular field, the default value is re-applied. Also added a test for this

>       * There's some nice info text below the old `AMI ID` option. If that's still relevant can it come across?
>       * User Data
>         
>         * Useful text in old world lost in new
>         * We've lost the `Read from a file` option for User Data. Super low priority though.

Both added now.


>   * Networking
>     
>     * Public / Private Networking options were grouping in an `API Server Endpoint Access` section with a nice blurb. If that's still relevant can we keep? Maybe even docs link optional
>     * The endpoints also have a small blurb that's missing

I readded these as an info banner



> * Edit Amazon EKS - Main page
>   


>   * Changes from previous disabled/enabled state (are these fixes?)
>     
>     * Credentials
>       
>       * Can change the cloud credential
>     * Node Pools
>       
>       * Unable to change the desired/min/max of an existing pool
>       * Can change the gpu enabled instance and request spot instances

There was a bug here with detecting the active/provisioning state of the underlying eks cluster...these inputs should match up with the old ui now. I also updated the launch template dropdown to clarify when a nodegroup has a rancher-managed template as that affects which fields can be edited per https://github.com/rancher/rancher/issues/30613#issue-773898047


>     * Labels and Annotations
>       
>       * Some of these (both types) are now editable. From memory i think there's a hardcoded list that defines them

I see that we do taht in some places with steve resources, it looks like the list is a little different with these clusters though. I've grabbed the list of protected labels/annotations from the old UI and brought them over for this resource.

> * Generally
>   
>   * I created a cluster and assigned a standard user as a cluster member
>   * The user can then see the cluster. They can't edit... but going to the cluster detail's `Config` page gives a number of warning notifications (below). I _think_ these shouldn't happen (user can make requests via meta/proxy with cloud credential)? It might be a general problem that applies to other cluster providers as well, if so this can be a different issue to agree a pattern?
>     
>     * Request is missing Authentication Token
>     * Error fetching regions: MissingParameter: The request must contain the parameter AWSAccessKeyId
>     * The request must contain the parameter AWSAccessKeyId

This is a more general problem. I tried a few other providers and they all have poor ux: showing errors instead of displaying the form if the user doesn't have permission to access the cloud credential. I tried aks and rke2 on ec2, digital ocean, and azure they all throw errors when a standard user/cluster member tries to 'view config'. I updated the EKS config view to not run any of those aws api calls in 'view' mode because they're uneeded anyway. I filed an issue for the other providers: https://github.com/rancher/dashboard/issues/10734

pkg/eks/components/CruEKS.vue
```diff
@@ -0,0 +1,766 @@
+<script lang='ts'>
```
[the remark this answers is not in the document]
richard-cox: updated inline

pkg/eks/components/CruEKS.vue
```diff
@@ -0,0 +1,778 @@
...
+        const diff = diffUpstreamSpec(upstreamConfig, this.config);
+
+        this.$set(this.normanCluster, 'eksConfig', diff);
+      }
+    },
+
+    async actuallySave(): Promise<void> {
+      await this.normanCluster.save();
+
+      return await this.normanCluster.waitForCondition('InitialRolesPopulated');
```
richard-cox, line 474: this one kicks off a TS error. given where `waitForCondition` comes from i'm not sure we can avoid anything other than an `as` here

pkg/eks/components/AccountAccess.vue
```diff
@@ -0,0 +1,178 @@
...
+    async fetchRegions() {
+      const { region, credential }: { region: string, credential: string} = this;
+
+      if (!!region && !!credential) {
+        try {
+          const ec2Client = await this.$store.dispatch('aws/ec2', { region, cloudCredentialId: credential });
+
+          const res = await ec2Client.describeRegions({});
+
+          this.regions = (res?.Regions || []).map((r) => r.RegionName);
```
richard-cox, line 76: r needs typing

pkg/eks/components/AccountAccess.vue
```diff
@@ -0,0 +1,178 @@
...
+      &>.select-credential-container{
+      display:flex;
+      flex-direction: column;
+      flex-grow: 1;
+      }
+    }
+
+    &>.select-credential-container{
+      flex-basis: 50%;
+      // display:flex;
```
richard-cox, line 167: comments

richard-cox (changes requested):
I've updated the original review comment. i think two new, most resolved, a few left over

pkg/eks/components/CruEKS.vue
mantis-toboggan-md: I don't see this error. Different vscode configuration? I'm using the official vue extension as well as eslint and prettier, and set vscode to use the workspace's ts version (v4.5.5)
Screenshot 2024-04-08 at 10 08 58 AM

mantis-toboggan-md:
> With no creds (or multiple?) the create creds page shows the cluster name, description and region inputs at top

Updated this so they're hidden when there's no credential. I kept these inputs above the credential/region dropdown though, as I do think they look better there.

>The endpoints also have a small blurb that's missing

I rolled this and the other networking blurb into one info banner because it seemed to work best with the input component options we have in the new UI, and they do seem like conceptually connected messages.

>Labels and Annotations
>This is now an empty drop down (no way to add either type)

Fixed this and added a couple unit tests in the area

>After creating a new cluster it failed to come up with the error...

I think there may actually be a new backend issue around here?? I see this bug now, and did not in the past. I also see it with the old ui. I've used the same name in the past, including with the default launch template option (specifically I have created many all-default clusters in the old ui and new ui named `nb-defaults-old` and `nb-defaults`). This morning I also tried reusing a cluster name with unique node group name, and still saw that error.

Selecting the 'default' option in the UI removes any launch template configuration from the cluster spec so I think this is going to be purely a backend fix. I opened a backend issue here:https://github.com/rancher/rancher/issues/45044. I didn't label that as blocking the UI, not sure it qualifies?

@richard-cox thanks again for another review, sorry I missed a few requests last time, that was unintentional.

pkg/eks/components/CruEKS.vue
richard-cox: ah, it's a vetur error (i always struggled with the official extension). It's picking up that the return value for `waitForCondition` is `Promise<unknown>` which  contradicts the fn's `Promise<void>`.

image

can we update the return object to match?

richard-cox (changes requested):
RE https://github.com/rancher/rancher/issues/45044 - perfect, thanks

I've updated https://github.com/rancher/dashboard/pull/10581#pullrequestreview-1960462594. One for me, one for you.

One inline comment remaining.

pkg/eks/components/CruEKS.vue
mantis-toboggan-md: Frustratingly I couldn't get this error to show up even switcher over to vetur. I've typed the return value for `waitForCondition` now though

mantis-toboggan-md:
>user data guide text overlaps button and link is not clickable (not sure why)

That threw me for a moment. Turns out all label tags within labeledinput have `pointer-events: none;`. I've added an exception for the sublabel.

richard-cox (changes requested):
Was just about to approve but noticed a possible regression.

Node groups aren't editable, in contrast to the old behaviour where a lot of the details were. Did we decided this was a bug in the old ui (given templates)?  Everything else looks good

image

image

mantis-toboggan-md:
Nice catch Richard, that is a regression. Node group details should generally be editable (besides name and role): I updated the desired/min/max size inputs as well as labels and tags. I also  verified that the Ember UI _is_ handling those fields correctly, by confirming that updating them in Rancher will update the EKS cluster accordingly if done either before the underlying EKS cluster is created; after the cluster is created, but before the node groups are done creating; or after the node groups are running.

pkg/eks/components/CruEKS.vue
```diff
@@ -0,0 +1,781 @@
...
+          <NodeGroup
+            :rules="{
+              nodegroupName: fvGetAndReportPathRules('nodegroupNames'),
+              maxSize: fvGetAndReportPathRules('maxSize'),
+              minSize: fvGetAndReportPathRules('minSize'),
+              desiredSize: fvGetAndReportPathRules('desiredSize'),
+              instanceType: fvGetAndReportPathRules('instanceType'),
+              diskSize: fvGetAndReportPathRules('diskSize')
+            }"
+            :node-role.sync="node.nodeRole"
```
eva-vashkevich, line 647: Won't .sync complicate our Vue3 migration?
https://eslint.vuejs.org/rules/no-deprecated-v-bind-sync.html

PR #10592, opened 2024-03-10
automation: disable retries in cluster prov aws/azure tests

yonasberhe23:
### Summary
Fixes #
This change is needed for debugging purposes. Cluster provisioning tests (aws and azure) are failing in Jenkins nightly job but they are passing when running the tests locally which makes it difficult to debug this issue. Disabling the retries for these tests will allow us to capture the error thrown after the first attempt (not the 3rd) which will give us a better idea of what the issue is.


yonasberhe23:
No longer needed

PR #10611, opened 2024-03-12
Fix flaky registries test

### LINKED ISSUE #10609: Flaky test - registries.spec.ts
`registries.spec.ts` - `"Should send the correct payload to the server"`

image

momesgin:
### Summary
Fixes #10609 

Needed a little wait because of a brief debounce effect when typing in the LabeledInput component.


nwmac (changes requested):
@momesgin We try to avoid using `cy.wait` - can we not use `should` to wait for the control to be in the expected state?

yonasberhe23 (approved):
LGTM

nwmac (approved):

PR #10650, opened 2024-03-19
Improve LabeledSelect performance

### LINKED ISSUE #9668: LabeledSelect performs poorly with >20 item lists

mantis-toboggan-md:
### Summary
Fixes #9668 

This PR re-introduces https://github.com/rancher/dashboard/issues/8149 

This PR removes code added to make LabeledSelect option labels more dynamic because it causes a substantial performance hit. I have attempted a partial fix in the past; see https://github.com/rancher/dashboard/pull/9689 for details. Previously I was reluctant to re-introduce #8149 so I improved labeledselect performance only when the options provided are strings, not objects...this solution turns out to be inadequate, as we have some more complex dropdown options in EC2 and EKS

As this PR removes code added to fix #8133 I have added an alternative fix, supplying that LabeledSelect with a key which will force it to re-render if the option label changes. Though it was added for the psa dropdown, it is possible that other code is relying on the functionality in #8149 -- it has been in master for over a year now -- I did my best to look through the codebase but LabeledSelect is used in so many places it is virtually impossible to be certain I haven't broken something. (hence my reluctance to make this change when I looked at the issue previously)

RKE2 ec2 cluster instance type dropdown is a good example of a massive list

Ensure https://github.com/rancher/dashboard/pull/8133 has not been re-introduced

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/form/LabeledSelect.vue
+++ b/shell/components/form/LabeledSelect.vue
@@ -153,18 +153,6 @@ export default {
         return;
       }
 
-      // This check is only needed if its possible for an option's label to change without the option's value changing - we can skip this if options are just strings or numbers
-      // HOWEVER even if strings are passed to v-select the 'option' in the slot is normalized to {label: <option>} so we have to check the options prop here instead of the 'option' itself
-      if (typeof this.options[0] === 'object') {
-        const newOption = this.getUpdatedOption(option);
-
-        if (newOption) {
-          const label = get(newOption, this.optionLabel);
-
-          return this.localizedLabel ? this.$store.getters['i18n/t'](label) || label : label;
-        }
-      }
-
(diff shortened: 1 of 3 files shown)


cnotv (approved):
As agreed we prioritize performances till we found a different approach and possibly test.

LGTM

PR #10654, opened 2024-03-19
Factoring plugin installation out of index.js.

codyrancher:
One of a series of changes to simplify the bootstrapping and authentication code.

### DIFF
--- a/shell/initialize/plugins.js
+++ b/shell/initialize/plugins.js
@@ -1,0 +1,39 @@
+import '@shell/plugins/extend-router';
+import '@shell/plugins/formatters';
+import '@shell/plugins/global-formatters';
+import '@shell/plugins/i18n';
+import '@shell/plugins/js-yaml';
+import '@shell/plugins/portal-vue.js';
+import '@shell/plugins/resize';
+import '@shell/plugins/shortkey';
+import '@shell/plugins/tooltip';
+import '@shell/plugins/trim-whitespace';
+import '@shell/plugins/v-select';
+
+import axios from '../utils/axios.js';
+import axiosShell from '@shell/plugins/axios';
+import backButton from '@shell/plugins/back-button';
+import codeMirror from '@shell/plugins/codemirror-loader';
+import cookieUniversalNuxt from '../utils/cookie-universal-nuxt.js';
+import * as intNumber from '@shell/plugins/int-number';
+import nuxtClientInit from '@shell/plugins/nuxt-client-init';
+import plugin from '@shell/plugins/plugin';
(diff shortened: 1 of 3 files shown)


shell/initialize/plugins.js
```diff
@@ -0,0 +1,39 @@
...
+import '@shell/plugins/tooltip';
+import '@shell/plugins/trim-whitespace';
+import '@shell/plugins/v-select';
+
+import axios from '../utils/axios.js';
+import axiosShell from '@shell/plugins/axios';
+import backButton from '@shell/plugins/back-button';
+import codeMirror from '@shell/plugins/codemirror-loader';
+import cookieUniversalNuxt from '../utils/cookie-universal-nuxt.js';
+import intNumber from '@shell/plugins/int-number';
```
cnotv, line 18: Could you take the chance to replace the import as follows? During the migration I got complained that these files do not have an export.
```
import * as intNumber from '@shell/plugins/int-number';
import * as positiveIntNumber from '@shell/plugins/positive-int-number.js';
```

shell/initialize/plugins.js
```diff
@@ -0,0 +1,39 @@
...
+import positiveIntNumber from '@shell/plugins/positive-int-number.js';
+import replaceAll from '@shell/plugins/replaceall';
+import steveCreateWorker from '@shell/plugins/steve-create-worker';
+import version from '@shell/plugins/version';
+import emberCookie from '@shell/plugins/ember-cookie';
+
+export async function installPlugins(app, inject) {
+  const pluginDefinitions = [cookieUniversalNuxt, axios, plugins, pluginsLoader, axiosShell, intNumber, positiveIntNumber, nuxtClientInit, replaceAll, backButton, plugin, codeMirror, version, steveCreateWorker, emberCookie];
+
+  const installations = pluginDefinitions.map(async(pd) => {
```
cnotv, line 32: Could you add a comment of why this is needed, if you are aware of it?
Also could you please use meaningful attribute names, otherwise we go back to the initial point that we cannot read the code. Something like `pluginFn`?

cnotv (changes requested):
Thank you very much for the initiative <3
If would you not mind, could we please adjust these 2 parts?

shell/initialize/plugins.js
codyrancher: Done

shell/initialize/plugins.js
codyrancher: Done

cnotv (approved):
LGTM

Issue #5966, opened 2022-05-17
Review flaky E2E tests for side navigation


cnotv:
**Detailed Description**
Restore disabled E2E tests for navigation:
- https://github.com/rancher/dashboard/blob/master/cypress/integration/tests/navigation/page-actions.spec.ts
- https://github.com/rancher/dashboard/blob/master/cypress/integration/tests/navigation/side-nav/main-side-menu.spec.ts
- https://github.com/rancher/dashboard/blob/master/cypress/integration/tests/navigation/side-nav/mproduct-side-nav.spec

**Context**
With CI integration (#4181) we have been forced to disable navigation tests as they kept failing randomly.

Read **Disabled tests** details in the PR for further analysis https://github.com/rancher/dashboard/pull/5451

cnotv:
This seems related to an existing issue: https://github.com/rancher/dashboard/issues/4425

cnotv:
During this time, more tests have been disabled and complained with time about lack of timeout.
Specifically @n313893254 mentioned the fact that tests may pass by increasing the timeout to 6000.
We should identify why there's such a long delay and performance issue in loading the UI.

cnotv:
Some cases do fail because the action menu is actually not loaded, even if the rest of the UI is there.
E.g. [E2E test #271:](https://github.com/rancher/dashboard/actions/runs/2608954388)

![Side navigation Cluster -- Can close first menu groups on click -- before each hook (failed)](https://user-images.githubusercontent.com/5009481/177149339-54b8e637-c1f3-41b0-aee9-8830de1c428a.png)

cnotv:
I was able to emulate this error by delaying the response in the browser by setting `Throttling to Fast 3G`. It is however possible to delay responses in Cypress using the method `delay()` for intercepted responses.
We may consider some operation to be tested in the same conditions or eventually match the CI constrains for local development.

https://docs.cypress.io/api/commands/intercept#Throttle-or-delay-response-all-incoming-responses

cnotv:
To be solved with https://github.com/rancher/dashboard/issues/8418 or related automation.

PR #10670, opened 2024-03-20
[FIX] Fixed clicking on a group header outside of text not working

### LINKED ISSUE #10558: Can't click on 'Workloads' group header to go to Workloads
This is a bug specific to group headers in the side nav.

Go to the local cluster and click on Workloads.

Click on CronJobs.

Now try and click on Workloads - not by clicking on the text, but in the area around it (see red line below) - although the cursor shows a hand icon and although this works on the first click into Workloads, this does not work when the group is expanded - you have to click on the text.

Screenshot 2024-03-05 at 12 00 15

You should be able to click anywhere in the sider nav group header to go back to it.

eva-vashkevich:
### Summary
Fixes #10558

The link element had a padding around that wasn't clickable. Making the link the same size as the tab header fixed the problem.

I wasn't able to create a test that was failing before the change, so manual testing is required.
Some tests in cypress/e2e/tests/navigation/side-nav/product-side-nav.spec.ts were disabled pending #5966, which is now closed, so i enabled them back.

1. Go Local -> Workloads
2. Click on "Jobs" subgroup
3. Click back on "Workloads" below from the text
4. Workloads page should be opened

N/A

### Screenshot/Video

https://github.com/rancher/dashboard/assets/133266076/d0154e67-9d22-45f0-8929-06cd71f67d7c

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/nav/Group.vue
+++ b/shell/components/nav/Group.vue
@@ -293,7 +293,8 @@ export default {
 
     > A {
       display: block;
-      padding-left: 16px;
+      box-sizing:border-box;
+      height: 33px;
       &:hover{
         text-decoration: none;
       }
@@ -302,6 +303,7 @@ export default {
       }
       > H6 {
         text-transform: none;
+        padding: 8px 0px 8px 16px;
       }
     }
   }
@@ -313,6 +315,7 @@ export default {
         background-color: var(--primary-hover-bg);
 
         h6 {
+          padding: 8px 0px 8px 16px;
           font-weight: bold;
           color: var(--primary-hover-text);
         }
@@ -330,7 +333,6 @@ export default {
   .accordion {
     &.depth-0 {
       > .header {
-        padding: 8px 0;
 
         &.noHover {
           cursor: default;
@@ -339,6 +341,7 @@ export default {
         > H6 {
           text-transform: none;
           padding-left: 16px;
+          padding: 8px 0px 8px 16px;
         }
 
         > I {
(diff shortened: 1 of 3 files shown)


shell/components/nav/Group.vue
```diff
@@ -339,6 +341,7 @@ export default {
         > H6 {
           text-transform: none;
           padding-left: 16px;
+          padding: 8px 0px 8px 16px;
```
rak-phillip, line 344: It looks like `padding-left` is made redundant now.

```suggestion
          padding: 8px 0px 8px 16px;
```

shell/components/nav/Group.vue
```diff
@@ -293,7 +293,8 @@ export default {
 
     > A {
       display: block;
-      padding-left: 16px;
+      box-sizing:border-box;
+      height: 33px;
```
rak-phillip, line 297: Would we be able to specify a height of `100%` since our header already has a height of `33px`?

rak-phillip (commented):
Looks good and functions as advertised 🙂 

I just have a few questions before merging.

shell/components/nav/Group.vue
```diff
@@ -302,6 +303,7 @@ export default {
       }
       > H6 {
         text-transform: none;
+        padding: 8px 0px 8px 16px;
```
rak-phillip, line 306: A bit of a nitpick: the uom is redundant with a 0 value

```suggestion
        padding: 8px 0 8px 16px;
```

shell/components/nav/Group.vue
eva-vashkevich: That works too, you are right that it's better in case outer height changes

rak-phillip (approved):

PR #10672, opened 2024-03-20
Migrating some of the authenticated middleware into utilities and plugins

codyrancher:
This is a series of smaller PRs where the end goal is to get to get to https://github.com/rancher/dashboard/pull/10525 where we remove all nuxt middleware including authenticated.

### DIFF
--- a/cypress/e2e/tests/setup/rancher-setup.spec.ts
+++ b/cypress/e2e/tests/setup/rancher-setup.spec.ts
@@ -15,6 +15,7 @@ describe('Rancher setup', { tags: ['@adminUserSetup', '@standardUserSetup', '@se
 
     rancherSetupLoginPage.waitForPage();
     rancherSetupLoginPage.hasInfoMessage();
+    cy.url().should('contain', `${ Cypress.config().baseUrl }/auth/login`);
   });
 
   it('Confirm correct number of settings requests made', () => {
--- a/shell/middleware/authenticated.js
+++ b/shell/middleware/authenticated.js
@@ -1,9 +1,5 @@
-import { NAME as EXPLORER } from '@shell/config/product/explorer';
-import { SETUP, TIMED_OUT } from '@shell/config/query-params';
-import { SETTING } from '@shell/config/settings';
-import { MANAGEMENT, NORMAN, DEFAULT_WORKSPACE } from '@shell/config/types';
+import { DEFAULT_WORKSPACE } from '@shell/config/types';
 import { applyProducts } from '@shell/store/type-map';
-import { findBy } from '@shell/utils/array';
 import { ClusterNotFoundError, RedirectToError } from '@shell/utils/error';
 import { get } from '@shell/utils/object';
 import { setFavIcon, haveSetFavIcon } from '@shell/utils/favicon';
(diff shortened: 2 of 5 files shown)


cypress/e2e/tests/setup/rancher-setup.spec.ts
```diff
@@ -8,7 +8,7 @@ describe('Rancher setup', { tags: ['@adminUserSetup', '@standardUserSetup', '@se
     cy.visit('');
 
     new RancherSetupPagePo().hasInfoMessage();
-    cy.url().should('equal', `${ Cypress.config().baseUrl }/auth/login`);
```
codyrancher, line 11: This appeared to have a race-condition which was being exacerbated by these changes causing this test to be more flaky. Depending on when the url is checked it may contain the timed-out query param.

shell/middleware/authenticated.js
```diff
@@ -239,26 +84,17 @@ export default async function({
...
 
       next();
     });
 
     // Call it for the initial pageload
-    const redirected = setProduct(store, route, redirect);
-
-    if (redirected) {
-      return redirected();
+    if (await store.dispatch('setProduct', route)) {
```
codyrancher, line 95: We don't return a redirect method anymore but we still need to exit from this middleware method

cypress/e2e/tests/setup/rancher-setup.spec.ts
cnotv: I guess there are query paramters added or?

shell/middleware/authenticated.js
cnotv: In the other returns it's used `false`, should we keep the same?

shell/plugins/back-button.js
```diff
@@ -1,3 +1,12 @@
-window.addEventListener('popstate', () => {
-  window._popStateDetected = true;
-});
+export default function({ app }) {
+  window.addEventListener('popstate', () => {
+    window._popStateDetected = true;
+  });
+
+  app.router.afterEach((to, from) => {
```
cnotv, line 6: Could we have router-related logic inside the router configuration?
Sparse code is quite hard to track.

shell/plugins/back-button.js
```diff
@@ -1,3 +1,12 @@
...
+    // Clear state used to record if back button was used for navigation
+    setTimeout(() => {
```
cnotv, line 8: Should this be a Vue `nextTick()`?

shell/middleware/authenticated.js
```diff
@@ -1,210 +1,55 @@
...
-      return redirect({ name: 'auth-setup' });
-    }
+    await store.dispatch('auth/handleUserMustChangePassword');
 
     // In newer versions the API calls return the auth state instead of having to make a new call all the time.
     const fromHeader = store.getters['auth/fromHeader'];
 
     if ( fromHeader === 'none' ) {
-      noAuth();
+      store.dispatch('auth/noAuth');
```
cnotv, line 36: Could we turn this whole chain of conditions into a switch for clarity?

shell/store/auth.js
```diff
@@ -130,10 +130,94 @@ export const actions = {
...
   },
 
+  async loggedIn({ dispatch, commit }) {
+    const me = await dispatch('findMe');
+
+    commit('hasAuth', true);
+    commit('loggedInAs', me.id);
+  },
+
+  notLoggedIn({ commit }) {
```
cnotv, line 140: I do not think we should handle redirects in the store.
The idea of the state management and actions is to commit changes to the store. If nothing is happening there, then it's either a helper or a service (if it contains its own variables) and it does not need to pass from the store.
The very fact that we pass the router logic as single actions and not value to a router logic set somewhere else makes it very complicated to find out where things happen.

shell/store/auth.js
```diff
@@ -130,10 +130,94 @@ export const actions = {
...
+  // First thing we do in loadManagement is fetch principals anyway.... so don't ?me=true here
+    const principals = await dispatch('rancher/findAll', {
+      type: NORMAN.PRINCIPAL,
+      opt:  {
+        url:                  '/v3/principals',
+        redirectUnauthorized: false,
+      }
+    }, { root: true });
+
+    const me = findBy(principals, 'me', true);
```
cnotv, line 203: me? 😅

shell/store/index.js
```diff
@@ -809,6 +812,131 @@ export const actions = {
...
+
+      const res = getters['management/byId'](MANAGEMENT.SETTING, SETTING.FIRST_LOGIN);
+      const plSetting = getters['management/byId'](MANAGEMENT.SETTING, SETTING.PL);
+
+      isFirstLogin = res?.value === 'true';
+
+      if (!defaultPassword && plSetting?.value === 'Harvester') {
+        defaultPassword = 'admin';
+      }
+    } catch (e) {
```
cnotv, line 908: Why no handling?

shell/store/index.js
```diff
@@ -809,6 +812,131 @@ export const actions = {
...
+        const plSetting = await dispatch('rancher/find', {
+          type: 'setting',
+          id:   SETTING.PL,
+          opt:  { url: `/v3/settings/${ SETTING.PL }` }
+        });
+
+        if (!defaultPassword && plSetting?.value === 'Harvester') {
+          defaultPassword = 'admin';
+        }
+      } catch (e) {
```
cnotv, line 930: Also here, what should we do in this case if it fails?

cnotv (changes requested):
I like how we cleaned up everything, definitely easier to read.

Before everything, are we testing all these changes here? If not, we should as a bare minimum create issues for handling tests of this whole logic.

Also, could we split this function into clearer actions/features, for `authenticated.js`?

cypress/e2e/tests/setup/rancher-setup.spec.ts
codyrancher: Correct ?timed-out.

image

shell/store/auth.js
rak-phillip: I also share similar concerns about handling routing from the store. Unfortunately, this is already a pattern that is pervasive throughout dashboard. We explicitly expose the router to the store for this very purpose

https://github.com/rancher/dashboard/blob/d8c0cf69bf1be6229ab7e58ac4b49dc3ab07d55d/shell/initialize/index.js#L46-L47

My initial instinct is that functions that handle dispatching actions and redirects  (`setProduct`, `invalidResource`, `handleFirstLogin`, etc..) would better fit as utility functions, but I can't say that's a better abstraction than what's represented here..

@codyrancher have we considered alternatives to using the store for these functions?

rak-phillip (commented):
I don't have much to add outside of what's already been mentioned. Handling redirects from within the store isn't my preferred approach and I'd like to better understand what approaches we have available to us.

shell/middleware/authenticated.js
codyrancher: Moved to a switch

shell/store/auth.js
codyrancher: I moved it to a utility.

shell/utils/authentication.js
codyrancher, line 1: I recognize that there's an auth.js but it's doing authorization and not authentication. I think that one should probably be renamed in a separate pr.

shell/middleware/authenticated.js
codyrancher: I'm not sure what you're asking here. This should still be the same behavior given that setProduct doesn't return a redirect method.

shell/store/auth.js
cnotv: hahah cmon 😄 
I guess also this case we were just moving the code.

shell/plugins/back-button.js
cnotv: Good idea

cnotv (approved):
LGTM

codyrancher:
I decided I'm just going to abandon this PR. We'll be handling the middleware migration differently.

Issue #10675, opened 2024-03-21
[Harvester compatibility] Missing `vue-js-modal` package


torchiaf:
**Setup**
- Rancher version: v2.9-head

**Describe the bug**

Modal dialogs in Harvester embedded-mode require `vue-js-modal` package

**To Reproduce**

- Import a Harvester cluster in Rancher
- Explorer -> Virtualization Manager -> click on Harvester cluster

**Result**

Console error + other side effects

```
vue.js:5108 [Vue warn]: Unknown custom element: <modal> - did you register the component correctly? For recursive components, make sure to provide the "name" option.

found in

---> <ModalWithCard>
       <HarvesterUpgrade>
         <Anonymous>
           <ResourceList>
             <HarvesterResourcedList>
               <Shell/components/templates/default.vue> at shell/components/templates/default.vue
                 <Root>

```

**Screenshots**

**Additional context**

`vue-js-modal` removed by https://github.com/rancher/dashboard/pull/10591/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519

PR #10676, opened 2024-03-21
Revert vue-js-modal removal as required in Harvester

torchiaf:
### Summary
Fixes https://github.com/rancher/dashboard/issues/10675

### Technical notes summary

Modal dialogs in Harvester embedded-mode require vue-js-modal package.

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,6 +110,7 @@
     "url-parse": "1.5.10",
     "v-tooltip": "2.0.3",
     "vue-codemirror": "4.0.6",
+    "vue-js-modal": "1.3.35",
     "vue-resize": "0.4.5",
     "vue-router": "3.6.5",
     "vue-select": "3.18.3",
--- a/shell/initialize/plugins.js
+++ b/shell/initialize/plugins.js
@@ -2,6 +2,7 @@ import '@shell/plugins/extend-router';
 import '@shell/plugins/formatters';
 import '@shell/plugins/global-formatters';
 import '@shell/plugins/i18n';
+import '@shell/plugins/vue-js-modal';
 import '@shell/plugins/js-yaml';
 import '@shell/plugins/portal-vue.js';
 import '@shell/plugins/resize';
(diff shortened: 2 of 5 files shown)


rak-phillip (commented):
We need to drop this dependency so that we'll have a path forward. Do you think that we'll be able to migrate Harvester away from `vue-js-modal` instead of reintroducing it into dashboard?

torchiaf:
> We need to drop this dependency so that we'll have a path forward with the Vue3 migration. Do you think that we'll be able to migrate Harvester away from `vue-js-modal` instead of reintroducing it into dashboard?

@rak-phillip yes, there will be a porting of the modals migration on Harvester, but this will work for newest Harvester versions.

In the case the users updates their Rancher version to `v2.9-head`, having an older Harvester version, they will still get the error.
I think this is the only way we have to preserve the compatibility until they upgrade Harvester.

rak-phillip (dismissed):
@torchiaf Thanks for the additional context. I think it makes sense that we merge this, but we'll need to take extra care not to introduce new usages of `vue-js-modal` as we develop new features moving forward.

torchiaf:
@rak-phillip I pushed a new change to fix a build error: https://github.com/rancher/dashboard/actions/runs/8417966453/job/23047462028
Could you have another a look ?

rak-phillip (approved):
LGTM

For reference, this appears to be the build error that we encountered[^1] 

```
 $ ./node_modules/.bin/vue-cli-service build
Build: Production
Version: undefined
API: 'http://localhost:8989'. Env: 'web'

-  Building for production...
 ERROR  Failed to compile with 1 error9:41:48 AM

This dependency was not found:

* vue-js-modal/dist/ssr.index in ./node_modules/@rancher/shell/plugins/vue-js-modal.js

To install it, you can run: npm install --save vue-js-modal/dist/ssr.index
 ERROR  Build failed with errors.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
Error: Process completed with exit code 1.
```

Which is quite curious, as this is how we previously imported the module for the plugin.. However, we don't rely on ssr anymore, and the fix makes sense. 

[^1]: https://github.com/rancher/dashboard/actions/runs/8417966453/job/23047462028

PR #10689, opened 2024-03-22
Replace `createNamespace` pattern with a simple store for sharing data

rak-phillip:
### Summary
This replaces the pattern of `shell/components/form/NameNsDescription.vue` invoking its parent to emit the `createNamespace` event with a simple store. This store is only used to share data with `CruResource.vue` and is used to determine if a namespace should be created.  

### Technical notes summary

I opted to use a store in this case in an effort to write something that will be Vue3 compatible without the requirement of additional libraries. An alternative would be to utilize an event bus, but that would require the addition of a dependency like mitt[^1] to remain Vue3 compatible[^2].

[^1]: https://www.npmjs.com/package/mitt
[^2]: https://v3-migration.vuejs.org/breaking-changes/events-api.html#migration-strategy

I also experimented with dynamically registering[^3] this store only when it was required by `CruResource.vue`, but this proved to be problematic; I found that the `flush()` function in `shell/plugins/steve/subscribe.js` was throwing several unhandled exceptions with stores that were not registered:

```
Error: Missing module "cru-resource" for path "cru-resource[vdt]".

...

An error occurred in hook 'getInspectorState' registered by plugin 'org.vuejs.vue2-internal' with payload:

...
```

I gave up on this approach, as it probably wasn't a good pattern in the first place. Dynamic registration is probably best left for plugin code.

[^3]: https://v3.vuex.vuejs.org/guide/modules.html#dynamic-module-registration

This mostly changes the behavior of the components `NameNsDescription.vue` and `CruResource.vue`.

- When the Namespace dropdown includes a "Create a new namespace" option, it should create a new namespace when that option is selected
- When the Namespace dropdown should not include a new namespace when the "Create a new namespace" options is not selected
 
There are some instances of `NameNsDescription` that are not composed via `CruResource`. These cases should still function as they did before. Interacting with them should not create a new namespace. For reference:

- shell/edit/catalog.cattle.io.clusterrepo.vue
- shell/pages/c/_cluster/apps/charts/install.vue

### Screenshot/Video

image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/form/NameNsDescription.vue
+++ b/shell/components/form/NameNsDescription.vue
@@ -1,6 +1,6 @@
 <script>
 import Vue from 'vue';
-import { mapGetters } from 'vuex';
+import { mapGetters, mapActions } from 'vuex';
 import { get, set } from '@shell/utils/object';
 import { sortBy } from '@shell/utils/sort';
 import { NAMESPACE } from '@shell/config/types';
@@ -215,6 +215,7 @@ export default {
 
   computed: {
     ...mapGetters(['currentProduct', 'currentCluster', 'namespaces', 'allowedNamespaces']),
+    ...mapActions('cru-resource', ['setCreateNamespace']),
     namespaceReallyDisabled() {
       return (
         !!this.forceNamespace || this.namespaceDisabled || this.mode === _EDIT
@@ -381,12 +382,18 @@ export default {
     selectNamespace(e) {
       if (!e || e.value === '') { // The blank value in the dropdown is labeled "Create a New Namespace"
         this.createNamespace = true;
-        this.$parent.$emit('createNamespace', true);
+        this.$store.dispatch(
+          'cru-resource/setCreateNamespace',
+          true,
+        );
         this.$emit('isNamespaceNew', true);
         Vue.nextTick(() => this.$refs.namespace.focus());
       } else {
         this.createNamespace = false;
-        this.$parent.$emit('createNamespace', false);
+        this.$store.dispatch(
+          'cru-resource/setCreateNamespace',
+          false,
+        );
         this.$emit('isNamespaceNew', false);
       }
     },
--- a/shell/edit/__tests__/management.cattle.io.clusterroletemplatebinding.test.ts
+++ b/shell/edit/__tests__/management.cattle.io.clusterroletemplatebinding.test.ts
@@ -25,7 +25,18 @@ describe('view: management.cattle.io.clusterroletemplatebinding should', () => {
           'i18n/t':                  (val) => val,
           'i18n/exists':             jest.fn(),
         },
-        dispatch: jest.fn(),
+        dispatch: jest.fn((action, payload) => {
+          const actions = {
+            'management/findAll':              () => [],
+            'cru-resource/setCreateNamespace': jest.fn(),
+          };
+
+          if (actions[action]) {
+            return actions[action](payload);
+          }
+
+          throw new Error(`Unknown action: ${ action }`);
+        }),
       },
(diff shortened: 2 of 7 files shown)


shell/components/form/NameNsDescription.vue
```diff
@@ -381,12 +382,18 @@ export default {
     selectNamespace(e) {
       if (!e || e.value === '') { // The blank value in the dropdown is labeled "Create a New Namespace"
         this.createNamespace = true;
-        this.$parent.$emit('createNamespace', true);
+        this.$store.dispatch(
```
cnotv, line 385: This solution is great, although should we maybe use an index in case of multiple `CruResource`? I remember we have some cases with more of them together at the same time.

codyrancher: I don't think we really want to modify behavior (when avoidable) given that these changes are really only to support getting vue3 in.

codyrancher (approved):

rak-phillip:
@cnotv @codyrancher I'll need another review - I had to push updates to our unit tests because they were failing with the following error:

```
  ● view: management.cattle.io.setting should › using predefined generic rule › generate extra rules based on settings

    TypeError: this.$store.dispatch is not a function
```

shell/edit/__tests__/management.cattle.io.clusterroletemplatebinding.test.ts
```diff
@@ -25,7 +25,7 @@ describe('view: management.cattle.io.clusterroletemplatebinding should', () => {
           'i18n/t':                  (val) => val,
           'i18n/exists':             jest.fn(),
         },
-        dispatch: { 'management/findAll': () => ([]) }
+        dispatch: jest.fn(),
```
cnotv, line 28: I would rather mock single keys, because it helps to understand what are we doing and because it's easier to edit afterward.
I completely missed that we did also in the other test. This is quite crucial to make it easier for other components, like in this case. E.g. think about how much time you had to spent to figure it out 😬 

Replacement should be something like:

```ts
dispatch: {
  'management/findAll': () => ([]),
  'cru-resource/setCreateNamespace': jest.fn()
}
```

cnotv (changes requested):

shell/edit/__tests__/management.cattle.io.clusterroletemplatebinding.test.ts
rak-phillip: Hmm, it seems that the suggested approach still fails in the same manner:

```
  ● view: management.cattle.io.clusterroletemplatebinding should › should only show one error banner

    TypeError: this.$store.dispatch is not a function
```

I'm unsure what the difference is at this point that's causing the error, but it seemed pretty clear to me that `this.$store.dispatch` is lacking a proper mock. I tried updating this so that we could retain individual keys without failing the test, but I'm still not too satisfied with it. 

I tried looking to the vue test utils docs for guidance[^1], but we utilize a different pattern for working with Vuex than what they recommend. 

[^1]: https://v1.test-utils.vuejs.org/guides/using-with-vuex.html#mocking-actions

cnotv: We have already a case for the `dispatch()`, let me check it out.

cnotv: Ehm, sorry I got confused with the `getter()` 🤦

cnotv:
Thanks again for finding a solution and sorry again for the mistake 🙇

cnotv (approved):
LGTM

PR #10700, opened 2024-03-26
Add new cache states to determine plugin init

### LINKED ISSUE #10639: Installing an Extension Can Break if the Cache Is Disabled in the CRD
If the cache is disabled for any reason, it can look like an extension didn't install or is waiting to install in perpetuity.

## Background

Rancher Dashboard uses the `isCached` property to confirm that the backend has acknowledged the presence of a newly installed extension[^1]. This implementation mitigated a race condition where an extension could appear like it didn't install; when a CR is created and triggeres a refresh, it's possible that the backend hasn't yet received the webhook to inform it of the new extension.

Changes recently introduced in rancher/rancher#44230 have modified the caching behavior. If any file is larger than 20MB, the cache is disabled and removes any recently saved files. Furthermore, users can manually disable the cache by changing the CRD, potentially breaking the plugin install process in Dashboard.

## Proposed Solution

Implement a polling mechanism to deal with the race condition, checking until the backend confirms that the extension installs successfully. 

## Investigation Needed

It's still unknown how the changes to caching might affect airgapped installations. 

## Related Jira

- SURE-6314
- SURE-6315

[^1]: https://github.com/rancher/dashboard/blob/bb66802a2dc9bbd724e5c1b5f60de0dd4cf31bc8/shell/pages/c/_cluster/uiplugins/index.vue#L464-L471

rak-phillip:
### Summary
This expands the cases of using the cache state to determine if a plugin has been initialized or not; an initialized plugin can either be in a "cached" or "disabled" state.

Fixes #10639

### Technical notes summary

- This also removes the getters for `willBeCached()` and `isCached()` as they are unused by Dashboard with this change
- `noCache` can be manually set in the CRD so this isn't a good indicator for determining plugin initialization

- Managing plugins should remain unchanged
- Installing plugins should succeed when cache is disabled.

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/models/catalog.cattle.io.uiplugin.js
+++ b/shell/models/catalog.cattle.io.uiplugin.js
@@ -1,6 +1,10 @@
 import SteveModel from '@shell/plugins/steve/steve-class';
 
-const CACHED_STATUS = 'cached';
+const CACHE_STATE = Object.freeze({
+  CACHED:   'cached',
+  DISABLED: 'disabled',
+  PENDING:  'pending',
+});
 
 export default class UIPlugin extends SteveModel {
   get name() {
@@ -15,13 +19,8 @@ export default class UIPlugin extends SteveModel {
     return this.spec?.plugin?.version;
   }
 
-  get willBeCached() {
-    return this.spec?.plugin?.noCache === false;
-  }
-
-  // Has the plugin been cached?
-  get isCached() {
-    return !this.willBeCached || (this.willBeCached && this.status?.cacheState === CACHED_STATUS);
+  get isInitialized() {
+    return this.status?.cacheState !== CACHE_STATE.PENDING;
   }
 
   get pluginMetadata() {
--- a/shell/pages/c/_cluster/uiplugins/index.vue
+++ b/shell/pages/c/_cluster/uiplugins/index.vue
@@ -461,7 +461,7 @@ export default {
       neu.forEach((plugin) => {
         const existing = installed.find((p) => !p.removed && p.name === plugin.name && p.version === plugin.version);
 
-        if (!existing && plugin.isCached) {
+        if (!existing && plugin.isInitialized) {
           if (!this.uiErrors[plugin.name]) {
             changes++;
           }


shell/models/catalog.cattle.io.uiplugin.js
```diff
@@ -1,6 +1,10 @@
 import SteveModel from '@shell/plugins/steve/steve-class';
 
-const CACHED_STATUS = 'cached';
+const CACHE_STATE = Object.freeze({
+  CACHED:   'cached',
+  DISABLED: 'disabled',
+  PENDING:  'pending',
```
rak-phillip, line 6: Cache states are derived from:

https://github.com/rancher/rancher/blob/12a48d56eb2e577ac386134aca1c90aadacd5069/pkg/controllers/dashboard/plugin/fscache.go#L29-L31

aalves08 (commented):
If `pending` is the only state meaningful in terms of failure for installing a plugin, then this does look good.

What about the pooling mechanism and race conditions mentioned in https://github.com/rancher/dashboard/issues/10639? Do we need to deal with that? 

Also, don't know how feasible it is, but creating a test that cover the installation of an extension bigger than 20mb would be good. Wdyt?

rak-phillip:
> What about the pooling mechanism and race conditions mentioned in https://github.com/rancher/dashboard/issues/10639? Do we need to deal with that?

Our current implementation handles the race conditions, but it does so with the assumption that a plugin will always report itself as being cached. This change preserves our current implementation by removing the requirement that a plugin reports itself as cached, instead requiring that it exits a "pending" state.

If we were unable to preserve our current implementation, then I think that we would need to implement a new polling mechanism. I think this approach is much more straightforward 🙂

> Also, don't know how feasible it is, but creating a test that cover the installation of an extension bigger than 20mb would be good. Wdyt?

I am open to this. Do we have documentation on how we can publish an extension so that it can be accessed by our e2e tests?

aalves08:
> > What about the pooling mechanism and race conditions mentioned in #10639? Do we need to deal with that?
> 
> Our current implementation handles the race conditions, but it does so with the assumption that a plugin will always report itself as being cached. This change preserves our current implementation by removing the requirement that a plugin reports itself as cached, instead requiring that it exits a "pending" state.
> 
> If we were unable to preserve our current implementation, then I think that we would need to implement a new polling mechanism. I think this approach is much more straightforward 🙂
> 
> > Also, don't know how feasible it is, but creating a test that cover the installation of an extension bigger than 20mb would be good. Wdyt?
> 
> I am open to this. Do we have documentation on how we can publish an extension so that it can be accessed by our e2e tests?

On the e2e we already add the `ui-plugin-examples` repo and some e2e tests with an extension https://github.com/rancher/dashboard/blob/master/cypress/e2e/tests/pages/extensions/extensions.spec.ts#L23

We could create another one there with the conditions for that "20mb usecase", publish it on the repo and it would be automatically available to run an e2e about installing that extension.

https://github.com/rancher/dashboard/blob/master/cypress/e2e/tests/pages/extensions/extensions.spec.ts covers install, uninstall, upgrade, rollback, etc, so a lot of code is already covered in that front.

aalves08:
@rak-phillip Let me know if you want to move this forward, I can approve it and we can capture another issue to cover this scenario as an e2e. At least this will allow us to "unblock" extensions running in against a `-head`backend, if I am not mistaken

aalves08 (approved):
LGTM

Issue #9822, opened 2023-02-11
[RFE] More specific page titles


CC007:
**Is your feature request related to a problem? Please describe.**
Right now the page title for rancher is always Rancher. If you want to go back to a previous page in your browser history, this makes it hard to find the right page to go back to.

**Describe the solution you'd like**
It would be nice if the page title was more specific, like:
 - `Rancher - <cluster name> - Deployments`
 - `Rancher - <cluster name> - Deployment: phpmyadmin`
 - `Rancher - CD - <git repo>`
 - `Rancher - VM mgmt - <harvester cluster name> - Hosts`
 - `Rancher - VM mgmt - <harvester cluster name> - Image: ubuntu-22.04-minimal-cloudimg-amd64`
 - etc.

**Describe alternatives you've considered**
An alternative is that Rancher keeps track of a page history itself, but that seems clunky and the browser doesn't provide the functionality without a reason.

**Additional context**
![afbeelding](https://user-images.githubusercontent.com/5381337/218258807-5c52c79c-47cb-40ca-b901-50c45e489f04.png)

gaktive:
Internal reference: SURE-6121

CC007:
Thanks for the response, @gaktive . I'm glad to see not just a comment, but actual actions taken to triage the issue.

gaktive:
#9538 will be a prerequisite for this.

nwmac:
@codyrancher I've re-opened this - there is a small bug - go to Workloads, then go to a workload type, like Deployments and then go back to Workloads and the title is now incorrect - is says `Rancher - local - ?` - first time in, it is correct.

I had a tiny look and I see a watch in `TabTitle`, but I don't think Vue allows watching of slot content like that.

nwmac:
If you add `:key="_typeDisplay"` to the `TabTitle` in MastHead.vue, this resolves the issue, but this forces the TabTitle to be recreated/rendered when _typeDisplay changes - which gets around the watch - but not sure why this is not an issue when you move to a different resource type.

yonasberhe23:
The page titles have been updated to be more specific however when clicking and holding the back button to see browser history, I only see the specific page titles there for a few areas. Going to reopen this for further investigation



https://github.com/rancher/dashboard/assets/127343932/c7752f2f-05ac-409e-bdf7-328023a8c2c2

codyrancher:
[...]

PR #10702, opened 2024-03-26
Making sure tabTitle updates if the content of a slot updates

codyrancher:
Watches and computed props didn't update based on the content of slots.

The render method would still be called when the slot was updated so I moved the update to there. I also just use a method to compute the title because the value would get cached and not updated even if you accessed it more than once.

https://github.com/rancher/dashboard/issues/9822#issuecomment-1976814164

### DIFF
--- a/shell/components/TabTitle.vue
+++ b/shell/components/TabTitle.vue
@@ -27,10 +27,11 @@ export default {
       default: true
     }
   },
-  computed: {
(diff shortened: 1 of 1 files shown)


momesgin (approved):
Tested and LGTM, just wondering if this needs a milestone tag

codyrancher:
Added the milestone. Thanks for looking.

PR #10707, opened 2024-03-27
Tweaks to new cluster switch feature

### LINKED ISSUE #10706: Tweaks to new cluster switch feature
1. One concrete ask - can we change the text in the tooltip, from:

image

to `Switch clusters and keep location`.

i.e. remove the plural on `keeps`

2. On the mac, I noticed if you press SHIFT+ALT and then also hold down COMMAND+F5, the arrow overlay gets stuck:

image

I don't know if there are other cases where it can get stuck.

3. I find ALT+SHIFT awkward - I can't see a reason why we can't just use ALT? Can we change that @kwwii ??

4. lodash - I notice the code changes use `lodash/isObject` - generally we should avoid lodash and looks like it is not really needed in this case - can we remove?

aalves08:
### Summary
Fixes #10706 

- add missing translation to hardcoded string and adjust of copy 
- change key shortcut to `ALT` key only 
- remove lodash usage
- adjust e2e test

**Note:** Pressing `COMMAND + F5` opens a system-like panel in mac OS, which might lead to the "icon" getting stuck and visible. Pressing `ALT` multiple times seems to get it unstuck. We cannot prevent system key combinations from possible overrides and it's side-effects, imo. If there's additional info that might change this, let me know.

https://github.com/rancher/dashboard/assets/97888974/0cb12512-db5f-455d-a802-26003ae52e25

- Bring up a second cluster in ready state (meaning you can navigate to cluster explorer)
- In that cluster, go to `project/namespaces`
- press the key `alt`
- while pressing the key, make sure the icons appear next to the cluster boxes in the main nav, and click on the `local` cluster
- assert that the route loaded is the `project/namespaces` but in the `local` cluster

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/nav/TopLevelMenu.vue
+++ b/shell/components/nav/TopLevelMenu.vue
@@ -16,7 +16,6 @@ import { filterOnlyKubernetesClusters, filterHiddenLocalCluster } from '@shell/u
 import { getProductFromRoute } from '@shell/utils/router';
 import { isRancherPrime } from '@shell/config/version';
 import Pinned from '@shell/components/nav/Pinned';
-import isObject from 'lodash/isObject';
 
 export default {
   components: {
@@ -365,8 +364,12 @@ export default {
         content = this.shown ? null : contentText;
 
       // if key combo is pressed, then we update the tooltip as well
-      } else if (this.routeCombo && isObject(item) && item.ready) {
-        contentText = 'Switch clusters and keeps location';
+      } else if (this.routeCombo &&
+        typeof item === 'object' &&
+        !Array.isArray(item) &&
+        item !== null &&
+        item.ready) {
+        contentText = this.t('nav.keyComboTooltip');
 
         if (showWhenClosed) {
           content = !this.shown ? contentText : null;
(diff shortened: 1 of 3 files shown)


shell/components/nav/TopLevelMenu.vue
```diff
@@ -365,8 +364,12 @@ export default {
         content = this.shown ? null : contentText;
 
       // if key combo is pressed, then we update the tooltip as well
-      } else if (this.routeCombo && isObject(item) && item.ready) {
-        contentText = 'Switch clusters and keeps location';
+      } else if (this.routeCombo &&
+        typeof item === 'object' &&
+        !Array.isArray(item) &&
```
eva-vashkevich, line 369: why wouldn't array be a valid type here? it looks like lodash recognized them as objects https://lodash.com/docs/4.17.15#isObject

eva-vashkevich:
I've tried switching between clusters and it works as expected. I did notice the "stuck" problem but I am also unsure how we can work around it.

shell/components/nav/TopLevelMenu.vue
aalves08: It can be either a `string` or an `object` and logic changes based on that (different applications of the same function). 

Ah 😅 , I was missusing the `isObject` lodash function here! I though it ruled out array's... I understand why they allow it, but it it's kinda misleading in the sense that they are usually helper functions for stuff just like this 😛

aalves08:
> I've tried switching between clusters and it works as expected. I did notice the "stuck" problem but I am also unsure how we can work around it.

Me neither... It's "system" stuff. You'll eventually find something system related that will leave you in a weird state.

eva-vashkevich (approved):
lgtm

PR #10716, opened 2024-03-27
Replace `nuxt-link` with `router-link`

rak-phillip:
### Summary
This removes the `NuxtLink` & `NLink` components from Dashboard. If an extension or outside component attempts to use `NuxtLink` or `NLink`, a deprecation warning should print to the console and `router-link` will be used instead.

### Technical notes summary

I'm attempting to make this change in a way that will be more backwards compatible with extensions that might relay on `NuxtLink` or `NLink`. This is done by replacing the `NuxtLink` component with an alias that will print a deprecation warning in the console.

This affects all routing throughout dashboard. 

### Checklist
- [x] The PR has a Milestone 


rak-phillip:
@torchiaf I'm requesting a review from you so that you can test with Harvester. I want to avoid another situation where we might need to revert a portion of this PR like we did in #10676

codyrancher (approved):
As long as tests pass this LGTM.

codyrancher (approved):

cnotv (approved):
I think you covered everything, but please wait for @torchiaf about Harvester, otherwise this will get reverted anyway.

LGTM

torchiaf (approved):
Harvester router seems to be working correctly.
If possible, I would remove the warning message when loading Harvester.
LGTM

rak-phillip:
> If possible, I would remove the warning message when loading Harvester.

@torchiaf can you expand on why we would want to make an exception for Harvester? If we are flagging features as deprecated in v2.9 of Dashboard, would we not want to warn all extension authors that this will break in later versions after said features are removed?

torchiaf:
> > If possible, I would remove the warning message when loading Harvester.
> 
> @torchiaf can you expand on why we would want to make an exception for Harvester? If we are flagging features as deprecated in v2.9 of Dashboard, would we not want to warn all extension authors that this will break in later versions after said features are removed?

 In general, I think you are right, but in the Harvester's case, those warnings would be visible only when using Rancher + Harvester older versions embedded.
 We will have to port this code in the next Harvester version, including the warning logic - in theory the Harvester plugin would take care of showing the warnings then (even in standalone mode).

Anyway, this is really a non-issue I think, sorry, I was a little pedantic here :smiley:

Issue #10784, opened 2024-04-11
Modals: scroll artifact bugs


torchiaf:
**Setup**
- Rancher version: 2.9-head
- Rancher UI Extensions:
- Browser type & version: Chrome 123.0.6312.105

**Describe the bug**

On chrome scroll artifacts are shown -> scrollbar arrows in dark theme.

**To Reproduce**

- Go to Project creation form
- Click on Add members

**Screenshots**

![image (9)](https://github.com/rancher/dashboard/assets/26394656/942e97f6-1f43-48e8-b145-1057433b3fd5)

**Additional context**

PR #10785, opened 2024-04-11
Fix AddProjectMemberDialog, remove scroll artifacts

torchiaf:
### Summary
Fixes https://github.com/rancher/dashboard/issues/10784

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/pkg/rancher-components/src/components/Card/Card.vue
+++ b/pkg/rancher-components/src/components/Card/Card.vue
@@ -151,7 +151,7 @@ export default defineComponent({
 
       .card-body {
         justify-content: flex-start;
-        overflow: scroll;
+        overflow: auto;
       }
 
       > * {


shell/dialog/AddProjectMemberDialog.vue
rak-phillip, line 1: Would it make sense to change `pkg/rancher-components/src/components/Card/Card.vue` instead of `AddProjectMemberDialog`?

shell/dialog/AddProjectMemberDialog.vue
```diff
@@ -177,4 +177,7 @@ export default {
     border: 1px solid var(--border);
     box-shadow: none;
   }
+  .card-container.card-sticky .card-wrap .card-body {
+    overflow: hidden
```
rak-phillip, line 181: Would `auto` be more appropriate than `hidden`? I don't want to end up with a scenario where we would require scrolling, but are unable to do so.

shell/dialog/AddProjectMemberDialog.vue
torchiaf: Right, but I was wondering if it can introduce side effects in other components where we are using Card.vue

rak-phillip (approved):
LGTM

PR #10793, opened 2024-04-11
Updating <router-link> to no longer use the deprecated :tag property

codyrancher:
### Summary
Updating <router-link> to no longer use the deprecated :tag property

Navigation

### Screenshot/Video
https://github.com/rancher/dashboard/assets/55104481/d134267a-a3eb-48b6-a3cd-5d33382a3336

### Checklist
- [ ] The PR has a Milestone 


shell/components/TabbedLinks/index.vue
codyrancher, line 1: This file wasn't used anywhere in dashboard.

shell/components/nav/Type.vue
```diff
@@ -28,65 +28,10 @@ export default {
...
-    return {
-      near:     false,
-      over:     false,
-      menuPath: this.type.route ? this.$router.resolve(this.type.route)?.route?.path : undefined,
-    };
+    return { near: false };
   },
 
   computed: {
-    isCurrent() {
```
codyrancher, line 39: This wasn't used in the file.

shell/components/nav/Type.vue
```diff
@@ -28,65 +28,10 @@ export default {
   },
 
   data() {
-    return {
-      near:     false,
-      over:     false,
-      menuPath: this.type.route ? this.$router.resolve(this.type.route)?.route?.path : undefined,
```
codyrancher, line 34: These weren't used.

shell/components/nav/Type.vue
```diff
@@ -28,65 +28,10 @@ export default {
   },
 
   data() {
-    return {
-      near:     false,
```
codyrancher, line 32: `near` really should be mouseover it seems.

shell/components/nav/Type.vue
```diff
@@ -116,14 +61,6 @@ export default {
       this.near = val;
     },
 
-    setOver(val) {
-      this.over = val;
-    },
-
-    removeFavorite() {
-      this.$store.dispatch('type-map/removeFavorite', this.type.name);
-    },
```
codyrancher, line 125: Unused

shell/components/nav/Type.vue
```diff
@@ -142,63 +79,70 @@ export default {
...
     class="child nav-type"
+    data-testid="link-type"
   >
     <a
       :href="type.link"
       :target="type.target"
       rel="noopener noreferrer nofollow"
-      @click="selectType"
-      @mouseenter="setNear(true)"
-      @mouseleave="setNear(false)"
```
codyrancher, line 201: - `selectType` is specific to types in our app while this is being treated as a link to an external page based on the `rel=".."`.
- `setNear` is only actually used in relation to showing the Favorite component which isn't possible in this link.

shell/components/nav/__tests__/Type.test.ts
codyrancher, line 1: The original tests here were mostly exercising and testing router-link instead of our implementation of the `Type` component.

The new slot API makes these tests non-functional because the RouterLinkStub doesn't fully support the new API. As such I decided to rewrite these tests to achieve the following:

- Test the `Type` component instead of RouterLink
- Test more of the functionality than was tested before
- Switch to shallowMount instead of mount

More tests could still be written but I think this puts things in a better place than things were. I do recognize that rewriting broken tests is a bit of a code smell though.

shell/utils/unit-tests/ChildRenderingRouterLinkStub.ts
codyrancher, line 1: It's in the comments of this file but I'd like to further point out that this is a workaround because the vue/test-utils RouterLinkStub doesn't actually support the new slot api yet. I took the idea from https://github.com/vuejs/vue-test-utils/issues/1803#issuecomment-940884170.

shell/components/TabbedLinks/index.vue
cnotv: Are you sure it's not also in other plugins?

shell/components/nav/__tests__/Type.test.ts
```diff
@@ -1,167 +1,365 @@
...
+    currentStore:    () => 'cluster',
+    'cluster/count': () => defaultCount,
+  }
+};
+const routerMock = { resolve: () => ({ route: { path: 'whatever' } }) };
+const localVue = createLocalVue();
 
-// Configuration text
-const className = 'router-link-active';
+localVue.directive('cleanHtml', (identity) => identity);
```
cnotv, line 35: Is this not already mocked in the setup? With this extra?

shell/components/nav/__tests__/Type.test.ts
```diff
@@ -1,167 +1,365 @@
-import { mount, RouterLinkStub } from '@vue/test-utils';
+import { shallowMount, RouterLinkStub, createLocalVue } from '@vue/test-utils';
 import Type from '@shell/components/nav/Type.vue';
+import { createChildRenderingRouterLinkStub } from '@shell/utils/unit-tests/ChildRenderingRouterLinkStub';
+import { TYPE_MODES } from '@shell/store/type-map';
+
+// Configuration
```
cnotv, line 6: Could we avoid having data outside the tests? It is very time-consuming to have all this list and try to figure out which test it is.

This is usually a code smell pointing to a convoluted structure of the unit tests.
Specifically if in your test you are addressing some common points, they must be shared in the same `describe()`.

I usually avoid comments like this and aim to deliver tests, but this one really took me too long, I can see already a lot of headaches in the migration phase.

shell/components/nav/__tests__/Type.test.ts
```diff
@@ -1,167 +1,365 @@
...
 // Mandatory to mock vue-router in this test
 jest.mock('vue-router');
+const storeMock = {
+  getters: {
+    currentStore:    () => 'cluster',
+    'cluster/count': () => defaultCount,
+  }
+};
+const routerMock = { resolve: () => ({ route: { path: 'whatever' } }) };
+const localVue = createLocalVue();
```
cnotv, line 33: This is not needed.

shell/utils/unit-tests/ChildRenderingRouterLinkStub.ts
```diff
@@ -0,0 +1,36 @@
...
+ * See @link https://github.com/vuejs/vue-test-utils/issues/1803#issuecomment-940884170
+ *
+ * @param slotProps Provide arguments that you want passed to the child rendered by router-link
+ * @returns A stub
+ */
+export function createChildRenderingRouterLinkStub(slotProps?: RouterLinkSlotArgumentOptional): typeof RouterLinkStub | any {
+  return {
+    ...RouterLinkStub,
+    render() {
+      return this.$scopedSlots.default({
```
cnotv, line 27: Can we use normal slots instead of scoped?

shell/utils/unit-tests/ChildRenderingRouterLinkStub.ts
```diff
@@ -0,0 +1,36 @@
...
+export interface RouterLinkSlotArgumentOptional {
+    href?: string;
+    route?: Route;
+    navigate?: (e?: MouseEvent) => Promise<undefined | NavigationFailure>;
+    isActive?: boolean;
+    isExactActive?: boolean;
+}
+
+/**
+ * This is a workaround because VueUtils RouterLinkStub doesn't currently support the slot api.
```
cnotv, line 16: The aim of unit test is to separate the functionalities of the code into more parts, simplify the troubleshooting, and ensure to repeat the same tests.
By providing this utility it seems, please correct me if I am wrong, we are leading in the opposite direction.
Is there anything that needs to be inspected multiple times?

cnotv (changes requested):
I can see a lot of changes that are unrelated to the `: tag` property.
May I ask you what are the changes required due to that?

rak-phillip (dismissed):
Functionally, I think this looks good and the refactoring aligns with recommendations for addressing the deprecated `tag` prop[^1]

[^1]: https://router.vuejs.org/guide/migration/#Removal-of-event-and-tag-props-in-router-link-

shell/components/nav/__tests__/Type.test.ts
codyrancher: It looks like you're correct about routerMock. LocalVue is still needed per my other comment regarding cleanHtml.

shell/utils/unit-tests/ChildRenderingRouterLinkStub.ts
codyrancher: No, it's empty.

image

shell/components/nav/__tests__/Type.test.ts
codyrancher: I think what you're suggesting is that we narrow the scope of the shared data, I think that's fair. I've gone ahead and I've done that. 

I extended a pattern which initially defined state outside of the tests https://github.com/rancher/dashboard/commit/ece9a33f2aaf94406bfb9b42255671bf5123f42d#diff-8c1a0fb993ec56591f5c80663d5ccf84607dee19981b5bc6ff54064fd23676edR8 I could see how you would see that as different but in my mind the pattern is consistent.

> I usually avoid comments like this and aim to deliver tests, but this one really took me too long, I can see already a lot of headaches in the migration phase.

Sorry if I'm misinterpreting this but given the context of the rest of this comment I think you're being impolite here. If you're frustrated with me and you'd like to discuss something, we can. If you're thinking I'm incompetent and don't want me working on vue3 or some other aspect of the app please let me and our managers know.

shell/utils/unit-tests/ChildRenderingRouterLinkStub.ts
codyrancher: Here's my definition of tests and unit tests just for understanding:

- Tests are meant to codify application invariants to ensure that changes in behavior do not go unnoticed.
- Unit tests are supposed to test small units (often functions or components) in isolation of the rest of the application

Because the vue test util's RouterLinkStub doesn't currently support the slots API we needed a way to test the Type component in isolation (without router-link). We also have other components that now use the slots API which we may want to add unit tests for. That makes me think this will allow us to improve and add more unit tests.

Ideally RouterLinkStub will add support for the slots API and we remove this.

codyrancher:
> I can see a lot of changes that are unrelated to the : tag property.
> May I ask you what are the changes required due to that?

This questions seems odd to me because I don't think this change meanders very far from just the :tag property. But a summary is:

- For each file that used `:tag` we need to switch to the slot api
  - First follow https://eslint.vuejs.org/rules/no-deprecated-router-link-tag-prop
  - Second, those docs were out of date and I also had to add the custom attribute to each area
  - Those changes would have to happen to these files
    - shell/components/SideNav.vue
    - shell/components/TabbedLinks/index.vue
      - Do we want to make changes here even though the file isn't used anywhere? That seemed like wasted effort to me. 
    - shell/components/nav/Header.vue
    - shell/components/nav/Type.vue
- After making those changes we also need to update the Type.tests.ts because they broke.
    - The lack of support for slots in RouterLinkStub makes all of the original tests useless because ultimately it would come down to what slot props get passed. We could either delete those tests or rewrite them to exercise the Type component.

What's notably missing from that list is the dead code that I deleted from Type.vue. That wasn't necessary but since I was testing the component it seemed like a good time to delete dead code that didn't need to be exercised by a test.

codyrancher:
@cnotv I know @rak-phillip already approved this but I think we need to get through the feedback on this and hash things out otherwise going forward with vue 3 is going to be rough.

shell/components/nav/__tests__/Type.test.ts
cnotv: I will have to remove this on the migration anyway because [`createLocalVue` does not exist anymore](https://test-utils.vuejs.org/migration/#No-more-createLocalVue).
This should be rather mocked globally and tested in an isolated way.

Since I've seen that we have a dozen of tests like this, I think we can leave this.

shell/components/nav/__tests__/Type.test.ts
cnotv: No it's not about you. I have a personal issue about this which I am trying to address, so I apologize I have caused pressure on you and made you feel uncomfortable. I unfortunately sometimes do not realize it.
I hope you can understand and accept my apologies and also thanks for letting me know.
Just to be clear, it was not something aimed at you but at how I usually do reviews in general.

I will try to be more careful in the future.

shell/components/nav/__tests__/Type.test.ts
cnotv: You are right, it is actually missing in the [global Jest setup](https://github.com/rancher/dashboard/blob/master/jest.setup.js). I got confused with the branch I was working on. I will double check next time.

cnotv:
> @cnotv I know @rak-phillip already approved this but I think we need to get through the feedback on this and hash things out otherwise going forward with vue 3 is going to be rough.

I think it would be sufficient for one of us to approve the PR to proceed. Sometimes I may be overzealous or confused on some points, so I will try to be more careful from my side later on. If there are some other points which you were thinking about, let's have a chat about it, with the hope I can be helpful.

cnotv (approved):
LGTM

PR #10796, opened 2024-04-12
 Rancher UI branding customisation is not persistent on the login screen

### LINKED ISSUE #10788: Rancher UI branding customisation is not persistent on the login screen
Internal reference: SURE-8175

**Setup**
- Rancher version: repro'ed in Rancher 2.8.2 and 2.9-head
- Rancher UI Extensions: n/a
- Browser type & version: n/a

**Describe the bug**

The customer reports that before logging in, the login page does not use the primary and link colours. For example, they set both colours to orange, and the login page stays blue, as with the default parameters.

**To Reproduce**

- In The UI > Global Settings > Branding > Set "Primary Color" to Orange + set "Link Color" to Orange > Apply.
- Notice the change to Orange
   

image-2024-04-11-12-22-12-190 (2)

- Now "Logout" and notice the colour is still orange:

image-2024-04-11-12-23-04-242

- If you refresh the browser page, you'll notice that the colour has changed to the default:

image-2024-04-11-12-23-48-336

- If you log back in, the colour scheme is orange:
image-2024-04-11-12-24-34-644

**Result**
- Branding does not persist on logout/login screen

**Expected Result**

- Branding should be consistent and persistent

**Screenshots**

**Additional context**

aalves08:
### Summary
Fixes #10788 

- update `brand` mixin to handle login scenario for branding application of `primary` color and `link` color 
- add e2e tests to cover possible regression scenario

### Technical notes summary

Needed to convert the `globalSettings` variable in a getter-based in order to get the settings information from the network request done last to settings, which is done from the `login` view. Had to convert all the watchers to `immediate: true` in order to pick up changes

- This needs a backend version `2.9-head` with date 12th of April 2024 or later, due to changes done in `rancher/rancher` https://github.com/rancher/rancher/pull/45108
- Go to `Global Settings` -> `branding` and change `primary` color and `link` color to custom ones
- Log out of Rancher and check that the branding changes are applied
- Do a refresh on the login page to check that changes persist
- Log back into Rancher and check that general branding changes are applied

### Screenshot/Video

Screenshot 2024-04-12 at 16 10 03
Screenshot 2024-04-12 at 16 10 09
Screenshot 2024-04-12 at 16 10 17

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/cypress/e2e/po/pages/login-page.po.ts
+++ b/cypress/e2e/po/pages/login-page.po.ts
@@ -44,7 +44,7 @@ export class LoginPagePo extends PagePo {
     });
   }
 
-  private submitButton(): AsyncButtonPo {
+  submitButton(): AsyncButtonPo {
     return new AsyncButtonPo('[data-testid="login-submit"]', this.self());
   }
 
--- a/cypress/e2e/tests/pages/global-settings/branding.spec.ts
+++ b/cypress/e2e/tests/pages/global-settings/branding.spec.ts
@@ -369,6 +369,18 @@ describe('Branding', { testIsolation: 'off' }, () => {
       expect(background).to.satisfy((b) => b.startsWith(settings.primaryColor.newRGB));
     });
 
+    // check that login page has new styles applied
+    // https://github.com/rancher/dashboard/issues/10788
+    loginPage.goTo();
+
+    loginPage.submitButton().self().should('have.css', 'background').should((background: string) => {
+      expect(background).to.satisfy((b) => b.startsWith(settings.primaryColor.newRGB));
+    });
+
+    cy.login();
+    HomePagePo.goToAndWaitForGet();
+    BrandingPagePo.navTo();
+
     // Reset
     brandingPage.primaryColorPicker().set(settings.primaryColor.original);
     brandingPage.primaryColorCheckbox().set();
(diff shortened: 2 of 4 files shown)


cypress/e2e/po/pages/login-page.po.ts
```diff
@@ -44,7 +44,7 @@ export class LoginPagePo extends PagePo {
     });
   }
 
-  private submitButton(): AsyncButtonPo {
+  submitButton(): AsyncButtonPo {
```
eva-vashkevich, line 47: why was this needed?

aalves08: so that I can use it on the spec, otherwise it would be unreachable outside it's PO:
```
loginPage.submitButton().self().should('have.css', 'background').should((background: string) => { ... etc
```

cypress/e2e/tests/pages/global-settings/branding.spec.ts
```diff
@@ -369,6 +369,18 @@ describe('Branding', { testIsolation: 'off' }, () => {
...
 
+    // check that login page has new styles applied
+    // https://github.com/rancher/dashboard/issues/10788
+    loginPage.goTo();
+
+    loginPage.submitButton().self().should('have.css', 'background').should((background: string) => {
+      expect(background).to.satisfy((b) => b.startsWith(settings.primaryColor.newRGB));
+    });
+
+    cy.login();
```
eva-vashkevich, line 380: I think it would be good to reload a page and check that the styles are still the same just in case?

aalves08: done

PR #10790, opened 2024-04-11
Feature/10774 plugins

rak-phillip:
### Summary
Fixes #10774 

### Technical notes summary

- `shell/plugins/i18n.js` contains the most substantial refactoring. After updating this plugin to function in our Vue3 branch, I realized that we can easily apply the same patterns in Vue2. Could this cause issues in older versions of Harvester that target Vue versions less than 2.7?
- `shell/composables/useStore.ts` is a utility function that allows us to access the store from `setup()` within Vue2. This can be replaced by the `useStore` method that's provided by Vuex once we properly target Vue3

### Checklist
- [x] The PR is linked to an issue and the linked issue has a Milestone, or no issue is needed
- [x] The PR has a Milestone 
- [ ] The PR template has been filled out
- [ ] The PR has been self reviewed 
- [ ] The PR has a reviewer assigned
- [ ] The PR has automated tests or clear instructions for manual tests and the linked issue has appropriate QA labels, or tests are not needed
- [ ] The PR has reviewed with UX and tested in light and dark mode, or there are no UX changes

PR #10801, opened 2024-04-12
Feature/10774 plugins i18n

### LINKED ISSUE #10800: Refactor `shell/plugins/i18n.js` plugin
Refactor to i18n plugin to use an install function. Using an install function will help us to properly manage our Vue instance as Rancher Dashboard migrates to Vue3. 

rancher/dashboard#10774

rak-phillip:
### Summary
This refactors the i18n plugin to use an install function[^1]. 

[^1]: https://v2.vuejs.org/v2/guide/plugins.html#Writing-a-Plugin

Fixes #10800

### Technical notes summary

Our original approach of writing plugins will not be compatible with Vue3, but we can rewrite our plugins today so that there will be less refactoring required when we make the migration.

This change also rewrites the `t` component so that it can take advantage of backported Vue3 apis[^2].

[^2]: https://v2.vuejs.org/v2/guide/migration-vue-2-7.html#Backported-Features

I split this out of #10790, as I found that to be growing in complexity. We can more readily identify potential errors with this change by changing one component at a time.

Ideally, unit and e2e tests will catch any exceptions that could be introduced with this change. In the event that they don't, we will want to find example of the i18n plugin that are representative of the entire scope. That it included the t component, option, and directive.

Unhandled exceptions as a result of this change might cause a string to not render on the page with the only indication that something went wrong being an error printed to the console.

### Checklist
- [ ] The PR has a Milestone 


shell/composables/useStore.ts
rak-phillip, line 1: I need to document why `useStore` exists and stress that this is a temporary measure until we port to Vue3

rak-phillip:
@aalves08 I'm requesting a review from you because I'm running into some issues with the Kubewarden plugin. We might need to discuss ideas for how we can move this forward in a backwards compatible way..

jest.setup.js
```diff
@@ -13,6 +14,7 @@ global.TextEncoder = TextEncoder;
 global.TextDecoder = TextDecoder;
 
 Vue.config.productionTip = false;
+Vue.use(i18n);
```
cnotv, line 17: In Jest configuration with Vue3 I'd to do something like:
```

config.global.directives['t'] = t;
config.global.directives['clean-tooltip'] = VCleanTooltip ;
config.global.components['v-select'] = vSelect;
config.global.plugins = [VTooltip];
config.global.mocks['t'] = key => `%${key}%`
```

How would you translate that using i18n?
Using just `config.global.plugins = i18n`? 
Would that include all the 3 parts (global, component, directive)?

shell/composables/useStore.ts
```diff
@@ -0,0 +1,15 @@
+import { getCurrentInstance } from 'vue';
+
+/**
+ * useStore allows for accessing Vuex stores from within a setup function. This is a temporary measure for working with
+ * Vuex in Vue2.
+ *
+ * TODO: #9541 Remove for Vue 3 migration
```
cnotv, line 7: Since you mentioned the issue I'm working on, may I ask you what is the plan on how to remove this? 😅

rak-phillip: That's a good question. The expectation is that we will delete this shim and replace all usages of 

```js
 import { useStore } from '@/shell/composables/useStore';
```

with 

```js
```

Will that suffice, or shall we open a new issue to deal with refactorings like this?

jest.setup.js
cnotv: That sounds good because I've to mock many parts instead of 1 only.

shell/composables/useStore.ts
cnotv: Obviously not, I just needed to know what was the plan, thanks 🍻

cnotv:
Mind you sometimes we have something like this one around.

https://github.com/rancher/dashboard/blob/master/shell/components/form/MatchExpressions.vue#L73-L87

In the case of the tests, it's expected to exist in the store to don't break everything with `TypeError: t is not a function`

rak-phillip:
> In the case of the tests, it's expected to exist in the store to don't break everything with TypeError: t is not a function

The `i18n/t` mock will remain intact for the store. 

Anything else related to the `i18n` module for the store should remain unchanged, as that's handled via `shell/store/i18n.js`.

cnotv (approved):
LGTM

codyrancher (approved):

aalves08:
I did https://github.com/rancher/dashboard/pull/10816 + https://drone-publish.rancher.io/rancher/dashboard/5205 to test these changes and noticed a couple of things:

**Descriptions over list views don't seem to be appearing (they are translations):**
Screenshot 2024-04-16 at 18 22 30

**Kamaji partner extension is not working....** (I don't think this is related to this work, but this a dashboard that derives from current master, so possibly some changes may happened that contributed to this - no new publishes occurred in these last few months)

https://github.com/rancher/dashboard/assets/97888974/f398303e-d704-499c-9e72-e52b4507cc5f

The rest of the official Rancher extensions seems to work good (didn't notice anything weird...)

FYI @rak-phillip @codyrancher

rak-phillip:
@aalves08 thanks for the review. I fixed the issue that you noted in 3365157 and also wrote a simple test to assert that there's content within the banner on an affected page.

rak-phillip:
I'm noting e2e test failures in CI, but they're passing when I test a few of them locally. I'll have to look more into why this might be.

rak-phillip:
@aalves08 this is ready for another review

aalves08 (approved):
LGTM

Issue #10358, opened 2024-01-30
Create automation to ensure presence of all i18n entries


cnotv:
# Description

Ensure presence of the 7500+ entries in the i18n file as CI automation on PRs (GH action workflow).

## Context

As occurred in more scenarios and specifically for rancher/dashboard#9968 an entry was missing and displayed as `unknown`, although this should not be verified as a single case.

An attempted single case has [been handled](https://github.com/rancher/dashboard/blob/102f4e63e2970b6f7fa573586997d1f5501f444e/shell/models/__tests__/storage.k8s.io.storageclass.test.ts) (thanks @aalves08), clashing with other configurations. 

The i18n could eventually be retrieved with:

```ts
const yaml = require('js-yaml');
const fs = require('fs');

const translationData = yaml.load(fs.readFileSync(`${ process.cwd() }/shell/assets/translations/en-us.yaml`, 'utf8'));
```

The static entries may be retrieved with a RegEx through the files which use `this.t()` or `t()`, Vue text interpolation, etc

aalves08:
FYI @aalves08

gaktive:
Updating version number here since the PR was marked 2.9.0 but this ticket was 2.9.x. Hopefully the QA none label still applies here.

cnotv:
> Updating version number here since the PR was marked 2.9.0 but this ticket was 2.9.x. Hopefully the QA none label still applies here.

It's a test automation, so yes 😄

PR #10805, opened 2024-04-15
Add i18n string usage checker script

nwmac:
Fixes https://github.com/rancher/dashboard/issues/10358

This PR adds a script that checks that we are not missing i18n strings that are referenced in code.

Currently the check added to our CI workflow does not fail if there are missing strings.

Added to the i18n docs to describe the script and the comments that can be used to correct detections.

In addition to the script, this PR uses the new i18n comments to correct some uses in our code base - more to be done and we need to address the 16 missing strings that are detected.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -200,7 +200,10 @@ jobs:
         run: yarn install:ci
       
       - name: Run i18n string check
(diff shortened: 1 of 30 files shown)


richard-cox (approved):
Worth another shout out to the team, if they add translations that are used indirectly outside of `I18N_REGEX_LIST` they will need to add i18n comment

Issue #9711, opened 2023-09-14
Remove fetch() usage as leftover from Nuxt 


cnotv:
# Description

As the Nuxt world is departed, we should replace data fetching in the Vue3 way: https://vuejs.org/guide/built-ins/suspense.html#async-setup

The feature is **not** supported in Vue 2.7 https://v2.vuejs.org/v2/guide/migration-vue-2-7

The pattern is so specifically intended to have it as essential rule in the linter: https://eslint.vuejs.org/rules/no-lifecycle-after-await.html

cnotv:
@eva-vashkevich suggested moving the logic of `fetch()` within the `beforeMount()` lifecycle hook by wrapping the evaluation within a promise as a proven solution in some cases.

cnotv:
Another possible option of existing hooks:
https://router.vuejs.org/guide/advanced/data-fetching

cnotv:
The whole `fetch()` deprecation would include these considerations:
- We would need to define async requests in another location
- The use of `async setup()` is not possible due these restrictions: 
  - access global vars
  - define custom lifecycle hooks
  - access to other properties from `async setup()`, e.g. from `data()`
- The global variable `$fetch()` is adopted in more circumstances and will need to be replaced as well
- [`fetch()` is added from a mixin](https://github.com/rancher/dashboard/blob/master/shell/mixins/fetch.client.js) altogether with other elements:
  - Data `state` which includes `errors`, `pending`, and `timestamp`
  - Computed `$fetchState` which returns the `state`
  - `beforeCreate()` is used to assign `$fetch()`
  - Some delay is added for fetching

cnotv:
After analyzing the hackish implementation of Nuxt to create the `fetch()` "hook", I have been trying to create a PoC on Vue3 and it seems feasible the possibility to have an `async fetch()` also there.

https://codesandbox.io/s/vue3-vuex-custom-hook-yjghfv?file=/src/views/Home.vue

cnotv:
I am postponing then this refactoring after migration to Vue3 and keeping it open to define guidelines for the alternatives of `fetch()`.

cnotv:
Did we not say we were going to remove this? Because of the way it's written, I am not so sure is going to work with Vue3.
What's the purpose of this code? I have seen it outside of the `fetch.js` file, so I considered was a secondary purpose.

cnotv:
[...]

PR #10839, opened 2024-04-18
Migrate the usage of asyncData to fetch

codyrancher:
### Summary
Migrate the usage of asyncData to fetch. I'll remove the nuxt implementation of asyncData in a separate pr (https://github.com/rancher/dashboard/pull/10896/commits/3cdf3eceb4468eb7e4433cfcab07c1ff7d2569cd).

This will give us one less thing that we have to consider while working on Vue3. After the migration we will then remove all of the initialization code. 

The affected pages are, login, logout, setup, auth/config, and auth/roles.

All of these pages appear to be exercised through existing e2e tests.

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/pages/auth/login.vue
+++ b/shell/pages/auth/login.vue
@@ -23,112 +23,49 @@ import { LOGIN_ERRORS } from '@shell/store/auth';
 import {
   getBrand,
   getVendor,
-  getProduct,
   setBrand,
   setVendor
 } from '@shell/config/private-label';
 import loadPlugins from '@shell/plugins/plugin';
 import { fetchInitialSettings } from '@shell/utils/settings';
+import Loading from '@shell/components/Loading';
 
 export default {
   name:       'Login',
   components: {
-    LabeledInput, AsyncButton, Checkbox, BrandImage, Banner, InfoBox, CopyCode, Password, LocaleSelector
+    LabeledInput, AsyncButton, Checkbox, BrandImage, Banner, InfoBox, CopyCode, Password, LocaleSelector, Loading
   },
 
-  async asyncData({ route, redirect, store }) {
-    const drivers = await store.dispatch('auth/getAuthProviders');
-    const providers = sortBy(drivers.map((x) => x.id), ['id']);
-
-    const hasLocal = providers.includes('local');
-    const hasOthers = hasLocal && !!providers.find((x) => x !== 'local');
-
-    if ( hasLocal ) {
-      // Local is special and handled here so that it can be toggled
-      removeObject(providers, 'local');
-    }
-
-    let firstLoginSetting, plSetting, brand;
-
-    // Load settings.
-    // For newer versions this will return all settings if you are somehow logged in,
-    // and just the public ones if you aren't.
-    try {
-      await fetchInitialSettings(store);
-
-      firstLoginSetting = store.getters['management/byId'](MANAGEMENT.SETTING, SETTING.FIRST_LOGIN);
-      plSetting = store.getters['management/byId'](MANAGEMENT.SETTING, SETTING.PL);
-      brand = store.getters['management/byId'](MANAGEMENT.SETTING, SETTING.BRAND);
-    } catch (e) {
-      // Older versions used Norman API to get these
-      firstLoginSetting = await store.dispatch('rancher/find', {
-        type: NORMAN.SETTING,
-        id:   SETTING.FIRST_LOGIN,
-        opt:  { url: `/v3/settings/${ SETTING.FIRST_LOGIN }` }
-      });
-
-      plSetting = await store.dispatch('rancher/find', {
-        type: NORMAN.SETTING,
-        id:   SETTING.PL,
-        opt:  { url: `/v3/settings/${ SETTING.PL }` }
-      });
-
-      brand = await store.dispatch('rancher/find', {
-        type: NORMAN.SETTING,
-        id:   SETTING.BRAND,
-        opt:  { url: `/v3/settings/${ SETTING.BRAND }` }
-      });
-    }
-
-    if (plSetting.value?.length && plSetting.value !== getVendor()) {
-      setVendor(plSetting.value);
-    }
-
-    if (brand?.value?.length && brand.value !== getBrand()) {
-      setBrand(brand.value);
-    }
-
-    let singleProvider;
-
-    if (providers.length === 1) {
-      singleProvider = providers[0];
-    }
+  data() {
+    const username = this.$cookies.get(USERNAME, { parseJSON: false }) || '';
 
     return {
-      vendor:             getVendor(),
-      providers,
-      hasOthers,
-      hasLocal,
-      showLocal:          !hasOthers || (route.query[LOCAL] === _FLAGGED),
-      firstLogin:         firstLoginSetting?.value === 'true',
-      singleProvider,
-      showLocaleSelector: !process.env.loginLocaleSelector || process.env.loginLocaleSelector === 'true'
-    };
-  },
-
-  data({ $cookies }) {
-    const username = $cookies.get(USERNAME, { parseJSON: false }) || '';
-
-    return {
-      product: getProduct(),
-
       username,
       remember: !!username,
       password: '',
 
-      timedOut:    this.$route.query[TIMED_OUT] === _FLAGGED,
-      loggedOut:   this.$route.query[LOGGED_OUT] === _FLAGGED,
-      isSsoLogout: this.$route.query[IS_SSO] === _FLAGGED,
-      err:         this.$route.query.err,
+      timedOut:           this.$route.query[TIMED_OUT] === _FLAGGED,
+      loggedOut:          this.$route.query[LOGGED_OUT] === _FLAGGED,
+      isSsoLogout:        this.$route.query[IS_SSO] === _FLAGGED,
+      err:                this.$route.query.err,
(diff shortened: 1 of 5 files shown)


shell/pages/auth/login.vue
codyrancher, line 1: This page should be scrutinized the most during review. I wasn't able to do a 1 to 1 translation between asyncData and fetch primarly due to the previously used `mounted()` hook. 

While figuring out what the issue was I broke up and shrunk the definition of asyncData and decided to leave it. If you'd like me to go back to a 1 to 1 definition let me know, I don't think it should be an issue this time around.

shell/pages/auth/login.vue
```diff
@@ -36,100 +35,40 @@ export default {
...
-    let singleProvider;
-
-    if (providers.length === 1) {
-      singleProvider = providers[0];
-    }
+  data() {
+    const username = this.$cookies.get(USERNAME, { parseJSON: false }) || '';
 
     return {
-      vendor:             getVendor(),
```
codyrancher, line 98: This state wasn't actually used anywhere. It does still appear that we need to setVendor on this page for use cases off this page.

shell/pages/auth/login.vue
```diff
@@ -36,100 +35,40 @@ export default {
...
-      setBrand(brand.value);
-    }
-
-    let singleProvider;
-
-    if (providers.length === 1) {
-      singleProvider = providers[0];
-    }
+  data() {
+    const username = this.$cookies.get(USERNAME, { parseJSON: false }) || '';
```
codyrancher, line 39: Switching the access pattern of $cookies to be consistent with $route. It doesn't matter here but it matters when defining fetch as the code is treated completely differently by nuxt.

shell/pages/auth/login.vue
```diff
@@ -179,19 +118,76 @@ export default {
...
+    this.customLoginError = JSON.parse(value).loginError;
+    this.firstLogin = firstLoginSetting?.value === 'true';
     this.username = this.firstLogin ? 'admin' : this.username;
     this.$nextTick(() => {
       this.focusSomething();
     });
   },
 
   methods: {
+    async loadInitialSettings() {
```
codyrancher, line 145: Pulled all of this out to reduce the amount I needed to understand while mentally parsing asyncData.

shell/pages/auth/login.vue
```diff
@@ -179,19 +118,76 @@ export default {
...
-  mounted() {
+    this.providers = providers;
+    this.hasLocal = hasLocal;
+    this.showLocal = !hasOthers || (this.$route.query[LOCAL] === _FLAGGED);
+    this.customLoginError = JSON.parse(value).loginError;
+    this.firstLogin = firstLoginSetting?.value === 'true';
     this.username = this.firstLogin ? 'admin' : this.username;
     this.$nextTick(() => {
       this.focusSomething();
     });
```
codyrancher, line 141: This had to be moved from mounted. Previously asyncData would be executed and the state set before mounted was called. This caused big headaches while debugging e2e tests.

shell/pages/auth/login.vue
```diff
@@ -179,19 +118,76 @@ export default {
...
+    const providers = sortBy(drivers.map((x) => x.id), ['id']);
+    const hasLocal = providers.includes('local');
+    const hasOthers = hasLocal && !!providers.find((x) => x !== 'local');
 
-    this.customLoginError = JSON.parse(value).loginError;
-  },
+    if ( hasLocal ) {
+      // Local is special and handled here so that it can be toggled
+      removeObject(providers, 'local');
+    }
```
codyrancher, line 131: I love how we modify const lists. I really think immutable should take place of referentially const practically everywhere.

shell/pages/auth/login.vue
```diff
@@ -36,100 +35,40 @@ export default {
...
-      singleProvider,
-      showLocaleSelector: !process.env.loginLocaleSelector || process.env.loginLocaleSelector === 'true'
-    };
-  },
-
-  data({ $cookies }) {
-    const username = $cookies.get(USERNAME, { parseJSON: false }) || '';
-
-    return {
-      product: getProduct(),
```
codyrancher, line 113: This wasn't used anywhere in login

shell/pages/auth/login.vue
```diff
@@ -36,100 +35,40 @@ export default {
...
 
     return {
-      vendor:             getVendor(),
-      providers,
-      hasOthers,
-      hasLocal,
-      showLocal:          !hasOthers || (route.query[LOCAL] === _FLAGGED),
-      firstLogin:         firstLoginSetting?.value === 'true',
-      singleProvider,
-      showLocaleSelector: !process.env.loginLocaleSelector || process.env.loginLocaleSelector === 'true'
```
codyrancher, line 105: Nothing about this code is async so I moved it to data.

shell/pages/auth/setup.vue
```diff
@@ -137,36 +158,20 @@ export default {
...
+    this.$set(this, 'haveCurrent', !!current);
+    this.$set(this, 'username', me?.loginName || 'admin');
+    this.$set(this, 'isFirstLogin', isFirstLogin);
+    this.$set(this, 'mustChangePassword', mustChangePassword);
+    this.$set(this, 'current', current);
+    this.$set(this, 'v3User', v3User);
+    this.$set(this, 'serverUrl', serverUrl);
+    this.$set(this, 'mcmEnabled', mcmEnabled);
+    this.$set(this, 'telemetry', telemetry);
+    this.$set(this, 'principals', principals);
```
codyrancher, line 174: Some places we use $set and others we just use assignment. Is one better or easier for migration?

shell/pages/auth/login.vue
codyrancher: Should've just highlighted `showLocaleSelector`

shell/pages/auth/login.vue
cnotv: Our utilities are aimed at doing that, so it may be worth starting the conversation from that point since it's simpler.
For the record, I think me you and @rak-phillip are all in for that and we can even use ESLint for it, including Vue specific cases.

shell/pages/auth/setup.vue
cnotv: We are going to remove all the `this.$set` in the migration script with simple assignments, although it works differently between Vue2 and Vue3. The only concern is that my regex rule may not capture this case `me?.loginName || 'admin'` and will need to be modified by hand, but it's not a big deal.

cnotv: Oh also that's a good point, no idea if we want specifically some cases to not be reactive, so a consistent assignment is preferred to avoid debugging later on.

shell/pages/auth/login.vue
cnotv: I would have expected all the data to come from one place only, which is meant to be the state management, while persistent data is to be collected together in one single object. We do not want to tackle this now though, so what you did for me is already totally fine and is a good start.

shell/pages/auth/login.vue
cnotv: That's good you found out, thanks.
There's one aspect related to vendors and providers, which I am not sure we have covered in anyhow, but I guess it should be fine till other related variables are there as well.

cnotv (approved):
I cannot see any issue or valuable point to discuss that you did not already mention.

A topic for the future out of scope and which I already expressed is the presence of async patterns all the time from the state, which is really not necessary and caused clearly headaches also here.

LGTM

rak-phillip (commented):
Do you think that we'll be able to safely remove these lines about `asyncData` now that it's no longer used throughout Dashboard?

https://github.com/rancher/dashboard/blob/08f11a5d813e216467b418961e12ec6ac6954e8c/shell/initialize/App.js#L109-L118

https://github.com/rancher/dashboard/blob/08f11a5d813e216467b418961e12ec6ac6954e8c/shell/initialize/client.js#L382-L393

cnotv:
> Do you think that we'll be able to safely remove these lines about `asyncData` now that it's no longer used throughout Dashboard?

As you mentioned we all agree there is a point for another issue and there are more cases, but since we do it I add a couple of comments.

There are more components using asyncData: https://github.com/search?type=code&auto_enroll=true&q=%22.asyncData%22+OR+%22+asyncData%22+repo%3Arancher%2Fdashboard

Also, let's not forget that something related to the routing can be handled with the related [vue-router function](https://github.com/rancher/dashboard/issues/9711#:~:text=vue%2Drouter%20data%20fetching%20hook). Albeit we may not want to fall in the same trap 😅

codyrancher:
> Do you think that we'll be able to safely remove these lines about `asyncData` now that it's no longer used throughout Dashboard?
> 
> https://github.com/rancher/dashboard/blob/08f11a5d813e216467b418961e12ec6ac6954e8c/shell/initialize/App.js#L109-L118
> 
> https://github.com/rancher/dashboard/blob/08f11a5d813e216467b418961e12ec6ac6954e8c/shell/initialize/client.js#L382-L393
> 
> edit: There's quite a few more references to `asyncData` in `shell/initialize/client.js`, but I don't think it makes sense to list them all here 🙂

Absolutely. I took care of that in this PR https://github.com/rancher/dashboard/commit/3cdf3eceb4468eb7e4433cfcab07c1ff7d2569cd.

rak-phillip (approved):
LGTM - I gave the changes a quick smoke test and everything functions as it did before.

PR #10848, opened 2024-04-19
Replace default values of props using global variables with computed case

### LINKED ISSUE #10829: Replace default values of props using global variables with computed case
# Description

Replace the existing default assignment with a computed case. 
TBD naming convention/logic.

## Context

In the Vue3 migration list it has been omitted [this case](https://vuejs.org/guide/components/props.html) and it needs to be handled manually, also possible before actual migration.

Screenshot 2024-04-17 at 11 55 59

cnotv:
### Summary
Fixes #10829

As mentioned in the issue, this is a safe migration to Vue3 with limited use cases.
Using computed values based on props using i18n default values should be sufficient to not break anywhere.

A first attempt to use the global variable `t` has been carried out, unfortunately, it's not exposed in a prop scope and cannot be used.

Finally, as you can see by the commits and logs, it's mandatory to set a default empty value regardless, to not break existing logic.

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/KeyValueView.vue
+++ b/shell/components/KeyValueView.vue
@@ -36,7 +36,7 @@ export default {
     keyLabel: {
       type: String,
       default() {
-        return this.$store.getters['i18n/t']('generic.key');
+        return this.t('generic.key');
       },
     },
 
@@ -54,7 +54,7 @@ export default {
     valueLabel: {
       type: String,
       default() {
-        return this.$store.getters['i18n/t']('generic.value');
+        return this.t('generic.value');
       },
     },
 
--- a/shell/components/KeyValueView.vue
+++ b/shell/components/KeyValueView.vue
@@ -35,6 +35,7 @@ export default {
 
     keyLabel: {
       type: String,
+      default: '',
     },
 
     separatorLabel: {
@@ -50,6 +51,7 @@ export default {
 
     valueLabel: {
       type: String,
+      default: '',
     },
 
     valueBase64: {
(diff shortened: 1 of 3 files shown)


shell/components/KeyValueView.vue
```diff
@@ -70,6 +66,14 @@ export default {
   },
 
   computed: {
+    _keyLabel() {
```
codyrancher, line 69: With the `_` prefix are you trying to denote private scope?

Stylistically I don't think I've run into this in our codebase.

codyrancher (approved):

shell/components/KeyValueView.vue
cnotv: Kind of. 
I just had to pair an exposed value to an internal one, so I tried to use the closest naming convention to this case.
I am aware that's usually denotes more of a private scope, so if you think is confusing I'm fine to change it. 
What would you use in that case?

rak-phillip: I've encountered the convention of using `_` to denote "private" scope in some of our models and classes, for example:

https://github.com/rancher/dashboard/blob/75c595f97f8cfe1d0c2a37242306863ec4d61486/shell/plugins/dashboard-store/resource-class.js#L1709-L1739

I know that I have encountered it elsewhere throughout Dashboard, and it's a common enough convention in javascript where I think that its meaning is clear.

rak-phillip (approved):

shell/components/KeyValueView.vue
codyrancher: It was just a question. It's fine to ship as is.

PR #10859, opened 2024-04-22
Pin rancher backend image version for e2e tests

nwmac:
Testing pinning backend image to older version.

Note, done in file rather than via a var on the repository, so it is very clear what version is being used and so that to change it, we have to use a PR, which will itself go through the gates.


cnotv (approved):
LGTM

PR #10920, opened 2024-04-30
Removing the unauthenticated middleware

codyrancher:
### Summary
Removing the unauthenticated middleware

### Technical notes summary
Migrated the favorite icon logic out of the un/authenticated middlewares into the brand mixin.

Even though I'm not a fan of mixins I think the grouping at least makes sense given the favorite icon is one of several brand settings.

Unauthenticated pages, really just the login page.

### Screenshot/Video
The flashing is present in master as well so this isn't a regression. We should consider caching all brand settings in cookies like we do with light/dark mode.
https://github.com/rancher/dashboard/assets/55104481/7a59c8f5-54b9-4109-95bf-ab8b3d172e4d

### Checklist
- [x] The PR has a Milestone 


shell/middleware/authenticated.js
```diff
@@ -91,11 +90,6 @@ export default async function({
     // Load settings, which will either be just the public ones if not logged in, or all if you are
     await fetchInitialSettings(store);
 
-    // Set the favicon - use custom one from store if set
-    if (!haveSetFavIcon()) {
-      setFavIcon(store);
-    }
```
codyrancher, line 97: The brand mixin is used everywhere that the authenticated middleware is used so this should be fine.

shell/components/templates/standalone.vue
```diff
@@ -1,7 +1,3 @@
-<script>
```
codyrancher, line 1: This is for standalone plugins. They should have their own branding so I didn't switch this with the Brand mixin.

rak-phillip (approved):
LGTM - The favicon conceptually fits wit the brand mixin. I agree with your assessment of mixins, but addressing those is definitely something that would fall outside of the scope of this PR

PR #10923, opened 2024-05-01
Removing the i18n middleware from the application

codyrancher:
### Summary
Removing the i18n middleware from the application

### Technical notes summary
The i18n middleware only initialized the i18n store. I moved the initialization elsewhere.

Translations on any page could be tested.

### Screenshot/Video
https://github.com/rancher/dashboard/assets/55104481/de5d38ff-95dd-4f20-86b2-c5bb928cdfe4

### Checklist
- [x] The PR has a Milestone 


shell/initialize/client.js
```diff
@@ -90,7 +90,7 @@ const errorHandler = Vue.config.errorHandler || console.error; // eslint-disable
 createApp(nuxt.publicRuntimeConfig).then(mountApp).catch(errorHandler); // eslint-disable-line no-undef
 
 function callMiddleware(Components, context) {
-  let midd = ['i18n'];
```
codyrancher, line 93: This was the only place I found it. If you can find it elsewhere let me know.

shell/initialize/App.js
```diff
@@ -30,6 +30,9 @@ export default {
   }),
 
   created() {
+    // Ensure we initialize i18n for all pages
+    this.$store.dispatch('i18n/init');
```
codyrancher, line 34: If you have a better idea of where we should init this let me know. 

I moved this from the store/index because the store/index didn't actually call it for every page. Notably, if you refreshed the login page after changing the language it would switch back to english.

shell/store/index.js
```diff
@@ -778,7 +778,6 @@ export const actions = {
     }
 
     res = await allHash(promises);
-    dispatch('i18n/init');
```
codyrancher, line 781: No point in having it here if we need it for every page and we're already making that call elsewhere.

shell/initialize/App.js
rak-phillip: Perhaps from within the install function of the i18n plugin itself? I think we have the store available to us at that point.

Honestly, I think that it fits here just as well. This is essentially the root of our app and dispatching some sort of init method makes sense at this point.

codyrancher: I like the idea. It doesn't look like we currently expose the store to the install function and useStore is only evailable in the setup function within that plugin. Is there a reason we shouldn't expose it?

cnotv: We must install the plugin always after the store, since that's how it works, plus that's something which belongs to the plugin only and it's easier to find.
Since in the plugin install we have the options (as `_options`), should we not be able to dispatch it there?

shell/plugins/i18n.js
```diff
@@ -129,6 +132,12 @@ const i18n = {
 
 export default i18n;
 
-console.warn('The implicit addition of i18n options has been deprecated in Rancher Shell and will be removed in a future version. Make sure to invoke `Vue.use(i18n)` to maintain compatibility.');
+// This is being done for backwards compatibility with our extensions that have written tests and didn't properly make use of Vue.use() when importing and mocking translations
+// Example failing test https://github.com/rancher/dashboard/actions/runs/8927503474/job/24521022320?pr=10923
+const isThisFileBeingExecutedInATest = process.env.JEST_WORKER_ID !== undefined || process.env.NODE_ENV === 'test';
```
codyrancher, line 137: I wasn't sure which would be true in our extensions so I used both.

Do we even want to use this pattern? It's a little odd but it provides the backwards compatibility while keeping our production environment in the ideal state.

cnotv: AFAIK `process.env.NODE_ENV === 'test'` is the standard.
This is what we use in other configs:
https://github.com/rancher/dashboard/blob/master/shell/babel.config.js#L23-L28

Also, I've never seen the other one, but I dunno everything 😅

rak-phillip: I did open an issue for Kubewarden to fix this[^1]. Maybe this is a scenario where we need to identify and update all extensions that are relying on this behavior. 

Ideally, we would be able to remove this pattern after that's done. 

[^1]: https://github.com/rancher/kubewarden-ui/issues/679

codyrancher:
Whenever you're ready, feel free to let me know if you'd like any additional changes or if you approve.

cnotv (approved):
LGTM

jest.setup.js
```diff
@@ -14,7 +14,7 @@ global.TextEncoder = TextEncoder;
 global.TextDecoder = TextDecoder;
 
 Vue.config.productionTip = false;
-Vue.use(i18n);
+Vue.use(i18n, { store: { dispatch() {} } });
```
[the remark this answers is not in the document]
codyrancher: I think it could actually just be entirely omitted now that I access it using the optional operator. If not, yeah we should mock it.

I'll address it in my next PR.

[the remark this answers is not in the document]
cnotv: If we mock the dispatch, then we should mock everything else too.
In both cases we are not that strict though, as we do not test the i18n, we just want to not break it.

PR #10929, opened 2024-05-01
Only logout if the user is logged in

### LINKED ISSUE #10926: Rancher UI reloads on initial load when not logged in
When not logged in, load the Rancher URL.

The page will load, show the login screen and then shortly after will reload and show the login screen with the timed out message shown:

Screen Recording 2024-05-01 at 09 57 26

codyrancher:
### Summary
This prevents us from redirecting multiple times when visiting the index on login 

Fixes #10926

### Technical notes summary
There is still a case on the logout page that we need to be able to call logout even if the user hasn't been authenticated yet because the auth pages don't use the authenticated middleware.

This caused issues with the login page in a roundabout way. 

1. On the index page we authenticate the user so we can redirect them to the appropriate page if the user is logged in.
2. If a steve api fails due to authentication (happens if the user isn't logged in when visiting the index page) we immediately log the user out and redirect them to the login page.
3. Between the index page and steve api we would redirect multiple times and we now do a full page refresh on logout to shed all of the loaded extensions. This change prevents logging out unless the user is logged in (unless they visit the logout page directly), which reduce the amount of redirecting. 

This change will also reduce the unnecessary timeout redirect we would previously see when visiting the index page.

Logging in and logging out

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/pages/auth/logout.vue
+++ b/shell/pages/auth/logout.vue
@@ -2,7 +2,7 @@
 
 export default {
   async fetch() {
-    await this.$store.dispatch('auth/logout', null, { root: true });
+    await this.$store.dispatch('auth/logout', { force: true }, { root: true });
   }
 };
 </script>
--- a/shell/store/auth.js
+++ b/shell/store/auth.js
@@ -344,7 +344,23 @@ export const actions = {
     }
   },
 
-  async logout({ dispatch, commit }) {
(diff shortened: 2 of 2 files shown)


rak-phillip (approved):
LGTM - tested and confirmed that the issue is resolved.

PR #10945, opened 2024-05-03
Adding e2e tests for pages which use middleware to redirect

codyrancher:
### Summary
Also fixed a couple routes, and removed some test setup code that was commented on in another review.

### Technical notes summary
I wanted to put these tests in place before changing or removing the middleware to make sure we keep the experience consistent.

These are the tests.

None

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/jest.setup.js
+++ b/jest.setup.js
@@ -14,7 +14,7 @@ global.TextEncoder = TextEncoder;
 global.TextDecoder = TextDecoder;
 
 Vue.config.productionTip = false;
-Vue.use(i18n, { store: { dispatch() {} } });
+Vue.use(i18n);
 Vue.use(VTooltip);
 Vue.use(VCleanTooltip);
 Vue.component('v-select', vSelect);
--- a/shell/config/router.js
+++ b/shell/config/router.js
@@ -89,7 +89,7 @@ export const routerOptions = {
           name:      'account-create-key'
         },
         {
-          path:      'auth',
+          path:      '/c/:cluster/auth',
           component: () => interopDefault(import('../pages/c/_cluster/auth/index.vue')),
           name:      'c-cluster-auth'
         },
@@ -99,7 +99,7 @@ export const routerOptions = {
           name:      'c-cluster-ecm'
         },
         {
-          path:      'settings',
+          path:      '/c/:cluster/settings',
           component: () => interopDefault(import('../pages/c/_cluster/settings/index.vue')),
           name:      'c-cluster-settings'
         }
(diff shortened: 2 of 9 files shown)


shell/config/router.js
```diff
@@ -89,7 +89,7 @@ export const routerOptions = {
           name:      'account-create-key'
         },
         {
-          path:      'auth',
```
codyrancher, line 92: This configuration wasn't actually being used correctly. It was falling back on `/c/:cluster/:product`

auth

shell/config/router.js
```diff
@@ -99,7 +99,7 @@ export const routerOptions = {
           name:      'c-cluster-ecm'
         },
         {
-          path:      'settings',
```
codyrancher, line 102: This configuration wasn't actually being used correctly. It was falling back on /c/:cluster/:product

settings

jest.setup.js
```diff
@@ -14,7 +14,7 @@ global.TextEncoder = TextEncoder;
 global.TextDecoder = TextDecoder;
 
 Vue.config.productionTip = false;
-Vue.use(i18n, { store: { dispatch() {} } });
```
codyrancher, line 17: Made this change in regards to https://github.com/rancher/dashboard/pull/10923#discussion_r1588528484

cnotv (commented):
For me, the test coverage seems fine, but I do not grasp why it should redirect and if this is common in all these pages, would it not be better a unit test at this point?
It obviously does not exclude having both or any issue to have just E2E, so it's just a simple curiosity.

codyrancher:
> For me, the test coverage seems fine, but I do not grasp why it should redirect and if this is common in all these pages, would it not be better a unit test at this point? It obviously does not exclude having both or any issue to have just E2E, so it's just a simple curiosity.

Most of these redirects don't have any logic so I was planning to move them as redirects defined in the router config. This would change the units all together so I'd have to rewrite all the unit tests and therefore I wouldn't be verifying that I didn't change behavior.

I could just make the changes and maybe write some router unit tests without this verification step. I'm okay with either. Let me know which you'd prefer.

cnotv:
> Most of these redirects don't have any logic so I was planning to move them as redirects defined in the router config. This would change the units all together so I'd have to rewrite all the unit tests and therefore I wouldn't be verifying that I didn't change behavior.

I would expect just to change the spied function, although I may completely miss the actual case.
Could we at least specify when the redirect happens in the test? Because as it's written it simply redirects for no reason.
Is it me overlooking something or would you agree? If not agree, please leave it as it is :)
 
> I could just make the changes and maybe write some router unit tests without this verification step. I'm okay with either. Let me know which you'd prefer.

I tend to stick to an already done work, especially if is also fine.

cnotv (approved):
LGTM

PR #10956, opened 2024-05-06
Remove inline defined middleware

codyrancher:
### Summary
Remove inline defined middleware. Replace it primarily with router configuration.

We already have adequate automated testing for this.

### Checklist
- [x] The PR has a Milestone 


shell/pages/c/_cluster/settings/index.vue
codyrancher, line 1: I couldn't migrate this file to the router config because it relies on the store already having schemas. 

The initialization sequence (even modified slightly) will not have schemas loaded before route resolution when re/loading a page.

shell/pages/c/_cluster/settings/index.vue
```diff
@@ -3,13 +3,13 @@ import { NAME as SETTINGS } from '@shell/config/product/settings';
 import { MANAGEMENT } from '@shell/config/types';
 
 export default {
-  middleware({ redirect, route, store } ) {
-    const hasSettings = !!store.getters[`management/schemaFor`](MANAGEMENT.SETTING);
+  beforeCreate() {
```
rak-phillip, line 6: I'll preface this comment by saying that I'm fine with the approach represented here, I'm more interested in exploring an alternative to see if it makes sense for this use case. 

Have we considered utilizing in-component guards[^1] instead of the `beforeCreate()` hook? I think that the guard would better communicate what we are doing within this component, but it comes with the tradeoff that we might not have access to `this` when we want it.. 

[^1]: https://v3.router.vuejs.org/guide/advanced/navigation-guards.html#in-component-guards

codyrancher: I haven't considered that. Do you know of a way to access the store from the guard? 

It's actually a pretty reasonable idea which may apply better in that it executes before the component is created just like middleware.

codyrancher: I did find https://github.com/vuejs/vue-router/issues/1103#issuecomment-938511312

rak-phillip: Yeah, it looks like we would have to do something like this instead

```
beforeRouteEnter (to, from, next) {
  next(vm => {
    const hasSettings = !!vm.$store.getters[`management/schemaFor`](MANAGEMENT.SETTING);

    ...
  })
}
```

codyrancher: I'll keep looking into this. It's interesting and not as straight forward as I would've thought. If we access the vm it looks like the component will already be created so not a big advantage.

image

There's another suggestion I found which mentions just importing our store and accessing it directly. With responses that would resemble our current predicament in regards to how we can access our store in the app.

image

Any thoughts on the idea of making the store global?

codyrancher: I gave the global store a shot and it's a no go. The schemas aren't present when re/loaded.

I'm fine with leaving it as is or using the next(vm => {}). Do you think one is better than the other?

rak-phillip: I think we can leave this as is. It sounds like we won't get much of a benefit and the syntax will end up being less direct with the in-component guard.

rak-phillip (approved):
LGTM

cnotv:
Nice cleanup 👍

PR #10971, opened 2024-05-07
Install redirect swap

codyrancher:
### Summary
Swap `InstallRedirect` middleware with a mixin.

There is a change in behavior. The mixin is non-blocking which requires consumers of the mixin to recognize and use `redirectPending` if they needed blocking behavior.

Index pages of gatekeeper, istio, logging, longhorn, monitoring, neuvector

### Checklist
- [x] The PR has a Milestone 


shell/config/router/routes.js
```diff
@@ -181,14 +181,6 @@ export default [
...
         component: () => interopDefault(import('@shell/pages/c/_cluster/explorer/index.vue')),
         name:      'c-cluster-explorer'
-      }, {
-        path:      '/c/:cluster/backup',
-        component: () => interopDefault(import('@shell/pages/c/_cluster/backup/index.vue')),
-        name:      'c-cluster-backup'
-      }, {
-        path:      '/c/:cluster/cis',
-        component: () => interopDefault(import('@shell/pages/c/_cluster/cis/index.vue')),
-        name:      'c-cluster-cis'
```
codyrancher, line 191: I couldn't find usage of these routes.

shell/mixins/__tests__/chart-product-redirect.test.ts
codyrancher, line 1: After the comments on https://github.com/rancher/dashboard/pull/10945 I think I should justify the switch to unit tests.

- To exercise this in e2e we'd have to install and uninstall charts which would be slow.
- I changed the behavior slightly so I wasn't trying to preserve the exact same behavior like I was last time.

shell/utils/install-redirect.js
```diff
@@ -1,55 +0,0 @@
-import { REPO_TYPE, REPO, CHART, VERSION } from '@shell/config/query-params';
-
-export default function(product, chartName, defaultResourceOrRoute, install = true) {
```
codyrancher, line 3: `defaultResourceOrRoute` was only used in backup and cis which I removed.

shell/mixins/chart-product-redirect.js
```diff
@@ -0,0 +1,49 @@
+import { REPO_TYPE, REPO, CHART, VERSION } from '@shell/config/query-params';
+
+export default function(product, chartName, install = true) {
+  return {
+    data() {
+      return { redirectPending: true };
```
codyrancher, line 6: I really don't like this, if you have suggestions to remove `redirectPending` I'd like to hear it.

This is in place because on the pages which have content to render switching from the middleware to the mixin would cause the content to flash before performing a redirect.

The reason for this is because the pages wouldn't render until middleware was fully executed. This version on the other hand doesn't block rendering implicitly so we needed a way to hide the content until all the redirect logic finished executing.

shell/mixins/__tests__/chart-product-redirect.test.ts
```diff
@@ -0,0 +1,93 @@
...
+  const PRODUCT = 'PRODUCT';
+  const CHART_NAME = 'CHART_NAME';
+
+  const createComponent = (...args: Parameters<typeof ChartProductRedirect>) => ({
+    props:  {},
+    render() { },
+    mixins: [ChartProductRedirect(...args)]
+  });
+
+  it('should exit early because the product is already active', async() => {
```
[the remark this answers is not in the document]
codyrancher: So, this test is mostly about coverage and it's verifying that we exit beforeCreate here https://github.com/rancher/dashboard/pull/10971/files#diff-902296d80668085dc0fe9c31e2d89c2ea9c00e5d9bdc08b479b83749eb5dbfdfR11-R13

I'm verifying that we don't execute any of the code below by checking to make sure the currentCluster mock getter is never called which implies we've exited early.

shell/mixins/chart-product-redirect.js
codyrancher: > About the suggestions, could we not take out all this logic and use the state management to trigger an action and get the state for the loading?
Why is this an exception in our loading state scenario using fetch()?
Also considering the repetition of the markup, should we not have some kind of component with slots or a routing guard for this?

I think you're right. It's related to the issue you opened a few weeks ago where we want to refactor loading. I think pushing the loading logic into state management is the right thing to do. I'm considering doing that when removing $loading from our codebase.

There is a subtle different in this case because the original code didn't actually have a loading screen for these page components. The page would just stay blank until the middleware finished and it would either render the current page or redirect.

I am still debating if changing beforeCreate to fetch would be better for this.

cnotv: Then probably better to take a second look later for the loading. It's a bit weird that we keep a blank page, but not doing UX at this round 😅 

I am personally indifferent between `fetch()` and `beforeCreate()`, plus in our code we have both, so I would rather stick to what you did, since it would be purely Vue functions at least.

shell/mixins/__tests__/chart-product-redirect.test.ts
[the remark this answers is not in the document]
cnotv: Yes, I've seen it later in the code. I was just saying that it was not clear by just reading the test and I wanted just to be sure of other possible cases. 
Otherwise it's fine like this 👍

shell/mixins/chart-product-redirect.js
rak-phillip: I don't have any direct recommendations for `redirectPending`, but if I'm reading this correctly, we might be able to consider implementing a per-route guard[^1] using `beforeEnter()` to keep the behavior similar to as it was before. It appears that access to the store in a route guard would be the primary blocker in accomplishing this, but our `useStore()` composable[^2] might be able to get the job done.  

This would require us to replace this mixin entirely and I'm not asking that we do this, but I can envision how it would remove the `redirectPending` requirement.

[^1]: https://v3.router.vuejs.org/guide/advanced/navigation-guards.html#per-route-guard

[^2]: https://github.com/rancher/dashboard/blob/master/shell/composables/useStore.ts

cnotv: The concept of having route guards in a dedicated section for routing would be optimal.

codyrancher: Since we now have that dedicated section I went ahead and gave it a try moving this to beforeEnter(). As of right now beforeEnter() gets called before we load the schemas via the authenticated middleware.

Also unfortunately the mixin doesn't work anymore because of recent changes which prevents the component from ever rendering and thus never executed. This still works in master because middleware doesn't require the component to be rendered before it's executed.

I'm going to update this PR with the beforeEnter solution once the loadManagement and product setting is moved to navigation guards in master.

Issue #10621, opened 2024-03-14
Missing doc link to "GKE cluster provisioning docs" while creating Private GKE


valaparthvi:
**Setup**
- Rancher version: 2.8.3-rc3
- Rancher UI Extensions: v2.8.0
- Browser type & version: Chrome

**Describe the bug**

The doc link to access "GKE cluster provisioning docs" is missing.

**To Reproduce**

1. Go to GKE cluster creation form.
2. In `Cluster Options`, set `Private Cluster` to true, and click the link "GKE cluster provisioning docs"(ref:https://rancher.com/docs/rancher/v2.8/en/cluster-provisioning/hosted-kubernetes-clusters/gke/)

**Result**
404 Not Found
**Expected Result**

**Screenshots**

![Screenshot from 2024-03-14 14-13-59](https://github.com/rancher/dashboard/assets/11015077/dc4729a4-3ada-43b4-8fc9-64b42b1b711b)

**Additional context**

nwmac:
@mantis-toboggan-md I suggest we ensure this is corrected in the Vue port

eva-vashkevich:
Confirmed that documentation opens correctly after migration.

valaparthvi:
I confirm it too on Rancher (v2.9-d2fa8a7d6116c48a814ee7983af522d614461e49-head) and Dashboard (master https://github.com/rancher/dashboard/commit/df60f440f41fe1b59c35b5181292504bb5e66059), thanks for checking @eva-vashkevich. I will close the issue now unless it needs to remain open.

PR #11025, opened 2024-05-14
Gke provisioning

### LINKED ISSUE #8967: Move GKE from Ember to Vue
In support of moving all public cloud provisioning off Ember (https://github.com/rancher/dashboard/issues/8083 and in a way, tied to work for finding all pages off Ember in https://github.com/rancher/dashboard/issues/8718), this ticket is to move the GKE pages to being in Vue. 

Currently, to provision GKE, it's iFramed from Ember into Vue.

Original embedding work: https://github.com/rancher/dashboard/issues/2749

mantis-toboggan-md:
### Summary
Fixes #8967 #10621 

Functionality should be on par with existing GKE provisioning in the Ember UI. Some Ember validation happens at same time: it's all form validation in this port. There are a handful of areas where I have deviated from Ember:

- logging and monitoring services are hardcoded lists with two options, one of which is none: I'm using a checkbox instead of a dropdown. 
- node pool 'local ssd count/local ssd disks' is a GB input in Ember - this is inaccurate, the google ui describes the field as: "For every node in a node-pool, the number of local SSD disks are attached."
- node pool preemptible is labeled as beta - it appears to be GA now https://cloud.google.com/kubernetes-engine/docs/how-to/preemptible-vms
- node pool has only 1 thing under 'show advanced options' so I just got rid of the show advanced toggle - seems excessive
- (autoscaling) node minimum is not configurable in Ember. The input component has a max value of...itself. The input is now validated so that the minimum number of nodes specified can't exceed the maximum number of nodes specified
- Ember had two 'description' inputs - one for the norman cluster object and one for the GKE cluster. I combined them in the same way that the 'name' input actually sets a norman cluster name and gkeconfig name (aks and eks have the same thing with names)
- image type options list is shorter and simpler - my ~excuses~ justification for this is detailed in inline comments
-    Ember UI had a section called 'config' with what appears to be entirely networking stuff, plus a kubernetes version dropdown. I did not care for this, particularly given that the networking stuff is the most complicated part of the form, and re-arranged the sections so that the k8s version dropdown appears with location and labels ('general stuff' I guess)
- in the Ember location can't be changed once you start configuring the rest of the form...I allow it to be changed and re-fetch relevant GCP data when it does
- unlike EKS and AKS I've left projectid/cloud credential un-changeable once the user starts configuring the rest of the cluster (more in line with the Ember UX). This is because the project id field is an input, while EKS/AKS are able to rely entirely on dropdowns...the ux where we just automatically verify the new cred selection and re-fetch any data from azure/aws does not work well here

- GKE provisioning generally
- editing clusters when the underlying GKE cluster was provisioned successfully - most fields can't be edited, though new node pools can be added/fully configured
- editing clusters when GKE provisioning fails (one way to achieve this state consistently is to use one of the spoofed 'shared subnetwork' options) - the user should be able to edit gke config fields they otherwise wouldn't be able to
- validation: the nodepool name validation, min/max node validation is the most deviant and probably worthy of double-checking
- adding/removing nodepools on edit
- adding cluster members on create or on edit
- in Ember I see a request for GKE service accounts but I don't see them used anywhere in the UI so I left it out. if you see something say something.

This PR is pretty self-contained and should not impact areas outside of GKE provisioning. There is a minor stylistic update to ArrayList so it matches KeyValue (both are used next to eachother in node pools and the difference in spacing was apparent, and ugly)

### Checklist
- [x] The PR has a Milestone 


shell/components/form/ArrayList.vue
```diff
@@ -339,7 +339,7 @@ export default {
     </div>
     <div
       v-if="showAdd && !isView"
-      class="footer"
+      class="footer mt-20"
```
mantis-toboggan-md, line 342: Added to make this consistent with keyvalue styling

pkg/gke/util/gcp.ts
```diff
@@ -0,0 +1,159 @@
...
+}
+
+export function getGKESubnetworks(store: Store<any>, cloudCredentialId: string, projectId: string, location: {zone?: string, region?: string}): Promise<getGKESubnetworksResponse> {
+  return getGKEOptions('gkeSubnetworks', store, cloudCredentialId, projectId, location);
+}
+
+export function getGKESharedSubnetworks(store: Store<any>, cloudCredentialId: string, projectId: string, location: {zone?: string, region?: string}): Promise<getGKESharedSubnetworksResponse> {
+  // return getGKEOptions('gkeSharedSubnets', store, cloudCredentialId, projectId, location);
+  // TODO nb remove this test code
+  return Promise.resolve({
```
mantis-toboggan-md, line 81: This should be removed before the PR is merged. Our dev GCP setup doesn't have permission to create shared subnets - this data was borrowed from https://github.com/rancher/ui/pull/4790

richard-cox (changes requested):
I've given the functional side a first pass, findings below

- Cloud Credentials
  - When creating a `Google` Cloud Credential directly via the `Cloud Credential` page no project id is requested, however whilst on the create cloud cred steps when creating a cluster (when no cred found) it is 
    - Is the project id something only storable / discernible in the cluster context?
  - When selecting a `Google` Cloud Credential via Cluster Create page 
    - clicking the `Create` button without going through `Authenticate` gives the error notification `cluster kubernetes version cannot be empty string`. Can we disable the button if the user hasn't authed?
    - clicking the `Create` button without going through `Authenticate`, and then supplying the project id --> authenticate results in the rest of the form showing however the node pools section is empty
  - the red notification at top of the create cluster create cloud cred page saying `cloud credential or project id missing` pops up in odd ways
    - on a system with no cloud creds, when going to the cluster create page it's shown before the user hasn't had a chance to enter a project id, and will then blip in/out when the user enters/clears field
  - After selecting a Cloud Cred and clicking on `Authenticate` the button width changes to show the spinner which wiggles the two input field widths
- Create
  - I created a cluster by adding just a name, no other changes, and it fails to come up
    - rancher status `Waiting on gke crd to be initialized`, i don't see any clusters in https://console.cloud.google.com/kubernetes/list/overview?authuser=1&project=<snip> (is that the right place?)
- Create (form)
  - `logging and monitoring services are hardcoded lists with two options, one of which is none: I'm using a checkbox instead of a dropdown.`
    - Not sure about this suggestion, but should we inform the user of the service names in a tooltip?
  - Old UI Cluster Options --> Pod secondary Range Name is missing (but Pod Secondary CIDR Block is shown). Is this related to the `getGKESharedSubnetworks` comment?
    - Same with missing `Services Secondary Range Name`
  - New UI Networking --> Advanced Options
    - Can the checkboxes be vertical instead of horizontal? They all seem to affect each other in some way, and it would tie together `CIDR Block` key value component exposed via `Master Authorized Network`. 
  - After filling in the form and clicking `Create` the contents of the Node Pool tabs is briefly shown with no content
  - Node Pools
    - Old UI `Image Type` has an additional option `Windows Long Term Service Channel with Containerd` (zonal, us-central1-c)
    - API Access Scopes --> `Set access for each API`. We've lost horizontal space in the UI given the tab bar. For me full screen this means only one column of select controls are shown. Given the content of these is small can we shrink them width wise to allow more column/s? Think maybe 2, 3 if they don't look squished?


Bonus Point (not related to PR)  - Create EKS --> Cluster Options --> shows a Kube Version of `[1.29`. Not sure where this comes from in hosted world, but it also happens in master

pkg/gke/components/CruGKE.vue
```diff
@@ -0,0 +1,840 @@
...
+
+    machineTypeOptions(): {label: string, kind?: string, value?: string, disabled?: boolean, [key: string]: any}[] {
+      const allTypes = this.machineTypesResponse?.items;
+      const out = [] as {label: string, kind?: string, value?: string, disabled?: boolean, [key: string]: any}[];
+
+      if (!allTypes) {
+        return out;
+      }
+
+      allTypes.forEach((type, i) => {
```
richard-cox, line 476: TS needs some typin' here, line 532 (`this.errors.push(err);`) and lines 548, 549 (`this.$refs.pools?.select` related)

pkg/gke/components/CruGKE.vue
```diff
@@ -0,0 +1,840 @@
...
+      if (this.config.projectID && this.config.googleCredentialSecret) {
+        this.getMachineTypes();
+      }
+    },
+
+    getMachineTypes() {
+      this.loadingMachineTypes = true;
+      const zone = this.config.zone || this.config.locations?.[0];
+
+      getGKEMachineTypes(this.$store, this.config.googleCredentialSecret, this.config.projectID, { zone, region: this.config.region }).then((res) => {
```
richard-cox, line 528: IMHO for this use case then/catch looks neater, but i think we're trying to avoid using this pattern. changing this to try/await/catch should still work (even where it's called doesn't `await getMachineTypes`)

pkg/gke/components/Networking.vue
```diff
@@ -0,0 +1,771 @@
...
+    loadGCPData() {
+      this.loadingNetworks = true;
+      this.loadingSubnetworks = true;
+      this.getNetworks();
+      this.getSubnetworks();
+      this.getSharedSubnetworks();
+    },
+
+    getNetworks() {
+      getGKENetworks(this.$store, this.cloudCredentialId, this.projectId, { zone: this.zone, region: this.region }).then((res) => {
```
richard-cox, line 277: Few more then/catch usages here

pkg/gke/components/Networking.vue
```diff
@@ -0,0 +1,771 @@
...
+    subnetworks() {
+      return this.subnetworksResponse.items || [];
+    },
+
+    sharedNetworks(): {[key: string]: GKESubnetwork[]} {
+      const allSharedSubnetworks = this.sharedSubnetworksResponse?.subnetworks || [];
+      const networks = {} as any;
+
+      allSharedSubnetworks.forEach((s) => {
+        const network = (s.network).split('/').pop() || s.network;
```
richard-cox, line 347: (s.network) --> s.network

pkg/gke/components/Networking.vue
```diff
@@ -0,0 +1,771 @@
...
+
+        if (network && !networks[network]) {
+          networks[network] = [];
+        }
+        networks[network].push(s);
```
richard-cox, line 352: what happens if `network` is undefined?

pkg/gke/components/Networking.vue
```diff
@@ -0,0 +1,771 @@
...
+  computed: {
+    ...mapGetters({ t: 'i18n/t' }),
+
+    subnetworks() {
+      return this.subnetworksResponse.items || [];
+    },
+
+    sharedNetworks(): {[key: string]: GKESubnetwork[]} {
+      const allSharedSubnetworks = this.sharedSubnetworksResponse?.subnetworks || [];
+      const networks = {} as any;
```
richard-cox, line 344: nitpick, --> `const networks: any = {};`

pkg/gke/components/Networking.vue
```diff
@@ -0,0 +1,771 @@
...
+    },
+
+    selectedClusterSecondaryRangeName: {
+      get(): {rangeName: string, ipCidrRange?: string, label: string} | undefined {
+        if (!this.clusterSecondaryRangeName) {
+          return {
+            label:     this.t('generic.none'),
+            rangeName: GKE_NONE_OPTION
+          };
+        } else return this.clusterSecondaryRangeOptions.find((opt) => opt.rangeName === this.clusterSecondaryRangeName);
```
richard-cox, line 486: nitpick, code style wise we always wrap in `{ }`. In this case though it could just be
```
if (this.clusterSecondaryRangeOptions) {
   return this.clusterSecondaryRangeOptions.find
}

return {
...
}
```

or super safe and simpler
```
        return this.clusterSecondaryRangeOptions.find((opt) => opt.rangeName === this.clusterSecondaryRangeName) || {
          label:     this.t('generic.none'),
          rangeName: GKE_NONE_OPTION
        };
```

pkg/gke/components/Networking.vue
```diff
@@ -0,0 +1,771 @@
...
+          this.$emit('update:clusterIpv4CidrBlock', '');
+        } else {
+          this.$emit('update:clusterSecondaryRangeName', neu.rangeName);
+          this.$emit('update:clusterIpv4CidrBlock', neu.ipCidrRange);
+        }
+      }
+    },
+
+    selectedServicesSecondaryRangeName: {
+      get(): {rangeName: string, ipCidrRange?: string, label: string} | undefined {
```
richard-cox, line 500: same as selectedClusterSecondaryRangeName above

pkg/gke/components/Networking.vue
```diff
@@ -0,0 +1,771 @@
...
+            data-testid="gke-master-authorized-network-cidr-keyvalue"
+            @input="$emit('update:masterAuthorizedNetworkCidrBlocks', $event)"
+          />
+        </div>
+      </div>
+    </template>
+  </div>
+</template>
+
+<style lang="scss">
```
richard-cox, line 765: needs a `scoped`

pkg/gke/components/GKENodePool.vue
```diff
@@ -0,0 +1,488 @@
...
+          value: 'pd-ssd',
+        }
+      ];
+    },
+
+    selectedImageType: {
+      get() {
+        return this.imageTypeOptions.find((opt) => opt.value === this.imageType) || { value: this.imageType, label: this.t(`gke.imageType.${ this.imageType }`, null, this.imageType) };
+      },
+      set(neu: {label: string, kind?: string, value?: string, disabled?: boolean, [key: string]: any}) {
```
richard-cox, line 239: `label: string, kind?: string, value?: string, disabled?: boolean, [key: string]: any` is used 6 times, would be worth being a `type` somewhere

pkg/gke/components/Config.vue
```diff
@@ -0,0 +1,484 @@
...
+              {{ t('gke.clusterLabels.label') }}
+            </h4>
+          </template>
+        </KeyValue>
+      </div>
+    </div>
+  </div>
+</template>
+
+<style lang="scss">
```
richard-cox, line 475: `scoped`

pkg/gke/components/AdvancedOptions.vue
```diff
@@ -0,0 +1,225 @@
...
+          color="warning"
+          label-key="gke.enableKubernetesAlpha.warning"
+          icon="icon-warning"
+        />
+      </div>
+    </div>
+  </div>
+</template>
+
+<style lang="scss">
```
richard-cox, line 216: `scoped`

pkg/gke/components/AccountAccess.vue
```diff
@@ -0,0 +1,161 @@
...
+        type="button"
+        class="btn role-secondary"
+        mode="authenticate"
+        @click="testProjectId"
+      />
+    </div>
+  </div>
+</template>
+
+<style lang="scss">
```
richard-cox, line 121: `scoped`

richard-cox (commented):
Given the code a once over, just minor things.

mantis-toboggan-md:
> * Cloud Credentials
>   
>   * When creating a `Google` Cloud Credential directly via the `Cloud Credential` page no project id is requested, however whilst on the create cloud cred steps when creating a cluster (when no cred found) it is

That's an oversight on my part - the project id isn't required for cloud credential creation, just cluster creation. I've hidden the input while the user is creating cloud credentials. 

>   * Is the project id something only storable / discernible in the cluster context?

The cloud credential json has its project id in it, so when the user creates one inline, I've updated the form to automatically populate the project id input with the value from that JSON blob (we can't do this in other scenarios because that data is not saved on the cloud credential itself). I opted not to automatically authenticate as well in the event that the user wants to use a different project id (honestly don't know how valid that is, our dev gcp access only has one project, but the old ui allows it)


>   * When selecting a `Google` Cloud Credential via Cluster Create page
>     
>     * clicking the `Create` button without going through `Authenticate` gives the error notification `cluster kubernetes version cannot be empty string`. Can we disable the button if the user hasn't authed?
>     * clicking the `Create` button without going through `Authenticate`, and then supplying the project id --> authenticate results in the rest of the form showing however the node pools section is empty
>   * the red notification at top of the create cluster create cloud cred page saying `cloud credential or project id missing` pops up in odd ways
>     
>     * on a system with no cloud creds, when going to the cluster create page it's shown before the user hasn't had a chance to enter a project id, and will then blip in/out when the user enters/clears field

Above bugs should all be resolved now

>   * After selecting a Cloud Cred and clicking on `Authenticate` the button width changes to show the spinner which wiggles the two input field widths

Honestly, I'm not sure what a good alternative would be here. Putting the button on a new line when the input widths are already huge looks weird, too. 

> * Create
>   
~ * I created a cluster by adding just a name, no other changes, and it fails to come up
>     
>     * rancher status `Waiting on gke crd to be initialized`, i don't see any clusters in https://console.cloud.google.com/kubernetes/list/overview?authuser=1&project= (is that the right place?)~

discussed above offline

> * Create (form)
>   
>   * `logging and monitoring services are hardcoded lists with two options, one of which is none: I'm using a checkbox instead of a dropdown.`
>     
>     * Not sure about this suggestion, but should we inform the user of the service names in a tooltip?

I wasn't sure either so I did a little investigating. Googling those service names doesn't turn up much helpful information (nor many direct hits at all) so I'm not sure there would be any value added in showing them to the user. I did however notice that there are a handful of ways that logging and monitoring can be configured in GKE, so I added an info banner clarifying which aspects are enabled with these Rancher options.

Screenshot 2024-06-10 at 4 27 41 PM

>   * Old UI Cluster Options --> Pod secondary Range Name is missing (but Pod Secondary CIDR Block is shown). Is this related to the `getGKESharedSubnetworks` comment?
>     
>     * Same with missing `Services Secondary Range Name`

My form logic was a bit off here. The Pod/Service Secondary Range Name inputs should be shown whenever a subnet is selected; I was showing them only when the subnet also had secondary ranges defined.

>   * New UI Networking --> Advanced Options
>     
>     * Can the checkboxes be vertical instead of horizontal? They all seem to affect each other in some way, and it would tie together `CIDR Block` key value component exposed via `Master Authorized Network`.

Done - also changed the other networking checkboxes to be vertical for consistency:


Screenshot 2024-06-10 at 4 29 21 PM


>   * After filling in the form and clicking `Create` the contents of the Node Pool tabs is briefly shown with no content

Should be fixed now

>   * Node Pools
>     
>     * Old UI `Image Type` has an additional option `Windows Long Term Service Channel with Containerd` (zonal, us-central1-c)

I see this option in the new UI as well - do you perhaps mean `Windows Semi-Annual Channel with Containerd` ? I didn't include that in the new UI as it is deprecated in all versions of kubernetes supported in Rancher 2.9.x ( I did include the translation in in the event that someone edits an ooold GKE cluster in an up-to-date Rancher instance)

>     * API Access Scopes --> `Set access for each API`. We've lost horizontal space in the UI given the tab bar. For me full screen this means only one column of select controls are shown. Given the content of these is small can we shrink them width wise to allow more column/s? Think maybe 2, 3 if they don't look squished?

 I've opted for two columns - I found a 2x9 matrix easier to scan through than 3x6

Screenshot 2024-06-10 at 4 37 09 PM


> * Detail View - Config
>   
>   * This shows a red banner `Network Error` and the cloud creds + authenticate button. Pressing `Authenticate` shows another network error banner. For this use case is there an _easy_ way to avoid the user having to press `Authenticate`?

I'll admit, I forgot about the 'view config' mode in general. I've made a handful of small improvements around this: in addition to hiding the authenticate button, I've prevented any of the gcp api calls from running in view mode (no need to show loading spinners for disabled inputs). I also corrected up the display mode of a few inputs when they're disabled.

> * Edit
>   
>   * Same as detail view config (red banner network error)

I think I fixed this as part of my view config fixes.

> Bonus Point (not related to PR) - Create EKS --> Cluster Options --> shows a Kube Version of `[1.29`. Not sure where this comes from in hosted world, but it also happens in master

There have already been two issues filed about this :( I took a peek and know what's going on, but will add to an EKS-scoped PR.

pkg/gke/components/GKENodePool.vue
mantis-toboggan-md: I added types for a handful of labeledselect option lists here and in networking.

pkg/gke/components/Networking.vue
mantis-toboggan-md: Yeah, that is a little sloppy looking. Logic tweaked slightly from what you suggested because of some changes I made to accomodate the view config mode, but I think I covered the gist of what you're saying.

pkg/gke/components/Networking.vue
mantis-toboggan-md: did ya one better and typed that badboy correctly

pkg/gke/components/Networking.vue
mantis-toboggan-md: Nice catch. Should be fixed now

pkg/gke/components/CruGKE.vue
mantis-toboggan-md: Intellisense thinks this _is_ typed, because `machineTypesResponse` is typed. Is there a benefit to explicitly calling out the type here as well?

Screenshot 2024-06-10 at 1 51 51 PM

I updated typing on `this.errors` and `this.$refs.pools?.select`...`errors` still throws errors in the template - it's an awkward one to type because its coming from a mixin, which simply do not play well with typescript. I looked into it previously and didn't find a nice clean solution for the template errors.

pkg/gke/components/CruGKE.vue
```diff
@@ -0,0 +1,840 @@
...
+      this.addPool();
+    }
+    const registerBeforeHook = this.registerBeforeHook as Function;
+    const registerAfterHook = this.registerAfterHook as Function;
+
+    registerBeforeHook(this.cleanPoolsForSave);
+    registerBeforeHook(this.removeUnchangedConfigFields);
+    registerAfterHook(this.saveRoleBindings, 'save-role-bindings');
+
+    this.debouncedLoadGCPData = debounce(this.loadGCPData, 500);
```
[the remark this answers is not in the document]
mantis-toboggan-md: I think happened here was that I put it into `created()` because that seems to be how we use debounce across the app, and I wanted to stick to the pattern. Then typescript complained at me, so I added a definition to data as well...moving to data would _probably_ be fine.  But I like the consistency with other uses of debounce around the dashboard.

pkg/gke/util/validators.ts
```diff
@@ -0,0 +1,71 @@
...
+    let toValidate = get(ctx.normanCluster, clusterPath);
+
+    if (!toValidate) {
+      toValidate = val || '';
+    }
+    let isValidCIDR = false;
+    let isIpv4 = false;
+
+    try {
+      isValidCIDR = ipaddr.isValidCIDR(toValidate);
```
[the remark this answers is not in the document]
mantis-toboggan-md: It looks like our `isValidCIDR` applies only to ipv4, whereas in this case, we need to check both ipv4 and ipv6 + cidr. I looked into ipv6 regex to do so and they are...quite something. Daunting in length. Bringing in a small library with no deps and a maintainer who appears to be at least paying attention to it if not _frequently_ updating seemed better than copy-pasting a regex I don't comprehend.  Tradeoffs either way, I could be persuaded otherwise.

pkg/gke/components/CruGKE.vue
richard-cox: I think we might have hit the same issue in a previous PR, looked into it a bit more. With this one i don't get the typing for `type` because of the imported type files end with `.d.ts`. Removing those from `types/gcp` and `types/index` solved the issue, however exposed an issues
- line 491 `name:     group,` (name is missing from `GKEMachineTypeOption`)

which just leaves two, both mixins and can be ignored
- pools.select 
- emitOrRoute

richard-cox: there's a couple of other references to .d.ts in the plugin as well

richard-cox (changes requested):
Given the functionality another pass. Almost all issues resolved (i'll start removing them in the next pass)

- Cloud Credentials
  - ~When creating a `Google` Cloud Credential directly via the `Cloud Credential` page no project id is requested, however whilst on the create cloud cred steps when creating a cluster (when no cred found) it is~
    - > The cloud credential json has its project id in it, so when the user creates one inline, I've updated the form to automatically populate the project id input with the value from that JSON data
     - Nice!
  - ~When selecting a `Google` Cloud Credential via Cluster Create page~
    - ~clicking the `Create` button without going through `Authenticate` gives the error notification `cluster kubernetes version cannot be empty string`. Can we disable the button if the user hasn't authed?~
    - ~clicking the `Create` button without going through `Authenticate`, and then supplying the project id --> authenticate results in the rest of the form showing however the node pools section is empty~
  - ~the red notification at top of the create cluster create cloud cred page saying `cloud credential or project id missing` pops up in odd ways~
    - ~on a system with no cloud creds, when going to the cluster create page it's shown before the user hasn't had a chance to enter a project id, and will then blip in/out when the user enters/clears field~
  - After selecting a Cloud Cred and clicking on `Authenticate` the button width changes to show the spinner which wiggles the two input field widths
    - > Honestly, I'm not sure what a good alternative would be here. Putting the button on a new line when the input widths are already huge looks weird, too.
    - We could add something like this below, which gives the area enough room so only the button changes and not forms
      ```
      min-width: 150px;
      display: flex;
      align-items: center;
      justify-content: center;
      ```
  - After adding junk creds the button shows an error state, but with incorrect text colour. Is this something within this PR or a general regression?
    - image

- Create
  - ~I created a cluster by adding just a name, no other changes, and it fails to come up~
- Create (form)
  - ~`logging and monitoring services are hardcoded lists with two options, one of which is none: I'm using a checkbox instead of a dropdown.`~
  - ~Old UI Cluster Options --> Pod secondary Range Name is missing (but Pod Secondary CIDR Block is shown). Is this related to the `getGKESharedSubnetworks` comment?~
  - New UI Networking --> Advanced Options
    - ~Can the checkboxes be vertical instead of horizontal? They all seem to affect each other in some way, and it would tie together `CIDR Block` key value component exposed via `Master Authorized Network`.~
    - Can the `Master Authorized Network` `CIDR Block` control go underneath the checkbox? just to keep them tied together
  - ~After filling in the form and clicking `Create` the contents of the Node Pool tabs is briefly shown with no content~
  - ~Node Pools~
    - ~Old UI `Image Type` has an additional option `Windows Long Term Service Channel with Containerd` (zonal, us-central1-c)~
    - ~API Access Scopes --> `Set access for each API`. We've lost horizontal space in the UI given the tab bar. For me full screen this means only one column of select controls are shown. Given the content of these is small can we shrink them width wise to allow more column/s? Think maybe 2, 3 if they don't look squished?~
- Detail View - Config
  - ~This shows a red banner `Network Error` and the cloud creds + authenticate button. Pressing `Authenticate` shows another network error banner. For this use case is there an _easy_ way to avoid the user having to press `Authenticate`?~
- Edit
  - ~Same as detail view config (red banner network error)~
  - On going to `Edit Config` the first page has some floating buttons
    - image
  - Think this has a similar issue to edit EKS. Existing node pools cannot have their name set (disabled - expected) however new node pools also can't (no node pool name)

~Bonus Point (not related to PR)  - Create EKS --> Cluster Options --> shows a Kube Version of `[1.29`. Not sure where this comes from in hosted world, but it also happens in master~

mantis-toboggan-md:
> We could add something like this below, which gives the area enough room so only the button changes and not forms

...why didn't I think of that? Done.

> After adding junk creds the button shows an error state, but with incorrect text colour. Is this something within this PR or a general regression?
> 

This was caused by using the`role-secondary` class on the authenticate button; I've updated it to use the default 'role' styling (blue with white text).


> Can the Master Authorized Network CIDR Block control go underneath the checkbox? just to keep them tied together

Sure. I played around with a few different arrangements and landed on this (screen recordings are ~1080p window, for context):

https://github.com/rancher/dashboard/assets/42977925/27b81497-480d-49cc-a897-fc6e8899569b



I don't have strong feelings on these conditionally rendered inputs being to the side vs below the checkboxes, but I will note that they don't work well stacked below the checkboxes; one or the other will be fully out out view when enabling private cluster:

https://github.com/rancher/dashboard/assets/42977925/cdeed4d7-498e-40a9-a462-d5fc50abcc48


I also updated the text on the add button to align better with what that keyvalue component corresponds to in the GCP ui.

> Console in error when going to page
>```
> entry-helpers.js:53 TypeError: Cannot read properties of undefined (reading 'createSubnetwork')
> at Proxy.render (CruGKE.vue:465:1)
>```

Fixed by adding the `Loading` component

> On going to Edit Config the first page has some floating buttons

I've added a fix to all 3 hosted providers. SelectCredential rendered before a `value` prop was provided and did not update itself accordingly...I 'fixed' this by adding a loading component to CruEKS CruGKE CruAKS, so SelectCredential doesn't render until cluster credential data has been fetched. I think SelectCredential is not handling the value prop correctly, but these 3 components should use the loading indicator anyway. 

> Think this has a similar issue to edit EKS. Existing node pools cannot have their name set (disabled - expected) however new node pools also can't (no node pool name)

I'm not seeing this issue in GKE - new node pools in a provisioned cluster can have their name altered; existing pools can have their name altered only if the cluster has not finished provisioning. I have however moved 'group details' to the top of the node tab, to be consistent with the other two providers.

mantis-toboggan-md:
> Scale down failed with Cannot read properties of undefined (reading 'includes'). This points to somewhere that isn't hit when adding a breakpoint to and trying again (handleSpoofedRequest

I don't think that button should be shown at all for kontainer driver clusters: these are node groups but they're being shown as individual nodes that can be scaled down individually. I don't think this PR introduced the bug; I see the issue on master as well. I _dont_ see the bug in 2.8-head however so it is some kinda regression. I'll keep investigating next week, but would prefer to tackle this in a separate PR since it's a shared component and the problem exists across kontainer driver clusters. https://github.com/rancher/dashboard/issues/11246

pkg/gke/components/CruGKE.vue
richard-cox: Given the update to my now standard vs code extension i don't get these

richard-cox (changes requested):
Just some new (in review, not regressions given pr changes) edit items remaining

- Cloud Credentials
  - Nowt
- Create (form)
  - Nowt
- Detail View - Config
  - Nowt
- Detail View - Node list
  - Nowt
- Edit
  - ~Think this has a similar issue to edit EKS. Existing node pools cannot have their name set (disabled - expected) however new node pools also can't (no node pool name)~
    - No longer seeing, or just dreamt it
  - Copying in the project id every time get's a bit tiresome. Is there anything we can do to improve this? 
    - does it exist anywhere in the cluster (i think you confirmed already it's not in the cloud cred)?
    - would we be able to provide a list for the user to select?
  - Going to edit a single node pool cluster (with default/highest kube version), add a new node pool, Node Details kube version is replaced by check box with missing data
    - `Upgrade the node pool version from to 1.29.5-gke.1121000`
    - Not sure on the mechanism, but for the no-op world we should hide the checkbox (but show an empty section?)
  - existing values of gkeConfig.locations are ignored
    - cluster in region europe-west1. in edit page only europe-west1-b is shown selected (even when gkeConfig.locations contains 3 entries "europe-west1-b", "europe-west1-c", "europe-west1-d"). 'detail view - config' no extra zones are shown.

richard-cox (dismissed):
Just some new (in review, not regressions given pr changes) edit items remaining


- Edit
  - Copying in the project id every time get's a bit tiresome. Is there anything we can do to improve this? 
    - Not sure what was happening before, but i see this there on edit now. for create it's fine leaving out for the mo
  - ~Going to edit a single node pool cluster (with default/highest kube version), add a new node pool, Node Details kube version is replaced by check box with missing data~
  - ~existing values of gkeConfig.locations are ignored~

mantis-toboggan-md:
> * Edit
>  
>   * Copying in the project id every time get's a bit tiresome. Is there anything we can do to improve this?
>     
>     * does it exist anywhere in the cluster (i think you confirmed already it's not in the cloud cred)?
>     * would we be able to provide a list for the user to select?
>   * Going to edit a single node pool cluster (with default/highest kube version), add a new node pool, Node Details kube version is replaced by check box with missing data

Thanks for reminding me; I've filed a backend issue https://github.com/rancher/rancher/issues/45880 hopefully this is something we can add for 2.10... It should be possible to list projects through the gcp API, but we're using a subset of available gcp api options through the norman UI (similar to AKS) so we need support added there. There's not much more we can do on the UI side without that.

richard-cox (approved):

PR #11026, opened 2024-05-14
Update the plugin inject method to accept a Vue instance

### LINKED ISSUE #11028: [Vue3] Refactor `inject()` function
The `inject()` function defined within `shell/initialize/index.js` has a few patterns that will make it difficult to utilize this method in Vue3. These patterns include:

- Using a global Vue instance to install plugins (the concept of a global Vue instance to modify is gone in Vue3)
- Mutating app and context outside of function scope (not a major problem, but can make refactoring difficult)

rak-phillip:
### Summary
This moves `inject()` into `shell/initialize/plugins.js` helps to better isolate the scoping of variables that are mutated by inject. This change also helps us to better prepare for utilizing `inject()` in Vue3, as the function now accepts a Vue instance. Finally, `inject()` is conceptually related to plugin installation, so it makes more sense to have this function more closely located to the root of the plugin install code.

Fixes #11028

Refactors the `inject()` function:

- Move from `shell/initialize/index.js` to `shell/initialize/plugins.js`
- Add the context and vue instance as parameters to `inject()`
- Refactor plugins that invoke `inject()` to pass the context and vue instance as arguments
- Update relative imports to use alias instead
 
### Technical notes summary

I considered a few alternatives before landing on this approach.

1. Creating a Vue plugin that would handle the work of inject and registering global properties. I ran into an issue with caching behavior of Vue plugins [^1]
   > Vue.use automatically prevents you from using the same plugin more than once, so calling it multiple times on the same plugin will install the plugin only once.
2. Creating a Vue plugin that accepts an array of plugins to inject. This was problematic because some plugin definitions require other plugins to be previously registered before they can be configured.. There are ways around this, but it was complicating the implementation in ways that wouldn't improve the current situation.
3. Writing a generic inject function that can be imported into plugins that want to be injected into he app context/store. This would work fine, but required more refactoring than what is present in this PR 

[^1]: https://v2.vuejs.org/v2/guide/plugins.html#Using-a-Plugin

- Application initialization
- Loading Rancher Extensions

- Application initialization
- Loading Rancher Extensions

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/initialize/index.js
+++ b/shell/initialize/index.js
@@ -38,7 +38,8 @@ loadDirectives(Vue);
 Vue.component(NuxtChild.name, NuxtChild);
 Vue.component('NChild', NuxtChild);
 
-async function createApp(config = {}) {
+async function createApp() {
+  const config = nuxt.publicRuntimeConfig;
   const router = await createRouter(config);
 
   const store = createStore();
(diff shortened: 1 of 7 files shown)


shell/initialize/plugins.js
```diff
@@ -52,3 +53,62 @@ export async function installPlugins(app, inject, Vue) {
...
+ * @param {object} Vue The Vue instance
+ * @returns {void}
+ *
+ * @property {object} context.app The app object within the context
+ * @property {object} context.store The store object within the context
+ *
+ * Note: This function mutates the context object, including the context app and
+ * store.
+ */
+function inject(key, value, context, Vue) {
```
codyrancher, line 74: Because the context and vue instance are both present in this app, would it instead make sense to pass and inject function to the plugins that already have these arguments captures and allow the plugin to just pass the key and value?

Do we want the plugins to be able to change the instance and context?

codyrancher (approved):

shell/initialize/plugins.js
cnotv: Yeah this is something I was thinking as well, to just export something to be imported in the entry file or expose the instance (and config?) 
I would find more easy to finally install, import and integrate anything inside the same place, plus I am not so sure how much is easy the other way around.

shell/initialize/index.js
```diff
@@ -38,7 +38,9 @@ loadDirectives(Vue);
 Vue.component(NuxtChild.name, NuxtChild);
 Vue.component('NChild', NuxtChild);
 
-async function createApp(config = {}) {
+async function createApp() {
+  // eslint-disable-next-line no-undef
+  const config = nuxt.publicRuntimeConfig;
```
cnotv, line 43: I bumped my head multiple times into this and I wondered: do we know where this `nuxt` is coming from?

cnotv (approved):
LGTM

shell/initialize/index.js
rak-phillip: Yeah, it's solely supplied by an environment variable. Instead of passing the config around, I thought it to be more clear to simply access the environment viariable when we require it.

codyrancher (approved):

PR #11037, opened 2024-05-15
Fix AKS cluster name and node pool count validation

### LINKED ISSUE #9416: AKS: Unable to create cluster with underscore in name
**Setup**
- Rancher version: `v2.7-67422772955d5830b237fe84dbf1a26d266b55c1-head`
- Browser type & version: Chrome Latest

**Describe the bug**
AKS: Unable to create cluster with underscore in name. The same is valid as per AKS Docs

**To Reproduce**
Create AKS cluster via Rancher UI with "_" in name eg. rancher_test

**Expected Result**
User must be able to create AKS cluster with underscore in name as it is supported
image

**Screenshots**
image

cc: @rancher/highlander

### LINKED ISSUE #10963: [v2.9] AKS: Nodepool values validation
**Setup**
Rancher version: `v2.9-6d87a11ea46b7571646d7c3d7af704584c39fd62-head`
Browser type & version: Chrome latest

**Describe the bug**
AKS (Vue UI): 
Nodepool values on UI like Count, Max surge need validation for nil, 0 value, as in old UI

**To Reproduce**
Provision AKS cluster from Rancher UI with Nodepool count 0 and max surge empty

**Screenshots**
image

mantis-toboggan-md:
### Summary
Fixes #9416 
Fixes #10963 

This PR updates AKS provisioning form validation:
* allow underscores in cluster name (not at beginning nor end of name)
* node count must be at least 1

if autoscaling is enabled:
* minimum nodes must not be greater than maximum nodes
* maximum nodes must not be less than minimum nodes
* minimum and maximum nodes must both be greater than 0 and at most 100
Additionally, I noticed that if I configured a node count <0 and enabled autoscaling, the aks api returned an error stating that node count must not be less than minimum nodes nor greater than maximum nodes. In Ember the node count input is disabled when autoscaling is enabled. Given that there still appear to be restrictions on node count when autoscaling is enabled, I have updated the AKS vue ui to keep the node count input enabled regardless of autoscaling.

### Technical notes summary
I also refactored node pool validation a little to make it a bit simpler to show the error icon in the tab title.

### Checklist
- [x] The PR has a Milestone 


mantis-toboggan-md:
I updated the PR to remove the aks node pool name validation unit test. With this refactor, the name input is now using the form validation mixin more conventionally and the mixin doesn't need to be retested here...the pool name validator function is covered by a test in the cruaks unit test suite

pkg/aks/components/AksNodePool.vue
```diff
@@ -215,7 +224,7 @@ export default defineComponent({
           type="number"
           :mode="mode"
           label-key="aks.nodePools.count.label"
-          :disabled="pool.enableAutoScaling"
+          :rules="validationRules.count"
```
momesgin, line 227: would it be a good idea to also add `min="1"` here?

mantis-toboggan-md: Good call. I added min and max attributes to node count, and node min/max inputs. It looks like that attribute stops the user from lowering the value below that point with the arrow buttons but its still possible to manually type in a lower number, so I have left in the form validation rules.

momesgin: You're right, it's also possible for the user to inspect the page's html and manually remove the attribute, but I think it's ok since we're just improving the ux here, and as you mentioned we already have the form validation in place.

momesgin (approved):
lgtm

PR #11045, opened 2024-05-16
Prevent errors in existing cases binding props with v-model

### LINKED ISSUE #11029: Prevent errors in existing cases binding props with v-model
# Description

Investigate issues printed out by the Vue3 builder about the use of props in `v-model`.

## Context

Since Vue3, it is not possible to rebind props with v-model. Compiler error: 

```
VueCompilerError: v-model cannot be used on a prop, because local prop bindings are not writable.
Use a v-bind binding combined with a v-on listener that emits update:x event instead.
```

The error is generated from `vue/core` and there are at the moment 54 occurrences.
Specifically to this codebase, it happens mostly when we define in the markup `v-model="value"`,  at the moment 86 occurrences.

E.g.:

Screenshot 2024-05-14 at 20 16 19

**Please note**: 
- In Vue2 props bound with `v-model` are defined as `value`, while in Vue3 as `modelValue`
- The error is generated exclusively on reassignment, so do not confuse the term "prop mutation" with "object mutation"

## Solution

In the current phase, the only solution seems to replace the prop with a computed property. This can be automated and solved by using a mixin that would add prop `modelValue` and computed `value`, then remove the prop `value` ([example PR](https://github.com/rak-phillip/vue-3-v-model/pull/1/files)).

Using existing migration script, we would need to remove all the occurrences where we define `value` as a prop, including mixins and components in .js format.

## References

- PoC:
  - https://codesandbox.io/s/vue3-vuex-custom-hook-yjghfv?file=/src/components/HelloWorld.vue
  - https://github.com/rak-phillip/vue-3-v-model/
- Docs: https://vuejs.org/guide/components/props.html#one-way-data-flow
- Linter: https://eslint.vuejs.org/rules/no-mutating-props.html

## Collaterals

As in Vue3 the `v-model` binding is generating a `modelValue` instead of `value`, we'll need to apply this mixing everywhere.

cnotv:
### Summary
Fixes #11029

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/mixins/model-value.js
+++ b/shell/mixins/model-value.js
@@ -1,0 +1,27 @@
+/**
+ * Allow to maintain current 2 way binding behavior for v-model in Vue 3.
+ * In Vue 3, props mutation is not allowed since they are read-only.
+ * v-model="foo" is short for :model-value="foo" @update:model-value="foo = $event"
+ * Docs: https://vuejs.org/guide/components/props.html#one-way-data-flow
+ * Linter: https://eslint.vuejs.org/rules/no-mutating-props.html
+ * IMPORTANT: This mixin is temporary and the logic should be replaced with something to be defined in the future.
+ */
+export default {
+  props: {
+    modelValue: {
+      type:     Object,
+      required: true,
+      default:  () => {}
+    }
+  },
+  computed: {
+    value: {
+      get() {
+        return this.modelValue;
+      },
+      set(value) {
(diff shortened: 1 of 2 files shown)


cnotv:
Closing in favor of a better approach in https://github.com/rancher/dashboard/pull/11597

cnotv:
Closing in favor of a better approach in https://github.com/rancher/dashboard/pull/11597

PR #11082, opened 2024-05-21
Add e2e tests for network policy

### LINKED ISSUE #10641: NetworkPolicy pod selectors are not displayed properly "Matches 0 of 0 pods"
Internal reference: SURE-7871
Reported in 2.7.9 and reproduced in 2.8-head

Issue description:
When UI is used to define NetworkPolicy, the whole dialogue behaves incorrectly for Pod and Namespace selectors in Ingress rules. Once the rule is saved, it shows "Matches 0 of 0 pods" while just before saving it was showing different numbers.

In "Create view", the display works correctly but once the policy is created the UI is broken. Policy YAML looks fine and works, so it's only a UI issue.

With multiple rules, the problem replicates with each rule.

Impact:
Defining NetworkPolicy in UI doesn't work as expected and may lead to incorrect decisions.

Troubleshooting steps:
After creating the following (YAML or UI) in cluster/Network Policies, this sample was used when trying to edit the ruleset:

```
---
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: httpd
  name: httpd
  namespace: webapp
spec:
  replicas: 3
  selector:
    matchLabels:
      app: httpd
  template:
    metadata:
      labels:
        app: httpd
    spec:
      containers:
      - image: httpd:2.4
        imagePullPolicy: IfNotPresent
        name: httpd
---
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: test
  namespace: webapp
spec:
  ingress:
    - from:
      - podSelector: {}
  podSelector: {}
  policyTypes:
    - Ingress
```

The editing page was showing matching object count but after saving it changed to "Matches 0 of 0 pods" and the list was only showing header column names Key/Operator/Value/In.

When editing:

When saving:

By changing the Rule type and operator, odd errors were being reported with combinations where an empty ruleset should be just saved.

The error was replicating with more rules added. Even when there was no error, multiple rules produced the same "Matches 0 of 0 pods" which is not true. 

Repro steps:

- Create deployment (see above)
- Create network policy
- Edit network policy/Ingress ruleset
- Save

Actual behavior:
In the green box, "Matches 0 of 0 pods" is there. There is a table below the green box which is something like this: Key Operator Value In

Expected behavior:
In the network policy UI (`/dashboard/c/local/explorer/networking.k8s.io.networkpolicy/webapp/test#rule-ingress0`) there is a green box. In the green box, `"Matches 3 of 3 pods, including ..."` should be there. The table below the green box should either say "No rules defined" or disappear. For multiple rules, the pod count in each green box should match according to each rule.

momesgin:
### Summary
Fixes #10641 

### Checklist
- [ ] The PR has a Milestone 


izaac (approved):
:shipit:

PR #11131, opened 2024-05-28
Improve keyboard mapping display

### LINKED ISSUE #11092: Replace 'Normal Human' with keyboard icon to be less intrusive
We made an update to show the keyboard mapping for the text editor (e.g. yaml editor).

Our default is called 'Norman Human' - I find it very distracting and intrusive, e.g.:
image

I would suggest we replace 'Norman Human' with a keyboard icon and a tooltip showing the current mapping - and I would suggest we ONLY show the icon when the mapping is something other than the default - the whole point of showing it was for people who had changed it to 'vim' and didn't appreciate that the keyboard would behave differently.

So - in the normal case, we would not show anything.

@edenhernandez-suse ^^ does this seem reasonable?

nwmac:
### Summary
Fixes #11092

This PR fixes the above issue.

Instead of always showing the current keyboard mapping overlay on the yaml editor, we only show it when set to something other that the default (i.e. Emacs or Vim). The default is called 'Norman Human', so showing it just looks weird.

When shown now, the display has been changed to only show an icon with a tooltip, so that even when shown, it is less intrusive.

Was:

image

Now:

image

With the exiting UI/UX, the close icon appears and the label jolts as a result. With this change, subtle animations have been applied to make this is a nicer transition that is less jarry.

Automated tests are included.

Video before:
Keyboard_mapping_existing

Video with this change:
Keyboard_mapping_new

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/cypress/e2e/po/components/keyboard-mapping-indicator.po.ts
+++ b/cypress/e2e/po/components/keyboard-mapping-indicator.po.ts
@@ -7,7 +7,7 @@ export default class KeyboardMappingIndicatorPo extends ComponentPo {
 
   showTooltip(): Cypress.Chainable {
     return this.self().trigger('mouseenter');
-  }  
+  }
 
   getTooltipContent(): Cypress.Chainable {
     return cy.get('.tooltip.vue-tooltip-theme .tooltip-inner');
--- a/cypress/e2e/po/components/keyboard-mapping-indicator.po.ts
+++ b/cypress/e2e/po/components/keyboard-mapping-indicator.po.ts
@@ -1,0 +1,15 @@
+import ComponentPo from '@/cypress/e2e/po/components/component.po';
+
+export default class KeyboardMappingIndicatorPo extends ComponentPo {
+  constructor() {
+    super('[data-testid="code-mirror-keymap"]');
+  }
+
+  showTooltip(): Cypress.Chainable {
+    return this.self().trigger('mouseenter');
+  }  
+
+  getTooltipContent(): Cypress.Chainable {
+    return cy.get('.tooltip.vue-tooltip-theme .tooltip-inner');
+  }
+}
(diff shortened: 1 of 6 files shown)


nwmac:
@edenhernandez-suse Can you review the videos above and let me know if you are okay with this change - I think its better than what we have.

edenhernandez-suse:
It's an improvement indeed. A couple of small remarks:

- I saw positioning changed from sticky to absolute, I can see a scenario where the YAML file is large enough to need scrolling, in that case I assume the absolute positioned overlay will go out of sight, am I wrong? I can't check it in our environment. Maybe the sticky position would work better (it will be visible after scrolling)

- The animation might look better than it seems to be in the video, but I would still reduce the timing to .15s for such a short interaction. No big deal, if you want to merge it already, I’m fine with it. At some point, I will write a transition/animation guide for the team so we can do this consistently.

nwmac:
@edenhernandez-suse I updated the animation time. On the scroll issue, it is fine because the absolute position is with respect to the nearest positioned element in the hierarchy, so the icon will move as the page scrolls.

edenhernandez-suse:
> the icon will move as the page scrolls
Is there a way to see that working? I can't use your branch in our environment.
Also, do you need my review approval to unblock the merge or are you waiting for another dev?

nwmac:
@edenhernandez-suse "just" need a review from Alex, so you're not blocking this.

Here's the scroll behaviour:

Screen Recording 2024-06-03 at 09 40 29

aalves08 (approved):
LGTM

Screenshot 2024-06-03 at 09 49 06

edenhernandez-suse:
> @edenhernandez-suse "just" need a review from Alex, so you're not blocking this.
> 
> Here's the scroll behaviour:
> 

Yes that's what I meant with the change, I think I didn't express it properly, the icon stays in an absolute position and "goes away" when scrolling. With the sticky behaviour, it would be always visible. I prefer to keep it always visible, but it's not a critical detail, especially if you merged it already.

PR #11164, opened 2024-06-03
improve shell api gating

### LINKED ISSUE #11163: Improve gating mechanism for extensions based on Extensions API version
With the upcoming release of Rancher we need to update the current mechanisms we have in place for Extensions loading so that they work correctly with `UI_EXTENSIONS_API_VERSION` version number.

In parallel, the `creators/pkg/init` script needs to be updated to include the latest `UI_EXTENSIONS_API_VERSION` version number as an annotation `catalog.cattle.io/ui-extensions-version`

aalves08:
### Summary
Fixes #11163 

- updated const name in `uiplugins.js` for better context
- updated path in `uiplugins.js` for fetching `UI_EXTENSIONS_API_VERSION` from the correct property in `UIPlugin` recource
- Update `creators/pkg` to include latest `UI_EXTENSIONS_API_VERSION` as the correct annotation for newly created pkg's

- run the init script locally: `./shell/creators/pkg/init test`

### Checklist
- [ ] The PR has a Milestone 

### DIFF
--- a/shell/creators/pkg/init
+++ b/shell/creators/pkg/init
@@ -3,6 +3,7 @@
 const fs = require('fs-extra');
 const path = require('path');
 const https = require('https');
+var semver = require('semver')
 
 const targets = {
   dev:  './node_modules/.bin/nuxt dev',
@@ -131,6 +132,21 @@ function fetchLatestVersion() {
           console.log(`  Adding rancher-version annotation '>= ${ release.tag_name }' to package.json`);
 
           pkg.rancher = { annotations: { 'catalog.cattle.io/rancher-version': `>= ${ release.tag_name }` } };
+
+          const extensionsApiVersionText = fs.readFileSync(path.join(__dirname, '../../config/extensions-api-version.js'), { encoding: 'utf8' });
+          const extensionsApiVersion = extensionsApiVersionText.substring(
+            extensionsApiVersionText.indexOf("'") + 1, 
+            extensionsApiVersionText.lastIndexOf("'")
+          );
+
+          if (!semver.valid(extensionsApiVersion)) {
+            console.log('Wrong format for version number for Extensions API', semver.valid(extensionsApiVersion));
+          } else {
+            console.log(`  Adding catalog.cattle.io/ui-extensions-version '>= ${extensionsApiVersion }' to package.json`);
+            
+            pkg.rancher.annotations['catalog.cattle.io/ui-extensions-version'] = `>= ${ extensionsApiVersion }`;
+          }
+
           writePackageJson();
         }
       } catch (e) {
--- a/shell/creators/pkg/init
+++ b/shell/creators/pkg/init
@@ -5,8 +5,6 @@ const path = require('path');
 const https = require('https');
 const semver = require('semver');
 
-const UI_EXTENSIONS_API_VERSION = '1.2.0';
-
 const targets = {
   dev:  './node_modules/.bin/nuxt dev',
   nuxt: './node_modules/.bin/nuxt',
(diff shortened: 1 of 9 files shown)


aalves08:
@nwmac I had to duplicate the const `UI_EXTENSIONS_API_VERSION`  because the logic I had before was tied to a file that doesn't exist in the context of the `creators/pkg` npm lib. I just realised this, sorry, it's different than the feedback I had provided earlier. 

I ran out of ideas to fix this in a more intelligent way 🙏

shell/creators/pkg/init
```diff
@@ -3,6 +3,9 @@
 const fs = require('fs-extra');
 const path = require('path');
 const https = require('https');
+const semver = require('semver');
+
+const UI_EXTENSIONS_API_VERSION = '1.2.0';
```
nwmac, line 8: Probably a bad idea to have this in two places.

I think this will be the same version as the version in `shell/package.json` - so reading it from there in both places in this PR would be preferable.

nwmac (changes requested):

jordojordo (changes requested):

shell/creators/pkg/init
```diff
@@ -135,12 +133,28 @@ function fetchLatestVersion() {
...
-            
-            pkg.rancher.annotations['catalog.cattle.io/ui-extensions-version'] = `>= ${ UI_EXTENSIONS_API_VERSION }`;
+            const splitShellVersion = latestShellVersion.split('.');
+
+            if (splitShellVersion.length <= 1) {
+              console.log('Wrong semver usage/pattern for shell version from dependencies on base package.json. Must follow x.x.x');
+            } else {
+              // use only major and minor version from the shell pkg version
+              const majorVersion = splitShellVersion[0].includes('^') ? splitShellVersion[0].substring(1, 100) : splitShellVersion[0];
+              const minorVersion = splitShellVersion[1];
```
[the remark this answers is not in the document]
aalves08: Unfortunately `semver.valid(latestShellVersion)` doesn't consider valid shell pkg notations such as `^1.2.0` or anything with special chars so we cannot use it. I've improved this part

shell/creators/pkg/init
```diff
@@ -131,6 +144,25 @@ function fetchLatestVersion() {
...
+          const shellPkg = JSON.parse(shellPkgRawData);
+          const latestShellVersion = shellPkg.version || null;
+
+          if (!latestShellVersion) {
+            console.log('Could not get a shell version from node_modules/@rancher/shell/package.json in app');
+          } else {
+            const splitShellVersion = latestShellVersion.split('.');
+            const majorVersion = splitShellVersion[0];
+            const minorVersion = splitShellVersion[1];
+            const parsedShellVersion = `${majorVersion}.${minorVersion}.0`;
```
jordojordo, line 159: Sorry if you've already mentioned this, but what's the reasoning behind the patch version always being equal to `0`?

jordojordo (commented):
One small comment, otherwise looks good to me.

shell/creators/pkg/init
aalves08: Because we should only employ breaking changes on the major version. Feature(s) might be related to minor versions but patch versions should not have any impacts at all, therefore we always start them at `x.x.0`

aalves08:
@nwmac @jordojordo requesting final approval for this PR. Latest change was the bumping of the shell pkg version to `2.0.0` in order to comply to our versioning document regarding breaking changes for the upcoming `2.9` release 🙏

nwmac (approved):

jordojordo (approved):

PR #11165, opened 2024-06-03
Pulling out authentication code into router navigation guards

codyrancher:
### Summary
Pulling out authentication code into router navigation guards

### Technical notes summary
I chose to migrate the code into a beforeEach navigation guard because it is the closest we have to maintaining the same execution timeline as the middleware. (Middleware is actually called from a beforeEach with the entry-helpers.js render method.

I do think it's a slightly kludgey implementation because we don't want it to execute for the `unauthenticated` routes. We may want to move this into a mixin at some point.

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/cypress/e2e/po/lists/management.cattle.io.user.po.ts
+++ b/cypress/e2e/po/lists/management.cattle.io.user.po.ts
@@ -6,7 +6,7 @@ import BaseResourceList from '@/cypress/e2e/po/lists/base-resource-list.po';
  */
 export default class MgmtUsersListPo extends BaseResourceList {
   create() {
-    return this.masthead().actions().eq(0).click();
+    return cy.get('[data-testid="masthead-create"]').click();
   }
 
   refreshGroupMembership(): AsyncButtonPo {
--- a/cypress/e2e/tests/pages/users-and-auth/roles.spec.ts
+++ b/cypress/e2e/tests/pages/users-and-auth/roles.spec.ts
@@ -11,10 +11,11 @@ const usersPo = new UsersPo('_');
 const userCreate = usersPo.createEdit();
 const sideNav = new ProductNavPo();
 
-const runTimestamp = +new Date();
-const runPrefix = `e2e-test-${ runTimestamp }`;
 const downloadsFolder = Cypress.config('downloadsFolder');
-const globalRoleName = `${ runPrefix }-my-global-role`;
+
+let runTimestamp;
+let runPrefix;
+let globalRoleName;
 
 describe('Roles', { tags: ['@usersAndAuths', '@adminUser'] }, () => {
   beforeEach(() => {
@@ -23,6 +24,12 @@ describe('Roles', { tags: ['@usersAndAuths', '@adminUser'] }, () => {
   });
 
   it('can create a Global Role', () => {
+    // We want to define these here because if this test fails after it created the global role all subsequent
+    // retries will reference the wrong global-role because a second roll will with the same name but different id will be created
+    runTimestamp = +new Date();
+    runPrefix = `e2e-test-${ runTimestamp }`;
+    globalRoleName = `${ runPrefix }-my-global-role`;
+
     // create global role
     const fragment = 'GLOBAL';
 
--- a/shell/config/router/navigation-guards/attempt-first-login.js
+++ b/shell/config/router/navigation-guards/attempt-first-login.js
@@ -1,0 +1,71 @@
+import { SETUP } from '@shell/config/query-params';
+import { SETTING } from '@shell/config/settings';
+import { MANAGEMENT, NORMAN } from '@shell/config/types';
+import { tryInitialSetup } from '@shell/utils/auth';
+import { routeNameMatched } from '@shell/utils/router';
+
+export function install(router, context) {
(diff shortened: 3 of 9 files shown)


cypress/e2e/po/lists/management.cattle.io.user.po.ts
```diff
@@ -6,7 +6,7 @@ import BaseResourceList from '@/cypress/e2e/po/lists/base-resource-list.po';
  */
 export default class MgmtUsersListPo extends BaseResourceList {
   create() {
-    return this.masthead().actions().eq(0).click();
+    return cy.get('[data-testid="masthead-create"]').click();
```
codyrancher, line 9: We already use this same selector in `resource-list-masthead.po.ts`. 

Our `role.spec.ts` was sometimes failing to click the create button and I'm not entirely sure why. I noticed that the Refresh Group Memberships button pops in so I think that could be related.

cypress/e2e/tests/pages/users-and-auth/roles.spec.ts
codyrancher, line 1: I wanted to improve the robustness of this test.  The `can create a Global Role` test would fail in different places retries explained a bit more in the code comment below.

cypress/e2e/tests/pages/users-and-auth/roles.spec.ts
```diff
@@ -11,10 +11,11 @@ const usersPo = new UsersPo('_');
...
 const sideNav = new ProductNavPo();
 
-const runTimestamp = +new Date();
-const runPrefix = `e2e-test-${ runTimestamp }`;
 const downloadsFolder = Cypress.config('downloadsFolder');
-const globalRoleName = `${ runPrefix }-my-global-role`;
+
+let runTimestamp;
+let runPrefix;
+let globalRoleName;
```
codyrancher, line 18: I left these where they were but I can push them into the describe if desired.

cypress/e2e/tests/pages/users-and-auth/roles.spec.ts
```diff
@@ -23,6 +24,12 @@ describe('Roles', { tags: ['@usersAndAuths', '@adminUser'] }, () => {
   });
 
   it('can create a Global Role', () => {
+    // We want to define these here because if this test fails after it created the global role all subsequent
+    // retries will reference the wrong global-role because a second roll will with the same name but different id will be created
+    runTimestamp = +new Date();
+    runPrefix = `e2e-test-${ runTimestamp }`;
+    globalRoleName = `${ runPrefix }-my-global-role`;
```
codyrancher, line 31: I wanted these to just be const within the scope of the describe  but describe scope wasn't rerunning on retry.

shell/config/router/navigation-guards/index.js
codyrancher, line 1: I couldn't decide if we wanted this in initialize or in config. If there's a preference to move this to initialize let me know.

shell/config/router/navigation-guards/attempt-first-login.js
```diff
@@ -0,0 +1,71 @@
...
+import { routeNameMatched } from '@shell/utils/router';
+
+export function install(router, context) {
+  router.beforeEach((from, to, next) => attemptFirstLogin(from, to, next, context));
+}
+
+export async function attemptFirstLogin(from, to, next, { store }) {
+  if (routeNameMatched(to, 'unauthenticated')) {
+    return next();
+  }
```
codyrancher, line 14: Ensures we only run when in a route that uses the 'authenticated' parent route.

I have a feeling we're going to want to move this to a mixin at some point. As of right now running this code in a route beforeEach() is the closest in the execution timeline to middleware.

shell/config/router/navigation-guards/attempt-first-login.js
```diff
@@ -0,0 +1,71 @@
...
+  try {
+    const res = store.getters['management/byId'](MANAGEMENT.SETTING, SETTING.FIRST_LOGIN);
+    const plSetting = store.getters['management/byId'](MANAGEMENT.SETTING, SETTING.PL);
+
+    firstLogin = res?.value === 'true';
+
+    if (!initialPass && plSetting?.value === 'Harvester') {
+      initialPass = 'admin';
+    }
+  } catch (e) {
```
codyrancher, line 29: This is a part of the original implementation. I think we'll want to fix this if we get around to breaking this apart further.

shell/config/router/navigation-guards/load-initial-settings.js
```diff
@@ -0,0 +1,13 @@
+import { fetchInitialSettings } from '@shell/utils/settings';
+
+export function install(router, context) {
+  router.beforeEach((from, to, next) => loadInitialSettings(from, to, next));
+}
+
+export async function loadInitialSettings(from, to, next) {
+  try {
+    await fetchInitialSettings();
+  } catch (ex) {}
```
codyrancher, line 10: The original implementation executed fetchInitialSettings() with a try catch which swallowed and ignored the exception. I think we should consider fixing when we have time to dig deeper.

rak-phillip (approved):
LGTM - I agree with the approach of lifting our existing middleware code and dropping it into a router guard. Everything appears to behave as expected.

PR #11172, opened 2024-06-04
Storybook: Fix i18n

### LINKED ISSUE #11171: Storybook is broken for components that use i18n
You'll get an error like this:

image

This is probably due to the re-factor for the i18n plugin as part of the Vue 3 work.

nwmac:
### Summary
Fixes #11171 

Fixes the way i18n is used in Storybook so that the pages are not broken.

Also disables telemetry.

Issue is marked as QA/None - PR reviewer to validate locally on built Storybook - this should be sufficient - this is for the Storybook, not code that makes it into the product UI.

### DIFF
--- a/storybook/.storybook/main.js
+++ b/storybook/.storybook/main.js
@@ -20,6 +20,10 @@ module.exports = {
     buildStoriesJson: true, // 👈 Enable this to build the stories.json file
   },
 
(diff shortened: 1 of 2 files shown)


aalves08 (approved):
LGTM

PR #11186, opened 2024-06-05
Extracted the authentication portion of the authenticated middleware into a navigation guard

codyrancher:
### Summary
Extracted the authentication portion of the authenticated middleware into a navigation guard

### Technical notes summary
This is the closest analogue for getting authentication to be executed at the same point in time it was when it was in middleware.

I did verify this works for both local and github authentication providers.

All authenticated area

All authenticated area

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/config/router/navigation-guards/authentication.js
+++ b/shell/config/router/navigation-guards/authentication.js
@@ -1,0 +1,63 @@
+import { routeRequiresAuthentication } from '@shell/utils/router';
+import { isLoggedIn, notLoggedIn, noAuth, findMe } from '@shell/utils/auth';
+
+export function install(router, context) {
+  router.beforeEach((to, from, next) => authenticate(to, from, next, context));
+}
+
+export async function authenticate(to, from, next, { store }) {
+  if (!routeRequiresAuthentication(to)) {
+    return next();
+  }
+
+  if ( store.getters['auth/enabled'] !== false && !store.getters['auth/loggedIn'] ) {
+    // `await` so we have one successfully request whilst possibly logged in (ensures fromHeader is populated from `x-api-cattle-auth`)
+    await store.dispatch('auth/getUser');
+
+    const v3User = store.getters['auth/v3User'] || {};
+
+    if (v3User?.mustChangePassword) {
+      return next({ name: 'auth-setup' });
+    }
+
+    // In newer versions the API calls return the auth state instead of having to make a new call all the time.
+    const fromHeader = store.getters['auth/fromHeader'];
+
+    if ( fromHeader === 'none' ) {
+      noAuth(store);
+    } else if ( fromHeader === 'true' ) {
+      const me = await findMe(store);
+
+      isLoggedIn(store, me);
+    } else if ( fromHeader === 'false' ) {
+      notLoggedIn(store, next, to);
+
+      return;
+    } else {
+      // Older versions look at principals and see what happens
+      try {
+        const me = await findMe(store);
+
+        isLoggedIn(store, me);
+      } catch (e) {
+        const status = e?._status;
+
+        if ( status === 404 ) {
+          noAuth(store);
+        } else {
+          if ( status === 401 ) {
+            notLoggedIn(store, next, to);
+          } else {
+            store.commit('setError', { error: e, locationError: new Error('Auth Middleware') });
+          }
+
+          return;
+        }
+      }
+    }
+
+    store.dispatch('gcStartIntervals');
+  }
+
+  next();
+}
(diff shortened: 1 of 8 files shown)


shell/utils/router.js
```diff
@@ -89,14 +89,13 @@ export const routeMatched = (to, fn) => {
...
+ * contains { requiresAuthentication: true }
  * @param {*} to a VueRouter Route object
- * @param {*} routeName the name of a route you're checking to see if it was matched.
- * @returns true if a matching route was found, false otherwise
+ * @returns true if the route requires authentication, false otherwise
  */
-export const routeNameMatched = (to, routeName) => {
-  return routeMatched(to, (matched) => (matched?.name === routeName));
+export const routeRequiresAuthentication = (to) => {
+  return routeMatched(to, (matched) => matched.meta?.requiresAuthentication);
```
codyrancher, line 98: I decided to update this to use metadata instead of names so that we don't have to maintain a separate list of authenticated pages.

The vue3 router docs also have this exact example for authentication: https://v3.router.vuejs.org/guide/advanced/meta.html#route-meta-fields so it should migrate better.

shell/utils/auth.js
```diff
@@ -297,9 +297,9 @@ export function notLoggedIn(store, redirect, route) {
   store.commit('auth/hasAuth', true);
 
   if ( route.name === 'index' ) {
-    return redirect(302, '/auth/login');
+    return redirect('/auth/login');
```
codyrancher, line 300: navigation guard next() doesn't support passing a status.

shell/middleware/authenticated.js
```diff
@@ -17,52 +15,7 @@ export default async function({
   route, store, redirect, from, $plugin, next
 }) {
   if ( store.getters['auth/enabled'] !== false && !store.getters['auth/loggedIn'] ) {
```
codyrancher, line 17: This is still necessary because we don't always proceed any further in the middleware depending on whether the user was authenticated or not.

shell/initialize/entry-helpers.js
```diff
@@ -265,7 +265,9 @@ export async function mountApp(appPartials, VueClass) {
   // Add beforeEach router hooks
   router.beforeEach(render.bind(vueApp));
   router.afterEach((from, to) => {
-    updatePageTitle(getVendor());
+    if (from?.name !== to?.name) {
+      updatePageTitle(getVendor());
+    }
```
codyrancher, line 270: When investigating a flaky test I noticed that the page title was flickering on pages where the query or hash in the url was updated (due to tabs for instance)

shell/utils/router.js
rak-phillip: I like this approach, it's straightforward and the idiomatic way to handle authentication in vue router.

shell/initialize/entry-helpers.js
rak-phillip: Could the flashing behavior also have anything to do with the TabTitle (`shell/components/TabTitle.vue`) component?

shell/config/router/navigation-guards/authentication.js
```diff
@@ -0,0 +1,63 @@
...
+          } else {
+            store.commit('setError', { error: e, locationError: new Error('Auth Middleware') });
+          }
+
+          return;
+        }
+      }
+    }
+
+    store.dispatch('gcStartIntervals');
```
rak-phillip, line 59: How is `gcStartIntervals` related to authentication?

rak-phillip (commented):
I think that this looks great! I just have one question about initializing garbage collection from within the authentication guard.

shell/config/router/navigation-guards/authentication.js
codyrancher: It's really not. But it's not related to setting product either and I was going to move it into one of the two lol.

Would you like to put it into a separate navigation guard? I think we can preserve order and timing that way

codyrancher:
> Could the flashing behavior also have anything to do with the TabTitle (shell/components/TabTitle.vue) component?

So it can cause flashing if the content changes but there isn't much we can do about that besides debouncing. This was constantly resetting it back to 'Rancher' as the url was modified even if the route hadn't actually changed.

shell/config/router/navigation-guards/authentication.js
rak-phillip: @codyrancher and I synced on this out of band. It looks like there's quite a bit of behavior in the authenticated middleware that isn't related to authentication. We will follow up with an issue to address the garbage collection in the future.

rak-phillip (approved):

shell/config/router/navigation-guards/authentication.js
codyrancher: Created an issue to track per our conversation https://github.com/rancher/dashboard/issues/11205

PR #11192, opened 2024-06-06
Trying forcing the Cypress binary install due to erros in CI

izaac:
### Summary
Fixes CI

Fixing cypress not found on tests running on 2.7

### Technical notes summary
this will force cypress to be installed before executing

Jenkins


yonasberhe23 (approved):
LGTM

PR #11292, opened 2024-06-24
[FEAT] Added JWT Authentication feature

### LINKED ISSUE #10402: Add support for service account tokens
Internal reference: SURE-6467

A new feature, taken from the parent backend ticket: https://github.com/rancher/rancher/issues/22417

Many 3rd party integrations available for Kubernetes (e.g. for Gitlab, Vault, Avi, etc.) involve giving an external process access to the Kubernetes API using a native Kubernetes Service Account token for authentication.

Rancher's Auth Proxy currently has no notion of SA tokens and sending any request to the Rancher endpoint that uses such token will result in an auth error, but work is being done to implement this in the backend, so this is the frontend portion.

Expected behaviour:

Ranchers auth proxy should support authentication of requests that specify a Service Account token in the Authorization Bearer header

# Updated UI Design/Approach (4th June)

Service Account tokens will need to be configured on a per-cluster basis. For each cluster where this feature is enabled, a `ProxyConfig` resource will be created. Note: The resource includes a single property of interest `enabled` - the presence of a ProxyConfig resource with `enabled` set to false is the same as the ProcyConfig resource not existing.

*UI*

Add a new list view for the new `ProxyConfig` resource under the `Advanced` section of cluster management - name TBD, but suggest `Service Account Access`:

image

This will show only the clusters for which a `ProxyConfig` resource exists and is set to be `enabled`.

Note: The `ProxyConfig` resource uses the cluster ID to reference a cluster, so we will need to fetch the cluster list and do the mapping, so that the list shows the cluster name and not the ID.

The list should show:

Name | Enabled | Age

Note: Enabled will always be true if shown in the list, so we could chose to show this differently. Name should NOT be a link - there is no value in seeing the underlying `ProxyConfig` resource, since the only value on it of use if the `enabled` flag.

You should be able to disable a given row and also bulk disable clusters using the standard table actions.

We should have an info banner at the top of the list page saying something like:

`Service Account Token proxying is configured for the clusters listed below`

The list page will include a call-to-action button in the top-right, labelled 'Configure' - this gives a screen that can be used to enable proxying for one of more clusters. This could be a modified version of the UX from the original edit design - the multi-select input field where you can choose one or more clusters.

Bonus points:

- Add a banner to the cluster creation screen in the Advanced tab saying that Service Account Token proxying can be configured after cluster creation from `Cluster Management > Advanced > Service Account Access`.

- Enhance the cluster detail screen to show a message/status at the top if Service Account Token proxying has been enabled for the cluster

- Future: We could allow this to be enabled at cluster creation time, if we are able to create the cluster and then get the ID and be able to create the resource needed (need to check backend has created the namespace for the cluster)

^^ FYI @kwwii 

Original design:

Proposed design workflow: 
Edit mode: https://xd.adobe.com/view/4eb8efa9-0a4c-447f-81d9-6e3e822bb236-1beb/?fullscreen
View mode: https://xd.adobe.com/view/4eb8efa9-0a4c-447f-81d9-6e3e822bb236-1beb/screen/e015bf0c-261f-4efc-b75a-9e2023817183?fullscreen

eva-vashkevich:
### Summary
Fixes #10402 

Added a new menu entry under "Advanced" tab. "JWT Authentication" page shows users if JWT Authentication has been enabled for different clusters.

Added a banner to "Advanced" tab of default cluster provisioning wizard (ie Digital Ocean) telling users where to enable JWT Authentication. We decided not to add it to the wizards that do not have Advanced tab.

### Technical notes summary

To enable JWT Authentication for a cluster, we need to create a ClusterProxyConfig resource in the namespace on a local cluster that corresponds to the the cluster we are enabling it for.
For this feature, we are displaying the list of all clusters and show if the ClusterProxyConfig exists and has "enabled: true" property.
To enable JWT authentication, ClusterProxyConfig is created. To disable, it is removed (this allows us to tell when it was enabled because we cannot track when "enabled" property was changed).

1. Check that when cluster is created, it appears with JWTA disabled.
2. Check that JWTA can be enabled for a cluster in bulk and individually.
3. Check that JWTA can be disabled for a cluster in bulk and individually.

N/A

### Screenshot/Video

Screenshot 2024-06-24 at 5 43 06 AM

### Checklist
- [x] The PR has a Milestone 


shell/assets/translations/en-us.yaml
```diff
@@ -2867,6 +2871,21 @@ istio:
...
+  actions:
+    activate: Enable
+    deactivate: Disable
+    refresh: Refresh
+  state:
+    enabled: Enabled
+    disabled: Disabled
+  headers:
+    clusterName: Cluster Name
+    updatedOn: JWT enabled on
```
mantis-toboggan-md, line 2885: This should be capitalized, JWT Enabled On

shell/pages/c/_cluster/manager/jwt.authentication/index.vue
```diff
@@ -0,0 +1,243 @@
...
+      schema:                           this.$store.getters['management/schemaFor'](MANAGEMENT.CLUSTER_PROXY_CONFIG),
+      useQueryParamsForSimpleFiltering: false,
+      errors:                           [],
+      forceUpdateLiveAndDelayed:        5
+    };
+  },
+  methods: {
+    formatRows() {
+
+    },
```
mantis-toboggan-md, line 49: It looks like this can be removed

shell/pages/c/_cluster/manager/jwt.authentication/index.vue
```diff
@@ -0,0 +1,243 @@
...
+
+    const res = await allHash(hash);
+
+    this.clusters = res.mgmtClusters;
+    this.proxyConfigs = res.proxyConfig;
+  },
+
+  data() {
+    return {
+      clusterRows:                      [],
```
mantis-toboggan-md, line 36: This is not being used

cypress/e2e/tests/pages/manager/jwt-authentication.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+  const namespace = 'fleet-default';
+
+  before(() => {
+    cy.login();
+    cy.createE2EResourceName('rke2cluster0').as('rke2Ec2ClusterName0');
+    cy.createE2EResourceName('rke2cluster1').as('rke2Ec2ClusterName1');
+
+    cy.get<string>('@rke2Ec2ClusterName0').then((name) => {
+      instance0 = name;
+      // create real cluster
```
mantis-toboggan-md, line 22: Do these tests really need to use actual third-party compute resources? It looks like the feature works with RKE2 custom clusters that don't have any nodes registered. Not a huge deal, just thinking about cost optimization

mantis-toboggan-md (changes requested):
I had a few very minor pieces of feedback, otherwise the PR looks good and works well.

cypress/e2e/tests/pages/manager/jwt-authentication.spec.ts
eva-vashkevich: Good catch, thank you

cypress/e2e/tests/pages/manager/jwt-authentication.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+        rke2ClusterAmazon: {
+          clusterName: name,
+          namespace,
+        }
+      }).then(() => {
+        removeCluster1 = true;
+      });
+    });
+  });
+  beforeEach(() => {
```
yonasberhe23, line 59: this can be `before()` instead of `beforeEach()`. since `testIsolation` is off you don't need to login before each test.

mantis-toboggan-md (approved):
Code looks good to me. I don't know enough about the aws e2e mechanism to comment on the `createAmazonRke2ClusterWithoutMachineConfig` implementation; I will cede to @yonasberhe23's expertise there.

mantis-toboggan-md (approved):

cypress/e2e/tests/pages/manager/jwt-authentication.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+    JWTAuthenticationPagePo.navTo();
+    jwtAuthenticationPage.waitForPage();
+    jwtAuthenticationPage.title().should('be.visible');
+    jwtAuthenticationPage.list().resourceTable().sortableTable().checkVisible();
+    jwtAuthenticationPage.list().resourceTable().sortableTable().checkLoadingIndicatorNotVisible();
+    jwtAuthenticationPage.list().state(instance0).should('contain', 'Disabled');
+    jwtAuthenticationPage.list().state(instance1).should('contain', 'Disabled');
+  });
+  it('should be able to enable JWT Authentication for a cluster', () => {
+    jwtAuthenticationPage.goTo();
```
yonasberhe23, line 71: apologies, i should have seen this earlier but this and below tests can use `navTo`s like you have done in the first test. aside from that, this e2e spec lgtm!

  ```
 JWTAuthenticationPagePo.navTo();
 jwtAuthenticationPage.waitForPage();
```

eva-vashkevich: Do I need to do that? I don't see a reason to test navigation at every test if we already did it once

rak-phillip (approved):
LGTM - the request made by @yonasberhe23 looks to be addressed and the previous review from @mantis-toboggan-md still stands.

PR #11356, opened 2024-07-04
Add shim for `defineComponent` for better backwards compatibility

nwmac:
### Summary
Fixes #

This PR adds a shim for `defineComponent`.

This means that extensions built with the newer shell will be able to be loaded into older versions of Rancher (2.8.0, 2.8.1, 2.8.2, 2.8.3 and 2.7).

The main benefit of `defineComponent` is in the typing - the actual implementation simply returns the options it is passed.

Note: The AsyncButton uses the `inject` function from `vue` - this is the only other use of function exports from Vue, so to be complete, we would need to replace the use of inject there (temporarily) - it is not easy to shim the `inject` function.

This change doesn't need to go into a Rancher release as such - it needs to be published into a shell version, which will then mean extensions built with that shell will work in older versions if they use components declared with `defineComponent`.

The implementation is a WebPack plugin - during compilation we process the abstract syntax tree looking at import statements - any from `vue` for `defineComponent` are updated to refer to a small shim, rather than the `vue` module.

### Checklist
- [ ] The PR has a Milestone 

### DIFF
--- a/shell/pkg/define-component-plugin.js
+++ b/shell/pkg/define-component-plugin.js
@@ -1,0 +1,42 @@
+/**
+ * Custom WebPack plugin that will detect 'defineComponent' imports from Vue and
+ * replace those with references to our shim - which will either use the Vue
+ * definition if avialble or it will use a fallback.
+ */
+
+const path = require('path');
+const PLUGIN_NAME = 'rancherDefineComponentPlugin';
+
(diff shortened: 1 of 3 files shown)


nwmac:
After discussion, to work around this issue, the recommendation is to compile extensions with the Vue library embedded when compatibility is required.

PR #11409, opened 2024-07-10
Remove the rest of middleware and implementation

codyrancher:
### Summary
This replaces the rest of the existing middleware and the removes all the corresponding implementations.

Charts routes when the charts are both installed and not installed.

Same as above

### Checklist
- [x] The PR has a Milestone 


shell/pages/c/_cluster/backup/index.vue
codyrancher, line 1: Not referenced/used

shell/pages/c/_cluster/cis/index.vue
codyrancher, line 1: Not referenced/used

shell/initialize/entry-helpers.js
```diff
@@ -520,7 +451,6 @@ export const setContext = async(app, context) => {
 
   app.context.next = context.next;
   app.context._redirected = false;
-  app.context._errored = false;
```
codyrancher, line 523: No read references

shell/config/router/routes.js
```diff
@@ -192,25 +197,27 @@ export default [
         component: () => interopDefault(import('@shell/pages/c/_cluster/explorer/index.vue')),
         name:      'c-cluster-explorer'
       }, {
-        path:      '/c/:cluster/backup',
-        component: () => interopDefault(import('@shell/pages/c/_cluster/backup/index.vue')),
-        name:      'c-cluster-backup'
+        path: '/c/:cluster/backup',
+        name: 'c-cluster-backup',
+        meta: { ...installRedirectRouteMeta(BACKUP_NAME, BACKUP_CHART_NAME, BACKUP_RESTORE.BACKUP) }
```
codyrancher, line 202: I first attempted this as a `beforeEnter()` route guard but it didn't pan out because we didn't have all of the store data loaded before we got to this point.

I also attempted to use `beforeRouteEnter` in our components but the the guard wouldn't get called I believe before we would redirect.

I don't particularly love this solution and I think we should consider getting rid of this concept entirely but for the time being I just want to preserve existing behavior.

rak-phillip (approved):
LGTM - I think that testing for this should include regressions for Harvester.

codyrancher:
Added @torchiaf for the Harvester perspective.

torchiaf:
I tested this change + the latest [fix](https://github.com/rancher/dashboard/pull/11568) for the switch versions issue . I couldn't see any regression :+1:

PR #11420, opened 2024-07-11
Remove Nuxt reference in extensions and creator 

### LINKED ISSUE #11407: Remove Nuxt reference in extensions and creator
# Description

Some scripts, logic, and configuration are still used around the extensions ecosystem but are outdated.

- Nuxt commands [from extensions in `/pkg`](https://github.com/search?q=%22nuxt+dev%22+repo%3Arancher%2Fdashboard&type=code&ref=advsearch) and [creator](https://github.com/rancher/dashboard/blob/master/shell/creators/pkg/init#L7-L10)
- Plugin check integration  https://github.com/rancher/dashboard/blob/master/shell/initialize/install-plugins.js#L104
- https://github.com/rancher/dashboard/blob/master/shell/core/plugins-loader.js#L17
- `nuxt.config.js` mentions
- `Nuxt` term as comment
- VSCode setting

## Findings

Shell still includes `nuxt` library as a dependency. When this is removed, issues start to appear around the CI, although the application itself seems working just fine, as well as the E2E.

cnotv:
### Summary
Fixes #11407

Some of the dependencies added seems were present in the `nuxt` library and not defined within the shell, therefore have been added.
https://github.com/rancher/dashboard/pull/11420/commits/c3a78783ba52ed800162ec6414ad1fec1e62ee01

### Checklist
- [x] The PR has a Milestone 


shell/initialize/app-extended.js
```diff
@@ -1,5 +1,3 @@
-// Taken from @nuxt/vue-app/template/index.js
```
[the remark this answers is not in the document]
cnotv: Yes sure.
This file is going to be removed with https://github.com/rancher/dashboard/issues/11421, so I will just put it into `entry.js` and in the [vue.config.js](https://github.com/rancher/dashboard/blob/master/shell/vue.config.js#L502) which maybe makes more sense.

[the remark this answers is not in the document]
cnotv: Is there an actual link? I guess it may make more sense.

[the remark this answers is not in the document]
cnotv: I'm leaving a TODO for when you are back and we can add more information, so it can be more helpful.

shell/package.json
```diff
@@ -103,7 +104,6 @@
     "lodash": "4.17.21",
     "marked": "4.0.17",
     "nodemon": "2.0.22",
-    "nuxt": "2.15.8",
```
richard-cox, line 106: @codyrancher can you remember why this was originally left in following the nuxt removal?

shell/package.json
```diff
@@ -122,9 +122,12 @@
     "v-tooltip": "2.0.3",
     "vue": "2.7.16",
     "clipboard-polyfill": "4.0.1",
+    "ufo": "0.7.11",
```
richard-cox, line 125: were ufo, unfetch and vue-router the deps that were brought in via nuxt, but given nuxt has been removed we have to manually add?

shell/vue.config.js
```diff
@@ -498,6 +499,8 @@ module.exports = function(dir, _appConfig) {
     },
     outputDir,
     pages: {
+      // Taken from @nuxt/vue-app/template/index.js
```
richard-cox, line 502: not sure this comment is in the right place

shell/initialize/entry.js
```diff
@@ -1,5 +1,5 @@
-// Taken from @nuxt/vue-app/template/client.js
-
+// Taken from @nuxt/vue-app/template/index.js
```
richard-cox, line 1: this is the wrong file

shell/initialize/app-extended.js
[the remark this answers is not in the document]
richard-cox: We shouldn't be merging TODOs. 

This file and the other were taken from nuxt during the migration and have been updated since. Would it be ok to just leave the comments unchanged until the file is removed?

richard-cox (changes requested):

shell/package.json
cnotv: Exactly 👍

shell/vue.config.js
cnotv: This is where we define the entry, so it's worth adding it here, it doesn't cause any harm to have some extra information.

shell/initialize/app-extended.js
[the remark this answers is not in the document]
cnotv: Since I'll have to remove the file, do you have any declaration related to this link? It may be useful for the migration and speed up the process.

shell/package.json
codyrancher: No, I think it was just an oversight.

yarn.lock
```diff
@@ -1943,7 +1943,7 @@
 
 "@babel/plugin-proposal-private-methods@^7.18.6":
   version "7.18.6"
-  resolved "https://registry.yarnpkg.com/@babel/plugin-proposal-private-methods/-/plugin-proposal-private-methods-7.18.6.tgz#5209de7d213457548a98436fa2882f52f4be6bea"
+  resolved "https://registry.npmjs.org/@babel/plugin-proposal-private-methods/-/plugin-proposal-private-methods-7.18.6.tgz#5209de7d213457548a98436fa2882f52f4be6bea"
```
richard-cox, line 1946: Could this be kept as yarnpkg.com?

shell/initialize/app-extended.js
```diff
@@ -1,5 +1,6 @@
 // Taken from @nuxt/vue-app/template/index.js
 // This file was generated during Nuxt migration
+// TODO: Add more clarification to this comment
```
richard-cox, line 3: There shouldn't be any TODOs in code. 

The code comes from nuxt, aka https://github.com/nuxt/nuxt/blob/v2.18.1/packages/vue-app/template/index.js --> which is the old node_modules/@nuxt/vue-app/template/index.js

shell/initialize/entry.js
```diff
@@ -1,5 +1,4 @@
-// Taken from @nuxt/vue-app/template/client.js
-
+// TODO: Add more clarification to this comment
```
richard-cox, line 1: - lost file reference
- errant todo

richard-cox (changes requested):
i'll leave the functional testing to @codyrancher and @rak-phillip

shell/initialize/app-extended.js
cnotv: You cannot just put a comment with a path, you need to add an explanation and something working.
Thanks for the link, that's better and I can add it 👍

yarn.lock
cnotv: As the top line states, this file is automatically generated and should not be modified by hand.

richard-cox: Sorry, as in can the method used which updated be fixed so that it stays yarnpkg? Has the package been removed from the yarn registry?

shell/initialize/app-extended.js
richard-cox: The path comes from the node package, hence the @nuxt. Ideally i was trying to avoid a direct reference. 

Can we just use `node_modules/@nuxt/vue-app/template/index.js`?

cnotv: would you mind to motivate the choice, so I can understand as well and we can find an agreement?

yarn.lock
cnotv: It sounds a bit out of topic for this PR maybe?

cnotv: We have 1236 occurrences on `master` like this

Screenshot 2024-08-06 at 22 05 08

And 2080 in the other case


Screenshot 2024-08-06 at 22 06 16

codyrancher (approved):
Functionality wise I'm happy with the changes. Global, cluster, login, logout and extension pages all seem to render fine without error.

Once e2e tests pass and issues are resolved with @richard-cox I think it's good to go.

cnotv:
> Once e2e tests pass 

Wow I pushed 4h ago 😅

shell/creators/pkg/init
```diff
@@ -4,11 +4,6 @@ const fs = require('fs-extra');
 const path = require('path');
 const https = require('https');
 
-const targets = {
-  dev:  './node_modules/.bin/nuxt dev',
```
rak-phillip, line 8: Should we be removing the `dev` script? Would it make more sense to replace it with the `vue-cli` equivalent - `"dev": "./node_modules/.bin/vue-cli-service dev",`?

cnotv: It seems like this script has never been used since we do not rely on Nuxt, or am I wrong?

rak-phillip: That's hard for me to say.. perhaps someone who is more familiar with the creators package can inform us. 

@jordojordo @aalves08 does it make sense to remove the `dev` script?

cnotv: (while waiting for a feedback)
If that would be the case, we would have also to restore all the

```
  "scripts": {
    "dev": "./node_modules/.bin/nuxt dev",
    "nuxt": "./node_modules/.bin/nuxt"
  },
```

As well as removing the `nuxt` dependency. But we have already stated that it's not needed (both @nwmac in chat and @codyrancher in another review)
If that would also not be the case, we have to keep `nuxt` dependency or make investigation on what means remove Nuxt globally from the extensions.

cnotv: Btw, If we cannot remove Nuxt from the extensions, then we are not able to migrate anymore in the current state and we'll have to tackle that issue. Nuxt2 is not compatible with Vue3 ecosystem.

aalves08: I don't think any of these scripts are ever used at the package level 🤔 Thoughts @jordojordo ? I don't we use any of these... It's just app-level yarn commands 🤔

cnotv: I went through the whole documentation of the extension and there's no mention on starting one, rather publishing one and that's it: https://extensions.rancher.io/extensions

nwmac:
@richard-cox Can we get this one in, as its needed for Vue 3 and then address things like TODOs later.

yarn.lock
richard-cox: The PR contains the change, so it's very much in topic.

There's 64 references in `yarn.lock` to `npmjs`, 52 are from `@aws-sdk`.

There's 1167 references in `pkg/eks/yarn.lock` which is a whole other story. It looks like that suffered the same issue as this change.

I've checked and `https://registry.yarnpkg.com/@babel/plugin-proposal-private-methods/-/plugin-proposal-private-methods-7.18.6.tgz#5209de7d213457548a98436fa2882f52f4be6bea` does exist. 

I've tested in a new project and `yarn add @babel/plugin-proposal-private-methods` creates a yarn lock entry containing `yarnpkg.com`.

I'm not sure there's any reason for the change?

shell/initialize/app-extended.js
richard-cox: Can you leave unchanged and create an issue?

richard-cox (approved):

shell/creators/pkg/init
jordojordo: @aalves08 is correct, we don't use the package level scripts at all. Should be safe to remove those.

yarn.lock
cnotv: As I said, `yarn.lock` is automatically generated.
When you have countless conflicts on rebase, the only way is to create a new one.

Is there any specific issue forcing you to block this automatic generation? If not, please it would be great to proceed working on the migration.
We have changes on it constantly: https://github.com/rancher/dashboard/commits/master/yarn.lock

richard-cox: So you're saying you had to regenerate yarn.lock in order for the changes in this pr to work, and that regeneration added the change? I couldn't reproduce. Also there's no other changes in this file, so i'm not sure what the conflict was.

This file does change, but 21 times in 7 months though.

From my point of view the PR introduces the change and the logic is not clear how it was brought it or needed.

At the end of the day the PR is approved and can be merged. I'll try and find some time tomorrow to understand what caused this specific issue.

cnotv: FYI I regenerate `yarn.lock` every time. 
This is since I fixed the manually edited version we had 2 years and a half ago and that is what I have done since then, which is what makes it possible to edit the dependencies now or generate a new one.

You are free to investigate the registry, although at the current state and without any research I have no suggestion for you about this.

PR #11490, opened 2024-07-18
Introduce additional form validation for User Retention

### LINKED ISSUE #11333: User Retention: Improve validation for `disable-inactive-user-after` and `delete-inactive-user-after` fields
Setting an inactivity period to delete user accounts that is less than `336h0m0s` returns the error: 

```
admission webhook "rancher.cattle.io.settings.management.cattle.io" denied the request: value: Invalid value: "30h": must be at least 336h0m0s 
```

## Expected Behavior

The user retention form will not validate when known minimum values are not satisfied, alerting an admin that there is an issue with the form before they can submit.

## Actual Behavior

The user retention form passes validation, allowing an admin to submit the form with invalid values.

## Acceptance Criteria

- [ ] The user retention form does not validate if the inactivity period for deleting user accounts is set to a value less than `336h0m0s`
- [ ] When the inactivity period for deleting a user account is not valid, a clear and informative error message should be displayed, indicating the minimum allowed value
- [ ] The hint for inactivity period for deleting a user account is updated to reflect real values that satisfy form validation
- [ ] Disable and Delete after validate that the value is greater than `auth-user-session-ttl-minutes`
- [ ] Disable and Delete after validate that the duration format adheres to `{h|m|s}`
- [ ] Disable and Delete after disallow the use of fractional seconds in durations

rak-phillip:
### Summary
This adds additional validation rules to the user retention form. The following validations have been added:

- Disable/Delete after
   - validate the duration format `{h|m|s}`
      - disallow fractional seconds
   - validate that the value is greater than `auth-user-session-ttl-minutes`
- Delete after
   - validate that the value is greater than 336h0m0s

Additional changes have been made to support the validation changes:

- Add `requireDirty` prop to the `useLabeledFormElement` composable under `shell/composables/useLabeledFormElement.ts`
- Emit `update:validation` events when validation fails for a given input

Fixes #11333 

- User Retention form
  - Disable user after inactivity period validation
  - Delete user after inactivity period validation
  - User retention process schedule validation

- Validations surrounding the user retention form
- Submitting the user retention form

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/composables/useLabeledFormElement.ts
+++ b/shell/composables/useLabeledFormElement.ts
@@ -104,8 +104,10 @@ export const useLabeledFormElement = (props: LabeledFormElementProps, emit: (eve
     }
 
     if (ruleMessages.length > 0 && (blurred.value || focused.value)) {
+      emit('update:validation', false);
       return ruleMessages.join(', ');
     } else {
+      emit('update:validation', true);
       return undefined;
     }
   });
--- a/shell/composables/useLabeledFormElement.ts
+++ b/shell/composables/useLabeledFormElement.ts
@@ -7,6 +7,7 @@ interface LabeledFormElementProps {
   required: boolean;
   disabled: boolean;
   rules: Array<any>;
+  requireDirty?: boolean;
 }
 
 interface UseLabeledFormElement {
@@ -63,6 +64,10 @@ export const labeledFormElementProps = {
     type:    Boolean,
     default: false,
   },
+  requireDirty: {
+    default: true,
+    type:    Boolean
+  }
 };
 
 export const useLabeledFormElement = (props: LabeledFormElementProps, emit: (event: string, ...args: any[]) => void): UseLabeledFormElement => {
@@ -91,6 +96,8 @@ export const useLabeledFormElement = (props: LabeledFormElementProps, emit: (eve
       const message = requiredRule(value);
 
       if (!!message) {
+        emit('update:validation', false);
+
         return message;
       }
     }
@@ -103,11 +110,13 @@ export const useLabeledFormElement = (props: LabeledFormElementProps, emit: (eve
       }
     }
 
-    if (ruleMessages.length > 0 && (blurred.value || focused.value)) {
+    if (ruleMessages.length > 0 && (blurred.value || focused.value || !props.requireDirty)) {
       emit('update:validation', false);
+
       return ruleMessages.join(', ');
     } else {
       emit('update:validation', true);
+
       return undefined;
     }
   });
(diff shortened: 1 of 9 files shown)


shell/composables/useLabeledFormElement.ts
```diff
@@ -63,6 +64,10 @@ export const labeledFormElementProps = {
     type:    Boolean,
     default: false,
   },
+  requireDirty: {
+    default: true,
+    type:    Boolean
+  }
```
rak-phillip, line 70: This is a clear sign that there's some drift between the mixin and the composable. I created #11491 so that we can address.

shell/composables/useUserRetentionValidation.ts
```diff
@@ -0,0 +1,175 @@
...
+  links: Links;
+  apiVersion: string;
+  customized: boolean;
+  default: string;
+  kind: string;
+  metadata: Metadata;
+  source: string;
+  value: string | null;
+  save: () => void;
+};
```
rak-phillip, line 64: I'm considering if there's a better place for our Setting type to live? I'm not a fan of keeping it colocated with code related to user retention.

rak-phillip: I opted for `shell/types/resources/settings.d.ts`

shell/composables/useLabeledFormElement.ts
```diff
@@ -91,6 +96,8 @@ export const useLabeledFormElement = (props: LabeledFormElementProps, emit: (eve
       const message = requiredRule(value);
 
       if (!!message) {
+        emit('update:validation', false);
```
rak-phillip, line 99: We might want to consider also updating `shell/mixins/labeled-form-element.ts` to avoid further drift between the two methods code sharing.

mantis-toboggan-md (changes requested):
I see duplicate error messages for the duration format validation on the deleteAfterPeriod input:
Screenshot 2024-07-26 at 11 40 12 AM

It looks like this is due to the format of duration strings being validated within other validators. One potential solution would be to give duration formatting its own validator function, and return error messages in `validateDeleteInactiveUserAfter` and `validateDurationAgainstAuthUserSession` only when a valid duration string is passed in.



Otherwise, the PR looks good to me. I think the addition to `useLabeledFormElement` is a solid way of making the composable function without the form validation mixin. I've attempted to use input-level error reporting without the form validation mixin in another repo myself, and this change would have made that considerably tidier. I would also be in favor of adding it to the `labeled-form-element` mixin.

rak-phillip:
@mantis-toboggan-md that's a good catch, thanks for the review! I made the following changes

- Create a new validator `validateDuratio()`
- Print warning instead of returning validation strings in `validateDeleteInactiveUserAfter()` & `validateDurationAgainstAuthUserSession()` 

I also updated the `labeled-form-element` mixin so that it emits validation errors in the same way as the composable. I think it's a good idea to keep these two in sync until we are able to remove the mixin usage in #11491

mantis-toboggan-md (approved):

PR #11522, opened 2024-07-24
Vue3 login

codyrancher:


shell/initialize/install-plugins.js
rak-phillip, line 1: Can we revert most of the changes here? I think that we would ideally only have `@shell/plugins/vue-js-modal` commented out at this point

shell/initialize/entry-helpers.js
rak-phillip, line 1: I think that this is good to keep, I'd rather not have our init bomb in the event that we don't match a route..

shell/initialize/app-extended.js
```diff
@@ -70,19 +70,18 @@ async function extendApp(vueApp) {
...
     // Ignore 404s rather than blindly replacing URL in browser
-    const { route } = router.resolve(appPartials.context.route.fullPath);
+    const route = router.resolve(router.currentRoute.value.fullPath);
 
     if (!route.matched.length) {
       return resolve();
     }
 
-    router.replace(appPartials.context.route.fullPath, resolve, (err) => {
+    router.replace(router.currentRoute.value.fullPath, resolve, (err) => {
```
rak-phillip, line 84: I don't know if it matters, but this can be a candidate for `master`. Potentially one less change to track when we make the break for Vue3.

shell/config/router/index.js
rak-phillip, line 1: Can we try reverting this change to see if everything still works for you?

shell/mixins/compact-input.ts
```diff
@@ -1,7 +1,5 @@
-import { createApp } from 'vue';
-const vueApp = createApp({});
 
-export default Vue.extend({
+export default {
```
rak-phillip, line 2: If we were previously using `Vue.extend`, we would now want to use `defineComponent`

shell/mixins/labeled-form-element.ts
```diff
@@ -1,15 +1,13 @@
...
-const vueApp = createApp({});
 
 interface LabeledFormElement {
   raised: boolean;
   focused: boolean;
   blurred: number | null;
 }
 
-export default Vue.extend({
+export default {
```
rak-phillip, line 10: Same here, `export default defineComponent({ ... })`

shell/initialize/app-extended.js
codyrancher: I'd certainly be willing to try and see how e2e tests deal with it.

shell/initialize/install-plugins.js
codyrancher: Looks like I can't without introducing more warnings. I reintroduced intNumber but the rest are problematic. Each will need to be addressed though.

shell/mixins/compact-input.ts
rak-phillip: NVM `defineComponent` is used primarily for components and not mixins.

codyrancher: Spoke offline

shell/mixins/labeled-form-element.ts
codyrancher: Spoke offline

shell/initialize/entry-helpers.js
codyrancher: Keeping

shell/config/router/index.js
codyrancher: Looks like it works. Reverted.

rak-phillip (approved):

PR #11568, opened 2024-07-30
Remove old extension from parentElement

torchiaf:
### Summary
Fixes https://github.com/rancher/dashboard/issues/11559

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/core/plugins.js
+++ b/shell/core/plugins.js
@@ -90,13 +90,15 @@ export default function(context, inject, vueApp) {
               delete window[oldPlugin.id];
 
               delete plugins[oldPlugin.id];
+
+              const oldElement = document.getElementById(oldPlugin.id);
+
+              oldElement.parentElement.removeChild(oldElement);
             });
           }
 
           removed.then(() => {
             element.onload = () => {
-              element.parentElement.removeChild(element);
-
               if (!window[id]) {
                 return reject(new Error('Could not load plugin code'));
               }
--- a/shell/core/plugins.js
+++ b/shell/core/plugins.js
@@ -95,6 +95,8 @@ export default function(context, inject, vueApp) {
 
           removed.then(() => {
             element.onload = () => {
+              element.parentElement.removeChild(element);
+
               if (!window[id]) {
                 return reject(new Error('Could not load plugin code'));
               }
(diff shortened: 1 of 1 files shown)


shell/core/plugins.js
```diff
@@ -95,6 +95,8 @@ export default function(context, inject, vueApp) {
 
           removed.then(() => {
             element.onload = () => {
+              element.parentElement.removeChild(element);
```
codyrancher, line 98: When I originally pushed the login change there was a conversation with @richard-cox on this.

I've asserted and continue to assert that this shouldn't have any affect on our application because it doesn't unload the extension since the bundle has already been executed. 

If this is fixing something can someone explain how it's being fixed? I'm very confused how removing the script tag would have any affect.

codyrancher: Just to demonstrate what I expect to happen https://jsbin.com/cejeyupele/edit?html,js,output

image

torchiaf: I tested the load mechanism of the plugins and put some console logs in the code.

- Given that the `element.onload()` fires _after_ the Harvester plugin's script is loaded, essentially the `removeChild` is called every time a new Harvester plugin is loaded in the browser, so the script is removed from the HTML in any case.
- As result of the assumption above, this check here in the `loadAsync()` function:

```
  if (document.getElementById(id)) {
    return resolve();
  }
```
will be always false when switching Harvester clusters, therefore , the function can continue executing the code to load the plugin, which is in our case, a new plugin version.
- Removing the removeChild call means that `asyncLoad` stops over the condition above even if  loadClusterPlugin function verified that the new plugin version is missing and should be loaded.
- Notice that the remove/add HTML `element` mechanism is completely different from the remove/add `plugin` checks implemented in the Vue store.
- I would add the removeChild in the removePlugin function (when element is not null) for consistency - see next commit.

Here is my testing [branch](https://github.com/torchiaf/dashboard/tree/load-plugins-load), there are some sparse and confused log(), but it helps to figure out what is the workflow.

codyrancher: > As result of the assumption above, this check here in the loadAsync() function:

I think that means that different versions of harvester are using the same `id` and that we're now loading the bundles over and over with this change. But it sounds like it's working fine for you and we can deal with any issues if they come up.

codyrancher (dismissed):

torchiaf:
In the last commit we are limiting the `removeChild` for `oldPlugins` only  https://github.com/rancher/dashboard/blob/d221ed9fd6b3cc4f1222f42dd14b88512d539073/shell/core/plugins.js#L94-L96 
 thanks for the suggestion @richard-cox

codyrancher (approved):

PR #11824, opened 2024-09-04
Fix problematic boolean attributes

### LINKED ISSUE #11746: Vue3: Fix problematic disabled boolean attributes
# Description

As identified by @momesgin, in Vue3 the disabled attributed is handled differently.

Quoting our notes:

> Stylings may not apply properly to some disabled, selected or focused elements(e.g. div)
 In Vue 2, setting false to the value of an attribute would remove the attribute from being rendered. When doing this in Vue 3, false is considered a valid value, so the attribute will be rendered, to fix this you can pass null to the attribute's value instead 
 E.g. 
 > ```
 > <div :disabled="!opt.enabled">...</div>
 > ```
>  should be converted to 
 > ```
 > <div :disabled="opt.enabled ? null : true">...</div>
 > ```
>  More info: https://v3-migration.vuejs.org/breaking-changes/attribute-coercion.html

## Affected components

The following components have been identified to have potentially problematic usages of `:disabled`. This isn't a comprehensive list, but is a good starting point for identifying which patterns we might need to look out for in order to completely resolve this issue.

- [ ] pkg/harvester-manager/list/harvesterhci.io.management.cluster.vue
- [ ] shell/components/ActionMenu.vue
- [ ] shell/pages/c/_cluster/istio/index.vue
- [ ] shell/pages/c/_cluster/monitoring/index.vue

momesgin:
### Summary
Fixes #11746 

I searched for the `:disabled` attribute throughout the app and didn't find any cases other than those already mentioned in the issue.

### Checklist
- [ ] The PR has a Milestone 

### DIFF
--- a/pkg/harvester-manager/list/harvesterhci.io.management.cluster.vue
+++ b/pkg/harvester-manager/list/harvesterhci.io.management.cluster.vue
@@ -148,7 +148,7 @@ export default {
             <a
               v-if="row.isReady"
               class="link"
-              :disabled="navigating"
+              :disabled="navigating ? true : null"
               @click="goToCluster(row)"
             >{{ row.nameDisplay }}</a>
             <span v-else>
--- a/shell/components/ActionMenu.vue
+++ b/shell/components/ActionMenu.vue
@@ -258,7 +258,7 @@ export default {
       <li
         v-for="(opt, i) in menuOptions"
         :key="i"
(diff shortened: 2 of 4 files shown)


cnotv (approved):
As I am not able to identify manually further cases, I would opt for merging this first.

LGTM

PR #11890, opened 2024-09-11
Allow for loading dynamic components defined in extensions

rak-phillip:
### Summary
This fixes an issue with rendering components that are lazy loaded via an extension configuration. For example:

```
  plugin.addCard(
    CardLocation.CLUSTER_DASHBOARD_CARD,
    { cluster: ['local'] },
    {
      label: 'some-label2',
      labelKey: 'plugin-examples.card-title-three',
      component: () => import('./components/MastheadDetailsComponent.vue')
    }
  );
```

`defineAsyncComponent` helps Vue handle the asynchronous nature of lazy loading and `markRaw` prevents Vue from making the component reactive, which can help with performance.

- wrap the component property for Card, Tab, and Panel in `defineAsyncComponent` and `markRaw`

- This has been tested with [extensions-api-demo](https://github.com/rancher/ui-plugin-examples/tree/main/pkg/extensions-api-demo)
- Other extensions should be tested and verified

- Loading extensions that might not lazy load components

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/core/plugin.ts
+++ b/shell/core/plugin.ts
@@ -15,6 +15,7 @@ import {
   PluginRouteRecordRaw, RegisterStore, UnregisterStore, CoreStoreSpecifics, CoreStoreConfig, OnNavToPackage, OnNavAwayFromPackage, OnLogOut
 } from './types';
 import coreStore, { coreStoreModule, coreStoreState } from '@shell/plugins/dashboard-store';
+import { defineAsyncComponent, markRaw, Component } from 'vue';
 
 export type ProductFunction = (plugin: IPlugin, store: any) => void;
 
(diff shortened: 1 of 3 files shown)


cnotv:
Is there any limitation to defining the component without reactivity for this purpose? If so it should probably be more informed in a way that it cannot be missed by other developers, as it may create big headaches, especially for non-Vue experts.

shell/core/plugin.ts
```diff
@@ -161,21 +162,47 @@ export class Plugin implements IPlugin {
...
+   * configuration, but with the component property wrapped in
+   * defineAsyncComponent and markRaw. If the extension configuration doesn't
+   * have a component property, it returns the extension configuration
+   * unchanged.
+   */
+  private _createAsyncComponent(extensionConfig: Card | Panel | Tab) {
+    const { component } = extensionConfig;
+
+    if (!component) {
+      return extensionConfig;
```
cnotv, line 199: What is the benefit of returning the configuration and how would you envision catching errors in this case given the fact we are not returning a falsy value?

shell/core/plugin.ts
```diff
@@ -161,21 +162,47 @@ export class Plugin implements IPlugin {
...
+    }
+
+    return {
+      ...extensionConfig,
+      component: markRaw(defineAsyncComponent(component as () => Promise<Component>)),
```
cnotv, line 204: (Moving my previous comment directly to the code as better conversation, but I forgot to copy it 🤦 )

Is there any side effect from removing reactivity with `markRaw`? If so, we should make it clear to the developers who intend to use the extensions, as they may not be familiar with Vue patterns.

shell/core/plugin.ts
rak-phillip: We're guarding against modifying the component prop in the event that it does not exist; we return the configuration unchanged if there is nothing to modify.

shell/core/plugin.ts
rak-phillip: No side effect that I'm aware of. Using `markRaw` prevents Vue from treating the component as a reactive object - this can happen when a component is passed to another component via props.

cnotv: Wait, are there not plenty of these cases around?

cnotv:
I took the initiative to shove my conversion to TS of the helper since I started to convert it to do some investigations. Please feel free to remove it if bothers you and let me know if I should avoid such actions.

cnotv (approved):
LGTM

rak-phillip:
> I took the initiative to shove my conversion to TS of the helper since I started to convert it to do some investigations. Please feel free to remove it if bothers you and let me know if I should avoid such actions.

This is fine by me, the included changes don't modify the behavior of this PR at all 🙂

PR #11902, opened 2024-09-12
update all instances of LabeledInput using v-model.number to use v-model:value.number

### LINKED ISSUE #11839: Vue3: Service: Create External name cannot be created
**Setup**
- Rancher version: 2.10
- Rancher UI Extensions: N/A
- Browser type & version: N/A

**Describe the bug**

DNS Name field input binding is broken

**To Reproduce**

Service -> Create ExternalName
Enter "Name"
Enter "DNS name"

**Result**
"DNS Name" input field is reset
"Create" button is disabled

**Expected Result**

"DNS Name" value should be maintained
"Create"button should be enabled

**Screenshots**

https://github.com/user-attachments/assets/c77432ec-08ee-4e3e-86be-f2984dcd1f07

**Additional context**

### LINKED ISSUE #11856: Vue3: instances of LabeledInput using v-model.number modifier need to be updated to vue3 syntax
**Describe the bug**
It looks like v-model was generally replaced with v-model:value, but not when the .number modifier was used.  UnitInputs appear to be working, but LabeledInputs using v-model.number are not. A quick code search shows we  have quite a few instances of this specific problem; we can probably apply the same fix across them all in one go.

**To Reproduce**
The performance page garbage collection resource count input is a good example of this bug:

https://github.com/user-attachments/assets/adcfc8d1-db3c-4a98-951c-9f3e13bc9782

**Expected Result**
Old uses of `v-model.number` need to be replaced with `v-model:value.number`
https://vuejs.org/guide/components/v-model.html#modifiers-for-v-model-with-arguments

mantis-toboggan-md:
### Summary
Fixes #11856 

Fixes #11839

I've also fixed small bugs I hit while trying to manually verify each LabeledInput I touched. 

- fix HookOption not rendering (workload creation)
- fix pod and node affinity weight inputs not updating parent component (workload creation)
- fix Probe not updating parent component (workload creation)
- fix codemirror blowing up the second tab in output creation (install logging to create 'output' resources)
- delete unused component PodSecurity
- delete unused component Weight

I don't think it's necessary to re-check _every_ component in this PR as the same fix was applied across all; however, a quick verification of the above additional fixes is warranted.

### Checklist
- [x] The PR has a Milestone 


eva-vashkevich:
It also fixes #11839

cnotv:
I've added issue #11932 to be included in the GH process.

cnotv:
> I don't think it's necessary to re-check every component in this PR as the same fix was applied across all; however, a quick verification of the above additional fixes is warranted.

Mass replacements like this one can only been guaranteed by the QA team when they will make a whole investigation, just like any other migrated case.

shell/components/CodeMirror.vue
```diff
@@ -102,9 +102,8 @@ export default {
   },
 
   methods: {
-
     focus() {
-      if ( this.$refs.codeMirrorRef ) {
+      if ( this.$refs.codeMirrorRef && this.$refs.codeMirrorRef.codemirror ) {
```
cnotv, line 106: This one seems to prevent the view from breaking, although it's still not focusing. Also in the issue, we have defined to provide a test automation. We should either change it to manual or add a unit test.

For the unit test, an implementation may be something similar used in the view, although I am not sure it will reproduce the issue with a mocked parent component. 
https://github.com/rancher/dashboard/blob/master/shell/edit/logging-flow/index.vue#L281

Would you prefer to keep the error prevention separate from the focus functionality?

Sorry for being pedantic.

Screenshot 2024-09-18 at 22 09 42

cnotv (commented):
All the initial and other issues seems to be addressed. The extra part for CodeMirror seems to still not behaving. We could either close this one without that (and remove the fixes) or take a second look.

shell/components/CodeMirror.vue
jordojordo: I have [a PR ](https://github.com/rancher/dashboard/pull/11973)which should fix the codeMirrorRef issue, I believe with the change from `vue-codemirror` to `codemirror-editor-vue3` might have caused this ref naming mismatch.

mantis-toboggan-md: @cnotv I think that's fair -- it should probably have a test or be marked for manual testing, but @jordojordo has spared me by addressing this bug in his own pr. I have confirmed his PR resolves the issue with logging outputs I (tried to) fix here. Consequently I have removed this change from this PR.

cnotv: Good then and thanks @jordojordo 😄

cnotv (approved):
LGTM

Issue #11961, opened 2024-09-18
Add explicit emits in the components


cnotv:
# Description

As already defined [in the issue related to the migration documentation](https://github.com/rancher/dashboard/issues/10658), emitters are preferred in Vue3 as they may be inherited altogether with other attributes.

Add emitters identified by the [vue linter rule](https://eslint.vuejs.org/rules/require-explicit-emits.html#options). To do so
- [Enable globally in the default configuration](https://github.com/rancher/dashboard/blob/master/.eslintrc.default.js#L192) with: `'vue/require-explicit-emits': 'error',`
- Remove definition in [root configuration](https://github.com/rancher/dashboard/blob/master/.eslintrc.default.js#L192)

Afterwards add the emitter as indicated.

Occurrences:

<img width="844" alt="Screenshot 2024-09-18 at 18 31 24" src="https://github.com/user-attachments/assets/91615729-5dd7-4629-a58f-bcb3596905e5">

yonasberhe23:
nothing for QA to do here. moving to done

PR #11964, opened 2024-09-18
Enabling `vue/require-explicit-emits` linting rule

codyrancher:
### Summary
Enabling `vue/require-explicit-emits` linting rule

https://github.com/rancher/dashboard/issues/11961

`yarn lint`

Shouldn't, this is really just an annotation for our code, not functional.

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -59,7 +59,6 @@ module.exports = {
     '@typescript-eslint/no-var-requires':         'off',
     'vue/one-component-per-file':                 'off',
     'vue/no-deprecated-slot-attribute':           'off',
-    'vue/require-explicit-emits':                 'off',
     'vue/v-on-event-hyphenation':                 'off'
   },
   overrides: [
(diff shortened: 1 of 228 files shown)


.eslintrc.js
```diff
@@ -59,7 +59,7 @@ module.exports = {
     '@typescript-eslint/no-var-requires':         'off',
     'vue/one-component-per-file':                 'off',
     'vue/no-deprecated-slot-attribute':           'off',
-    'vue/require-explicit-emits':                 'off',
+    'vue/require-explicit-emits':                 'error',
```
cnotv, line 62: Sorry for being picky, but could we remove this line and set it as errors in `.eslintrc.default.js`, so it's used in all the projects of the monorepo?

cnotv (changes requested):
Looks good, just this small point to be sure it's used also in `@rancher/component` and other pkgs.

.eslintrc.js
codyrancher: I updated it.

It looks like the original change still caught everything in the monorepo with `yarn lint`. What scenario are you wanting to cover because I didn't verify anything else.

cnotv: It's important to have separate common rules adopted everywhere and linting rules strictly applied for the rest of the project. `@rancher/components` has its own linting scripts, can't say the same for other `pkg`. There's no need for you to try, since they are already included in CI.

PR #12000, opened 2024-09-23
Simplify InputOrDisplay.vue component

torchiaf:
### Summary

On Harvester, we have some errors when using the `InputOrDisplay` component as a container - probably something related to the use of `setup` function or dynamic HTML build.

https://github.com/torchiaf/harvester/blob/62da7658669d156d63b0d385a2978ab9d37637de/pkg/harvester/edit/kubevirt.io.virtualmachine/VirtualMachineCpuMemory.vue#L81-L94

image

The current PR just simplify the implementation and fixes the issue.

After

image

### DIFF
--- a/shell/components/InputOrDisplay.vue
+++ b/shell/components/InputOrDisplay.vue
@@ -1,5 +1,4 @@
 <script>
-import { h, computed } from 'vue';
 import { _VIEW } from '@shell/config/query-params';
 
 export default {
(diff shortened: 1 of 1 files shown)


shell/components/InputOrDisplay.vue
```diff
@@ -18,31 +17,37 @@ export default {
...
-    });
-
-    return () => {
-      if (isView.value) {
-        return h('div', { class: 'label' }, [
-          h('div', { class: 'text-label' }, slots.name ? slots.name : props.name),
-          h('div', { class: 'value' }, slots.value ? slots.value : displayValue.value)
-        ]);
-      } else {
-        return slots.default;
```
rak-phillip, line 39: Alternatively, we could have invoked the slot as a function via:

```
return slots.default();
```

The same would go for the two slots above.

rak-phillip (approved):
This change looks good to me.

To be clear about the issue, this issue was related to Slot Unification in Vue 3 now exposing slots as a function[^1]:

[^1]: https://v3-migration.vuejs.org/breaking-changes/slots-unification.html#overview

I left a comment on an alternative approach, but I think that we should ultimately use the solution that will be easiest to understand for future maintenance.

shell/components/InputOrDisplay.vue
torchiaf: Thanks Phil, better to know what the error was, I found another case where we need to update the slots API. https://github.com/rancher/dashboard/pull/12005

PR #12012, opened 2024-09-24
Fix Ingress Rule Path CSS

### LINKED ISSUE #12011: Ingress Rule Path input not displayed correctly
# Description

Ingress Rule Path input seems to have not the correct size and behavior. 

This one of the cases identified when investigating migration issues while using `:deep()`: https://github.com/rancher/dashboard/issues/11952

image

image

cnotv:
### Summary
Fixes #12011

Although I have not been able to reproduce a case where the definition breaks (https://codesandbox.io/p/sandbox/yy4cgk), this PR addresses the styling issue by correcting the `:deep()` definition got lost during the migration.

Navigate to `Ingress edit > Rule path` (e.g.: `/c/local/explorer/networking.k8s.io.ingress/create#rules`

### Screenshot/Video

Screenshot 2024-09-23 at 18 05 46

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/edit/networking.k8s.io.ingress/RulePath.vue
+++ b/shell/edit/networking.k8s.io.ingress/RulePath.vue
@@ -181,36 +181,27 @@ export default {
   </div>
 </template>
 <style lang="scss" scoped>
+// TODO #11952: Correct deep statement
 $row-height: 40px;
-
-.labeled-input :deep(), :deep() .labeled-input {
-  padding: 0 !important;
-  height: 100%;
-  input.no-label {
-    height: calc($row-height - 2px);
-    padding: 10px;
-  }
-}
-.rule-path :deep() {
-  .col, INPUT {
-    height: $row-height;
-  }
-
-  .unlabeled-select {
+.rule-path {
+  :deep(.labeled-input), :deep(.labeled-input) {
+    padding: 0 !important;
     height: 100%;
-  }
 
-  .path-type {
-    .unlabeled-select {
-      min-width: 200px;
+    input.no-label {
+      height: calc($row-height - 2px);
+      padding: 10px;
     }
   }
 
-  &, .input-container {
+  :deep(.col), INPUT {
+    height: $row-height;
+  }
+  &, :deep(.input-container) {
     height: $row-height;
   }
 
-  .input-container .in-input.unlabeled-select {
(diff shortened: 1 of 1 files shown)


shell/edit/networking.k8s.io.ingress/RulePath.vue
```diff
@@ -181,48 +181,44 @@ export default {
...
-  }
-}
-.rule-path :deep() {
-  .col, INPUT {
-    height: $row-height;
-  }
-
-  .unlabeled-select {
+.rule-path {
+  :deep(.labeled-input), :deep(.labeled-input) {
```
rak-phillip, line 187: I'm curious as to why we need the double `:deep()` selectors 🤔

Something tells me that we leave this as is for now and we can circle back to better understand the intent in the future..

rak-phillip (approved):
LGTM - I'm only noticing a minor spacing difference in 2.10 vs 2.9. We might want to figure it out if we intend for this to be a 1:1 match with what we had before.

## v2.10
image

## v2.9
image

cnotv:
Wait why do you see it like this? It should not have gaps. Is it with Firefox?

shell/edit/networking.k8s.io.ingress/RulePath.vue
cnotv: Maybe it had a reason before, but not now 😅

rak-phillip:
> Wait why do you see it like this? It should not have gaps. Is it with Firefox?

That's correct, I'm on Firefox. I'm willing to bet that this might have been an unintentional rendering bug that was left unreported.

cnotv:
> > Wait why do you see it like this? It should not have gaps. Is it with Firefox?
> 
> That's correct, I'm on Firefox. I'm willing to bet that this might have been an unintentional rendering bug that was left unreported.

This is how it looks for me on FF 130.0 (64-bit)
Screenshot 2024-09-27 at 14 32 07

cnotv:
> > Wait why do you see it like this? It should not have gaps. Is it with Firefox?
> 
> That's correct, I'm on Firefox. I'm willing to bet that this might have been an unintentional rendering bug that was left unreported.

Ok found out the issue: it's the UI to is not responsive, the same problem happens on Chrome. We probably are full of these types of issues which we are not going to tackle now.

rak-phillip (approved):

PR #12024, opened 2024-09-25
Allow users to specify that an array list is required

codyrancher:
### Summary
This is in support of Harvester. The forked version of ArrayListSelect was using this field in this way:
image

You can add the required field to any array list which has a title.

Shouldn't be any. Respects the existing default of false.

### Screenshot/Video
This change adds the exact same visual that was present in harvester even though I don't love it:
arraylist-required

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/form/ArrayList.vue
+++ b/shell/components/form/ArrayList.vue
@@ -80,6 +80,10 @@ export default {
       type:    Boolean,
       default: false,
     },
+    required: {
+      type:    Boolean,
+      default: false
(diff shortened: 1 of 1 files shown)


rak-phillip (approved):
> This change adds the exact same visual that was present in harvester even though I don't love it:

Yeah, I don't love it either.. since this is an existing implementation, I think that we make do with what we have and enhance later.

PR #12045, opened 2024-09-26
Create new `ButtonMultiAction.vue` component

### LINKED ISSUE #12007: Project/Namespace menu action is not borderless
# Description

Something which initially seemed a `:deep()` issue (https://github.com/rancher/dashboard/issues/11952), seems to be related to some other changes due Vue3.

Seems like the class `role-multi-action` is not added in the markup of the same file/component and therefore requires not a non-scoped definition.

https://github.com/rancher/dashboard/blob/master/shell/components/SortableTable/index.vue#L1664C1-L1668

```scss
  // Remove colors from multi-action buttons in the table
  td {
    .actions.role-multi-action {
      background-color: transparent;
      border: none;
```

Screenshot 2024-09-24 at 18 10 39

rak-phillip:
### Summary
This creates a new button component, `ButtonMultiAction.vue`, and replaces two targeted instances to address styling issues with Projects/Namespaces. 

There are other instances that can be replaced, but they have been left untouched to limit the scope of this change.

Fixes #12007 

- Create new component, `ButtonMultiAction.vue`
- Replace in-table multi action button
- Replace in-table multi action button in `shell/components/ExplorerProjectsNamespaces.vue` 

### Technical notes summary

While investigating the issue, it became clear that there is a difference between how scoped styles are applied to slot content in Vue3 when compared to previous versions of Vue. Looking at the similarities between the two implementations told me that a component could be created.

An alternative approach would be to duplicate the class in `shell/components/ExplorerProjectsNamespaces.vue`. 

Relying on unscoped styles is another option, but I prefer the component-driven approach. 

- Projects/Namespaces
- Sortable Table

- Projects/Namespaces
- Sortable Table

#### This PR

image

#### v2.9

image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/ButtonMultiAction.vue
+++ b/shell/components/ButtonMultiAction.vue
@@ -1,0 +1,41 @@
+<script setup lang="ts">
+import { computed, defineEmits } from 'vue';
+
+defineEmits(['click']);
+
+type Props = {
+  borderless?: boolean;
+  invisible?: boolean;
+}
+
+const props = defineProps<Props>();
+
+const buttonClass = computed(() => {
+  return {
+    borderless: props?.borderless,
+    invisible:  props?.invisible,
+  };
+});
+</script>
+
+<template>
+  <button
+    type="button"
+    class="btn btn-sm role-multi-action actions"
+    :class="buttonClass"
+    @click="(e: Event) => $emit('click', e)"
+  >
+    <i class="icon icon-actions" />
+  </button>
+</template>
+
+<style lang="scss" scoped>
+.borderless {
+  background-color: transparent;
+  border: none;
+  &:hover, &:focus {
+    background-color: var(--accent-btn);
+    box-shadow: none;
+  }
+}
+</style>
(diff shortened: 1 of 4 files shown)


shell/components/ButtonMultiAction.vue
```diff
@@ -0,0 +1,41 @@
...
+}
+
+const props = defineProps<Props>();
+
+const buttonClass = computed(() => {
+  return {
+    borderless: props?.borderless,
+    invisible:  props?.invisible,
+  };
+});
```
cnotv, line 18: Sorry for asking, but would this not be more concise?
Not sure why `prop` may be undefined in your case as well, since it's already initialized with `defineProps`.

```suggestion
const props = defineProps<{
  borderless?: boolean | undefined;
  invisible?: boolean | undefined;
}>();

const buttonClass = computed(() => {
  return {
    borderless: props.borderless,
    invisible:  props.invisible,
  };
});
```

cnotv (approved):
Thanks for the initiative to pick the issue up and use the composition API connotation 🥇 
I added a comment but it's intended in a conversational manner, not a blocker.

LGTM

shell/components/ButtonMultiAction.vue
rak-phillip: It really just comes down to preference whether we want to move props into a separate type or not. I'm open to either approach.

cnotv: Yeah not a blocker as I said 🍻

PR #12071, opened 2024-09-30
Fix: Vue3: Installed extensions throw unhandled exception

### LINKED ISSUE #11695: Vue3: Installed extensions throw unhandled exception
Navigating to the local cluster while an extension is installed can throw an unhandled error, causing Dashboard to stop functioning.

## Reproduction steps

In a stable Rancher instance running dashboard with Vue2, install the following extensions from the [ui-plugin-examples](https://github.com/rancher/ui-plugin-examples) repository. The following extensions were installed for this test:

- Clock
- Extensions-Api-Demo
- Large Extension

When starting a development server, make sure that you use the stable instance of Rancher as the backend for testing.

In the development version of Rancher running dashboard with Vue3, navigate to the Local Cluster Explorer.

## Actual behavior

An error overlay is rendered with the following:

```
Uncaught (in promise) TypeError: can't access property "extend", c.a is undefined
    eb32 Banner.vue:6
    Webpack 128
[Banner.vue:6:15](webpack://extensions-api-demo-0.1.0//Users/aalves/Desktop/PROJECTS/alex-ui-plugin-examples/node_modules/@rancher/shell/rancher-components/Banner/Banner.vue?ca18)
```

Other warnings are logged to the console:

```
[Vue warn]: Invalid VNode type: undefined (undefined) 
  at <Component resource= 
Object { id: "local", type: "management.cattle.io.cluster", links: {…}, actions: {…}, apiVersion: "management.cattle.io/v3", kind: "Cluster", metadata: {…}, spec: {…}, status: {…}, … }
 > 
  at <SimpleBox key="extensionCards0" class="extension-card" style=undefined > 
  at <ClusterExplorerIndexPage class="outlet" onVnodeUnmounted=fn<onVnodeUnmounted> ref=Ref< undefined > > 
  at <RouterView key="/c/local/explorer" class="outlet" > 
  at <Default onVnodeUnmounted=fn<onVnodeUnmounted> ref=Ref< null > > 
  at <RouterView key=1 > 
  at <App>
```

## Expected behavior

The Local Cluster Explorer renders the same way that it does without extensions installed.

### LINKED ISSUE #12034: Improve feedback given when we forcefully don't load an extension
This issue  is dependent on https://github.com/rancher/dashboard/issues/12001 (and possibly https://github.com/rancher/dashboard/pull/12018 which I would include in that forward-port, imho)

We currently only provide feedback to the root cause of an extension not loaded by the UI in case the kube version check is not met: 
https://github.com/rancher/dashboard/blob/release-2.9/shell/pages/c/_cluster/uiplugins/index.vue#L406-L408

despite the checks done cover more scenarios: 
https://github.com/rancher/dashboard/blob/release-2.9/shell/config/uiplugins.js#L114-L159

We should improve this feedback mechanism to cover the rest of checks being done, along with a bit of refactoring inspired in https://github.com/rancher/dashboard/pull/12018 in order to abstract the messaging from being too dispersed throughout the extensions code

### LINKED ISSUE #12075: Extensions install/uninstall loading process stuck
**Describe the bug**
When installing or uninstalling any extension, the loader is stuck and requires a force upload to display and access the configuration.

**To Reproduce**
- Navigate to the extensions page
- Install any extension
- The loader should start spin and/or redirect to the new tab

**Result**
The extension still not listed as installed, or still displays uninstall status

**Expected Result**
Process is completed with clear UX and data

**Screenshots**

https://github.com/user-attachments/assets/b42c07ab-856a-453f-994b-d37289f2e4f9

**Additional context**

E2E tests are failing as result.

aalves08:
### Summary
Fixes #11695 
Fixes #12075 
Fixes #12034 (we are now covering all annotation cases and providing proper feedback)

- Add missing connection in `vue.config` to allow population of `process.env.UI_EXTENSIONS_API_VERSION` variable
- Add proxy for `/version` endpoint in `vue.config`
- Remove duplicated mechanism of error display in slide-in extensions panel for details `PluginInfoPanel`
- add default fallback for `ui-extensions-version` annotation to be `> 3.0.0`
- Make watches in `uniplugins/index.vue` as `deep: true` in order to reconnect proper triggers for reload banner and status update that were lost with the Vue3 migration
- Remove “duplicate” check for kube version that wasn’t preventing loading of extension but simply informing the user that it wasn’t compatible in `uniplugins/index.vue`
- Move `/rancherversion` network request to `shell/plugins/plugin` where loading order for extensions in order to correctly populate the rancher version for `shouldNotLoadPlugin` check
- Add `/version` network request to `shell/plugins/plugin` in order to correctly populate the kube version for `shouldNotLoadPlugin` check
- Extensions code refactoring
- Add custom message in case the Extensions API annotation is missing `catalog.cattle.io/ui-extensions-version`
- Address graceful error degradation because of too many retries fetching `secrets` for Catalog part
 of the main extensions screen - not blocking the UI (check screenshot)
 
Screenshot 2024-10-01 at 16 00 46

- Load an extension with the required API version `>= 3.0.0`
- Load an extension with the required API version `>= 2.0.0` 
- Load an extension without the `catalog.cattle.io/ui-extensions-version` annotation

### Screenshot/Video

**Deprecated error banner in slide-in panel**
Screenshot 2024-09-27 at 12 08 16

**Output of /version endpoint**
Screenshot 2024-09-30 at 09 38 50

**Error display with loading prevented  - `shouldNotLoadPlugin` check **
Screenshot 2024-09-30 at 09 40 43
Screenshot 2024-09-30 at 16 04 41

**Warning display when there's a newer version of an extension BUT it's not compatible**
Screenshot 2024-09-30 at 11 05 37
Screenshot 2024-09-30 at 16 03 38

**Custom message if no `catalog.cattle.io/ui-extensions-version` annotation is present**
Screenshot 2024-10-02 at 08 19 48

### Checklist
- [x] The PR has a Milestone 


shell/config/uiplugins.js
```diff
@@ -3,44 +3,37 @@ import semver from 'semver';
...
+};
 
-export const UI_PLUGINS_COMMUNITY_REPO_URL = 'https://github.com/rancher/community-extensions';
-export const UI_PLUGINS_COMMUNITY_REPO_BRANCH = 'main';
+// Info for the Helm Chart Repositories
+export const UI_PLUGINS_REPOS = {
+  OFFICIAL: {
+    NAME:   'rancher-ui-plugins',
+    URL:    'https://github.com/rancher/ui-plugin-charts',
+    BRANCH: 'main',
```
richard-cox, line 24: Question only. in `master` these will point to `main`. In future versions these will change to one related to rancher version branches of those repos?

shell/config/uiplugins.js
```diff
@@ -57,123 +50,132 @@ export const UI_PLUGIN_CHART_ANNOTATIONS = {
...
- * Get value of the annotation from teh latest version for a chart
+ * Get value of the annotation from the latest version for a chart
  */
 export function uiPluginAnnotation(chart, name) {
-  if (chart?.versions?.length > 0) {
-    return chart.versions[0].annotations?.[name];
-  }
-
-  return undefined;
+  return chart?.versions?.[0]?.annotations?.[name];
```
richard-cox, line 99: Comment only.

:heavy_check_mark: :heavy_check_mark:  :heavy_check_mark:

shell/config/uiplugins.js
```diff
@@ -57,123 +50,132 @@ export const UI_PLUGIN_CHART_ANNOTATIONS = {
...
   if (v.includes('-') && splitArr?.length === 3) {
-    parsedRancherVersion = `${ splitArr[0] }.${ splitArr[1] }.999`;
+    parsedVersion = `${ splitArr[0] }.${ splitArr[1] }.999`;
   }
 
-  return parsedRancherVersion;
+  return parsedVersion;
 }
 
-// i18n-uses plugins.error.generic, plugins.error.api, plugins.error.host
```
richard-cox, line 144: This has been lost. This informs the process that checks strings are used... that these strings are used here. Looks like there still are, plus some new

shell/config/uiplugins.js
```diff
@@ -3,44 +3,37 @@ import semver from 'semver';
 // Version of the plugin API supported
 // here we inject the current shell version that we read in vue.config
 export const UI_EXTENSIONS_API_VERSION = process.env.UI_EXTENSIONS_API_VERSION;
+const MINIMUM_MANDATORY_EXTENSIONS_API_VERSION = `> ${ UI_EXTENSIONS_API_VERSION }`;
```
richard-cox, line 6: Shouldn't this be `>=`?

cnotv:
Just testing the UI: the issue with install/uninstall seems fine. Once only the uninstall did not bulge, but then it was ok.

https://github.com/user-attachments/assets/269596b6-0ccc-466d-9ca7-808b32ab76e6

shell/config/uiplugins.js
```diff
@@ -57,123 +50,132 @@ export const UI_PLUGIN_CHART_ANNOTATIONS = {
...
     return 'plugins.error.generic';
   }
 
   // Extension chart specified a required extension API version
   // we are propagating the annotations in pkg/package.json for any extension
   // inside the "spec.plugin.metadata" property of UIPlugin resource
-  const requiredUiExtensionsVersion = UIPluginResource.spec?.plugin?.metadata?.[UI_PLUGIN_CHART_ANNOTATIONS.EXTENSIONS_VERSION];
+  // the minimum version is enforced to prevent loading of extensions
+  // that don't have the annotation 'catalog.cattle.io/ui-extensions-version' present
+  const requiredUiExtensionsVersion = UIPluginResource.spec?.plugin?.metadata?.[UI_PLUGIN_CHART_ANNOTATIONS.EXTENSIONS_VERSION] || MINIMUM_MANDATORY_EXTENSIONS_API_VERSION;
```
richard-cox, line 150: Are we sure this is right? We're falling back to the minimum required version being the current shell package rather than a hardcoded 3.0.0

shell/plugins/plugin.js
```diff
@@ -26,38 +27,69 @@ export default async function(context) {
...
+    // Fetch rancher version metadata
+    reqs.push(context.store.dispatch('rancher/request', {
+      url:                  '/rancherversion',
+      method:               'get',
+      redirectUnauthorized: false
+    }));
+
+    // Fetch kubernetes version metadata
+    reqs.push(context.store.dispatch('rancher/request', {
+      url:                  '/version',
```
richard-cox, line 44: We need to bear in mind both this and /rancherversion could always fail on some systems (if not already) given the advise given out in [docs](https://documentation.suse.com/external-tree/en-us/cloudnative/rancher-manager/v2.9/en/security/rancher-security-best-practices.html#_restrict_public_access_to_version_and_rancherversion_path) (we were not consulted). To confirm if either are missing this process won't fall over, we just might load extensions we shouldn't?

richard-cox (changes requested):
Haven't given it a functional test yet, but scanned through code

shell/assets/translations/en-us.yaml
```diff
@@ -4286,9 +4286,10 @@ plugins:
     title: Error loading extension
     message: Could not load extension code
     generic: Extension error
-    api: This Extension is not compatible with the current Extensions API version
-    host: This Extension is not compatible with this application
-    version: This Extension is not compatible with this version of Rancher
+    api: Loading error! The version installed is not compatible with the current Extensions API version
```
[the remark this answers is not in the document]
aalves08: Sounds good mate. Will change to `Unable to load Extension`

shell/plugins/plugin.js
```diff
@@ -26,38 +27,69 @@ export default async function(context) {
...
-    // TODO: Get rancher version using the new API (can't use setting as we have not loading the store)
-    const rancherVersion = undefined;
+    let rancherVersion;
+    let kubeVersion;
+
+    const reqs = [];
+
+    // Fetch rancher version metadata
+    reqs.push(context.store.dispatch('rancher/request', {
+      url:                  '/rancherversion',
```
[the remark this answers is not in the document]
aalves08: The problem is that the loading of the extensions is unaware of the promise/await in the intialisation of the `version` vue plugin (file that I deleted).

They are executed "kinda" simultaneously (check here: https://github.com/rancher/dashboard/blob/master/shell/initialize/install-plugins.js#L46-L57) and `rancherVersion` will never get populated in the context of `shell/plugins/plugin.js`.

I've looked into this and I didn't see any kind of problem, since the data for `/rancherversion` is populating a JS variable for propagation/usage and it's network request is "kinda" done at the same time, still.

Worth to note that the JS variable is still being populated though, so transmission of data should be assured.

Is there any other problem you see with this that I am not aware? Because the way it was being done (which I kept) didn't not assure that the data was there at a given point in time because of the data awareness. Wdyt?

shell/vue.config.js
```diff
@@ -362,6 +363,10 @@ const getVirtualModulesAutoImport = (dir) => {
   return new VirtualModulesPlugin(autoImportTypes);
 };
 
+// Get current shell version
+const shellPkgRawData = fs.readFileSync(path.join(__dirname, 'package.json'));
```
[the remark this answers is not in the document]
aalves08: Seems to work just fine. Thanks for the feedback! 🙏

shell/config/uiplugins.js
```diff
@@ -3,44 +3,37 @@ import semver from 'semver';
...
-export const UI_PLUGINS_COMMUNITY_REPO_NAME = 'community-extensions';
+export const UI_PLUGIN_ANNOTATION = {
+  NAME:  'catalog.cattle.io/ui-component',
+  VALUE: 'plugins',
+};
 
-export const UI_PLUGINS_COMMUNITY_REPO_URL = 'https://github.com/rancher/community-extensions';
-export const UI_PLUGINS_COMMUNITY_REPO_BRANCH = 'main';
+// Info for the Helm Chart Repositories
+export const UI_PLUGINS_REPOS = {
```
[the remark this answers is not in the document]
aalves08: Kudos to @jordojordo for that one! 👏 👏 👏

shell/config/uiplugins.js
aalves08: Will add it back 🙏

shell/config/uiplugins.js
aalves08: Tbh, I don't see it evert changing in the future. 

These are consumed by our end users and if we couldn't change the `ui-plugin-examples` repo from `main` to `gh-pages`, these won't be allowed for sure.

shell/config/uiplugins.js
aalves08: We can go both ways here:

Either we hardcode it to `>= 3.0.0`, but have to be mindful to change this in the future whenever we change the `UI_EXTENSIONS_API_VERSION` or we keep it as I've done where we don't need to worry about it `>= ${ UI_EXTENSIONS_API_VERSION }`.

The goal here is to prevent loading of an extension that doesn't have the annotation `catalog.cattle.io/ui-extensions-version`, right @nwmac ?

If you think we should hardcode it, it's a 30sec fix but there's a drawback that you should be aware.

shell/plugins/plugin.js
aalves08: Unfortunately, through other places, I was already aware of that 😢 

Since the script is based on a `Promise.allSettled`, as long as 

```
reqs.push(context.store.dispatch('management/request', {
      url:                  `${ UI_PLUGIN_BASE_URL }`,
      method:               'GET',
      headers:              { accept: 'application/json' },
      redirectUnauthorized: false,
    }));
```
doesn't fail, extensions are still loaded. The side-effect is that we won't be able to prevent loading of extensions that are already installed based on `rancherVersion` and `kubeVersion`.

Bear in mind that if `/rancherversion` get blocked, we have the `prime` mechanism that get's broken as well and that's much more important, imo 😛

aalves08:
@richard-cox , update:

We've decided to go for a custom message in case the `catalog.cattle.io/ui-extensions-version` annotation is missing since the previous one lacked information that the annotation is mandatory from now on.

_Sample message on tooltip with button disabled on slide-in panel_
Screenshot 2024-10-02 at 08 19 48

With this we got rid of the `MINIMUM_MANDATORY_EXTENSIONS_API_VERSION ` as it's not needed 🙏

shell/plugins/plugin.js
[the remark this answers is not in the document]
richard-cox: If `loadPlugins` is false we never make the request, so anything that depends on rancherversion will be broken

shell/config/uiplugins.js
richard-cox: Could find these, or the new entries, sure they were added?

shell/config/uiplugins.js
richard-cox: The question is should it be 3.0.0 forever, or should it always be the shell that rancher was shipped with. I thought it was the former, but i can see how the later would force people to set an annotation.

shell/plugins/plugin.js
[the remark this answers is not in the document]
aalves08: That's true @richard-cox . Forgot about that. Let me adjust the script to always do the `/rancherversion` network request 🙏

shell/assets/translations/en-us.yaml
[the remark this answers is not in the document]
nwmac: Thanks @aalves08 - I think I agree with Richard

shell/config/uiplugins.js
aalves08: Sure thing. Just check the changes in `en-us.yaml` around line 4290

richard-cox: as in the comment has been added back in and updated with the new strings.

it would help if i could type as well. `Couldn't find these, or the new entries, sure they were added?`

shell/config/uiplugins.js
richard-cox: Chatted with Neil (see slack).

If the annotation is missing we should assume it's `< 3.0.0` (i.e not compatible with the vue3 bump)

richard-cox: Ok, got to the bottom comment.

So now if the annotation is missing, in a shell 3.0.0+ world, we won't load the extension and instead show a specific warning. Suggestion on that wording incoming, but overall good solution

shell/assets/translations/en-us.yaml
```diff
@@ -4286,9 +4287,11 @@ plugins:
     title: Error loading extension
     message: Could not load extension code
     generic: Extension error
-    api: This Extension is not compatible with the current Extensions API version
-    host: This Extension is not compatible with this application
-    version: This Extension is not compatible with this version of Rancher
+    apiAnnotationMissing: Unable to load Extension. The version installed does not have the mandatory annotation 'catalog.cattle.io/ui-extensions-version' from Rancher 2.10 and onwards
```
richard-cox, line 4290:
```suggestion
    apiAnnotationMissing: Unable to load Extension. The version installed is not compatible with the current Extensions API Version (Extension is missing the annotation 'catalog.cattle.io/ui-extensions-version' which implies incompatibility with Rancher 2.10 and onwards)
```

This will steer extension developers away from just sticking the annotation in

shell/assets/translations/en-us.yaml
```diff
@@ -4308,6 +4311,7 @@ plugins:
     versionError: Could not load version information
     requiresRancherVersion: "Requires Rancher {required}"
     requiresKubeVersion: "Requires Kube version {required}"
+    requiresExtensionApiVersionMissing: "Missing the mandatory annotation 'catalog.cattle.io/ui-extensions-version' from Rancher 2.10 and onwards"
```
richard-cox, line 4314:
```suggestion
    requiresExtensionApiVersionMissing: "Missing the annotation 'catalog.cattle.io/ui-extensions-version' which implies incompatibility with Rancher 2.10 and onwards"
```

shell/plugins/plugin.js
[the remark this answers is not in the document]
aalves08: PR updated

aalves08:
@richard-cox did a final round of tests with your changes and only spotted 1 problem with the paths for the annotations in `shouldNotLoadPlugins` function. After that got fixed, it all checked out 🙏  Just pushed the fix.

From my side, LGTM!

richard-cox (approved):

PR #12098, opened 2024-10-02
Mutate growl stack in a reactive way

### LINKED ISSUE #12097: Vue3: Cluster Management: Clusters: Copy KubeConfig to Clipboard: Toast message does not disappear after 3-4 seconds
**Setup**
- Rancher version: v2.10
- Rancher UI Extensions:
- Browser type & version: Firefox, Chrome

**Describe the bug**

**To Reproduce**
1. go to Cluster Management > Clusters
2. click ellipsis for any cluster within the cluster list
3. click 'Copy KubeConfig to Clipboard'
4. notice the toast message does not disappear after 3-4 seconds (unless the user interacts with the page)

**Result**
The toast message does not disappear after 3-4 seconds.
- unless the user interacts with the page (hover/click another page element)
- user clicks close button

Note: once the toast message disappears, the issue subsides but is repro'able after a page reload.

**Expected Result**
The toast message should disappear after 3-4 seconds (unless the user interacts with the page).

**Screenshots**

https://github.com/user-attachments/assets/a10d9c17-7532-4cf1-88ed-c375cb1410f5

**Additional context**
e2e test (`can copy config to clipboard`) consistently fails with `Timed out retrying after 4000ms: Expected <div.growl-text> not to exist in the DOM, but it was continuously found.`

rak-phillip:
### Summary
This replaces the growl `stack` array so that it's guaranteed the watcher in `shell/components/GrowlManager.vue` will more reliably trigger when changes are made.

https://github.com/rancher/dashboard/blob/9489da3568db485d75b2dae49b580df4c6099f2b/shell/components/GrowlManager.vue#L17-L25

Fixes #12097 

- Use the spread operator to build a new array instead of using push to reliably trigger reactivity

### Technical notes summary

While testing, I noted that the behavior was very inconsistent. There were times that the countdown timer would init and others where nothing would happen. 

We've seen this behavior in other areas where mutations methods aren't reliably triggering reactivity. I think that it makes sense to follow-up with a minimal repro to understand why exactly this can be an issue. The Vue docs explicitly state that we should be able to use `Array.push()` to mutate an array in a reactive way[^1].

[^1]: https://vuejs.org/guide/essentials/list.html#mutation-methods

- Growls

- Growls

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/store/growl.js
+++ b/shell/store/growl.js
@@ -18,11 +18,14 @@ export const getters = {
 
 export const mutations = {
   add(state, data) {
-    state.stack.push({
-      id:      state.nextId++,
-      started: (new Date().getTime()),
-      ...data
-    });
+    state.stack = [
+      ...state.stack,
+      {
+        id:      state.nextId++,
+        started: (new Date().getTime()),
+        ...data
+      }
+    ];
   },
 
   remove(state, id) {


shell/store/growl.js
```diff
@@ -18,11 +18,14 @@ export const getters = {
 
 export const mutations = {
   add(state, data) {
-    state.stack.push({
-      id:      state.nextId++,
-      started: (new Date().getTime()),
-      ...data
-    });
+    state.stack = [
```
eva-vashkevich, line 21: Push was appending elements to the end of the state stack, while this is appending it to the beginning. is this change intentional?

rak-phillip: Definitely not intentional, good catch!

eva-vashkevich (approved):
lgtm

PR #12129, opened 2024-10-04
fix bug with LabeledSelect with "multiple" attribute

### LINKED ISSUE #12117: Vue3: Bug with LabeledSelect with "multiple" attribute
**Setup**
- Rancher version: Any + Vue3 frontend
- Rancher UI Extensions:
- Browser type & version:

**Describe the bug**

Using `LabeledSelect` with option `multiple="true"` is not working properly

**To Reproduce**

- Create some secrets of type `kubernetes.io/dockercfg` (at least a couple)
- Create Registry Secret if missing
- Go to Deployments -> Create -> go to the `Pull Secrets` select on the `General` tab and select an option

**Result**
- Throws an error

**Expected Result**

- Should work fine

**Screenshots**

**Vue3** (fails)
https://github.com/user-attachments/assets/ede5a1ec-432a-4ba9-9562-1131ca6b5276

**2.9-head** (works)
https://github.com/user-attachments/assets/52ba106a-4e41-4921-8dca-515210b9d017

**Additional context**

I believe it's related to the option `multiple="true"`
https://vue-select.org/guide/values.html#single-multiple

If we add the attribute `multiple` to  the `v-select`, the output is an array and it works fine, so the binding of the attributes is where the problem lies

aalves08:
### Summary
Fixes #12117 

- fix issue with `LabeledSelect` in mode multiple by allowing the binding of the attribute to the element `v-select` that was incorrectly being filtered out

- Test repro case(s) in issue + test some other areas that use `LabeledSelect` with `multiple="true"`

- Just to discard any regressions, do a quick test of different instances of `LabeledSelect`, even if they don't use  `multiple="true"`

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/form/LabeledSelect.vue
+++ b/shell/components/form/LabeledSelect.vue
@@ -140,7 +140,6 @@ export default {
       const {
         class: _class,
         taggable,
-        multiple,
         ...rest
       } = this.$attrs;
 
--- a/shell/components/form/__tests__/LabeledSelect.test.ts
+++ b/shell/components/form/__tests__/LabeledSelect.test.ts
@@ -1,5 +1,6 @@
 import { mount } from '@vue/test-utils';
 import LabeledSelect from '@shell/components/form/LabeledSelect.vue';
+import { defineComponent } from 'vue';
 
 describe('component: LabeledSelect', () => {
   describe('should display correct label', () => {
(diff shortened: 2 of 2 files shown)


cnotv:
I am not so sure this is the issue. When I searched for other cases, I could not reproduce them equally. Investigating the issue specifically, resulted in trying to map a string of the term:

Screenshot 2024-10-04 at 12 51 50

In your PR this issue is fixed, although I suspect there's something further around the attributes clashing here.

Screenshot 2024-10-04 at 12 54 39

cnotv:
I've added a unit test that seems to validate the behavior change from string to array in case of multiple.
The other case has a different event handler which is not a bidirectional binding.

cnotv (approved):
LGTM

Storybook seems not achievable atm

Issue #12099, opened 2024-10-02
Improve e2e test retry behavior


mantis-toboggan-md:
When e2e tests fail they are automatically retried twice. Some tests are structured in such a way that these retries typically do not work. We regularly have to manually re-run these tests. We may also have tests that tend to fail in before/after hooks, which do not re-run on failure. We need to determine which tests are flaky and re-examine the logic in their before/after hooks, potentially moving more logic to beforeEach/afterEach. Initial investigation suggests these specs are failing regularly and should be examined:

- pages/manager/repositories.spec.ts 
- explorer/apps/repositories.spec.ts 
- pages/explorer2/resource-search.spec.ts xx 
- users-and-auth/index.spec.ts 
- explorer/more-resources/rbac/roles.spec.ts 
- pages/fleet/fleet-clusters.spec.ts  
- cluster-manager.spec.ts 
- pages/charts/rancher-backup.spec.ts

codyrancher:
@mantis-toboggan-md would it be okay to close this given the improved retry logic that's landed in `master` in the past couple months?

mantis-toboggan-md:
Yeah let's call it resolved.

PR #12178, opened 2024-10-08
make cluster repository tests more repeatable

mantis-toboggan-md:
### Summary
Partially addresses #12099 

### Technical notes summary
Refactored cluster repository tests into two categories:
1. Tests that verify some aspect of creating repositories
2. Tests that verify some aspect of interacting with existing repositories

Category 2 tests now have a beforeEach block that creates a new cluster repository for each test, and an after hook to delete all repositories created during these test runs.

I also fixed an issue with the repo+ssh key test -- that repo may very briefly have an 'active' state before switching to 'in progress' so verifying its creation by checking for 'active' status was flaky. I rewrote the test to verify that the right network requests are sent when saving the repository instead of waiting for the repository's status to change.

### Checklist
- [ ] The PR has a Milestone 

### DIFF
--- a/cypress/e2e/tests/pages/manager/repositories.spec.ts
+++ b/cypress/e2e/tests/pages/manager/repositories.spec.ts
@@ -17,11 +17,6 @@ describe('Cluster Management Helm Repositories', { testIsolation: 'off', tags: [
 
   describe('Create', () => {
     beforeEach(() => {
-      // const name = `e2e-repo-${ Math.random().toString(36).substr(2, 6) }`;
-      // cy.wrap(name).as('repoName');
-
-      // testRepos.push(name);
-
       cy.createE2EResourceName('repo').as('repoName');
       cy.get('@repoName').then((name) => testRepos.push(name));
     });
--- a/cypress/e2e/tests/pages/manager/repositories.spec.ts
+++ b/cypress/e2e/tests/pages/manager/repositories.spec.ts
@@ -17,10 +17,13 @@ describe('Cluster Management Helm Repositories', { testIsolation: 'off', tags: [
 
   describe('Create', () => {
     beforeEach(() => {
-      const name = `e2e-repo-${ Math.random().toString(36).substr(2, 6) }`;
+      // const name = `e2e-repo-${ Math.random().toString(36).substr(2, 6) }`;
+      // cy.wrap(name).as('repoName');
 
-      cy.wrap(name).as('repoName');
-      testRepos.push(name);
+      // testRepos.push(name);
+
+      cy.createE2EResourceName('repo').as('repoName');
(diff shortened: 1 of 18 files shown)


cypress/e2e/tests/pages/manager/repositories.spec.ts
```diff
@@ -2,252 +2,271 @@ import PromptRemove from '@/cypress/e2e/po/prompts/promptRemove.po';
...
   before(() => {
     cy.login();
   });
 
-  beforeEach(() => {
-    cy.createE2EResourceName('repo').as('repoName');
-  });
+  describe('Create', () => {
+    beforeEach(() => {
+      const name = `e2e-repo-${ Math.random().toString(36).substr(2, 6) }`;
```
mantis-toboggan-md, line 20: Looking at some failed/repeated sorry cypress runs I noticed that the name created by `createE2EResourceName` was the same when tests were retried, leading tests that involved creating a resource with repoName to error out on all subsequent retries.

I wasn't sure if that behavior was an intentional and/or necessary aspect of `createE2EResourceName` so I didn't modify that method. If it is intentional perhaps we need another convenience method to randomly generate names that _wont_ be repeated on test retry.

cypress/support/commands/commands.ts
```diff
@@ -27,6 +27,15 @@ Cypress.Commands.add('keyboardControls', (triggerKeys: any = {}, count = 1) => {
   }
 });
 
+/**
+ * wrapper around cypress.contains to search for exact strings eg find ONLY spans containing 'test-resource' and not 'test-resource-123-abc'
+ */
+Cypress.Commands.add('containsExactly', (searchString) => {
```
mantis-toboggan-md, line 33: We have PO methods that should probably be updated to use this, eg https://github.com/rancher/dashboard/blob/master/cypress/e2e/po/components/sortable-table.po.ts#L108-L114 but I could see that being its own PR

cypress/support/commands/commands.ts
```diff
@@ -67,7 +76,9 @@ Cypress.Commands.add('iFrame', () => {
 const runTimestamp = +new Date();
 
 Cypress.Commands.add('createE2EResourceName', (context) => {
-  return cy.wrap(`e2e-test-${ runTimestamp }-${ context }`);
+  const randomStr = Math.random().toString(36).substr(2, 6);
```
mantis-toboggan-md, line 79: Cluster repository tests were calling `createE2EResourceName` in a `beforeEach` hook but I noticed that the same value was generated each time. I think this means that tests which use this method to name the resources they're attempting to create wont retry correctly, as they'll be trying to create resources with duplicate names

nwmac:
@mantis-toboggan-md Do we still aim to get this in for 2.10?

PR #12263, opened 2024-10-15
update vsphere machine config to use new networks endpoint

### LINKED ISSUE #9817: Allow provisioning Vsphere RKE2 DS clusters with duplicate networks names
Internal reference: SURE-6853 & SURE-6177

Request description:
A user uses VMWare as infrastructure to provision rke2 downstream clusters. The Vsphere Distributed Switches, VDS, get networks from NSX-T. There is only one Transport zone so all VDS get the same networks, therefore network names are duplicated. 

Actual behavior:
The Rancher UI stalls when provisioning at networks tab and the provisioning process fails.

Expected behavior:
Provisioning a new rke2 cluster when the network names are duplicated

Workaround:
Provisioning the  cluster with cluster templates. The networks can be specified

https://github.com/rancher/cluster-template-examples/blob/main/charts/values-vsphere.yaml
 
```
  # vSphere network where the virtual machine will be attached
  # network: ""
``` 

A different workaround is using terraform.

### LINKED ISSUE #12271: [BUG] vsphere cluster creation fails when using vapp network protocol profiles
Using the latest rancher version, when creating a new vsphere cluster, it fails upon cloning the vm template to a virtual machine with the following error:

image

settings:
image

Network protocol profile is configured as following:

image

have also tried with empty domain / search but same issue

mantis-toboggan-md:
### Summary
Fixes #9817 
Fixes #12271 

This PR updates the vsphere machine config to use networks' `moid` instead of `name`.  This requires the use a new endpoint for networks, `meta/vsphere/networks-extended`, which contains extra information.

### Technical notes summary
This PR also corrects a reactivity issue in ArrayList:

https://github.com/user-attachments/assets/a0ff1d3f-e66d-40d7-b761-87e2d983192d

Create an RKE2 vsphere cluster and verify that:
 1. the machine pool is saved with the networks' moid instead of name
 2. you are able to select a network other than the default (first) option
 
Create an RKE2 vsphere cluster with at least 1 network and 'vapp' auto mode selected ('Use vapp to configure networks with network protocol profiles'), verify that the save request vapp properties reference the network name, not full path (and not moid), see https://github.com/rancher/dashboard/issues/12271#issuecomment-2415312325
 
I only had 1 network in my test environment so I added fake ones in the loadNetworks method to verify that I could add/remove different networks:

```
    async loadNetworks() {
      set(this, 'networksResults', null);

      const options = await this.requestOptions('networks-extended', this.value.datacenter);
      const content = options.map((opt) => {
        return { label: `${ opt.name } (${ opt.moid })`, value: opt.moid };
      });

      content.push({ label: 'abc-123-name (moid-123-abc)', value: 'moid-123-abc' });
      content.push({ label: 'abc-123-name (moid-456-def)', value: 'moid-456-def' });

      this.resetValueIfNecessary('network', content, options, true);

      set(this, 'networksResults', content);
    }
    
```

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/form/ArrayList.vue
+++ b/shell/components/form/ArrayList.vue
@@ -138,10 +138,14 @@ export default {
     }
   },
   watch: {
-    value() {
-      this.lastUpdateWasFromValue = true;
-      this.rows = (this.value || []).map((v) => ({ value: v }));
+    value: {
+      deep: true,
+      handler() {
+        this.lastUpdateWasFromValue = true;
+        this.rows = (this.value || []).map((v) => ({ value: v }));
+      }
     },
+
     rows: {
       deep: true,
       handler(newValue, oldValue) {
--- a/shell/machine-config/vmwarevsphere.vue
+++ b/shell/machine-config/vmwarevsphere.vue
@@ -546,8 +546,10 @@ export default {
     async loadNetworks() {
       set(this, 'networksResults', null);
 
-      const options = await this.requestOptions('networks', this.value.datacenter);
-      const content = this.mapPathOptionsToContent(options);
+      const options = await this.requestOptions('networks-extended', this.value.datacenter);
+      const content = options.map((opt) => {
+        return { label: `${ opt.name } (${ opt.moid })`, value: opt.moid };
+      });
 
       this.resetValueIfNecessary('network', content, options, true);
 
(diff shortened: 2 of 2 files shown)


momesgin (approved):
lgtm

PR #12277, opened 2024-10-16
Refactor Vue3 migration script for extensions

### LINKED ISSUE #12276: Vue3 migration script for extensions has multiple issues
**Describe the bug**

There are a number of issues with the migration script for extensions found in `shell/scripts/vue-migrate.js`.

When running the script within the Kubewarden UI Extension, these were noted:

- The script failed initially. On line 534 there is a `forEach` loop on the `eslintConfig`  `rules` property. If the `eslintConfig.rules` object does not exist the script will fail as `eslintConfig.rules[rule]` will return `eslintConfig.rules is undefined`.
- It changed files in the `/tests` directory.
  - This is outside of the scope of extensions and should be taken care of manually
  - Updated a utility file which imported Vue originally, left `Vue` artifacts within the file which will cause errors when attempting to run
  - Changed any mention of "Vue" to "vueApp". There is a constant created in this file called `localVue = createdLocalVue()` - the original constant name is the same, but subsequent method calls on this class were changed. For example:
	  ```typescript
	  // <<<<< Current
	  localVue.use(Vuex);
	  // >>>>>
	  // <<<<< Incoming
	  localvueApp.use(Vuex);
	  // >>>>>
		```
- All workflows within the `./.github/workflows/*` directory have been updated, this includes ones that are not created by the `@rancher/extension` creator.
  - Only the original files (if found) should be updated
	- Perhaps we should just show a warning in the cli output which notifies which files _should_ be updated or at least checked.
- Not all of the dependencies listed in the creator are added/updated in the root level `package.json`
	- Do these dependencies matter? Should they be removed from the creator if not?
- `.eslintrc.js` should not be modified, this is specific to each extension developer's preference.
	- We could suggest changes to this file but ultimately it should be up to the developer
- The annotations in a pkg's `package.json` file were not updated
	- If these are left unmodified, this could render their extension unusable with 2.10 or shell version 3.
- Any elements that contain a `v-for` within an element will have any additional attributes removed if it's a single line element.
	- Also the `key` will be modified from the original. For example in `./pkg/kubewarden/chart/kubewarden/admission/ContextAware/index.vue` there is a div that contained a key: 
		- ```:key="'filtered-resource-' + index"```
		- This has been modified to `:key="index"`
	- ^ Similar behavior when an element with a `v-for` loop has the `key` property in the 1st child component. For example `./pkg/kubewarden/components/InstallWizard.vue` had a `li` element with the key: `:key="step.name + 'li'"`
- An element that contains a `v-for` loop that is a multi-line element will have broken syntax.
  - For example in `./pkg/kubewarden/detail/policies.kubewarden.io.policyserver.vue:
  ```vue
  <!-- Current -->
  <CountGauge
    v-for="(group, key) in policyGauges"
    :key="key"
    :total="relatedPoliciesTotal"
    :useful="group.count || 0"
    :graphical="false"
    :primary-color-var="`--sizzle-${group.color}`"
    :name="key"
  />
  <!-- Modified -->
  <CountGauge
    v-for="(group, key) in policyGauges" :key="key":total="relatedPoliciesTotal"
    :useful="group.count || 0"
    :graphical="false"
    :primary-color-var="`--sizzle-${group.color}`"
    :name="key"
  />
  ```

- Any element that uses a `v-for` loop and has a child element that also has a `v-for` loop will be broken in the child component as a `key` must be added to the element. The nested `v-for` will contain the same variable for the `index` property, causing the error:
	  - `Variable 'i' is already declared in the upper scope`
- If there is a `$set` method that uses a dynamic value in the `key` argument, the additional `[]` will be added to the modified method.
	- For example in `./pkg/kubewarden/chart/kubewarden/policy-server/Registry/index.vue` there is a set for:
		- `this.$set(this.chartValues, [prop], this[prop]);`
		- this has been modified to: 
			- `this.chartValues[[prop]] = this[prop];`
	- Same goes for `this.$delete`, for example:
		- `this.$delete(this.value, [key]);`
		- modified to: `delete this.value[[key]];`
- Any use of `::v-deep` that used the method syntax is off. For example:
	- In `./pkg/kubewarden/components/Policies/PolicyReadmePanel.vue` there is a style for:
	  - `::v-deep(.btn-sm)`
	  - Modified to: `:deep()(.btn-sm)` - This returns a syntax error
- A component that contained a `key` property that is not within a `v-for` loop has the key property deleted.
	- For example `./pkg/kubewarden/components/Questions/Array.vue` 
- `v-popover` elements were not updated to `VDropdown`
	- Subsequent `<template #popover>` elements not updated to `<template #popper>`

jordojordo:
### Summary
Fixes #12276 

### Checklist
- [ ] The PR has a Milestone 

### DIFF
--- a/shell/scripts/vue-migrate.js
+++ b/shell/scripts/vue-migrate.js
@@ -348,6 +348,33 @@ const vueSyntaxUpdates = () => {
   const isSimpleIdentifier = (str) => /^[a-zA-Z_$][0-9a-zA-Z_$]*$/.test(str.trim());
   const isBracketedExpression = (str) => str.trim().startsWith('[') && str.trim().endsWith(']');
   const isStringLiteral = (str) => /^['"].*['"]$/.test(str.trim());
+  // Extracts the key expression from a v-for directive.
+  const extractKeyExpression = (vForContent) => {
+    const vForMatch = vForContent.match(/^\s*\(([^,]+),\s*([^)]+)\)\s+in\s+(.*)$/);
+    let keyExpression = null;
+
+    if (vForMatch) {
+      // v-for="(item, key) in items"
+      keyExpression = vForMatch[2].trim();
+    } else {
+      const simpleVForMatch = vForContent.match(/^\s*([^\s]+)\s+in\s+(.*)$/);
+
+      if (simpleVForMatch) {
+        // v-for="item in items"
+        // Use 'item' as key if it's a simple identifier
+        keyExpression = isSimpleIdentifier(simpleVForMatch[1].trim()) ? simpleVForMatch[1].trim() : null;
+      }
+    }
+
+    return keyExpression;
+  };
+  // Adds the :key attribute to a tag if it doesn't already have one.
+  const addKeyAttribute = (tag, keyExpression) => {
+    // Add space if necessary
+    const space = tag.endsWith(' ') ? '' : ' ';
+
+    return `${ tag }${ space }:key="${ keyExpression }"`;
+  };
 
   const replacementCases = [
     // Handle Vue.set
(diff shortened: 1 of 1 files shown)


shell/scripts/vue-migrate.js
```diff
@@ -312,110 +338,284 @@ const vueConfigUpdates = () => {
...
-    [`Vue.filter(`, `vueApp.filter(`, `https://v3-migration.vuejs.org/breaking-changes/global-api.html#a-new-global-api-createapp`],
-    [`Vue.mixin(`, `vueApp.mixin(`, `https://v3-migration.vuejs.org/breaking-changes/global-api.html#a-new-global-api-createapp`],
-    [`Vue.component(`, `vueApp.component(`, `https://v3-migration.vuejs.org/breaking-changes/global-api.html#a-new-global-api-createapp`],
-    [`Vue.use(`, `vueApp.use(`, `https://v3-migration.vuejs.org/breaking-changes/global-api.html#a-new-global-api-createapp`],
-    [`Vue.prototype`, `vueApp.config.globalProperties`, `https://v3-migration.vuejs.org/breaking-changes/global-api.html#a-new-global-api-createapp`],
-    ['Vue.util', '', 'Vue.util is private and no longer available https://v3-migration.vuejs.org/migration-build.html#partially-compatible-with-caveats'],
-    // [`Vue.extend`, removePlaceholder, 'https://v3-migration.vuejs.org/breaking-changes/global-api.html#vue-extend-removed'],
-    // [`Vue.extend`, `createApp({})`], //  (mixins)
+    // Handle Vue.set
+    [/\bVue\.set\(([^,]+),\s*([^,]+),\s*([^)]+)\)/g, (match, obj, prop, val) => {
```
cnotv, line 381: The idea is to have a simple regex replacement and to use rule sorting to avoid conditions. A separate function should always be outside this main function if there's a complex case that we want to handle. Comments can be added on top of the function in that case.

The whole code is meant to be one line each for readability.

shell/scripts/vue-migrate.js
```diff
@@ -312,110 +338,284 @@ const vueConfigUpdates = () => {
...
+      } else if (isStringLiteral(prop)) {
+        return `delete ${ obj }[${ prop }]`;
+      } else if (isSimpleIdentifier(prop)) {
+        return `delete ${ obj }.${ prop }`;
+      } else {
+        return `delete ${ obj }[${ prop }]`;
+      }
+    }, 'Replace this.$delete with delete operator - https://vuejs.org/guide/extras/reactivity-in-depth.html'],
+
+    // Replace Vue import with createApp and initialize vueApp
```
cnotv, line 445: The comment and the text output should be the same, and that's a repetition. It's not the end of the world, but I do not understand the purpose.

The rest of the optimization looks good.

shell/scripts/vue-migrate.js
```diff
@@ -312,110 +338,284 @@ const vueConfigUpdates = () => {
...
 const vueSyntaxUpdates = () => {
-  const files = glob.sync(params.paths || '**/*.{vue,js,ts}', { ignore: [...ignore, '**/*.spec.ts', '**/__tests__/**', '**/*.test.ts', 'jest.setup.js', '**/*.d.ts', '**/vue-shim.ts'] });
+  const files = glob.sync(params.paths || '**/*.{vue,js,ts}', { ignore: [...ignore, '**/*.spec.ts', '**/*.spec.js', '**/__tests__/**', '**/*.test.ts', 'jest.setup.js', '**/*.d.ts', '**/vue-shim.ts'] });
+
+  // Helper functions
+  const isSimpleIdentifier = (str) => /^[a-zA-Z_$][0-9a-zA-Z_$]*$/.test(str.trim());
+  const isBracketedExpression = (str) => str.trim().startsWith('[') && str.trim().endsWith(']');
+  const isStringLiteral = (str) => /^['"].*['"]$/.test(str.trim());
+  // Extracts the key expression from a v-for directive.
+  const extractKeyExpression = (vForContent) => {
```
cnotv, line 352: Functions should be set outside.

shell/scripts/vue-migrate.js
```diff
@@ -590,40 +813,39 @@ const printContent = (...args) => {
...
-    let content = fs.readFileSync(file, 'utf8');
 
-    replacementCases.forEach(([text, replacement, notes]) => {
-      // Simple text
-      if (typeof text === 'string') {
-        if (content.includes(text)) {
-          // Exclude cases without replacement
-          if (replacement) {
-            // Remove discontinued functionalities which do not break
-            content = content.replaceAll(text, replacement === removePlaceholder ? '' : replacement);
```
cnotv, line 607: This is missing and cannot remove any more libraries that are discontinued.

cnotv (approved):
The aim is to prioritize fixes, so I would have blocker requests.

There's some comments which would be nice to follow. I've to give a second look about the whole script. However this has been planned initially to migrate Rancher Dashboard and probably would not make sense to test it there again.
Till it works for your use case, I would be fine and trust the outcome, which I would consider it as test as well.

jordojordo:
Closing in favor of https://github.com/rancher/dashboard/pull/12293

PR #12286, opened 2024-10-17
Enabling the Kontainer-Driver's `can delete drivers in bulk` e2e test.

codyrancher:
### Summary
 Enabling the Kontainer-Driver's can delete drivers in bulk e2e test. #12286 

### Technical notes summary
This was disabled because it would block the workflow when we first migrated to vue3. We're now re-enabling those tests.

None, only an e2e test was modified.

None, only an e2e test was modified.

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/cypress/e2e/tests/pages/manager/kontainer-drivers.spec.ts
+++ b/cypress/e2e/tests/pages/manager/kontainer-drivers.spec.ts
@@ -207,7 +207,7 @@ describe('Kontainer Drivers', { testIsolation: 'off', tags: ['@manager', '@admin
     createCluster.gridElementExistanceByName(linodeDriver, 'not.exist');
   });
 
-  it.skip('[Vue3 Skip]: can delete drivers in bulk', () => {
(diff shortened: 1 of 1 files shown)


cypress/e2e/tests/pages/manager/kontainer-drivers.spec.ts
```diff
@@ -231,7 +231,6 @@ describe('Kontainer Drivers', { testIsolation: 'off', tags: ['@manager', '@admin
         cy.wait('@deleteOracleDriver').its('response.statusCode').should('eq', 200);
 
         driversPage.waitForPage();
-        driversPage.list().details(linodeDriver, 1).contains('Removing');
```
codyrancher, line 234: I only saw this status update once in 2.9-head. In general it seems like the resources are deleted before the status is updated. There's still validation to ensure that the drivers are removed from the rows so I don't think it's necessary to keep this.

rak-phillip (approved):

PR #12293, opened 2024-10-18
Add Vue3 migration helper to extension creator

### LINKED ISSUE #12276: Vue3 migration script for extensions has multiple issues

jordojordo:
### Summary
Fixes #12276 

The script found in `./shell/scripts/vue-migrate.js` was originally intended for migrating the Dashboard to Vue 3 compliance. In order to leverage this script with extensions, it makes the most sense to modularize it and add it as a creator within the npm library for `@rancher/create-extension`.

This will allow extension developers to run the migrate command within their extensions to have most of the heavy lifting applied for them, with some options to limit the scope of the migration and provide suggested changes before applying any modifications.

- Added an option for migration to the `@rancher/extension` creator
- Allows for `--suggestion` option to create a diff file containing the suggested changes without modifying the extension
- Updates the annotations necessary for the extension to be loaded in Vue 3
- Allows for ignore paths to be passed when calling the migrate script

### Technical notes summary
This is essentially what was written in #12277 - following a discussion I decided to move to a modular approach for the creator to make this easier to maintain and debug.

- Migrate the Kubewarden UI extension and preform a build of the extension
- Same as above but for the Elemental UI extension

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/creators/extension/migrate/utils/content.js
+++ b/creators/extension/migrate/utils/content.js
@@ -8,6 +8,23 @@ const {
 
 const diffOutput = [];
 
+function printUsage() {
+  console.log(`
+Usage: node index.js [options]
+
+Options:
+
+  --dry                 Dry Run Mode: Run the script without making any changes to your files.
+  --verbose             Verbose Output: Enable detailed logging.
+  --suggest             Suggest Mode: Generate a 'suggested_changes.diff' file with proposed changes.
+  --paths=<path>        Specify Paths: Limit migration to specific paths or files (accepts glob patterns).
+  --ignore=<patterns>   Ignore Patterns: Exclude specific files or directories (accepts comma-separated glob patterns).
+  --files               Output Modified Files: List all files modified during the migration.
+  --log                 Generate Log File: Write detailed migration statistics to 'stats.json'.
+  --help, -h            Display this help message and exit.
+  `);
+}
+
 function writeContent(file, content, originalContent) {
   if (!isDry && !isSuggest) {
     fs.writeFileSync(file, content);
@@ -138,6 +155,7 @@ function replaceCases(fileType, files, replacementCases, printText, params) {
 }
 
 module.exports = {
+  printUsage,
   writeContent,
   printContent,
   escapeRegExp,
--- a/creators/extension/migrate/utils/content.js
+++ b/creators/extension/migrate/utils/content.js
@@ -1,0 +1,147 @@
+/* eslint-disable no-console */
+const fs = require('fs');
+const { createPatch } = require('diff');
+const stats = require('../stats');
+const {
+  isDry, isSuggest, isVerbose, removePlaceholder
+} = require('../config');
+
+const diffOutput = [];
+
+function writeContent(file, content, originalContent) {
+  if (!isDry && !isSuggest) {
+    fs.writeFileSync(file, content);
+  } else if (isSuggest) {
+    if (typeof originalContent === 'undefined') {
+      originalContent = fs.readFileSync(file, 'utf8');
+    }
+    const diff = createPatch(file, originalContent, content, '', '', { context: 3 });
+
+    if (diff.trim()) {
+      diffOutput.push({ file, diff });
+    }
+  }
+}
+
+function printContent(...args) {
+  if (isVerbose) {
+    console.log(...args);
+  }
+}
+
+function escapeRegExp(string) {
+  return string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+}
+
+function setParams(params) {
+  const args = process.argv.slice(2);
+  const paramKeys = ['paths', 'ignore'];
+
+  args.forEach((val) => {
+    paramKeys.forEach((key) => {
+      if (val.startsWith(`--${ key }=`)) {
+        const value = val.split('=')[1];
+
+        if (key === 'ignore') {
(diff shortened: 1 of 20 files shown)


cnotv (approved):
I just gave a quick walk-through but I don't see any issue.
Very good the idea to split in files, it makes it a lot cleaner. Thanks for moving it also within the creator.
I think we do not need anymore the other migration script at this point, or? We should update the commands in that case then, but we can do in another PR dunno.

LGTM

EDIT: I guess on master there are some issues and we need a rebase?

aalves08 (approved):
LGTM mate!

creators/extension/migrate/utils/content.js
```diff
@@ -0,0 +1,165 @@
...
+const fs = require('fs');
+const { createPatch } = require('diff');
+const stats = require('../stats');
+const {
+  isDry, isSuggest, isVerbose, removePlaceholder
+} = require('../config');
+
+const diffOutput = [];
+
+function printUsage() {
```
cnotv, line 11: I really liked this idea.

PR #12313, opened 2024-10-21
AKS - import existing cluster

mantis-toboggan-md:
### Summary
Fixes #

### Checklist
- [ ] The PR has a Milestone 


pkg/aks/components/Config.vue
```diff
@@ -152,7 +160,7 @@ export default defineComponent({
       loadingVersions:        false,
       loadingVmSizes:         false,
       loadingVirtualNetworks: false,
-      setAuthorizedIPRanges:  false,
+      setAuthorizedIpRanges:  false,
```
jordojordo, line 163: Should this maintain the same case `setAuthorizedIPRanges`?

pkg/aks/components/Import.vue
```diff
@@ -0,0 +1,129 @@
...
+      immediate: true
+    }
+  },
+
+  methods: {
+    async getClusters() {
+      this.loadingClusters = true;
+      this.allClusters = await getAKSClusters(this.$store, this.azureCredentialSecret);
+      this.loadingClusters = false;
+    },
```
jordojordo, line 79: If `getAKSClusters` fails, `this.loadingClusters = false` won't be executed. We should probably wrap this in a try/catch block.

jordojordo (changes requested):

pkg/aks/components/Import.vue
mantis-toboggan-md: Good call - I've updated the method to use a try/catch block and emit an error if `getAksClusters` fails

jordojordo (approved):
LGTM

PR #12321, opened 2024-10-22
update shell dependencies to prevent adding them on the extensions side since they are dashboard related 

### LINKED ISSUE #12319: Optimize dependencies on extensions related to the migration of vue3

With the progression of the work in the migration of vue3, which has reached a stability point as well the extensions compatibility side, there is room for optimisation on the dependencies we are adding to extensions since there's quite a few that could be included on the `shell` side instead of being added to the extension side

Ex: Stackstate https://github.com/StackVista/rancher-extension-stackstate/pull/22 or https://github.com/aalves08/rancher-extension-stackstate/blob/migrate-vue3/package.json#L15-L39

Optmise dependencies added to every extension

**Additional context**

aalves08:
### Summary
Fixes #12319 

- update shell dependencies to prevent adding them on the extensions side since they are dashboard related 
- bump shell version

I have some doubts in relation to:
`cache-loader`, since on the extensions migration we use `^4.1.0` and root dashboard uses `4.1.0`. Should we also pin it?
`node-polyfill-webpack-plugin`, since on the extensions migration we use `^3.0.0` and root dashboard uses `3.0.0`. Again, should we also pin it?

I've noticed quite a few resolutions that were already present on `shell` so I've only added the ones missing

### Checklist
- [x] The PR has a Milestone 


aalves08:
@jordojordo this will introduce changes on the migration script 🤔

shell/package.json
```diff
@@ -153,7 +158,9 @@
     "qs": "6.11.1",
     "roarr": "7.0.4",
     "semver": "7.5.4",
-    "@vue/cli-service/html-webpack-plugin": "^5.0.0"
+    "@types/lodash": "4.17.5",
```
cnotv, line 161: Other resolutions were not needed, they have been removed as unnecessary.

shell/package.json
```diff
@@ -105,6 +109,7 @@
     "marked": "4.0.17",
     "nodemon": "2.0.22",
     "nyc": "15.1.0",
+    "node-polyfill-webpack-plugin": "^3.0.0",
```
cnotv, line 112: You are right, I should have been pinning the versions as well. Would you mind to do that? 😬

shell/package.json
```diff
@@ -41,6 +41,7 @@
     "@rancher/icons": "2.0.29",
     "@types/is-url": "1.2.30",
     "@types/node": "16.4.3",
+    "@types/semver": "^7.5.8",
```
cnotv, line 44: I don't think we need this, or?

aalves08: I think we do @cnotv , as Stackstate extension is not using the `semver` library explicitly but it throws error on build related to that

cnotv: Let's keep it then. Thanks for the update!

cnotv (approved):
LGTM

jordojordo:
@aalves08 This shouldn't change anything on the migration script, we're only updating the [dependencies listed here](https://github.com/rancher/dashboard/blob/af3cc8f1f506b27144d0cdc696db15d74f92dd6f/creators/extension/migrate/tasks/packageUpdates.js#L68-L100) _if_ they already are present in the package.json.

jordojordo (approved):

PR #12326, opened 2024-10-22
Clone `defaultContainer` to apply defaults for workload forms

### LINKED ISSUE #12243: Workloads: Unable to create basic workload after error is thrown
**Setup**
- Rancher version: v2.10
- Rancher UI Extensions:
- Browser type & version: Chrome, FF

**Describe the bug**
User is unable to create a basic workload after error is thrown.
> Deployment in version "v1" cannot be handled as a Deployment: json: cannot unmarshal string into Go struct field Container.spec.template.spec.containers.volumeMounts of type v1.VolumeMount

**To Reproduce**
1. go to workloads
2. click create
3. click Deployment
4. input name and image (nginx:latest)
5. go to Pod tab (in horizontal tab)
6. click Storage (in verticle tab)
7. click Add Volume > select volume option
8. go to container tab (in horizontal tab) and select Storage tab
9. click Select Volume > select volume option
10. go back to Pod tab > storage > remove volume
11. click create (you should see an error)
12. cancel create form
13. click create
14. input name and image (nginx:latest)
15. click create

**Result**
User is unable to create a basic workload.

**Expected Result**
User is able to create a basic workload.

**Screenshots**

https://github.com/user-attachments/assets/3caf4f07-28bf-495c-b23c-a82b6c3c866b

**Additional context**

rak-phillip:
### Summary
This clones `defaultContainer` to prevent this object from being modified by reference in container forms.

`defaultContainer` retains its values even after navigating away from the workload create form. This behavior was not observed in Vue2, and it requires further investigation to fully understand.

To address this issue in the short term, using `structuredClone` to create a deep copy of defaultContainer ensures that modifications to the container do not persist after navigating away from the form.

Fixes #12243 

- clone `defaultContainer` when applying defaults to workload forms

See case in attached issue.

This fixes an existing regression, I don't foresee this causing issues elsewhere. 

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/edit/workload/mixins/workload.js
+++ b/shell/edit/workload/mixins/workload.js
@@ -947,7 +947,7 @@ export default {
         nameNumber++;
       }
       const container = {
-        ...defaultContainer,
+        ...structuredClone(defaultContainer),
         name:   `container-${ nameNumber }`,
         active: true
       };
--- a/shell/models/workload.js
+++ b/shell/models/workload.js
@@ -110,7 +110,7 @@ export default class Workload extends WorkloadService {
         spec.template = {
           spec: {
             restartPolicy:  this.type === WORKLOAD_TYPES.JOB ? 'Never' : 'Always',
-            containers:     [{ ...defaultContainer }],
+            containers:     [{ ...structuredClone(defaultContainer) }],
             initContainers: []
           }
         };


codyrancher (approved):
I don't think it's a temporary fix. I think this is probably how we should've handled default values from the beginning.

We could've also made a createDefaultValues function which returns a new object instance on each invocation but I think both are fine.

PR #12392, opened 2024-10-29
Fix namespace create banner in basics step

### LINKED ISSUE #12205: Helm Install: New namespace creation banner is obscured
Go to create a helm application from Apps and Marketplace by selecting a chart.

For namespace, choose the option to create a new namespace and specify a name.

A banner is shown at the bottom of the page to indicate that a new namespace will be created - this banner is obscured - see below - it should not be.

image

momesgin:
### Summary
Fixes #12205 

### Technical notes summary

Added a class to fix the styling issue by taking into account the footer height.

Same as described on the issue's repro steps.

Since the styling is applied to the banner itself only, there shouldn't be any side effects, hopefully.
### Screenshot/Video

https://github.com/user-attachments/assets/d81d8657-6e14-42a7-bfb4-78d2c1c5ff93

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/pages/c/_cluster/apps/charts/install.vue
+++ b/shell/pages/c/_cluster/apps/charts/install.vue
@@ -1457,6 +1457,7 @@ export default {
           <Banner
             v-if="isNamespaceNew && value.metadata.namespace.length"
             color="info"
+            class="namespace-create-banner"
           >
             <div v-clean-html="t('catalog.install.steps.basics.createNamespace', {namespace: value.metadata.namespace}, true) " />
           </Banner>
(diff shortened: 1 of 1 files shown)


eva-vashkevich (approved):
lgtm
Screenshot 2024-10-29 at 12 03 49 AM

nwmac:
@momesgin I think we can merge this one in - but do we have a constant we can use for the 70px, rather than hard-coding, in case we were to change that?

momesgin:
> @momesgin I think we can merge this one in - but do we have a constant we can use for the 70px, rather than hard-coding, in case we were to change that?

Unfortunately the footer is not using a variable for its height, we should create one and re-use it for these kinds of cases. I'll create a new issue to address that.

PR #12426, opened 2024-10-30
Fix issue with Workspace dropdown in Firefox

### LINKED ISSUE #12388: Fleet: Workspace switcher: Dropdown menu overlaps select box
**Setup**
- Rancher version: 2.10
- Rancher UI Extensions:
- Browser type & version: FF, ~~Chrome~~

**Describe the bug**
Dropdown menu overlaps select box.

**To Reproduce**
1. go to Fleet
2. click workspace switcher in the header

**Result**
image
**Expected Result**
image

**Screenshots**

**Additional context**

aalves08:
### Summary
Fixes #12388 

- remove `append-to-body="false"` from the Fleet workspace switcher to fix render issue of dropdown in Firefox

### Technical notes summary

The problem is that Firefox doesn't calculate the width well with this `inline` class presenthttps://github.com/rancher/dashboard/blob/master/shell/components/form/Select.vue#L244 when appending `v-select` dropdown to the element.
We could have removed the `inline` class, but I think it's less prone to risks if we just let the dropdown to be appended to the `body` element rather than the dropdown as it forces `v-select` to calculate both width and position and apply it to the element without any relative measurements like `%`.

Original work done in https://github.com/rancher/dashboard/pull/2170

https://vue-select.org/api/props.html#appendtobody

- check workspace switcher for Fleet in Firefox

- I can't find any scenario where we would have regressions, but better to test with all other browsers

### Screenshot/Video

Screenshot 2024-10-30 at 10 54 04

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/nav/WorkspaceSwitcher.vue
+++ b/shell/components/nav/WorkspaceSwitcher.vue
@@ -92,7 +92,6 @@ export default {
       ref="select"
       v-model:value="value"
       label="label"
-      :append-to-body="false"
       :options="options"
       :clearable="false"
       :reduce="(opt) => opt.value"


richard-cox (approved):
General things of note
- This fixes the gap between options and input which has been bugging me for ages
- The bobble on hover still happens (not caused by or changed by PR)

PR #12450, opened 2024-10-31
Fix tagging not working when creating a logging flow

### LINKED ISSUE #12342: Logging: Flows: Creation: Unable to specify container names
**Setup**
- Rancher version: v2.10
- Rancher UI Extensions:
- Browser type & version: FF, Chrome

**Describe the bug**
User is unable to specify container names when creating a Flow. After entering text for the 'Exclude specific container names' field, the text becomes undefined/null. Also, an error is thrown when attempting to create a Flow with this invalid entry which makes this part of the feature unusable outside of using the default value.
> Flow.logging.banzaicloud.io "test1" is invalid: spec.match[0].select.container_names[0]: Invalid value: "null": spec.match[0].select.container_names[0] in body must be of type string: "null"

**To Reproduce**
1. install Logging (Cluster Explorer --> Apps --> Charts --> Logging --> Install)
2. Cluster Explorer --> Logging --> Flow --> Create --> Matches tab
3. press either 'Add Include Rule' or 'Add Exclude Rule'
4. for 'Exclude specific container names' type some value
5. hit enter

**Result**

https://github.com/user-attachments/assets/e3e32673-2af9-41ce-99a0-92e3a850f1d5

**Expected Result**

https://github.com/user-attachments/assets/e8581d62-f872-4fbf-82a2-73e7fb875bec

**Screenshots**

**Additional context**

momesgin:
### Summary
Fixes #12342 

Fix the issue with `LabeledSelect` when `taggable` is `true` by allowing the binding of the attribute to the element v-select that was being filtered out.

I traced back the previous change for filtering `taggable` to [the original issue](https://github.com/rancher/dashboard/issues/11954) and checked if this change broke that fix, and it's not breaking it. To discard any regressions, a quick test of different instances of `LabeledSelect` would be great, even if they don't use `:taggable="true"`

### Screenshot/Video

https://github.com/user-attachments/assets/321f31c9-cf21-4b59-b89e-d4d828c67339

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/form/LabeledSelect.vue
+++ b/shell/components/form/LabeledSelect.vue
@@ -139,7 +139,6 @@ export default {
     filteredAttrs() {
       const {
         class: _class,
-        taggable,
         ...rest
       } = this.$attrs;
 


rak-phillip (approved):
LGTM - I think that we should call out that `taggable` is a prop specified by Vue Select to enable creating options [^1].

[^1]: https://vue-select.org/api/props.html#taggable

codyrancher (approved):

PR #12456, opened 2024-11-01
Enabling the branding e2e tests and fixing what's broken

codyrancher:
### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/cypress/e2e/tests/pages/global-settings/branding.spec.ts
+++ b/cypress/e2e/tests/pages/global-settings/branding.spec.ts
@@ -97,7 +97,7 @@ describe('Branding', { testIsolation: 'off' }, () => {
     cy.title({ timeout: 2000 }).should('eq', settings.privateLabel.original);
   });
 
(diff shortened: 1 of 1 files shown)


cypress/e2e/tests/pages/global-settings/branding.spec.ts
```diff
@@ -165,16 +165,16 @@ describe('Branding', { testIsolation: 'off' }, () => {
 
     HomePagePo.navTo();
     burgerMenu.headerBrandLogoImage().should('be.visible').then((el) => {
-      expect(el).to.have.attr('src').includes('/img/rancher-logo.66cf5910.svg');
+      expect(el).to.have.attr('src').includes('/img/rancher-logo');
```
codyrancher, line 168: We were referencing a hash which was different from what we now have.

I didn't switch the hash because the hash appears to be different when running locally vs running e2e locally so I don't believe it's a stable value we can rely on.

Ultimately here and all the other references I've updated are really only making sure that a file name is present for the image src instead of a base64.

momesgin (approved):

PR #12532, opened 2024-11-06
Use `semver.coerce()` to get valid kubernetes version

### LINKED ISSUE #12525: Unable to Edit Config of imported Azure AKS cluster
**Setup**
- Rancher version: `v2.10.0-alpha7` Dashboard: v2.10.0-alpha5 (also on current v2.10-head)
- aks-operator version: `rancher-aks-operator:105.0.0+up1.10.0-rc.2`
- Browser type & version: any

**Describe the bug**
When attempting to edit/upgrade an imported Azure AKS cluster via `Edit Config`, the dialog is blank. This dialog is sometimes shown but in this case it offers an empty list of Kubernetes versions to upgrade to with only entry `No options` instead of any k8s version.

I was able to reproduce this twice in a row on a fresh clusters and with AKS clusters in different locations / k8s versions installed.

It works fine for rancher provisioned AKS clusters. 

I believe this is related to the UI, but I may be wrong, in this case please move to `CAPI & HP` project .

**To Reproduce**
* Provision an AKS cluster via az CLI:
```
az group create -l eastus -n th-rg
az aks create --resource-group th-rg --no-ssh-key --kubernetes-version 1.29 --name th-az-imported-293 --node-count 1 --location eastus
```
* Import the AKS cluster into rancher -> Cluster Management -> Import Existing -> Azure AKS, select/create the credentials -> select the cluster you created from the list -> press `Import` button
* After a short while the cluster will be imported
* From `Cluster Management` select the cluster's three dots vertical menu and select `Edit Config`

**Result**
* The Edit dialog is blank - only sometimes offer a list of k8s versions to upgrade to but there is only `No options` "version" available.

**Expected Result**
The Edit dialog should be shown and it should offer a list of valid k8s versions for upgrade

**Screenshots**
aks-imported-upgrade

image

rak-phillip:
### Summary
Fixes #12525 

This uses `semver.coerce()` to get a valid semantic version for the original Kubernetes version in `CruAks.vue`. The form could fail to render if the Kubernetes version is not a valid semver and `semver.coerce()` will correctly evaluate to a semantic version.

### Technical notes summary

In the production example, 1.29 is used for the Kubernetes version:

```
az aks create --resource-group th-rg --no-ssh-key --kubernetes-version 1.29 --name th-az-imported-293 --node-count 1 --location eastus
```

1.29 is a valid Kubernetes version and will default to the latest release[^1] (1.29.10 at the time of writing).

1.29 is stored in `aksConfig.kubernetesVersion`; this is not a valid `semver`, so an unhandled exception is thrown and the page fails to render correctly. 

It's entirely possible that `current-cluster-controllers-version` is the incorrect value to use for this case, but it was the only one that I found that evaluates to a valid semver with the expected kubernetes version of 1.29.10.

[^1]: https://kubernetes.io/releases/#release-v1-29

- See reproduction steps in issue

We are changing the Kubernetes version used for the AKS form. If this annotation doesn't exist for any reason, we will experience the same behavior that's described in the issue. We can consider adding some sort of failure mode if the original Kubernetes isn't present or a valid semver. We can also consider adding some sort of fallback value in the event that it fails. 

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/pkg/aks/components/CruAks.vue
+++ b/pkg/aks/components/CruAks.vue
@@ -160,7 +160,9 @@ export default defineComponent({
       }
 
       // track original version on edit to ensure we don't offer k8s downgrades
-      this.originalVersion = this.normanCluster?.aksConfig?.kubernetesVersion;
+      const kubernetesVersion = semver.coerce(this.normanCluster?.aksConfig?.kubernetesVersion);
+
+      this.originalVersion = kubernetesVersion;
     } else {
       this.normanCluster = await store.dispatch('rancher/create', { type: NORMAN.CLUSTER, ...defaultCluster }, { root: true });
 


pkg/aks/components/CruAks.vue
```diff
@@ -160,7 +160,9 @@ export default defineComponent({
       }
 
       // track original version on edit to ensure we don't offer k8s downgrades
-      this.originalVersion = this.normanCluster?.aksConfig?.kubernetesVersion;
+      const kubernetesVersion = this.normanCluster?.annotations['management.cattle.io/current-cluster-controllers-version'];
```
mantis-toboggan-md, line 163: It looks like this annotation wont be defined while a cluster is provisioning. Could we fall back here to `semver.coerce(this.normanCluster?.aksConfig?.kubernetesVersion)`

mantis-toboggan-md (changes requested):

pkg/aks/components/CruAks.vue
rak-phillip: I'm unsure if this will be correct.. `semver.coerce()` will evaluate to `1.29.0` when provided with 1.29 for a version; 1.29.10 is the actual Kubernetes version.

rak-phillip: We spoke OOB on this topic. The best we can reliably do with the information available to us at this point is to attempt to infer some Kubernetes version via `semver.coerce()`. It's not 100% correct, but this is at least stable and reliable when compared to the original approach.

mantis-toboggan-md (approved):

Issue #12506, opened 2024-11-05
Clusters: Unable to edit local cluster


yonasberhe23:
**Setup**
- Rancher version: v2.10-e59c50a79bff2b6cc023bdbdb4e825190b7833fb-head
- Rancher UI: `master f878279`
- Browser type & version: Chrome

**Describe the bug**
User is unable to edit local cluster.

**To Reproduce**
1. go to cluster management
2. click ellipsis for local cluster
3. click Edit Config
4. notice Create button (should be Save) is disabled
5. add description
6. notice Create button is still disabled

**Result**
User is unable to edit local cluster.

**Expected Result**
User is able to edit local cluster.

**Screenshots**

https://github.com/user-attachments/assets/092c7ed3-c583-4072-b371-7797922782ac

**Additional context**

gaktive:
Problem not happening on RKE2 & k3s. QA to confirm if this is happening elsewhere; if this is only local, then it's a release note for 2.10.0 until we figure out what to do since code change here will touch a lot of things.

mantis-toboggan-md:
Looks related to `[this](https://github.com/rancher/dashboard/blob/master/shell/models/management.cattle.io.cluster.js#L117)` getter. There is a possiblity for regression in the provider/distro column of cluster list views, as well as routing to/from the 'edit config' of clusters generally. We'll need to verify behavior for each of the following cluster types:

    local clusters:
        rke1
        rke2
        k3s
        hosted

    downstream clusters:
        rke1 provisioned
        rke2 provisioned
        k3s provisioned
        rke1 imported
        rke2 imported
        k3s imported
        hosted imported

yonasberhe23:
Closing this out as I have already tested this in 2.10.

Issue #12339, opened 2024-10-23
[BUG]Edit Config on local EKS cluster 


rolfbartels:
**Rancher Server Setup**
- Rancher version: 2.92
- Installation option (Docker install/Helm Chart): Helm
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): EKS
- Proxy/Cert Details: Rancher Cert from Cert Manager

**Information about the Cluster**
- Kubernetes version: 1.30
- Cluster Type (Local/Downstream): local
* Hosted = EKS

**User Information**
- What is the role of the user logged in? Admin/Cluster Owner

**Describe the bug**
On a new install or old install on Rancher 2.9.2 I want to add labels to the local cluster.
When I edit the config I am prompted that atleast 1 node group must be configured.
I insert the missing field (Cluster Name, Cloud Creds (Same Creds use to eksctl create), add a node group and try and save.

I am presented on and error
{"data":""}

**To Reproduce**
Cluster Management
Select the ellipsis for local cluster
Edit Config
Enter in required details
Save 

**Result**
{"data":""}
**Expected Result**
Cluster Config saves

**Screenshots**
![image](https://github.com/user-attachments/assets/50141fb6-e1c3-4a89-a7de-32bfce4cf08b)

**Additional context**

rak-phillip:
I can repro this issue in v2.9.3. The form behaves exactly as described when editing the local cluster for Rancher hosted on EKS

rak-phillip:
Following up with some information based on OOB discussions. We've identified that the behavior described in the issue is related to `eksStatus.upstreamSpec` evaluating to `null` and `eksConfig` is `undefined`. This has been reported to be expected behavior:

> on local cluster where Rancher is running, and even if it was deployed on EKS eksStatus should be:
>
>  ```
>  eksStatus:
>    generatedNodeRole: ''
>    managedLaunchTemplateID: ''
>    managedLaunchTemplateVersions: null
>    privateRequiresTunnel: null
>    securityGroups: null
>    subnets: null
>    upstreamSpec: null
>    virtualNetwork: ''
>  ```
>
> it is because local cluster is not provisioned or managed by eks-operator

Based on the information provided, I think that we will need to define the edge-cases for working with local clusters and handle them accordingly so that the form behaves as expected for this use case.

izaac:
[...]

PR #12177, opened 2024-10-08
Fix RKE1 upgrading via template

mantis-toboggan-md:
### Summary
Fixes #11874 

### Technical notes summary
The dashboard is setting the wrong value for the `provider` query param when loading some ember pages. It looks like the bug came in here https://github.com/rancher/dashboard/pull/9868/files#diff-9be0247bdd17368e0b020b8ab4933b0c53f2f4bbdb0b1936066fa67703e519d4R91 

It seems that #11874 is happening because when a new cluster template revision is selected, some ember components re-initialize, and at that second partial load, the incorrect `provider` query param causes ember to not load rke1 form components. There's probably something messy going on over in Ember, but fixing the query param being sent over seems the safest solution, vs digging into ember route-changing logic.

### Areas or cases that should be tested
Verify that the right create/edit pages load for the following cluster types:

- RKE1 custom
- RKE2 custom
- RKE1 provisioned
- RKE2 provisioned
- Generic imported
- AKS/EKS/GKE imported (1 of the 3)
- AKS/EKS/GKE provisioned (1 of the 3)

For RKE1 clusters, verify that when using an RKE1 template users can select new template versions and the form is updated accordingly

### Areas which could experience regressions
- Creating/editing any cluster types that use ember pages
- the 'provider' column in the cluster list

### Screenshot/Video
this pr:
Screenshot 2024-10-08 at 10 47 49 AM

release-2.9:
Screenshot 2024-10-08 at 10 47 42 AM

### Checklist
- [x] The PR is linked to an issue and the linked issue has a Milestone, or no issue is needed
- [x] The PR has a Milestone 
- [x] The PR template has been filled out
- [x] The PR has been self reviewed 
- [x] The PR has a reviewer assigned
- [x] The PR has automated tests or clear instructions for manual tests and the linked issue has appropriate QA labels, or tests are not needed
- [x] The PR has reviewed with UX and tested in light and dark mode, or there are no UX changes

PR #12548, opened 2024-11-08
[WIP] Fix issues with iFramed cluster pages

mantis-toboggan-md:
3 open issues all related to rendering the wrong iframed Ember page, or rendering the page with missing content:
https://github.com/rancher/dashboard/issues/12506
https://github.com/rancher/dashboard/issues/12541

possibly https://github.com/rancher/dashboard/issues/12339

The code used to determine which component to render when editing any given type of cluster is convoluted. There are inexplicable mappings required to keep iframed ember pages functional (why must aks map to azureaks? why is the provider query param sometimes import and sometimes import*ed*? Why is there an importProvider query param not used on imported clusters?) I am not even sure the above 3 issues have exactly the same root cause, but this area of dashboard code is so convoluted and prone to churn, I think they need to be addressed in tandem.

I think the above bugs came in via some combination of [this](https://github.com/mantis-toboggan-md/dashboard/commit/4cca5faf81dd658464b086c67894642f909ee6fc#diff-7655093720dda329bd1d1b72936a782cd7178edadbf55a7c1b8b58fb394f77cfR190) and [this](https://github.com/rancher/dashboard/pull/12177/files#diff-9be0247bdd17368e0b020b8ab4933b0c53f2f4bbdb0b1936066fa67703e519d4R121) PR. Limiting the scope of those changes seems to address #12506 and #12541  but not fully #12339 ( with this PR, I see the ember page as expected, but the 'mode' is wrong, create button shown instead of save)

Because I think the bugs are conntected to ember query params, I've collected data on the fields used to generate those query params, and what those query params are in 2.9-head ember standalone world - ember standalone has been removed from 2.10 but we can assume that these are the correct params the dashboard should be passing to ember. All below fields are on the relevant management cluster object, not the provisioning cluster object. If [providerForEmberParam](https://github.com/rancher/dashboard/blob/master/shell/models/management.cattle.io.cluster.js#L117) needs to be fixed, we need to verify that given the below status/config fields, the correct `provider` and `importProvider` query params are still set in every scenario.

| cluster type | status.provider | status.driver | spec.aksConfig/eksConfig/gkeConfig.imported | ember provider qp\* | ember importProvider qp\* |
|----|-----|-----|-----|-----|----|
local rke1 | rke | imported | n/a | import | other |
local rke2 | rke2 | rke2 | n/a | import | other |
local k3s | k3s | k3s | n/a | k3s | undefined | 
local hosted (eks)| eks | imported | undefined | import | other |
local MB ovh-downstream test cluster\** | '' | imported | undefined | import | other |
rke1 provisioned | 'rke' | rancherKubernetesEngine | n/a | digitalocean | undefined |
rke2 provisioned | rke2 | imported | n/a | n/a |  n/a |
k3s provisioned | k3s | imported | n/a |  n/a |  n/a |
hosted provisioned (aks) | aks | AKS | false | n/a | n/a | 
rke1 imported | rke | imported | n/a | imported | undefined |
rke2 imported | rke2 | rke2 | n/a | rke2 | undefined |
k3s imported | k3s | k3s | n/a | k3s | undefined |
hosted imported (aks) | aks | AKS |  true | azureaks | undefined |
3rd party kontainer provisioned | '' | ovhcloudmks | n/a | ovhcloudmks | undefined |

\* undefined => an iframed page should be used for this cluster, but this query param should be present; n/a=>iframed ember pages should not be used for this cluster
\** need to find out what this cluster actually is

Regarding ovh, once I get the right page to render I am seeing a console error that could be related
Screenshot 2024-11-07 at 12 46 44 PM

### DIFF
--- a/shell/edit/provisioning.cattle.io.cluster/index.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/index.vue
@@ -211,14 +211,15 @@ export default {
 
     emberLink() {
       if (this.value) {
-        if (this.value.provisioner) {
-          const matchingSubtype = this.subTypes.find((st) => st.id.toLowerCase() === this.value.provisioner.toLowerCase() || DRIVER_TO_IMPORT[st.id.toLowerCase()] === this.value.provisioner.toLowerCase());
+        // set subtype if editing non-local EKS/GKE/AKS cluster -- this ensures that the component provided by extension is loaded instead of iframing old ember ui
+        if (this.value.provisioner && !this.value.isLocal) {
+          const matchingSubtype = this.subTypes.find((st) => DRIVER_TO_IMPORT[st.id.toLowerCase()] === this.value.provisioner.toLowerCase());
 
           if (matchingSubtype) {
             this.selectType(matchingSubtype.id, false);
           }
         }
-        // For custom RKE2 clusters, don't load an Ember page.
+        // // For custom RKE2 clusters, don't load an Ember page.
         // It should be the dashboard.
         if ( this.value.isRke2 && ((this.value.isCustom && this.mode === _EDIT) || (this.value.isCustom && this.as === _CONFIG && this.mode === _VIEW) || (this.subType || '').toLowerCase() === 'custom')) {
           // For admins, this.value.isCustom is used to check if it is a custom cluster.
--- a/shell/models/management.cattle.io.cluster.js
+++ b/shell/models/management.cattle.io.cluster.js
@@ -120,7 +120,7 @@ export default class MgmtCluster extends SteveModel {
     // Provisioner is the "<something>Config" in the model
     const provisioner = KONTAINER_TO_DRIVER[(this.provisioner || '').toLowerCase()] || this.provisioner;
 
-    if ( provisioner === 'rancherKubernetesEngine' || provisioner === 'rke') {
+    if ( provisioner === 'rancherKubernetesEngine' || (provisioner === 'rke' && !this.isLocal )) {
       // Look for a cloud provider in one of the node templates
       if ( this.machinePools?.[0] ) {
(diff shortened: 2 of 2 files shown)


gaktive:
This would need a back port to v2.10.x -- .0 ideally but we're running out of time.

codyrancher:
> I think the above bugs came in via some combination of [this](https://github.com/mantis-toboggan-md/dashboard/commit/4cca5faf81dd658464b086c67894642f909ee6fc#diff-7655093720dda329bd1d1b72936a782cd7178edadbf55a7c1b8b58fb394f77cfR190) and [this](https://github.com/rancher/dashboard/pull/12177/files#diff-9be0247bdd17368e0b020b8ab4933b0c53f2f4bbdb0b1936066fa67703e519d4R121) PR. Limiting the scope of those changes seems to address https://github.com/rancher/dashboard/issues/12506 and https://github.com/rancher/dashboard/issues/12541 but not fully https://github.com/rancher/dashboard/issues/12339 ( with this PR, I see the ember page as expected, but the 'mode' is wrong, create button shown instead of save)

Because this all appears to be working in 2.9 I decided to diff dashboard between master and 2.9.

When diffing `master` with `2.9` I saw https://github.com/rancher/dashboard/pull/12177 and when reverting it seems to resolve the issues we're having with editing. The above pr was created in response to https://github.com/rancher/dashboard/pull/9868.

I'm leaning toward reverting https://github.com/rancher/dashboard/pull/12177 and changing the fix for https://github.com/rancher/dashboard/pull/9868 because ultimately it supposed to update a display property but seems to be having cascading consequences. I'll see if I can have some luck going down this path.

mantis-toboggan-md:
Closing in favor of https://github.com/rancher/dashboard/pull/12583

PR #12608, opened 2024-11-14
fix e2e chrome installation

aalves08:
We are replacing the download server because of frequent unavailability.


torchiaf (approved):
LGTM

PR #12630, opened 2024-11-20
Fix empty GitRepos list on Cluster details

aruiz14:
### Summary
Follow up of #12521

Fixes the GitRepos list being empty in the Cluster details view.

### Technical notes summary
The cluster lists used to get the `Cluster` object data is based on `targetClusters` spec, which may vary depending on the context. Since `resourcesStatuses` now iterates on `BundleDeployments` resources directly, it should skip those deployed in a cluster that is not present in the `targetClusters`.

### DIFF
--- a/shell/models/fleet.cattle.io.gitrepo.js
+++ b/shell/models/fleet.cattle.io.gitrepo.js
@@ -338,6 +338,11 @@ export default class GitRepo extends SteveModel {
     for (const bd of bundleDeployments) {
       const clusterId = FleetUtils.clusterIdFromBundleDeploymentLabels(bd.metadata?.labels);
       const c = clusters[clusterId];
+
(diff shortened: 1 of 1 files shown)


richard-cox (approved):
Worked through this with Alejandro. 

a git repo targets clusters, it then uses those clusters when mapping bundledeployments associated with the git repo. later on it'll summaries those mapped objects by cluster and on the cluster detail page show them for a specific cluster.

problems arise on the cluster detail page where we don't fetch all clusters, but just the 1 associated with the page.... so when mapping bundledeployements we don't have some of the clusters... which throws an undefined error

so only applicable to instances where there are git repos that target multiple clusters (could be quite common)... plus a refresh on the the fleet cluster detail page

PR #12633, opened 2024-11-20
Adding a custom lint rule to ensure we use v-clean-tooltip instead of v-tooltip directives.

codyrancher:
### Summary
Wanted to automate the ability to detect an issue with using v-tooltip over v-clean-tooltip. I wrote a custom linter rule so we can catch this in CI.

### Technical notes summary
I noticed that we currently exclude linting rancher-components from our root directory. I'd think we'd just want to include everything but I'm not sure.

None, it's in a way an additional test.

None

### Screenshot/Video
**Example:**
image

**You can disable the rule in the template using:**
image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "serve-pkgs": "./shell/scripts/serve-pkgs",
     "publish-shell-reset-reg": "cd shell && npm publish",
     "clean": "./shell/scripts/clean",
-    "lint": "./node_modules/.bin/eslint --max-warnings 0 --ext .js,.ts,.vue .",
+    "lint": "./node_modules/.bin/eslint --rulesdir ./eslint-local-rules --max-warnings 0 --ext .js,.ts,.vue .",
     "lint:lib": "cd pkg/rancher-components && yarn lint",
     "lint-l10n": "./node_modules/.bin/yamllint ./shell/assets/translations",
     "test": "NODE_OPTIONS=--max_old_space_size=8192 jest --watch",
--- a/.eslintrc.default.js
+++ b/.eslintrc.default.js
@@ -190,7 +190,10 @@ module.exports = {
     'vue/one-component-per-file':       'off',
     'vue/no-deprecated-slot-attribute': 'off',
     'vue/require-explicit-emits':       'error',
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -18,7 +18,7 @@
     "serve-pkgs": "./shell/scripts/serve-pkgs",
     "publish-shell-reset-reg": "cd shell && npm publish",
     "clean": "./shell/scripts/clean",
-    "lint": "./node_modules/.bin/eslint --max-warnings 0 --ext .js,.ts,.vue .",
+    "lint": "./node_modules/.bin/eslint --rulesdir ./eslint-local-rules --max-warnings 0 --ext .js,.ts,.vue .",
```
rak-phillip, line 21: Do you know if we can specify `rulesdir` in `eslintrc` instead of with a cli flag? 

My assumption is that it's best to stick with the approach in this PR, but my curiousity was piqued when I found some old conversations that suggests eslint deprecating the usage of `ruledir` and recommending `eslint-plugin-rulesdir` as the preferred approach[^1][^2]. I'm guessing that this never materialized in any meaningful way, seeing how we are using `rulesdir` today.

[^1]: https://github.com/eslint/eslint/issues/2715
[^2]: https://github.com/eslint/eslint/issues/8769

codyrancher: I'm not sure. I'll give it a try.

codyrancher: We cannot specify `rulesdir` in `eslintrc`.

It failed with 
image

Which lead me to go look at the config-validator code and there isn't a `rulesdir` option.
https://github.com/eslint/eslintrc/blob/v0.4.3/conf/config-schema.js

When we upgrade eslint we will likely want to switch from `--rulesdir` to `--plugin` which will cause us to create one extra file to create the plugin with all of our rules. https://eslint.org/docs/latest/use/command-line-interface#--plugin

rak-phillip (approved):
LGTM

> I noticed that we currently exclude linting rancher-components from our root directory. I'd think we'd just want to include everything but I'm not sure.

IIRC we should be linting Rancher Components along with root.. We might want to open up an issue to investigate and get a more concrete answer on why it is excluded at this point.

Issue #12670, opened 2024-11-25
[Tech Debt]: Investigate why we don't run the linter on `rancher-components` from the root of the dashboard project


codyrancher:
If the investigation shows that we can, we should begin running the linter on rancher-components from the root of dashboard. 

This originated from: https://github.com/rancher/dashboard/pull/12633#pullrequestreview-2458963583

PR #12746, opened 2024-12-03
Fixing an issue with eslint not working in vscode

codyrancher:
- Using rulesdir in the package scripts made it so the rules were only present when running them in the cli and showed errors in editors. This now allows the rules to run in the editor since I made a plugin using a local dependency
- I also made is so lint rules can be run from root on pkg/rancher-components

Fixes https://github.com/rancher/dashboard/issues/12670

### Summary
Fixes #

### Checklist
- [x] The PR has a Milestone 


pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.vue
```diff
@@ -116,7 +118,7 @@ export default defineComponent({
       isDisabled,
       validationMessage,
       requiredField
-    } = useLabeledFormElement(props, emit);
+    } = useLabeledFormElement(props, emit as emitMethodType);
```
codyrancher, line 121: Adding the emits will change the type of emit passed to the setup method. In this case it will be a union type of `'change' | 'update:value' | 'blur' | 'update:validation'` which surprisingly doesn't match the string type expected by `useLabeledFormElement`. So I extracted the type and then cast appropriately here. I couldn't figure out a type that would work.

pkg/rancher-components/src/components/StringList/StringList.test.ts
```diff
@@ -297,7 +297,7 @@ describe('stringList.vue', () => {
       await inputField.trigger('keydown.enter');
       await wrapper.vm.$nextTick();
 
-      const emitted = (wrapper.emitted('change') || [])[1][0][0];
+      const emitted = (wrapper.emitted('change') || [])[0][0][0];
```
codyrancher, line 300: When `emits` is specified the value returned by emitted() omits the event type so I had to go through and subtract 1 from each index to account for the lack of event type.

pkg/rancher-components/types/index.d.ts
```diff
@@ -19,4 +19,4 @@ type ArrayListRow = {
 export type ArrayListData = {
     lastUpdateWasFromValue: boolean;
     rows: Array<ArrayListRow>,
-}
\ No newline at end of file
+}
```
codyrancher, line 22: The lint rules in root require us to have an ending black newline at the end of each file.

shell/composables/useLabeledFormElement.ts
```diff
@@ -70,7 +70,9 @@ export const labeledFormElementProps = {
   }
 };
 
-export const useLabeledFormElement = (props: LabeledFormElementProps, emit: (event: string, ...args: any[]) => void): UseLabeledFormElement => {
+export type emitMethodType = (event: string, ...args: any[]) => void;
```
codyrancher, line 73: Explained above, I needed to be able to type case the emit method elsewhere so I extracted the definition.

package.json
```diff
@@ -200,7 +200,8 @@
     "webpack-virtual-modules": "0.4.3",
     "worker-loader": "3.0.8",
     "yaml-lint": "1.7.0",
-    "yarn": "1.22.18"
+    "yarn": "1.22.18",
+    "eslint-plugin-local-rules": "link:./eslint-plugin-local-rules"
```
codyrancher, line 204: We make the local rules a linked dev dependency. This allows eslint to search for the rules as a plugin.

shell/composables/useLabeledFormElement.ts
```diff
@@ -70,7 +70,9 @@ export const labeledFormElementProps = {
...
+
+export const useLabeledFormElement = (props: LabeledFormElementProps, emit: emitMethodType): UseLabeledFormElement => {
```
rak-phillip, line 75: I think that we might be able to more explicitly type the emits by using `defineEmits`:

```suggestion
const labeledFormElementEmits = defineEmits(['update:validation']);

export const useLabeledFormElement = (props: LabeledFormElementProps, emit: typeof labeledFormElementEmits): UseLabeledFormElement => {
```

This should accomplish two things:

1. Simplify consumption of `useLabeledFormElement` by removing the requirement to type emits as `emitMethodType`
2. Enforces that consumers define `update:validation` in their emits

pkg/rancher-components/src/components/StringList/StringList.test.ts
rak-phillip: This is a good development. This behavior was quite confusing while we were fixing the unit tests and it was unclear that omitting emits options was the root cause.

package.json
rak-phillip: We might want to consider documenting this somewhere. I can see how someone might encounter this link in the future and have questions about why we are doing this.

rak-phillip (commented):
This looks great - I have one suggestion about typing emits options in composables.

shell/composables/useLabeledFormElement.ts
codyrancher: Added, I had to also add the defineEmits import and remove the emitMethodType in the other file so I added it as a part of my commit instead of committing the suggestion.

codyrancher:
Added the `link` documentation to `eslint-plugin-local-rules/index.js`

rak-phillip (approved):

PR #12827, opened 2024-12-05
[FIX] Fixed app icon aligntment for extensions with image icon

### LINKED ISSUE #12758: App bar icon alignment broken
If you have an extension that uses an image for an icon, the alignment is broken.

This appears to be due to this change: https://github.com/rancher/dashboard/pull/12591

Before: (in 2.10)

image

Now: (in 2.11)

image

The css used to have the width set for the image, but now this was moved to only apply if the icon is provided via the icon font.

eva-vashkevich:
### Summary
Fixes #12758 

Changed css to exclude pins instead of specifying which element to apply to include more cases.

1. Install an extension from https://nwmac.github.io/rancher-dev-portal/
2. All items on app bar (top level menu) should be positioned the same
3. Create a new imported cluster without actually importing it.
4. Pin the new cluster and local cluster. Pin positioning should be the same.

Top level menu application bar

### Screenshot/Video

Screenshot 2024-12-05 at 10 41 06 AM
Screenshot 2024-12-05 at 10 40 57 AM

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/nav/TopLevelMenu.vue
+++ b/shell/components/nav/TopLevelMenu.vue
@@ -1122,7 +1122,6 @@ export default {
           &:not(.pin){
             width: 42px;
           }
-
         }
 
         .rancher-provider-icon,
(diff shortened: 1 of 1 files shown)


momesgin (approved):
LGTM

PR #12884, opened 2024-12-13
Resolving some issues around rendering pagination loading on resource list pages

codyrancher:
### Summary
It wasn't readily apparent what the expected behavior was around things like alt-loading and the manual refresh so I did what made sense to me. See the video to see if it's what we want or if adjustments need to be made.

fixes https://github.com/rancher/dashboard/issues/12748

We no longer flicker the no-rows messaging when first loading one of the server-side pagination pages.

SSP and non-SSP resource list pages.

I think that this should be manually tested, it seems a little difficult to test given the nature of network latency and debouncing we have.

### Screenshot/Video
The behavior seen in this video has changed some since the latest rebase. https://github.com/rancher/dashboard/pull/12884/files#diff-16199a1483a9651b7f40bfcb61ed890469ea35d889aee82b0f50b95d2ed2057eR506-R510 Causes the refresh button to no longer trigger the altLoading overlay/opacity. This appears to be intended but maybe we'll want to update it because the lack of feedback makes me think nothing is happening.

https://github.com/user-attachments/assets/ccaef498-ed43-4b8e-bfb4-82c8953d417c

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/ResourceList/index.vue
+++ b/shell/components/ResourceList/index.vue
@@ -271,7 +271,7 @@ export default {
       v-else
       :schema="schema"
       :rows="rows"
-      :alt-loading="canPaginate"
+      :alt-loading="canPaginate && !isFirstLoad"
       :loading="loading"
       :headers="headers"
       :group-by="groupBy"
--- a/shell/components/SortableTable/THead.vue
+++ b/shell/components/SortableTable/THead.vue
@@ -428,10 +428,7 @@ export default {
         background-color: var(--sortable-table-header-bg);
         color: var(--body-text);
         text-align: left;
-
-        &:not(.loading) {
-          border-bottom: 1px solid var(--sortable-table-top-divider);
-        }
+        border-bottom: 1px solid var(--sortable-table-top-divider);
       }
     }
 
--- a/shell/components/SortableTable/index.vue
+++ b/shell/components/SortableTable/index.vue
@@ -380,8 +380,10 @@ export default {
       eventualSearchQuery = this.$route.query?.q;
     }
 
+    const isLoading = this.loading || false;
+
     return {
-      refreshButtonPhase:         ASYNC_BUTTON_STATES.WAITING,
+      refreshButtonPhase:         isLoading ? ASYNC_BUTTON_STATES.WAITING : ASYNC_BUTTON_STATES.ACTION,
       expanded:                   {},
       searchQuery,
       eventualSearchQuery,
@@ -392,7 +394,7 @@ export default {
       /**
        * The is the bool the DOM uses to show loading state. it's proxied from `loading` to avoid blipping the indicator (see usages)
        */
-      isLoading:                  false,
+      isLoading
     };
   },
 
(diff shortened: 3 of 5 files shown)


shell/components/SortableTable/THead.vue
```diff
@@ -415,10 +415,7 @@ export default {
         background-color: var(--sortable-table-header-bg);
         color: var(--body-text);
         text-align: left;
-
-        &:not(.loading) {
-          border-bottom: 1px solid var(--sortable-table-top-divider);
-        }
+        border-bottom: 1px solid var(--sortable-table-top-divider);
```
codyrancher, line 418: We get a flicker without a border momentarily if we keep the .loading negation.

shell/components/SortableTable/index.vue
```diff
@@ -374,8 +374,10 @@ export default {
       eventualSearchQuery = this.$route.query?.q;
     }
 
+    const isLoading = this.loading || false;
```
codyrancher, line 377: Sometimes loading has a value and that's what we should initiate our state with.

shell/components/SortableTable/index.vue
```diff
@@ -512,7 +514,7 @@ export default {
   },
 
   created() {
-    this.debouncedRefreshTableData = debounce(this.refreshTableData, 500);
+    this.debouncedRefreshTableData = debounce(this.refreshTableData, 500, { leading: true });
```
codyrancher, line 517: This forces us to trigger at the beginning of the timer instead of the end which makes the visual for loading appear sooner.

shell/components/SortableTable/index.vue
```diff
@@ -530,9 +532,6 @@ export default {
     manualRefreshLoadingFinished() {
       const res = !!(!this.isLoading && this._didinit && this.rows?.length && !this.isManualRefreshLoading);
 
-      // Always ensure the Refresh button phase aligns with loading state (regardless of if manualRefreshLoadingFinished has changed or not)
-      this.refreshButtonPhase = !res || this.loading ? ASYNC_BUTTON_STATES.WAITING : ASYNC_BUTTON_STATES.ACTION;
-
```
codyrancher, line 535: This would cause the refresh button to get stuck on page load and refreshButtonPhase expects a phase value not a boolean true or false value.

You'll also notice that we have a manualRefreshLoadingFinished watcher which updates the phase as well so it seems safe to remove.

shell/components/ResourceList/index.vue
```diff
@@ -270,7 +270,7 @@ export default {
       v-else
       :schema="schema"
       :rows="rows"
-      :alt-loading="canPaginate"
+      :alt-loading="canPaginate && !isFirstLoad"
```
codyrancher, line 273: We really only want alt-loading to happen on manual refresh and when changing pages. It seemed easiest to base this on whether or not it was the first load or not.

richard-cox (commented):
Little bit nervous getting this one in before the two big PRs, can we wait for them or base this on on top of the later?

codyrancher:
> Little bit nervous getting this one in before the two big PRs, can we wait for them or base this on on top of the later?

Yeah, I'm in favor of waiting until after they're merged.

shell/components/ResourceList/index.vue
codyrancher: Added the same logic to PaginatedResourceTable

shell/components/SortableTable/index.vue
codyrancher: I went ahead and just removed the leading logic. The benefits were minor and fixing the async button instead is more involved.

richard-cox (approved):

Issue #12895, opened 2024-12-17
Follow-up on improving Fleet performance


aruiz14:
Follow-up of #12424, where a few areas of improvement were identified but was mainly focused on the resources table, which suffered the biggest slowdown.

**Setup**
- Rancher version: 2.9 - 2.10

**Describe the bug**
See original issue for details.

**Additional context**
Areas of improvement:
 - Continuous Delivery dashboard iterates over all bundles and clusters multiple times *per repository* (for calculating the badge, tooltip, etc.)
   - This information is already available in the GitRepo resource status
 - While `resourcesStatuses` was improved as part of the original issue, `clusterResourceStatus` is still inefficient.
   - Many of it usages use `find` to pick only one cluster. We need to review and address them.
 - Cluster details pages iterate over all clusters, to later filter by the one being displayed, which is not ideal.

IsaSih:
Satisfactory performance. Validated in 
Rancher | v2.11-3d729db8a848fb2523bbc2f3f21aa5367dcdcc69-head
Dashboard | master df18acc
Rancher CLI | v2.11.0-rc.1

PR #12896, opened 2024-12-17
Improve Fleet performance

aruiz14:
### Summary
Issue #12895

Performance improvements while keeping the same behavior.

(Changes are [logically split into different commits](https://github.com/rancher/dashboard/pull/12896/commits))

- Standardize on `clusterID` vs. "clusterLabel" (which normally refers to the *management* Cluster resource name, while the Fleet Cluster name is usually the same as the "display" name).
  - The `clusterId` is compound by `${namespace}/${name}` of the `Cluster` resource, making it easier to work with.
  - The `management` "Cluster" object (`clusterLabel`) is only really needed for calculating the `detailLocation` of deployed resouces, as they will link to the Explorer, which uses the management clusters' names.
- Refactor usages of `clusterResourceStatus`, which is always combined with a `find` by `clusterLabel` (now `clusterId`).
  - Used to obtain the resource statuses count per cluster.
  - Used to get the cluster state (based in the resource statuses aggregation).
  - This was being performed for every cluster individually, so it impacts performance exponentially to the number downstream clusters.
  - Reworking this allows to finally get rid of this function completely 🎉 .
- The Continuous Delivery dashboard iterates over all gitrepos/bundles/clusters in order obtain the badges and tooltip for every single entry in the table.
  - I've computed the necessary information in a single place and passed it to the relevant functions so they can easily pick what they need, hence greatly reducing the number of iterations (see screenshot below)
- Small optimizations (e.g. use a variable to avoid calling the same getters multiple times within the same function).
 

### Technical notes summary

Above changes were driven by capturing and analyzing Performance traces with the Chrome Dev Tools, on a local Rancher installation where many sample workloads were installed.
See a summary of CPU usage over 30s of loading the CD dashboard:
 - Before:
Screenshot 2024-12-10 at 12 42 50
 - After:
Screenshot 2024-12-10 at 12 42 59

 - Continuous Delivery:
   - Dashboard
   - Cluster list
   - Cluster details
   - GitRepo list
   - GitRepo details

These changes aim for improving performance while not introducing any changes to the information being displayed.
I've performed several tests using a local UI build and couldn't find any difference, but I may have missed something.

### Checklist
- [X] The PR has a Milestone 


richard-cox:
@aruiz14 i had to re-run all jobs given the build job output had expired, this has brought in an issue in check-i18n-links that should be fixed with a rebase

torchiaf:
@aruiz14 I found a difference in status color between main and your branch, I'm not sure if it is a wanted change or implicit bug fix:

image

shell/models/fleet.cattle.io.gitrepo.js
```diff
@@ -325,6 +329,37 @@ export default class GitRepo extends SteveModel {
     return bds.filter((bd) => bd.metadata?.labels?.['fleet.cattle.io/repo-name'] === this.name);
   }
 
+  get allBundlesStatuses() {
+    const { nonReadyResources, ...bundlesSummary } = this.status?.summary || {};
```
torchiaf, line 333: Is it correct to use GitRepo `status` to calculate Bundles `statuses` ?
I have an active bundle associated to my `test` GitRepo in my test environment but the UI shows `0/0`, I think this could be one of the root causes.

GitRepo `test`

image

Dashboard page - 'Ready Bundles' column is `0/0`:

image

torchiaf (changes requested):
Left some comments about Dashboard -> Status columns.

shell/models/fleet.cattle.io.gitrepo.js
aruiz14: Yes, GitRepo's `[status.summary](https://github.com/rancher/fleet/blob/main/pkg/apis/fleet.cattle.io/v1alpha1/status.go#L5)` is indeed a `[BundleSummary](https://github.com/rancher/fleet/blob/main/pkg/apis/fleet.cattle.io/v1alpha1/bundle_types.go#L247)` type, which aggregates the status of all the Bundles that were deployed.

From the screenshots, I can see that you are using just a Local cluster with no downstreams configured, while creating a GitRepo within the `fleet-default` workspace, which excludes the local cluster. In that situation, it makes sense that the number of matching clusters and bundles is zero.

shell/pages/c/_cluster/fleet/index.vue
```diff
@@ -170,96 +185,59 @@ export default {
...
-            icon:       STATES[STATES_ENUM.ACTIVE].compoundIcon
-          };
-        }
-      }
+      const state = primaryDisplayStatusFromCount(group.states);
 
       return {
-        badgeClass: `${ STATES[STATES_ENUM.NOT_READY].color } badge-class-area-${ area }`,
-        icon:       STATES[STATES_ENUM.NOT_READY].compoundIcon
+        badgeClass: STATES[state].color ? STATES[state].color : `${ STATES[STATES_ENUM.UNKNOWN].color } bg-unmapped-state`,
```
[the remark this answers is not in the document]
aruiz14: Good catch, thanks. Addressed in 1d7a0e230f0c2194728646a0778bb1368d04cebc

shell/models/fleet.cattle.io.gitrepo.js
aruiz14: I was wrong, despite its name, `BundleSummary` contains `BundleDeployment` counts, so this won't work.
I reworked it to obtain the Bundles statuses from BundleDeployments directly (still avoiding to use `Bundles` directly just for this, since they are expensive to download/process).

shell/utils/fleet.ts
```diff
@@ -28,6 +30,14 @@ function resourceKey(r: BundleResourceKey): string {
   return `${ r.kind }/${ r.namespace }/${ r.name }`;
 }
 
+function conditionIsTrue(conditions: Condition[], type: string): boolean {
```
torchiaf, line 33:
```suggestion
function conditionIsTrue(conditions: Condition[] | undefined, type: string): boolean {
```

torchiaf (approved):
LGTM, I added a commit suggestion just to fix a parameter type.

torchiaf (approved):

PR #12966, opened 2025-01-02
fix keyboard nav for home page

### LINKED ISSUE #12776: a11y: Fix keyboard navigation and focus display issues on the home screen
a11y Issues: 62, 64, 65, 69, 60

Fix keyboard navigation and focus issues on the home screen:

**In the clusters table** - sorting icons do not support keyboard navigation or show focus - these should be fixed in the Sortable Table component.

image

**Preferences link in the banner** - link does not support keyboard navigation or show focus

image

**Banner Close Button** - close button does not support keyboard navigation or show focus

image

**Banner Image Close Button** - close button does not support keyboard navigation or show focus

image

** Issue 60 - Textual description missing for interactive image**

These two images are missing the `aria-label` attribute:

image
image

aalves08:
### Summary
Fixes #12776 

- Fix keyboard navigation on `home` page
- checkbox is now visibly focusable
- `Banner` component has correct roles, labels and is focusable
- `SortableTable` component has been fixed for basic keyboard nav (sorting, checkboxes, links, open/close row action menu)

- Go to the `home` page and browse the page with the tab key
- use `enter` or `space` key to trigger UI elements on the page

### Screenshot/Video

https://github.com/user-attachments/assets/47ee2472-9fb8-445d-b808-a4330f7ad9d2

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/pkg/rancher-components/src/components/Form/Checkbox/Checkbox.vue
+++ b/pkg/rancher-components/src/components/Form/Checkbox/Checkbox.vue
@@ -227,7 +227,7 @@ export default defineComponent({
         :checked="isChecked"
         :value="valueWhenTrue"
         type="checkbox"
-        :tabindex="isDisabled ? -1 : 0"
+        tabindex="-1"
         :name="id"
         @click.stop.prevent
         @keyup.enter.stop.prevent
@@ -235,7 +235,7 @@ export default defineComponent({
       <span
         class="checkbox-custom"
         :class="{indeterminate: indeterminate}"
-        :tabindex="-1"
+        :tabindex="isDisabled ? -1 : 0"
         :aria-label="label"
         :aria-checked="!!value"
         role="checkbox"
--- a/pkg/rancher-components/src/components/Form/Checkbox/Checkbox.vue
+++ b/pkg/rancher-components/src/components/Form/Checkbox/Checkbox.vue
@@ -230,11 +230,12 @@ export default defineComponent({
         tabindex="-1"
         :name="id"
         @click.stop.prevent
+        @keyup.enter.stop.prevent
       >
       <span
         class="checkbox-custom"
         :class="{indeterminate: indeterminate}"
-        :tabindex="isDisabled ? -1 : 0"
+        :tabindex="-1"
         :aria-label="label"
         :aria-checked="!!value"
         role="checkbox"
@@ -340,6 +341,12 @@ $fontColor: var(--input-label);
     opacity: 0;
     position: absolute;
     z-index: -1;
+  }
+
+  input:focus-visible ~ .checkbox-custom {
+    @include focus-outline;
+    outline-offset: 2px;
+    border-radius: 0;
   }
 
   input:checked ~ .checkbox-custom {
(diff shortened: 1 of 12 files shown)


shell/components/SortableTable/THead.vue
```diff
@@ -235,7 +235,11 @@ export default {
         :align="col.align || 'left'"
         :width="col.width"
         :class="{ sortable: col.sort, [col.breakpoint]: !!col.breakpoint}"
+        :tabindex="col.sort ? 0 : -1"
+        class="sortable-table-head-element"
         @click.prevent="changeSort($event, col)"
+        @keyup.enter="changeSort($event, col)"
+        @keyup.space="changeSort($event, col)"
```
rak-phillip, line 242: Table headers should also indicate sort order with `aria-sort`[^1], but it doesn't look like there's a readily available way access sort order from within the component..

[^1]: https://www.w3.org/WAI/ARIA/apg/practices/grid-and-table-properties/#gridAndTableProperties_sort

shell/pages/home.vue
```diff
@@ -331,16 +331,19 @@ export default {
...
-              class="hand"
+              class="hand banner-link"
               :href="releaseNotesUrl"
+              role="link"
               target="_blank"
               rel="noopener noreferrer nofollow"
+              :aria-label="t('landing.whatsNewLink')"
               @click.stop="showWhatsNew"
+              @keyup.stop.enter="showWhatsNew"
+              @keyup.stop.space="showWhatsNew"
```
rak-phillip, line 342: Avoid triggering links with `space`.

rak-phillip (commented):
This looks good. I think all the comments I left are fairly minor and have been addressed in other open accessibility PRs.

pkg/rancher-components/src/components/Form/Checkbox/Checkbox.vue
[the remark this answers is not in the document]
aalves08: Agree

shell/components/SortableTable/THead.vue
aalves08: Was able to create a function that give us the sort string according to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-sort#values

shell/pages/home.vue
```diff
@@ -599,6 +611,9 @@ export default defineComponent({
                       :to="createLocation"
                       class="btn btn-sm role-primary"
                       data-testid="cluster-create-button"
+                      role="link"
+                      :aria-label="t('generic.create')"
+                      @keyup.space="$router.push(createLocation)"
```
rak-phillip, line 616: There's quite a few links that still use `space` for activation. I think these can be removed.

nwmac: But this one is a link being used as a button, so in that case I believe it is correct to add `role=button` and to handle the space key.

https://www.w3.org/WAI/ARIA/apg/patterns/button/

rak-phillip: Right.. I think that it makes more sense to just use a button with a single click event so we don't need to worry about handling edge cases like activating on space. I won't argue the point too much since this is an existing implementation, but I will advocate against promoting future usage of this pattern.

rak-phillip (approved):
LGTM

PR #13008, opened 2025-01-07
Add tests for workflows shipped with the creators package

### LINKED ISSUE #12657: Create tests for workflows that we ship from the creators package
Create tests for workflows that we ship from the creators package, like the `build-extension-chart`, which triggers the reusable workflow. This covers problems with the shipped workflow, which we found recently as per:

https://github.com/rancher/dashboard/pull/12636
https://github.com/rancher/dashboard/pull/12637
https://github.com/rancher/dashboard/pull/12638

We are already covering the actual reusable workflow with https://github.com/rancher/dashboard/issues/12032 (dependent on that work)

aalves08:
### Summary
Fixes #12657 

- add script to do content match of creators workflows files 
- update workflows for extensions tests to run this new check

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/.github/workflows/scripts/check-creators-workflows-code.sh
+++ b/.github/workflows/scripts/check-creators-workflows-code.sh
@@ -1,0 +1,79 @@
+#!/bin/bash
+
+set -eo pipefail
+
+SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)"
+BASE_DIR="$(
+  cd $SCRIPT_DIR && cd ../.. &
+  pwd
+)"
+
+CREATOR_WORKFLOWS_DIR=$BASE_DIR/creators/extension/app/files/.github/workflows
+
+WORKFLOW_BRANCH=$1
+
+PROPER_WORKFLOW_CHARTS="name: Build and Release Extension Charts
+
+on:
+  workflow_dispatch:
+  release:  
+    types: [released]
+
+defaults:
+  run:
+    shell: bash
+    working-directory: ./
+
+jobs:
+  build-extension-charts:
+    uses: rancher/dashboard/.github/workflows/build-extension-charts.yml@$WORKFLOW_BRANCH
+    permissions:
+      actions: write
+      contents: write
+      deployments: write
+      pages: write
+    with:
+      target_branch: gh-pages
+      tagged_release: \${{ github.ref_name }}"
+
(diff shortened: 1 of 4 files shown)


cnotv (approved):
I think it may be worth to try the CI on your fork on push, just to be sure.
Anyhow, if it works for you, I see no reason to block this.

LGTM

cnotv (dismissed):

.github/workflows/scripts/check-creators-workflows-code.sh
```diff
@@ -0,0 +1,79 @@
...
+    permissions:
+      actions: write
+      contents: read
+      packages: write
+    with:
+      registry_target: ghcr.io
+      registry_user: \${{ github.actor }}
+      tagged_release: \${{ github.ref_name }}
+    secrets: 
+      registry_token: \${{ secrets.GITHUB_TOKEN }}"
```
jordojordo, line 63: Wouldn't we need to update this script every time we update those workflow files? There should be an easier way to just pull the workflow files directly from the git repo.

I might be misunderstanding the point of this PR, but it seems to be superfluous to just ensure our workflow matches a set scaffold.

cnotv: Given this insight, I may recommend you use GH Actions for this purpose, so they can just import it and bump the version: https://docs.github.com/en/actions/sharing-automations/creating-actions/about-custom-actions

Not sure if we also need to publish it, but I guess that would not be bad. The principle seems similar to the template by using release tags: https://docs.github.com/en/actions/sharing-automations/creating-actions/publishing-actions-in-github-marketplace

jordojordo: Alex and I spoke about this, I think the way he has implemented should be fine for the time being.

jordojordo (dismissed):

jordojordo:
I've simplified this workflow configuration while maintaining all existing functionality. Here's what changed:

- Consolidated Workflows
  - Merged 3 separate workflows (`master`, `release-2.9`, `release-2.8`) into a single workflow using a matrix strategy
- Refactored check-creators-workflows-code.sh to:
  - Use `HEREDOC` for cleaner template management
  - Added explicit error messaging with `::error::` commands
  - Implemented proper diff-based comparison
- Added a dedicated aggregation job to maintain compatibility with existing Slack notifications

jordojordo (approved):

PR #13051, opened 2025-01-10
Allow HMR in Rancher Components and Harvester

rak-phillip:
### Summary
This allows HMR to trigger when changes are made to Rancher Components and Harvester by omitting the `excludes` argument from `getWatcherIgnored()`. Watch and WatchOptions work to allow Webpack to watch files and recompile whenever they change - I don't see any benefit in excluding Rancher Components and Harvester from this process by default.

As I understand, the `excludes` variable exists to control production builds; this change will only affect development.

This also includes a change to clean up the `appConfig` param in `shell/vue.config`.

The prefix `_` is a convention to indicate that a parameter is expected, but unused within a function. We make use of `appConfig` in the default export, so we can remove the `_` and supply a default value for `appConfig` if nothing is passed.

- HMR triggers when running development builds and modifying Rancher Components/Harvester

Perhaps, we were intentionally excluding these from HMR for a reason that I'm not aware of.

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/vue.config.js
+++ b/shell/vue.config.js
@@ -456,7 +456,7 @@ const printLogs = (dev, dashboardVersion, resourceBase, routerBasePath, pl, ranc
  * - as list: [/.shell/, /dist-pkg/, /scripts\/standalone/, /\/pkg.test-pkg/, /\/pkg.harvester/]
  * - as chained regex rule: /.shell|dist-pkg|scripts\/standalone|\/pkg.test-pkg|\/pkg.harvester/
  */
-const getWatcherIgnored = (excludes) => {
+const getWatcherIgnored = (excludes = []) => {
   const paths = [
     /node_modules/,
     /dist-pkg/,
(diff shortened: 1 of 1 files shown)


rak-phillip:
@torchiaf I'm requesting a review from you because harvester is included the the default exports as well[^1]. I'm unsure if there are some historical reasons why we would want to omit Harvester directories from HMR.

[^1]: https://github.com/rancher/dashboard/blob/05462046a3c3680745ff997c378d7be8686e6ffa/vue.config.js#L5

codyrancher (approved):

torchiaf:
> @torchiaf I'm requesting a review from you because harvester is included the the default excludes as well[1](#user-content-fn-1-da775774aede893b8884ed17a0f93e07). I'm unsure if there are some historical reasons why we would want to omit Harvester directories from HMR.
> 

I don't know if we ever had the `pkg/harvester` in the dashboard repository as base repository.
For sure we are already including `harvester-manager` in HMR.

In Harvester, we are not using the `excludes` param, so LGTM

```
module.exports = config(__dirname, {
  excludes: [],
  // excludes: ['harvester']
});
```

torchiaf (approved):
LGTM, I

PR #13085, opened 2025-01-13
Bump Vue to 3.5.13

### LINKED ISSUE #13084: Update Vue to 3.5.x
Dashboard's Vue version is rather out of date since the version was originally pinned during the Vue 3 migration. 

## Acceptance criteria

- [x] Update Vue and related dependencies to 3.5.x
- [x] Issues related to the bump are resolved
- [x] No regressions are introduced as a result of the bump

rak-phillip:
### Summary
This bumps Vue and related project dependencies to versions that are compatible with 3.5.13. 

Fixes #13084

- Bump Vue and related dependencies to versions compatible with 3.5.13
- Resolve error in routes as a result of the bump
- Resolve issues with Vue type augments as a result of the bump

- Basic regression testing of existing features and build processes

### Checklist
- [x] The PR has a Milestone 


creators/extension/pkg/vue-shim.ts
```diff
@@ -1,4 +1,15 @@
-declare module '*.vue' {
-  import Vue from 'vue';
-  export default Vue;
+export {}
```
rak-phillip, line 1: This empty export is in line with guidance from the Vue docs[^1] on the topic of augmenting vue types.

[^1]: https://vuejs.org/guide/typescript/options-api#type-augmentation-placement

pkg/rancher-components/src/shim-vue.d.ts
```diff
@@ -1,11 +1,6 @@
-import { DefineComponent } from 'vue'
+export {}
 
-declare module '*.vue' {
-  const component: DefineComponent<{}, {}, any>
-  export default component
-}
-
-declare module '@vue/runtime-core' {
+declare module 'vue' {
```
rak-phillip, line 3: It appears that we want to provide type augments for `vue` rather than `@vue/runtime-core`[^1][^2]

[^1]: https://vuejs.org/guide/typescript/options-api#type-augmentation-placement
[^2]: https://github.com/lovetingyuan/router/blob/97be7c3462b1d44bf7767a76b7290b6764150895/packages/router/src/globalExtensions.ts#L11

shell/config/router/routes.js
rak-phillip, line 1: The duplicated/nested `index` name was causing errors in vue-router:

```
Error: A route named "index" has been added as a child of a route with the same name. Route names must be unique and a nested route cannot use the same name as an ancestor.
```

I understand that the child with with the same name of `index` is the default route, so removing this parent name should be safe.

shell/package.json
```diff
@@ -41,51 +41,51 @@
...
+    "core-js": "3.40.0",
     "cron-validator": "1.3.1",
     "cronstrue": "2.50.0",
     "cross-env": "6.0.3",
     "css-loader": "6.7.3",
     "csv-loader": "3.0.3",
     "custom-event-polyfill": "^1.0.7",
-    "cypress": "10.3.1",
-    "d3": "7.3.0",
+    "cypress": "9.7.0",
```
cnotv, line 72: Sorry for showing the question like this but, do we need Cypress in shell?

cnotv (commented):
I see some version downgraded, mismatching or not pinned, is that intended?

package.json
```diff
@@ -165,34 +165,34 @@
...
-    "core-js": "3.25.3",
+    "core-js": "3.40.0",
     "css-loader": "6.7.3",
     "csv-loader": "3.0.3",
     "cy2": "4.0.9",
-    "cypress": "11.1.0",
     "cypress-delete-downloads-folder": "0.0.4",
     "cypress-mochawesome-reporter": "^3.8.2",
-    "eslint": "7.32.0",
+    "cypress": "9.7.0",
```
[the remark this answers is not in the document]
cnotv: I think that may make CI fails as well.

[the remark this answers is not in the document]
rak-phillip: The downgraded packages should already be resolved at this point.

shell/package.json
rak-phillip: I cannot answer that and I don't intend to fix it in this PR. Perhaps @aalves08 can provide more details.

pkg/rancher-components/package.json
```diff
@@ -24,34 +24,34 @@
...
+    "eslint-plugin-vue": "^8.0.3",
+    "eslint": "^7.32.0",
     "jsonpath-plus": "10.0.7",
-    "sass": "1.55.0",
-    "sass-loader": "~12.0.0",
+    "sass-loader": "^12.0.0",
+    "sass": "^1.32.7",
     "typescript": "5.6.3",
-    "vue": "~3.2.13",
+    "vue": "^3.2.13",
```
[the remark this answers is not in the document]
rak-phillip: Yes, Rancher Components should be API-compatible with 3.2.13 at this point.

shell/config/router/routes.js
cnotv: Oh I was wondering that part.

rak-phillip:
Regarding the failing e2e tests, these are the test results when running locally on my workstation:

- ✅ [explorer/apps/repositories.spec.ts](https://github.com/rancher/dashboard/actions/runs/12756082246/job/35598678231?pr=13085): Repo Refresh results in correct api requests
- ✅ [workloads/daemonsets.spec.ts](https://github.com/rancher/dashboard/actions/runs/12756082246/job/35598678753?pr=13085): modifying "Scaling and Upgrade Policy" to "On Delete" should use the correct property "OnDelete"
- ✅ [workloads/deployments.spec.ts](https://github.com/rancher/dashboard/actions/runs/12756082246/job/35598678753?pr=13085)
   - ✅ Should be able to scale the number of pods
   - ✅ Should be able to view and edit configuration of pod volumes with no custom component
   - ✅ should be able to add and remove container volume mounts
- ✅ [machine-deployments.spec.ts](https://github.com/rancher/dashboard/actions/runs/12756082246/job/35598679834?pr=13085): can edit a MachineDeployments
- ✅ [machine-sets.spec.ts](https://github.com/rancher/dashboard/actions/runs/12756082246/job/35598679834?pr=13085): can edit a MachineSet
- [machines.spec.ts](https://github.com/rancher/dashboard/actions/runs/12756082246/job/35598679834?pr=13085): can edit a Machine
- ✅ [repositories.spec.ts](https://github.com/rancher/dashboard/actions/runs/12756082246/job/35598679834?pr=13085)
   - ✅ can edit a repository
   - ✅ can clone a repository
   - ✅ can download YAML
   - ✅ can refresh a repository
   - ✅ can delete a repository
   - ✅ can create a repository with basic auth
   - ✅ can create a repository with SSH key
   - ✅ can delete repositories via bulk actions
   - ✅ can create an oci repository with basic auth
- ⚠️ [pages/explorer2/workloads/deployments.spec.ts](https://github.com/rancher/dashboard/actions/runs/12756082246/job/35598682243?pr=13085)
   - ❌ Should be able to scale the number of pods
   - ✅ Should be able to view and edit configuration of pod volumes with no custom component
   - ✅ should be able to add and remove container volume mounts

rak-phillip:
@cnotv @codyrancher @aalves08 I think that this vue bump is in a state that is ready for review.

@aalves08 there is an issue related to `dayjs` and the NeuVector extension. I can't say that I'm quite sure that I understand how this change could cause the failure.. Are you able to provide any insight on this?

aalves08:
> @cnotv @codyrancher @aalves08 I think that this vue bump is in a state that is ready for review.
> 
> @aalves08 there is an issue related to `dayjs` and the NeuVector extension. I can't say that I understand how this change could cause the failure.. Are you able to provide any insight on this?

So, I did a bit of investigation and turns out that neuvector is using `dayjs` coming from `shell` (meaning that there is no mention on their `package.json`), as per https://github.com/neuvector/manager-ext/blob/main/package.json

their yarn.lock have several mentions on `dayjs` but they are mainly as subdependencies: https://github.com/neuvector/manager-ext/blob/main/yarn.lock

I tried to replicate this issue locally by yarn linking shell from this PR to neuvector (don't forget to run `./shell/scripts/typegen.sh` first on dashboard before yarn linking shell!) but I couldn't reproduce it.

neuvector's version in branch `main`, when doing yarn install, uses dayjs `1.11.13`.

from a "normal" shell (from dashboard's current `master`), it installs dayjs `1.8.29`. From this PR, we also get dayjs `1.8.29`, so that doesn't seem to be the problem.

All in all, I don't really understand where the problem really lies. I think that adding a resolution to neuvector might go a long way to fix this, but I've also noted another problem: neuvector's `main` branch is not their latest code 😱 

Screenshot 2025-01-16 at 10 19 55

They do it a bit differently, where they create branches per version, meaning that branch `2.1.0` might have the most updated code on their repo. That subverts what we do in the `check-plugins-build` check 😩 

I would just say to open an issue on their repo, talking about adding a resolution to their extension, but also mentioning that they need to keep `main` up-to-date, otherwise the tests that we have for extensions are meaningless on their context. We can't change the branch we use every time they update their extension... No other "official" extension does that. They all keep `main` as the latest code. After that, ping me with the issue number, so that I can put on slack on a channel specific for the neuvector extension.

For now I would just comment out the test for their extension (also create an issue with label `area/extensions` in dashboard so that we can re-enable it in the future - we don't forget about it) https://github.com/rancher/dashboard/blob/master/shell/scripts/test-plugins-build.sh#L239 and move forward if it's important for this change to go in.

cnotv:
Maybe they can do what we proposed for Rancher a time ago, to have another branch for bookkeeping as stable. 
A common configuration is `development`, `release-x.x.x`, and `master`, where this last one is stable.
Since we do not want to mess with their branching, maybe you can ask them to use another branch and call it as they want (`qa`, `stable`, or whatever) so you could add the fixed name there.

cnotv:
Could we not just install `dayjs` version `1.11.13` and fix it in the shell? I think it is better to have the latest version, which is also the goal of this PR.

rak-phillip:
@aalves08 I commented out the neuvector plugin test, and now capi fails with the following error

```
ERROR  Failed to compile with 1 error9:38:42 PM

 error  in ./index.ts

Syntax Error: Thread Loader (Worker 0)
[BABEL] /home/runner/work/dashboard/dashboard/capi-ui-extension/pkg/capi/index.ts: decode_js.EntityDecoder is not a constructor (While processing: "/home/runner/work/dashboard/dashboard/capi-ui-extension/node_modules/@vue/cli-plugin-babel/preset.js")
    at Generator.next (<anonymous>)
    at Generator.next (<anonymous>)
    at Generator.next (<anonymous>)
```

`decode_js.EntityDecoder is not a constructor` leads me to believe that there are two different versions of Vue present in `node_modules` during the test. Can we schedule some time to sync to work through this? The relationship between plugins and changes in the current PR for these tests isn't 100% clear to me.

rak-phillip:
> Could we not just install dayjs version 1.11.13 and fix it in the shell? I think it is better to have the latest version, which is also the goal of this PR.

This could be an option as well. I would prefer to keep this change as focused on Vue as possible, but I'm willing to bump other dependencies if we find that to be required.

aalves08:
> @aalves08 I commented out the neuvector plugin test, and now capi fails with the following error
> 
> ```
> ERROR  Failed to compile with 1 error9:38:42 PM
> 
>  error  in ./index.ts
> 
> Syntax Error: Thread Loader (Worker 0)
> [BABEL] /home/runner/work/dashboard/dashboard/capi-ui-extension/pkg/capi/index.ts: decode_js.EntityDecoder is not a constructor (While processing: "/home/runner/work/dashboard/dashboard/capi-ui-extension/node_modules/@vue/cli-plugin-babel/preset.js")
>     at Generator.next (<anonymous>)
>     at Generator.next (<anonymous>)
>     at Generator.next (<anonymous>)
> ```
> 
> `decode_js.EntityDecoder is not a constructor` leads me to believe that there are two different versions of Vue present in `node_modules` during the test. Can we schedule some time to sync to work through this? The relationship between plugins and changes in the current PR for these tests isn't 100% clear to me.
> 
> edit: see also
> 
> * [Add resolution for dayjs neuvector/manager-ext#104](https://github.com/neuvector/manager-ext/issues/104)
> * [Enable Neuvector in `test-plugins-build` #13141](https://github.com/rancher/dashboard/issues/13141)

Sure thing @rak-phillip . Just drop me an invite and we can chat about it 🙏

rak-phillip:
In an OOB discussion with @aalves08, we decided to disable the remaining failing extensions from the tests. It was discovered that the failure is related to the version of yarn that Dashboard uses and is not reproducible via other package manager like npm or bun. The issue can be mitigated by setting `"entities": "4.5.0"` as a resolution in affected packages.

Opened issue #13173 so that we can re-enable the tests after this PR merges and the extensions have been patched.

aalves08 (approved):
from my side, as we have came up with a gameplan and all associated issue are created to track the re-enablement of tests in `test-plugins-build` I would say that from the extensions standpoint, this LGTM

codyrancher (approved):
Locally this appears to be working and based on the comment from @aalves08 it looks like we have a path forward.

rak-phillip:
With two approvals, I am going to merge this PR. 

@cnotv let me know if you have additional feedback and I will follow up with a separate issue/PR to address anything that might have been missed.

cnotv:
No no, I still stick to what I said, so also for me ok as we discussed :)

Issue #11680, opened 2024-08-22
Helm App Upgrade: UI should not offer an upgrade if multiple charts match


nwmac:
EPIC: https://github.com/rancher/dashboard/issues/11465

Backend: https://github.com/rancher/rancher/pull/45997

Goal:

Currently, if more than one chart matches for an app, we select the first chart and then if there are updates, we show to the user that an update is available.

Issue https://github.com/rancher/dashboard/issues/11679 should be done before this one.

For this issue, if we still can not detect the unique chart for an app, we should NOT show that an update is available for the app. We had talked about showing a more generic update message that takes the user to a page where they can choose - but I feel this is confusing the user to indicate to them that an update is available in the first place.

For an application where we can not unqiuely find the chart (i.e. there are multiple matching), instead of showing the orange upgrade indicator:

![image](https://github.com/user-attachments/assets/cc626a09-689b-4f68-83b7-df9492b5c84a)

we should show a less obtrusive info icon with a tooltip that says `Unable to identify source chart and repository for this application in order to determine if an upgrade is available`

izaac:
This is closely related to https://github.com/rancher/dashboard/issues/11679

yonasberhe23:
testing cover [here](https://github.com/rancher/dashboard/issues/11679#issuecomment-2719259476)

PR #13086, opened 2025-01-14
Restrict installed apps upgrade

### LINKED ISSUE #11679: Helm App Upgrade: Use more fields when trying to find the correct matching chart for an app
EPIC: https://github.com/rancher/dashboard/issues/11465

For Helm Apps, we attempt to find the chart for the app and then look to see if upgrades are available - if so, we indicate this to the user. A chart may exist in two (or more) Helm repositories. At the moment, we will just chose the first one, which means we might show an upgrade that the user probably doesn't want to make.

We should do a better job of finding the matching chart.

For reference, see: https://github.com/rancher/dashboard/blob/master/shell/models/catalog.cattle.io.app.js#L50

The store has a getter `catalog/chart` () - I suggest we add a new getter `catalog/matchingCharts` or alike that returns all of hte matching charts.

The code here - https://github.com/rancher/dashboard/blob/master/shell/models/catalog.cattle.io.app.js#L50 - should use that new getter AND if there are multiple charts returned, it should check the following additional chart metadata properties to get a better match if possible:

- home
- sources

If it finds a better match it should return that one - otherwise it should return the first, as it does today.

momesgin:
### Summary
Fixes #11679 and #11680

- Updated the return value of `upgradeAvailable` to a more meaningful and readable format
- Added an option to `catalog/chart` in the store to let returning multiple matching charts
- `matchingChart` now looks into the versions array of a matched chart to make sure it's a right match. Since a version is like a snapshot in time, it'll be safe to check the `home` value against it for an installed app

Unit tests are added to cover different cases for the new approach, but manual tests would be great to make sure it's working as expected.

Make sure `cert-manager` 1.7.1 is installed by CLI. Then add OCI repo for AppCo. Go to the installed apps page and make sure there's no upgrade available.

Install an old version of any apps. Go to the Installed apps page, and check the upgrade is available.

### Checklist
- [x] The PR has a Milestone 


shell/store/catalog.js
```diff
@@ -111,26 +118,26 @@ export const getters = {
...
-      if ( !matching.length ) {
+      if ( !matchingCharts.length ) {
         return;
       }
 
-      if ( preferRepoType && preferRepoName ) {
-        preferSameRepo(matching, preferRepoType, preferRepoName);
+      if (multiple) {
+        return matchingCharts;
       }
```
momesgin, line 138: this's been removed since it was not being used anymore

shell/models/catalog.cattle.io.app.js
```diff
@@ -56,45 +58,66 @@ export default class CatalogApp extends SteveModel {
...
+      multiple: true
+    }) || [];
+
+    if (matchingCharts.length === 0) {
+      return;
+    }
+
+    // Filtering matches by verifying if the current version is in the matched chart's available versions, and that the home value matches as well
+    const thisHome = chart?.metadata?.home;
+    const bestMatches = matchingCharts.filter((m) => m.versions.some((v) => v.version === this.currentVersion && v.home === thisHome));
```
momesgin, line 75: @nwmac as we discussion, I've updated the filtering logic, and it seems to be working well. The only drawback is that it's a bit expensive since it requires iterating through the versions array.

momesgin: I've updated the iteration through versions a bit

nwmac (changes requested):
See comments

shell/models/catalog.cattle.io.app.js
```diff
@@ -49,54 +50,99 @@ export default class CatalogApp extends SteveModel {
...
+      if (thisHome === versions[0]?.home) {
+        return true;
+      }
+
+      for (let i = 1; i < versions.length; i++) {
+        const { version, home } = versions[i];
+
+        // Finding the exact version, if the version is not there, then most likely it's not a match
+        // if the exact version is found, then we can compare the home values when they exist
+        if (version === this.currentVersion && (!home || !thisHome || home === thisHome)) {
```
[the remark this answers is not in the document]
momesgin: @nwmac So there are two steps before reaching this point,

First we filter the matches by repo names that now includes the new label added by the backend:
```
const repoName = chart.metadata?.annotations?.[CATALOG_ANNOTATIONS.SOURCE_REPO_NAME] || this.metadata?.labels?.[CATALOG_ANNOTATIONS.CLUSTER_REPO_NAME];

const matchingCharts = this.$rootGetters['catalog/chart']({
      chartName,
      repoName,
      includeHidden,
      multiple: true
  }) || [];
```
Next, we check the latest version's home URL, this was absolutely needed since AppCo's version history is mutable. I could find cases(like cert-manager if I'm not mistaken) where the exact matched version from AppCo didn't include the home URL, even though the app was originally installed from it:
```
      if (thisHome === versions[0]?.home) {
        return true;
      }
```

After these two steps, we've already significantly narrowed down the cases, especially considering that, previously we didn't filter matches at all. Given this, having a slightly looser condition here shouldn't be a major issue.

I haven't found any problematic cases, but we should definitely keep an eye on this in case any issues arise.

shell/assets/translations/en-us.yaml
```diff
@@ -922,6 +922,8 @@ catalog:
         label: Resources
         busy: The related resources will appear when {app} is fully installed.
       values: Values YAML
+    upgrade:
+      uncertainUpgradeWarningTooltip: Unable to identify source chart and repository for this application in order to determine if an upgrade is available
```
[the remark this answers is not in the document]
momesgin: done

nwmac (approved):

PR #13113, opened 2025-01-14
automation: add step to ensure aks is enabled

yonasberhe23:
### Summary
Fixes #https://github.com/rancher/qa-tasks/issues/1638

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/cypress/e2e/tests/pages/manager/cluster-manager.spec.ts
+++ b/cypress/e2e/tests/pages/manager/cluster-manager.spec.ts
@@ -39,6 +39,7 @@ const clusterNamePartial = `${ runPrefix }-create`;
 const rke1CustomName = `${ clusterNamePartial }-rke1-custom`;
 const rke2CustomName = `${ clusterNamePartial }-rke2-custom`;
 const importGenericName = `${ clusterNamePartial }-import-generic`;
+let reenableAKS = false;
 
 const downloadsFolder = Cypress.config('downloadsFolder');
 
@@ -91,16 +92,35 @@ describe('Cluster Manager', { testIsolation: 'off', tags: ['@manager', '@adminUs
   });
 
   it('deactivating a kontainer driver should hide its card from the cluster creation page', () => {
+    cy.intercept('GET', '/v3/kontainerdrivers').as('getKontainerDrivers');
+    cy.intercept('POST', 'v3/kontainerDrivers/azurekubernetesservice?action=deactivate').as('deactivateDriver');
+    cy.intercept('POST', 'v3/kontainerDrivers/azurekubernetesservice?action=activate').as('activateDriver');
+
     const driversPage = new KontainerDriversPagePo();
     const clusterCreatePage = new ClusterManagerCreatePagePo();
 
-    // deactivate the AKS driver
     KontainerDriversPagePo.navTo();
     driversPage.waitForPage();
+
+    // assert AKS kontainer driver is in Active state
+    cy.wait('@getKontainerDrivers').then(({ response }) => {
+      response.body.data.forEach((item: any) => {
+        if (item.id === 'azurekubernetesservice') {
+          const state = item['active'];
+
+          expect(state).to.eq(true);
+        }
+      });
+    });
+
+    // deactivate the AKS driver
     driversPage.list().actionMenu('Azure AKS').getMenuItem('Deactivate').click();
     const deactivateDialog = new DeactivateDriverDialogPo();
 
     deactivateDialog.deactivate();
+    cy.wait('@deactivateDriver').its('response.statusCode').should('eq', 200).then(() => {
+      reenableAKS = true;
+    });
 
     // verify that the AKS card is not shown
     clusterList.goTo();
@@ -112,6 +132,9 @@ describe('Cluster Manager', { testIsolation: 'off', tags: ['@manager', '@adminUs
     KontainerDriversPagePo.navTo();
     driversPage.waitForPage();
     driversPage.list().actionMenu('Azure AKS').getMenuItem('Activate').click();
+    cy.wait('@activateDriver').its('response.statusCode').should('eq', 200).then(() => {
+      reenableAKS = false;
+    });
 
     // verify that the AKS card is back
     clusterList.goTo();
@@ -827,5 +850,11 @@ describe('Cluster Manager', { testIsolation: 'off', tags: ['@manager', '@adminUs
         clusterCreate.credentialsBanner().checkNotExists();
       });
     });
+  });
+
+  after(() => {
+    if (reenableAKS) {
+      cy.createRancherResource('v3', 'kontainerDrivers/azurekubernetesservice?action=activate', {});
+    }
   });
 });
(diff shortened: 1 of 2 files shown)


izaac:
@IsaSih could you take a look at this too? 👍🏽

cypress/e2e/tests/pages/manager/cluster-manager.spec.ts
```diff
@@ -828,4 +831,10 @@ describe('Cluster Manager', { testIsolation: 'off', tags: ['@manager', '@adminUs
       });
     });
   });
+
+  after(() => {
+    if (reenableAKS) {
```
IsaSih, line 836: I believe that the after-hook will fail if the test fails after the AKS container driver is reactivated or if the whole run is successful. I verified that when this POST request is submitted for the case where AKS is already activated, the following error occurs: 
```
{
"baseType": "error",
"code": "Conflict",
"message": "Driver URL already in use: ",
"status": 409,
"type": "error"
} 
``` 

so the request will have HTTP status different than 200 or 201. I'd like to request that you make the API call just after verifying what is the current state of the AKS kontainer driver.

IsaSih (changes requested):

cypress/e2e/tests/pages/manager/cluster-manager.spec.ts
yonasberhe23: the aftershock passes even when AKS container driver is in Active state. see below screenshot
Screenshot 2025-01-16 at 2 57 54 PM

but I do agree that it should only run if its still deactivated. I've updated the test to set `reenableAKS` to false if it activates AKS successfully that way the request isn't made unnecessarily.

izaac (dismissed):
:shipit:

cypress/e2e/tests/pages/manager/cluster-manager.spec.ts
```diff
@@ -39,6 +39,7 @@ const clusterNamePartial = `${ runPrefix }-create`;
 const rke1CustomName = `${ clusterNamePartial }-rke1-custom`;
 const rke2CustomName = `${ clusterNamePartial }-rke2-custom`;
 const importGenericName = `${ clusterNamePartial }-import-generic`;
+let reenableAKS = false;
```
IsaSih, line 43: Could you please add a before-hook that ensures that the state of the AKS kontainer driver is activated as that's the pre-condition for the test?

cypress/e2e/tests/pages/manager/cluster-manager.spec.ts
```diff
@@ -100,7 +101,9 @@ describe('Cluster Manager', { testIsolation: 'off', tags: ['@manager', '@adminUs
     driversPage.list().actionMenu('Azure AKS').getMenuItem('Deactivate').click();
     const deactivateDialog = new DeactivateDriverDialogPo();
 
-    deactivateDialog.deactivate();
+    deactivateDialog.deactivate().then(() => {
+      reenableAKS = true;
```
IsaSih, line 105: Please, add an API call to ensure that the state of the container driver has successfully changed with the click() before changing `reenableAKS`

cypress/e2e/tests/pages/manager/cluster-manager.spec.ts
IsaSih: After analysis, we verified that the referred error occurs with a POST request, and the test is executing a POST request

cypress/e2e/tests/pages/manager/cluster-manager.spec.ts
```diff
@@ -111,7 +114,10 @@ describe('Cluster Manager', { testIsolation: 'off', tags: ['@manager', '@adminUs
     // re-enable the AKS kontainer driver
     KontainerDriversPagePo.navTo();
     driversPage.waitForPage();
-    driversPage.list().actionMenu('Azure AKS').getMenuItem('Activate').click();
+    driversPage.list().actionMenu('Azure AKS').getMenuItem('Activate').click()
+      .then(() => {
+        reenableAKS = false;
```
IsaSih, line 119: Please, add an API call to ensure that the state of the container driver has successfully changed with the click() before changing `reenableAKS`

IsaSih (changes requested):

cypress/e2e/tests/pages/manager/cluster-manager.spec.ts
yonasberhe23: I was able to accomplish checking the state of the kontainer driver using cy.intercept() so no need to add a step in the beforeAll hook for this. The check asserts kontainer driver is in Active state, if not the test will fail. 

The kontainer driver should be Active by default so the test shouldn't fail there, but if it does then we need to investigate the cause; could be a valid bug or caused by previous AKS related test.

izaac (approved):
:shipit:

cypress/e2e/tests/pages/manager/cluster-manager.spec.ts
```diff
@@ -91,16 +92,35 @@ describe('Cluster Manager', { testIsolation: 'off', tags: ['@manager', '@adminUs
...
 
     // deactivate the AKS driver
     KontainerDriversPagePo.navTo();
     driversPage.waitForPage();
+
+    // assert kontainer driver is in Active state
+    cy.wait('@getKontainerDrivers').then(({ response }) => {
+      response.body.data.forEach((item: any) => {
+        if (item.id === 'azurekubernetesservice') {
+          const state = item['active'];
```
IsaSih, line 110: Please, observe that in the` body.data` of this object, there's another attribute `status:active `. Can you please, verify if there's any difference in how this data is being consumed that could affect the test execution?

IsaSih (changes requested):

cypress/e2e/tests/pages/manager/cluster-manager.spec.ts
izaac: @IsaSih could you elaborate the cases that might impact the execution here and why is it relevant for the scenario?

cypress/e2e/tests/pages/manager/cluster-manager.spec.ts
```diff
@@ -91,16 +92,35 @@ describe('Cluster Manager', { testIsolation: 'off', tags: ['@manager', '@adminUs
...
 
   it('deactivating a kontainer driver should hide its card from the cluster creation page', () => {
+    cy.intercept('GET', '/v3/kontainerdrivers').as('getKontainerDrivers');
+    cy.intercept('POST', 'v3/kontainerDrivers/azurekubernetesservice?action=deactivate').as('deactivateDriver');
+    cy.intercept('POST', 'v3/kontainerDrivers/azurekubernetesservice?action=activate').as('activateDriver');
+
     const driversPage = new KontainerDriversPagePo();
     const clusterCreatePage = new ClusterManagerCreatePagePo();
 
     // deactivate the AKS driver
```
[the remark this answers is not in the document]
yonasberhe23: good call. I moved `deactivate the AKS driver` to 116 but not adding the navigate one bc its not needed imo

cypress/e2e/tests/pages/manager/cluster-manager.spec.ts
IsaSih: Me and @yonasberhe23 verified that both attributes change when the element changes its state so it seems like any of them can be used. The API payloads are just steps to ensure that the state of the element being tested changed, so that we then set `reenableAKS` accordingly

IsaSih (approved):

izaac (approved):
:shipit:

PR #13152, opened 2025-01-18
automation: services - external name tests

yonasberhe23:
### Summary
Fixes #https://github.com/rancher/qa-tasks/issues/1581

This PR also includes small fixes to address flaky Extensions tests.

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/cypress/e2e/po/components/growl-manager.po.ts
+++ b/cypress/e2e/po/components/growl-manager.po.ts
@@ -1,0 +1,23 @@
+import ComponentPo from '@/cypress/e2e/po/components/component.po';
+
+export class GrowlManagerPo extends ComponentPo {
+  constructor() {
+    super('.growl-container');
+  }
+
+  growlList() {
+    return this.self().find('.growl-list');
+  }
+
+  growlMessage() {
+    return this.self().find('.growl-message');
+  }
+
+  dismissWarning() {
+    return this.self().find('.icon-close').click();
+  }
+
+  dismissAllWarnings() {
+    return this.self().find('button.btn').contains('Clear All Notifications').click();
+  }
+}
--- a/cypress/e2e/po/components/tabbed.po.ts
+++ b/cypress/e2e/po/components/tabbed.po.ts
@@ -17,6 +17,10 @@ export default class TabbedPo extends ComponentPo {
     return this.self().get('[data-testid="tabbed-block"] > li');
   }
 
+  assertTabIsActive(selector: string) {
+    return this.self().find(`${ selector }`).should('have.class', 'active');
+  }
+
   /**
    * Get tab labels
    * @param tabLabelsSelector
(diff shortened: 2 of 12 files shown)


izaac:
@a-arias could you take a look at this too?  👍🏽

izaac:
@yonasberhe23 Will this only cover external names ? In that case the task can be break down to smaller ones for `cluster ip` `Headless` `loadBalancer` `nodePort`

Other than that it looks good to me.

cypress/e2e/po/edit/services.po.ts
```diff
@@ -1,18 +1,76 @@
...
-  constructor(clusterId = '_', id?: string) {
-    super(WorkloadsCreateEditPo.createPath(clusterId, id));
+  constructor(clusterId = 'local', namespace?: string, id?: string) {
+    super(ServicesCreateEditPo.createPath(clusterId, namespace, id));
+  }
+
+  resourceDetail() {
+    return new ResourceDetailPo(this.self());
+  }
+
```
a-arias, line 28: it would be good to add a little documentation on these methods to get a quick idea of what UI element they return

cypress/e2e/po/components/growl-manager.po.ts
```diff
@@ -0,0 +1,23 @@
+import ComponentPo from '@/cypress/e2e/po/components/component.po';
+
+export class GrowlManagerPo extends ComponentPo {
+  constructor() {
+    super('.growl-container');
+  }
+
+  growlList() {
+    return this.self().find('.growl-list');
```
a-arias, line 9: is the classes the only way to get these elements? , we could use ids or a combination of classes and names with xpath to create stronger locator strategy.

a-arias (approved):
great job, I added 2 observations that could be potential improvements.

yonasberhe23:
> @yonasberhe23 Will this only cover external names ? In that case the task can be break down to smaller ones for `cluster ip` `Headless` `loadBalancer` `nodePort`
> 
> Other than that it looks good to me.

@izaac ah good point. i can create some sub task tickets for those other areas

cypress/e2e/po/edit/services.po.ts
yonasberhe23: good call Abraham, I added comments for the functions within the ResourceDetailPo file itself (cypress/e2e/po/edit/resource-detail.po.ts) so that we can see the details of the class when its being used

cypress/e2e/po/components/growl-manager.po.ts
yonasberhe23: its already being used in combination with the parent class which is used in the class constructor `'.growl-container'`

cypress/e2e/po/components/tabbed.po.ts
```diff
@@ -17,6 +17,10 @@ export default class TabbedPo extends ComponentPo {
     return this.self().get('[data-testid="tabbed-block"] > li');
   }
 
+  checkSelectedTab(selector: string) {
```
izaac, line 20: I think the method name should include what's being checked here or move the chai assertion to the spec

izaac (changes requested):
Just a nit

cypress/e2e/po/components/tabbed.po.ts
yonasberhe23: ah that's confusing. how about `assertTabIsActive`?

izaac (approved):
:shipit:

PR #13176, opened 2025-01-22
Keyboard nav for extensions main screen

### LINKED ISSUE #12785: a11y: Fix keyboard navigation and focus display for Extensions grid view
a11y Issue: 215

The extensions grid (see below) does not show keyboard focus and does not response to keyboard navigation.

image

aalves08:
### Summary
Fixes #12785 

- Add keyboard navigation for Extensions main screen (normal extensions interface + image catalog load -> check extensions menu)
- Refactor focus traps to composables
- Add focus traps to various dialogs/modals on Extensions screen
- Improve `LabeledInput` accessibility

### Checklist
- [x] The PR has a Milestone 


shell/components/AppModal.vue
```diff
@@ -85,6 +100,11 @@ export default defineComponent({
       };
     }
   },
+  created() {
+    if (this.triggerFocusTrap) {
+      watcherBasedSetupFocusTrap(() => this.modalVisibility, '.modal-container');
+    }
+  },
```
rak-phillip, line 107: In `pkg/rancher-components/src/components/Card/Card.vue`, we do the following

```
  setup(props) {
    if (props.triggerFocusTrap) {
      basicSetupFocusTrap('#focus-trap-card-container');
    }
```

Is there any reason why we can't do the same here? It looks like we would be able to drop the `modalVisibility` prop if we were to achieve this.

pkg/rancher-components/src/components/Card/Card.vue
```diff
@@ -1,8 +1,9 @@
 <script lang="ts">
 import { defineComponent, PropType } from 'vue';
-import { createFocusTrap, FocusTrap } from 'focus-trap';
+import { basicSetupFocusTrap } from '@shell/composables/focusTrap';
```
rak-phillip, line 3: It's a common Vue convention for composable functions to start with `use`. I think that `useFocusTrap` would be a more idiomatic name for this function.

pkg/rancher-components/src/components/Card/Card.vue
```diff
@@ -56,32 +57,17 @@ export default defineComponent({
...
-    }
-  },
+  }
 });
 </script>
 
 <template>
   <div
-    ref="cardContainer"
+    id="focus-trap-card-container"
```
rak-phillip, line 70: Is there any reason we can't stick with the ref? I think that it is preferred over an `id` when working with component instances.

shell/components/AppModal.vue
```diff
@@ -56,6 +57,20 @@ export default defineComponent({
...
       type:    String,
       default: '',
+    },
+    /**
+     * Modal visibility (used for focus-trap)
+     */
+    modalVisibility: {
+      type:    Boolean,
+      default: false,
+    },
```
rak-phillip, line 67: The `modalVisibility` looks redundant to me.. Shouldn't modal instances be conditionally rendered already?

shell/composables/focusTrap.ts
```diff
@@ -0,0 +1,71 @@
+/**
+ * focusTrap is a composable based on the "focus-trap" package that allows us to implement focus traps
+ * on components for keyboard navigation is a safe and reusable way
+ */
+import { watch, nextTick, onMounted, onBeforeUnmount } from 'vue';
+import { createFocusTrap, FocusTrap } from 'focus-trap';
+
+export function basicSetupFocusTrap(containerSelector: string) {
+  let focusTrapInstance = {} as FocusTrap;
```
rak-phillip, line 9: I think that we can just specify the type:

```suggestion
  let focusTrapInstance: FocusTrap;
```

shell/components/AppModal.vue
aalves08: `Card` component uses `basicSetupFocusTrap`, which only relies on Vue lifecycle hooks like `onMounted` and `onBeforeUnmount`.

This is a watcher-based implementation of the focus trap, because the base component here is already mounted (not triggered by a `v-if`) meaning that it needs some variable to be watched in order to achieve the correct behaviour.

pkg/rancher-components/src/components/Card/Card.vue
aalves08: updated function names in focus-trap composable to start with `use`

shell/components/AppModal.vue
aalves08: It's not about the actual render of the modal itself, but we need to propagate a variable related to the visibility of the modal so it can act as a "trigger" (in this case a watcher) for the focus trap, as explained in https://github.com/rancher/dashboard/pull/13176#discussion_r1945097081

pkg/rancher-components/src/components/Card/Card.vue
aalves08: I've expanded all of the composable so that it would use an `HTMLElement` as well as selector (In theory it should work, according to https://github.com/focus-trap/focus-trap).

In this particular case, the actual usage of `useBasicSetupFocusTrap` depends on the `setup` function on which, after multiple tries, I could never get the `ref` to work 😬  . Let me know if you wanna take a stab at this

shell/components/AppModal.vue
rak-phillip: I honestly don't think that we need the watcher-based approach. Each instance of `app-modal` that specifies `:trigger-focus-trap="true"` is conditionally rendered, making the `:modal-visibility` prop redundant. For example, `shell/components/Dialog.vue`:

```html
  <app-modal
    v-if="!closed"
    :name="name"
    height="auto"
    :scrollable="true"
    :modal-visibility="!closed"
    :trigger-focus-trap="true"
    @close="closeDialog(false)"
    @before-open="beforeOpen"
  >
```

Since `app-modal` is conditionally rendered in each instance represented in this PR. We should be able to rely on the simpler `basicSetupFocusTrap` implementation. I gave this a quick test and it appears to be working:

```diff
diff --git a/shell/components/AppModal.vue b/shell/components/AppModal.vue
index 6877bdb505..04f4281c60 100644
--- a/shell/components/AppModal.vue
+++ b/shell/components/AppModal.vue
@@ -1,6 +1,6 @@
 <script lang="ts">
 import { defineComponent } from 'vue';
-import { watcherBasedSetupFocusTrap } from '@shell/composables/focusTrap';
+import { basicSetupFocusTrap } from '@shell/composables/focusTrap';

 export default defineComponent({
   name: 'AppModal',
@@ -58,13 +58,6 @@ export default defineComponent({
       type:    String,
       default: '',
     },
-    /**
-     * Modal visibility (used for focus-trap)
-     */
-    modalVisibility: {
-      type:    Boolean,
-      default: false,
-    },
     /**
      * trigger focus trap
      */
@@ -100,9 +93,9 @@ export default defineComponent({
       };
     }
   },
-  created() {
-    if (this.triggerFocusTrap) {
-      watcherBasedSetupFocusTrap(() => this.modalVisibility, '.modal-container');
+  setup(props) {
+    if (props.triggerFocusTrap) {
+      basicSetupFocusTrap('.modal-container');
     }
   },
   mounted() {
```

Is there any modal in this PR that would fail with this approach?

pkg/rancher-components/src/components/Card/Card.vue
rak-phillip: It looks like it _could_ work if we did something like this in the composable:

```js
export function basicSetupFocusTrap(containerRef: string) {
  let focusTrapInstance = {} as FocusTrap;
  const modalRef = useTemplateRef<HTMLElement>(containerRef);

  onMounted(() => {
    if (!modalRef.value) {
      return;
    }

    focusTrapInstance = createFocusTrap(modalRef.value, {
   ...
```

But this is completely optional if we have a universally unique ID that we can reference instead.

shell/components/AppModal.vue
aalves08: You are right! I must have switched this stuff so many times that I just got confused. AppModal IS conditionally rendered! therefore this works! we can do the change

aalves08: We do have an issue which I wonder if it's due to the way `AppModal` works (teleport).

By default, the focus-trap lib detects where the focus was before activating the actual focus trap. When the trap get's deactivated, it transfers the focus back to this element. It's a really nice way of getting back to the previous context of where the key navigation was before it got "sucked" into focus trap element.

There's this option for `activate` called `setReturnFocus` where we can override this auto-detected starting focus, if you want to send the key navigation back to some other context.

If we don't explicitly set `setReturnFocus` on the `AppModal` `activate` function, what the lib auto-detects is `document.body`, meaning at activation time it has no idea where the focus initially was 💦 

I had to tamper with the source code of the lib to see what target he detected and here's the console.log (check `state.nodeFocusedBeforeActivation` property):
Screenshot 2025-02-07 at 11 24 53


The problem we are seeing here is that even if we override it with `setReturnFocus` on these extension modals, we might end in one of two scenarios:
- if the user performs an action (say install) the modal is dismissed but the install button also disappears, meaning that if you want to return to that install button, it can't and returns an error.
- however, when a user goes into a modal, he can dismiss it  without performing any action. In the case of install, the install button would be there and would be a valid target 💦 💦 💦 💦 💦 

I've spent hours between yesterday and today looking into this and frankly there's no solution without involving what I deem as over-the-top logic for something that should be simpler and kind of straightforward.

My suggestion here is that I create an issue to follow-up on this topic and move this forward.

At this moment I see no easy way out.

Let me know if you want to have a quick sync on this.

rak-phillip: > The problem we are seeing here is that even if we override it with setReturnFocus on these extension modals, we might end in one of two scenarios:

    if the user performs an action (say install) the modal is dismissed but the install button also disappears, meaning that if you want to return to that install button, it can't and returns an error.
    however, when a user goes into a modal, he can dismiss it without performing any action. In the case of install, the install button would be there and would be a valid target 💦 💦 💦 💦 💦

We might require some shared state to specify a default target if the underlying DOM output changes as a result of activating a modal or taking action in a modal. I haven't tested this, but we might be able to utilize the `checkCanReturnFocus()` method in conjunction with `setReturnFocus()` to specify a fallback. In your example above, that might be the active tab or the first element in the list. For action buttons, we would need to return focus to the button the triggered the menu.

aalves08: Ok, so I've updated the PR with a solution, which is the best that we can do.

After we set a `setReturnFocus` on trap activation, we can't change it anymore. Given that fact, the condition to evaluate where to move the focus to has to be baked in this method. I thought to redirect the user for the install/uninstall scenario to the `reload` button if it was visible, but at the time of trap deactivation the button is never visible due to the async nature of helm charts operations.

The library has some limitations and our architecture poses some challenges as well.

Next best thing is to send the user to the first iterable key nav item in the extensions page 🤷 I've given so many hours to this and I can't come up with anything that would work "ideally". I think we will just need to accept the fact

pkg/rancher-components/src/components/Card/Card.vue
aalves08: Created https://github.com/rancher/dashboard/issues/13344 to track this. Assigned it to `2.12` milestone as it's definitely not a priority

shell/components/AppModal.vue
rak-phillip: I still see some inconsistent behavior with dropdown menus that trigger modals. For example: 

- Navigate to Extensions
- Click on Extensions Menu and click on Add Rancher Repositories
- Dismiss the menu

Instead of the Extensions Menu trigger receiving focus, it looks like this gets reset to the root document. I think that it makes sense to create an issue so that we can follow-up and explicitly address these scenarios.

shell/composables/focusTrap.ts
```diff
@@ -0,0 +1,92 @@
...
+      });
+    }
+  }, { immediate: true });
+
+  onBeforeUnmount(() => {
+    if (Object.keys(focusTrapInstance).length) {
+      focusTrapInstance.deactivate();
+    }
+  });
+}
```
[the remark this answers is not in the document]
aalves08: fixed

aalves08:
https://github.com/rancher/dashboard/pull/13176#discussion_r1949594134

I've updated the `Developer install` and `add extension repos` modals. I think that covers it all if I am not mistaken 🙏

shell/pages/c/_cluster/uiplugins/DeveloperInstallDialog.vue
```diff
@@ -18,17 +18,18 @@ export default {
...
-      canModifyName:     true,
-      canModifyLocation: true,
-      showModal:         false,
+      name:                '',
+      location:            '',
+      persist:             false,
+      canModifyName:       true,
+      canModifyLocation:   true,
+      showModal:           false,
+      returnFocusSelector: '[data-testid="extensions-page-menu"]'
```
rak-phillip, line 27: This is fine for now, but I still think there's a design problem that we need to solve if we plan on using this more in the future. `returnFocusSelector` requires that components like `DeveloperInstallDialog` have knowledge of parent or sibling components. IMO it would be best if we could have some localized data store that will allow us to specify the `returnFocusSelector` closer the the action that will be triggering the modal.

rak-phillip (approved):
Functions well and does what we need for now. Can we create an issue to follow up the design of how we set return focus? I think we're missing some component interaction that will pull everything together in a clean and predictable way.

shell/pages/c/_cluster/uiplugins/DeveloperInstallDialog.vue
aalves08: I've created https://github.com/rancher/dashboard/issues/13367 to follow-up with this topic. 

It also covers the part above about the return target for the focus.

PR #13184, opened 2025-01-23
CloudCredential Model: Fix CC expiry display

m-ildefons:
Fix CloudCredential expiry date not being displayed if it's more than a week in the future.
In this case, the behavior was the same as if there wasn't an expiry date at all. This was confusing to users.
Fix it by always displaying the expiry date for CloudCredentials if it's set.

### Summary
Fixes https://github.com/rancher/dashboard/issues/13186

### Technical notes summary

This fix changes the model such that if an expiration-timestamp is set, `expireData` will always return an object. It will only return `null`, if there is no expiration-timestamp set at all, which can be the case for CloudCredential objects for infrastructure providers other than Harvester.

Cloud Credential List

Case 1)
Cloud Credentials for non-Harvester infrastructure providers should be correctly displayed without an expiration date, if they don't have an expiration date

Case 2)
Cloud Credentials for Harvester should be correctly displayed with their expiration date clearly visible regardless of when the expiration date is

Case 3)
Cloud Credentials for Harvester should be displayed with a info/highlighting if the expiration date is within the next 7 days

Case 4)
Cloud Credentials for Harvester should be displayed with an warning/highlighting if the expiration date is in the past

Cloud Credential List

### Screenshot/Video

Screenshot at 2025-01-23 13-08-50
### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/models/cloudcredential.js
+++ b/shell/models/cloudcredential.js
@@ -277,7 +277,10 @@ export default class CloudCredential extends NormanModel {
       };
     }
 
-    return null;
+    return {
+      expired:  false,
+      expiring: false,
+    };
   }
 
   get expiresString() {


richard-cox (changes requested):
As per slack conversation.
- This is probably a bug, rather than expected for other reasons, but i'd like to double check
- There's a much better (safer, less impacting) fix where the formatter alone is updated. This reduces possible regressions and also makes review and testing (automation / manual) easier

Checklist needs filling out.

We should probably look into this when addressing non-expiring harv creds

m-ildefons:
I agree, we need to look into this some more and make sure changing the model does not introduce a regression elsewhere, but I'm not convinced that a fix to the formatter alone is going to cut it.
The model conflates CloudCredentials that don't have an expiration date at all with CloudCredentials that are not expired and not about to expire.

PR #13204, opened 2025-01-24
Add error handling to extension load

### LINKED ISSUE #13205: Errors are not handled on extension load
Internal reference: SURE-9597

We don't handle errors during plugin load, so if you try and load a Vue 2 extension into Vue 3, the UI breaks.

We rely on correctly annotated extensions to avoid this situation, but we should look to see if we can improve the error handling. 

For QA: Dev testing should be adequate, just install an extension and verify that it loads correctly.

nwmac:
### Summary
Fixes #13205

### Technical notes summary

This PR improves the error handling when we load extensions.

I checked out the elemental-ui extension with the 2.9 branch, built it and served it up locally and developer loaded it into a 2.11 UI running locally - to reproduce trying to load a Vue 2 extension into Vue 3.

We don't have any exception handling in the load - the script for the extension loads, so our onloadhandler gets called. The script has an exception when it is evaluated at load time (error reading 'extend'), but this does not affect our code path - we check window[id]`  is set, but because the injected script errored, it is an empty object and does not have the default function, so the real error for us is when we call that method here - https://github.com/rancher/dashboard/blob/master/shell/core/plugins.js#L116 - this throws an unhandled exception.

In the PR, I have added a check for the default function and added a try/catch around the call of it.

With this, the UI handles the load error. Note when running in dev, you will get the red overlay saying there is an unhandled exception (this is due to the error with the script load), but if you close it, our UI is working fine - before, you'd be stuck with the loading spinner. In production you wont' get the red overlay.

Extension loading

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/core/plugins.js
+++ b/shell/core/plugins.js
@@ -92,7 +92,7 @@ export default function(context, inject, vueApp) {
           element.dataset.purpose = 'extension';
 
           element.onload = () => {
-            if (!window[id]) {
+              if (!window[id] || (typeof window[id].default !== 'function')) {
               return reject(new Error('Could not load plugin code'));
             }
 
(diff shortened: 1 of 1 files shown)


jordojordo (approved):
lgtm - I was unable to trigger the "Could not initialize plugin" error but I did test with multiple incompatible extensions and noticed I could still navigate around without the `window[id].default is not a function` error.

nwmac:
@jordojordo I had to rebase and other changes went in to remove extension unload. Can you eyeball again and approve if okay. Thanks

jordojordo (approved):
lgtm

PR #13228, opened 2025-01-29
make home banner image close button always visible because of accessibility

### LINKED ISSUE #12776: a11y: Fix keyboard navigation and focus display issues on the home screen

aalves08:
### Summary
Fixes #12776 

Based on the comment from QA: https://github.com/rancher/dashboard/issues/12776#issuecomment-2620335461

- make home banner image close button always visible because of accessibility (was only visible on hover)

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/BannerGraphic.vue
+++ b/shell/components/BannerGraphic.vue
@@ -84,7 +84,7 @@ export default {
       }
     }
 
-    &:hover .close-button {
+    .close-button {
       visibility: visible;
       position: absolute;
       right: 4px;
@@ -96,7 +96,7 @@ export default {
       cursor: pointer;
       opacity: 0.4;
 
-      &:hover {
+      &:hover, &:focus, &:focus-visible {
         background-color: var(--accent-btn-hover);
         color: var(--accent-btn-hover-text);
         opacity: 1;


cnotv (approved):
I guess you mean on focus too, because otherwise now it's still 0.4 opacity.
Either case for me it's fine. If you do not mind to add a screenshot next time, sorry for asking.

LGTM

nwmac:
We should check this with UX - I don't like always showing the close button - there must be a pattern for hover close. I'd suggest the banner should be tabbable and pressing enter/space will close the banner.

aalves08:
@nwmac we already have a solution for banners, which allows us to focus on the close button. 

Imho, I don't think it's not a good idea to create yet another pattern to handle this since it's even more code for us to maintain/handle, on a situation that is an exception and not the "rule" for banners (almost all of the banners have the close button always visible - I'm sure we may find a couple that are like this one on the whole UI).

Any thoughts @rak-phillip ?

aalves08:
May I suggest we move this forward and create an issue to track another solution for this edge-case?

rak-phillip:
I might have a compromise that can satisfy both constraints of maintain the current behavior of showing the close button on hover while allowing the close button to be navigable via keyboard. We can control visibility with opacity via:

```diff
diff --git a/shell/components/BannerGraphic.vue b/shell/components/BannerGraphic.vue
index 4e1dc69982..ffa01c88ed 100644
--- a/shell/components/BannerGraphic.vue
+++ b/shell/components/BannerGraphic.vue
@@ -76,23 +76,27 @@ export default {

     .close-button {
       position: absolute;
-      visibility: hidden;
+      right: 4px;
+      top: 4px;
+      font-size: 16px;
+      padding: 4px;
+      display: flex;
+      align-items: center;
+      visibility: visible;
+      opacity: 0;
+      transition: opacity 0.3s;

       &:focus-visible {
         @include focus-outline;
         outline-offset: 2px;
+        visibility: visible;
+        opacity: 0.4;
       }
     }

     &:hover .close-button {
       visibility: visible;
-      position: absolute;
-      right: 4px;
-      top: 4px;
-      font-size: 16px;
-      padding: 4px;
-      display: flex;
-      align-items: center;
+      opacity: 1;
       cursor: pointer;
       opacity: 0.4;
```

My preferred approach is to always display the button, but this approach can move things forward while we organize a more in-depth UX review.

aalves08:
Thanks for the reply @nwmac . Going to close this PR 🙏

PR #13375, opened 2025-02-11
Fix select behavior for ResourceTable when the enter key is pressed

### LINKED ISSUE #13239: a11y: keyboard navigation performing incorrect action on resources list
**Setup**
Rancher version:
Rancher | v2.11-a1bda936c809e0394520c9d40d2998529223401e-head
Dashboard | master https://github.com/rancher/dashboard/commit/6eaaa6aa9cceca28dea5385bfea4de228ce09c49
Rancher CLI | v2.11.0-rc.1

**Describe the bug**
In pages with list of resources, in this example, the cluster dashboard, when the user navigates to the certificates tab using the keyboard navigation, then hit enter/return key to select the checkbox of a certificate in the table, instead of the resource being selected, the user is redirected to the secrets page. Performing the same action with the space bar, simply selects the resource, as expected.

**To Reproduce**
In the cluster dashboard page, navigate to the certificates tab using the keyboard navigation, then use enter/return key to select the checkbox of a certificate in the table. Observe that the UI redirects the user to another page.

**Result**
The enter/return key to select the checkbox of a certificate in the table, redirects the user to another page instead of performing the action to select the resource.

**Expected Result**
The enter/return key must select the checkbox of a given resource.

**Screenshots**

https://github.com/user-attachments/assets/e0e15453-de03-4138-9abf-efbc7472378b

rak-phillip:
### Summary

This refactors the event listener that was attached to the window so that it is now attached the sortable table. The table needs to listen to the keyup event and the row needs to receive focus in order for this to work; the row's tabindex is set to -1 so that it is not navigable via keyboard.

Fixes #13239

- Refactor event listener for the enter key so that it is no longer attached to the window object, but the sortable table instead
- Prevent event propagation when the action menu button is activated via the enter key

### Technical notes summary

The design of the Checkbox component is quite interesting in that it customizes how events are emitted, rendering common strategies to stop event propagation via `keydown.enter.stop` useless. Rather than attempt a refactor of the Checkbox implementation at this time, I thought it to be easier to resolve the problem within the listener.

- Keyboard interactions with the ResourceTable
  - Pressing enter on the selection checkbox will no longer route to the selected line item
  - Pressing enter on the actions menu will no longer route to the selected line item
- Mouse & Keyboard interactions for selecting a line item
  - Clicking on a row to select it should still work
  - Pressing enter should route to the selected row

It was technically possible to select a row with the mouse, remove focus from the table, and press enter to route to the selected item. I don't think that this use case makes sense.. I think that the table should minimally be the current active element in order to register the enter key action.

### Screenshot/Video

NA

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/SortableTable/index.vue
+++ b/shell/components/SortableTable/index.vue
@@ -1080,6 +1080,7 @@ export default {
                 :disabled="!act.enabled"
                 :data-testid="componentTestid + '-' + act.action"
                 @click="applyTableAction(act, null, $event)"
+                @keydown.enter.stop
                 @mouseover="setBulkActionOfInterest(act)"
                 @mouseleave="setBulkActionOfInterest(null)"
               >
--- a/shell/components/SortableTable/index.vue
+++ b/shell/components/SortableTable/index.vue
@@ -1,6 +1,6 @@
 <script>
 import { mapGetters } from 'vuex';
-import { defineAsyncComponent } from 'vue';
+import { defineAsyncComponent, useTemplateRef, onMounted, onBeforeUnmount } from 'vue';
 import day from 'dayjs';
 import isEmpty from 'lodash/isEmpty';
 import { dasherize, ucFirst } from '@shell/utils/string';
@@ -42,7 +42,14 @@ import ButtonMultiAction from '@shell/components/ButtonMultiAction.vue';
 export default {
   name: 'SortableTable',
 
-  emits: ['clickedActionButton', 'pagination-changed', 'group-value-change', 'selection', 'rowClick'],
+  emits: [
+    'clickedActionButton',
+    'pagination-changed',
+    'group-value-change',
+    'selection',
+    'rowClick',
+    'enter',
+  ],
 
   components: {
     THead,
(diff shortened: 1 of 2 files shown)


shell/components/SortableTable/index.vue
```diff
@@ -518,6 +525,23 @@ export default {
       immediate: true
     },
   },
+  setup(_props, { emit }) {
+    const table = useTemplateRef('table');
+
+    const handleEnterKey = (event) => {
+      if (event.key === 'Enter' && !event.explicitOriginalTarget.classList.contains('checkbox-custom')) {
```
aalves08, line 532: This yields an error for me:
Screenshot 2025-02-12 at 09 56 11

I didn't know about `explicitOriginalTarget` property... Apparently in non-standard and that's probably why this is not working on Chrome.

May I suggest we do `!event.target?.classList?.contains` instead?

aalves08 (changes requested):
Thank you for tackling this one @rak-phillip 🙇 🙇 🙇 

Left a comment

aalves08 (approved):
LGTM

PR #13401, opened 2025-02-13
fix focus return async btn

### LINKED ISSUE #12793: a11y: Fix focus and role on error banner control
a11y Issue: 54

When users submit "Login" form empty or with errors, error message get dynamically displayed on the page and user's focus moves to top of the page that resulted in screen reader users finding difficult to find the errors occurred for the form fields.

image

Affects these pages:

Login 
- Invalid username or password. Please try again. 

User: Create
- Password must be at least 12 characters

Import Extension Catalog
-  Unable to determine image name

aalves08:
### Summary
Fixes #12793 

- fix focus return async btn

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/AsyncButton.vue
+++ b/shell/components/AsyncButton.vue
@@ -117,7 +117,10 @@ export default defineComponent({
       type:    Boolean,
       default: false,
     },
-
+    returnFocusOnError: {
+      type:    Boolean,
+      default: false,
+    },
   },
 
   setup() {
@@ -231,7 +234,7 @@ export default defineComponent({
   },
 
   methods: {
-    clicked() {
+    clicked(ev: any) {
       if ( this.isDisabled ) {
         return;
       }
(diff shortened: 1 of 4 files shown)


shell/components/AsyncButton.vue
```diff
@@ -246,16 +249,21 @@ export default defineComponent({
...
 
-    done(success: boolean | 'cancelled') {
+    done(success: boolean | 'cancelled', ev?: any) {
       if (success === 'cancelled') {
         this.phase = ASYNC_BUTTON_STATES.ACTION;
       } else {
+        // ev.detail = 0 means that the trigger was a key press
+        if (!success && this.returnFocusOnError && ev && ev.detail === 0) {
+          // set the focus back on the async button. We lose it when saving a "form" HTML element
+          setTimeout(() => this.focus(), 0);
```
rak-phillip, line 265: I'm not 100% sure how I feel about this approach. A user can shift focus to different inputs while waiting for an async button to complete. We will steal focus back to the button in the event of an error. I can see this being frustrating in the following scenario:

1. I improperly fill out a form 
2. Click on the async button to submit
3. While the async button is waiting, I realize that I made an error and focus an input to make a correct
4. The async button enters an error state and steals focus away from me while I'm trying to make a correction

rak-phillip (commented):
The source of focus resetting in the async button has to deal with how we manage the disabled state to display the spinner: 

https://github.com/rancher/dashboard/blob/1f865169226cc3940232900e67d3d2ff2af5f881/shell/components/AsyncButton.vue#L293

I think we need to consider alternate approaches that will improve the usability of this button. We might be able to consider `AsyncButton` as a "compsable widget"[^1] and express a "disabled" state while loading in a different way.

[^1]: https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#focusabilityofdisabledcontrols

aalves08:
@rak-phillip I am going to close this PR. I have to agree with you. Solution is hacky and not ideal.

I don't think there's an "easy" way out of this. Probably the ideal scenario is for this to work with form validation and set the focus on the "first" input that is invalid/with error, but we just don't have the code in place to do that. It would be a great amount of work for tacking something "simple" as it's the scope of this original issue #12793 .

I think we should bump this issue out of `2.11`. I reached out to Neil about this and there are plans to tackle form validation in `2.12` scope/timeframe.

I think this issue should be evaluated in conjunction with the scope of that work.

Let me know your thoughts 🙏

Issue #11224, opened 2024-06-13
No proper redirection to original URL after login


asdil12:
**Setup**
- Rancher version: v2.8.3
- Rancher UI Extensions: v2.8.3
- Browser type & version: Vivaldi (current version)

**Describe the bug**

**To Reproduce**
* Open https://op-prg2.openplatform.suse.com/dashboard/c/c-m-t7lmqchx/explorer/apps.deployment/qem-bot/qem-bot
* Log in

**Result**
* A different URL than https://op-prg2.openplatform.suse.com/dashboard/c/c-m-t7lmqchx/explorer/apps.deployment/qem-bot/qem-bot is shown

**Expected Result**

* https://op-prg2.openplatform.suse.com/dashboard/c/c-m-t7lmqchx/explorer/apps.deployment/qem-bot/qem-bot is shown

gaktive:
Candidate to backport, unless it's somehow quick to sort out for 2.9.0

gaktive:
The scenario points to a bookmark to a specific page that requires login to access. Sounds like a lost preference.

codyrancher:
There are already several ways to specify how to redirect the user after login. I decided to support the `bookmark` case only if the user has selected the `Take me to the area I last visited` preference. This will prevent us from conflicting with the other preferences but we can always override that in the future if we'd like to.

![Image](https://github.com/user-attachments/assets/38960a0b-b9cb-4777-a9ce-607bf4cd4059)

gaktive:
Reopening for QA to evaluate automation.

yonasberhe23:
Requires manual validation.

PR #13450, opened 2025-02-19
Adding a after login redirect for the case when a user enters a url but gets redirected to the login page

codyrancher:
### Summary
Fixes https://github.com/rancher/dashboard/issues/11224

I fixed and simplified a few of the code-paths regarding the `Take me to the area I last visited` user preference. This should help prevent future breakages of this feature and reduce a small amount of a our tech debt.

Login landing page. Verifying that the user it taken to the correct landing page after logging in.

Login landing page. Verifying that the user it taken to the correct landing page after logging in.

### Screenshot/Video
https://github.com/user-attachments/assets/1f634715-49a9-4ad2-840e-dc9d954f8270

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/cypress/e2e/tests/pages/user-menu/preferences.spec.ts
+++ b/cypress/e2e/tests/pages/user-menu/preferences.spec.ts
@@ -104,59 +104,6 @@ describe('User can update their preferences', () => {
     }
   });
 
-  // it.skip('[Vue3 Skip]: Can select login landing page', { tags: ['@userMenu', '@adminUser'] }, () => {
-  //   /*
-  //   Select each radio button and verify its highlighted
-  //   Validate http request's payload & response contain correct values per selection
-  //   Verify user is landing on correct page after login
-  //   Verify selection is preserved after logout/login
-  //   */
-
-  //   const landingPageOptions = [
-  //     {
-  //       index: '0', value: '"home"', page: '/home'
-  //     },
-  //     {
-  //       index: '1', value: '"last-visited"', page: 'c/_/manager/provisioning.cattle.io.cluster'
-  //     },
-  //     { // This option only works when there is an existing local cluster
-  //       index: '2', value: '{\"name\":\"c-cluster\",\"params\":{\"cluster\":\"local\"}}', page: '/explore'
-  //     },
-  //   ];
-
-  //   prefPage.goTo();
-  //   prefPage.landingPageRadioBtn().checkVisible();
-  //   landingPageOptions.forEach((key) => {
-  //     cy.intercept('PUT', 'v1/userpreferences/*').as(`prefUpdate${ key.value }`);
-  //     prefPage.landingPageRadioBtn().set(parseInt(key.index));
-  //     cy.wait(`@prefUpdate${ key.value }`).then(({ request, response }) => {
-  //       expect(response?.statusCode).to.eq(200);
-  //       expect(request.body.data).to.have.property('after-login-route', key.value);
-  //       expect(response?.body.data).to.have.property('after-login-route', key.value);
-  //     });
-  //     prefPage.landingPageRadioBtn().isChecked(parseInt(key.index));
-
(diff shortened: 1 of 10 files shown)


cypress/e2e/tests/pages/user-menu/preferences.spec.ts
```diff
@@ -413,45 +360,43 @@ describe('User can update their preferences', () => {
     }
   });
 
-  describe('Check keyboard mapping on YAML Editor', { tags: ['@userMenu', '@adminUser', '@standardUser'] }, () => {
```
codyrancher, line 416: I moved all of this to be in the middle and out of the describe block because adding the login/logout tests for the login landing page was causing problems with the sessioned login attempts and I this was the only way I thought of resolving it without disabling all of the sessioning for these tests.

cypress/support/commands/rancher-api-commands.ts
```diff
@@ -14,15 +14,18 @@ Cypress.Commands.add('login', (
...
   cacheSession = true,
+  skipNavigation = false,
 ) => {
   const login = () => {
     cy.intercept('POST', '/v3-public/localProviders/local*').as('loginReq');
 
-    LoginPagePo.goTo(); // Needs to happen before the page element is created/located
+    if (!skipNavigation) {
+      LoginPagePo.goTo(); // Needs to happen before the page element is created/located
+    }
```
codyrancher, line 24: Maintained the expected default behavior. This was needed for the login landing page tests. Without it we would logout and when attempted to login we'd get redirected to a brand new login page. This meant that we would lose the record of what page the user was on before getting redirected to the login page.

cypress/support/commands/rancher-api-commands.ts
```diff
@@ -14,15 +14,18 @@ Cypress.Commands.add('login', (
...
     const loginPage = new LoginPagePo();
 
     loginPage
-      .checkIsCurrentPage();
+      .checkIsCurrentPage(!skipNavigation);
```
codyrancher, line 28: When we would logout we'd have a query parameter `?logged-out` which would cause the exact match to fail. I added this but maintained the default behavior. Maybe we'd want another argument in the future but for now the only time we want this to occur is if we skip navigation.

shell/components/LandingPagePreference.vue
```diff
@@ -120,7 +120,7 @@ export default {
             :val="false"
             :value="afterLoginRoute=== 'home' || afterLoginRoute === 'last-visited'"
             :v-bind="$attrs"
-            @update:value="afterLoginRoute = false"
+            @update:value="afterLoginRoute = routeFromDropdown.value"
```
codyrancher, line 123: We weren't automatically selecting the first cluster when selecting the cluster option.

shell/components/templates/default.vue
```diff
@@ -115,35 +113,6 @@ export default {
 
   },
 
-  watch: {
```
codyrancher, line 118: We now record all of the setLastVisited using the navigationGuard. This means we only have to record in one location simplifying the logic of the overall feature.

shell/store/prefs.js
```diff
@@ -523,28 +529,3 @@ export const actions = {
     }
   }
 };
-
-function getLoginRoute(route) {
```
codyrancher, line 527: This was erroneously modifying the route name and omitting parameters. This meant only certain pages could be returned to but not all.

shell/utils/auth.js
```diff
@@ -315,6 +315,10 @@ export function isLoggedIn(store, me) {
 export function notLoggedIn(store, redirect, route) {
   store.commit('auth/hasAuth', true);
 
+  if (!route.name.includes('auth')) {
+    store.commit('prefs/setAuthRedirect', route);
+  }
```
codyrancher, line 320: This area of the code is responsible for redirecting the user to the login page if they aren't logged in. The perfect place to get the url that was used before redirecting to login.

momesgin (approved):
Tested and LGTM

PR #13453, opened 2025-02-20
Fleet improve Workspaces delete dialog

### LINKED ISSUE #11501: Better UX around deleting Fleet Workspace and Consequences
SURE-8441

- When deleting a fleet workspace the UI shows a basic 'Are you sure?' style modal
- Deleting a Fleet workspace also deletes clusters within it, which contradics the light, generic modal
- We need to job around handling this

Possible Solution
- We have more specific and better warnings when deleting other resources (maybe deleting a namespace, deleting a role that has bindings, etc). We could expand that concept to the fleet workspace delete modal

torchiaf:
### Summary
Fixes #11501

### Screenshot/Video

image

New

image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -2610,8 +2610,7 @@ fleet:
       restrictions: Allowed Target Namespaces
     timeout: Workspace creation timeout. It's possible the workspace was created. We suggest checking the workspace page before trying to create another.
     remove:
-      warning1: "Deleting a Workspace also deletes all resources in the Workspace. This includes GitRepos but also Clusters."
-      warning2: "Deleted Clusters will be removed from {vendor}, and if {vendor} manages the life-cycle of the Cluster it's underlying infrastructure will also be taken down."
+      warning: "Deleting a Workspace also deletes all resources in the Workspace. This includes GitRepos but also Clusters.<br><br>Deleted Clusters will be removed from {vendor}, and if {vendor} manages the life-cycle of the Cluster it's underlying infrastructure will also be deleted."
   restrictions:
     addTitle: 'allowedTargetNamespaces'
     addLabel: Add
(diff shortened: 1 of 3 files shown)


shell/assets/translations/en-us.yaml
```diff
@@ -2568,6 +2568,16 @@ fleet:
     tabs:
       restrictions: Allowed Target Namespaces
     timeout: Workspace creation timeout. It's possible the workspace was created. We suggest checking the workspace page before trying to create another.
+    remove:
```
richard-cox, line 2571: I'm not quite sure this is clear enough (and one part is wrong).

Priority # 1 - This will delete cluster objects. If rancher manages the cluster lifecycle it will also delete the actual cluster (not just rancher CRs). For example a GKE cluster would be deleted and cannot be re-imported. I think we need to consult our local innovation champion / @manno ... this is correct right? if i were to delete the `fleet-default` namespace and it contained a v1 prov cluster it would be the same as just deleting the v1 prov cluster which would delete the underlying cluster and nodes?

Priority # 2 - The workspace contains a set of fleet resources, like gitrepos, which will all be deleted when the workspace is deleted.

richard-cox (changes requested):

shell/assets/translations/en-us.yaml
torchiaf: @richard-cox I changed the message:

image

We can change the label to include more information, but I don't know what exactly.
cc @manno

torchiaf: image

richard-cox: @kinarashah Would you be able to advise? What happens to provisioned artefacts (vm's, hosted cluster, etc) if a the namespace containing the v1 provisioning cluster is deleted? Would they be removed or left orphaned?

richard-cox: Kinara believes so, and Brad has confirmed. Deleting the namespace a v1 prov cluster is in is the same as deleting each cluster individually (infrastructure is removed).

```
Deleting a Workspace also deletes all resources in the Workspace. This includes GitRepos but also Clusters.

Deleted Clusters will be removed from Rancher, and if Rancher manages the life-cycle of the Cluster it's underlying infrastructure will also be taken down.
```

i think we replace Rancher with a string substitute for white labelling, so it wouldn't be the exact `Rancher` word in the above quote

richard-cox (approved):
I've updated (see last commit message)

image

image

PR #13456, opened 2025-02-20
Delete dialogs, fix list of items message and Vue warnings

### LINKED ISSUE #13455: Delete dialogs - the list of items is incomplete or with wrong syntax
**Setup**
- Rancher version: `head-58aa6e361f`
- Rancher UI Extensions:
- Browser type & version:

**Describe the bug**

- `Projects/Namespaces` remove: If the items are N and N > 5, there are 2 'and' conjunctions at the end of the message:

  Image

-  `Pods` remove: Always shows the first 5 elements and the `plusMore` string is missing

  Image

- Vue3 warnings generated in PromptRemove.vue component

  Image

**To Reproduce**

**Result**

**Expected Result**

**Screenshots**

**Additional context**

torchiaf:
### Summary
Fixes #13455

- Any of Remove dialogs using the default `shell/components/PromptRemove.vue` component.

  - Go to Fleet -> Advanced -> Workspaces
  - Create some new workspaces, even empty
  - Select N workspaces and click on Delete action

- DeactivateDriverDialog: `shell/dialog/DeactivateDriverDialog.vue`

  - Go to Cluster Management -> Drivers -> Node Drivers
  - Select N active drivers
  - Click on Delete action

- Each dialogs defined in `shell/promptRemove`

  - Go to Projects/Namespaces _or_ Users-> RoleTemplate _or_ Workspaces -> Pods
  - Select N items
  - Click on Delete action

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/PromptRemove.vue
+++ b/shell/components/PromptRemove.vue
@@ -47,7 +47,7 @@ export default {
   },
   computed: {
     names() {
-      return this.toRemove.map((obj) => obj.nameDisplay).slice(0, 5);
+      return this.toRemove.map((obj) => obj.nameDisplay);
     },
 
     nameToMatchPosition() {
@@ -91,12 +91,6 @@ export default {
       const first = this.toRemove[0];
 
       return first?.confirmRemove;
-    },
-
-    plusMore() {
-      const remaining = this.toRemove.length - this.names.length;
-
-      return this.t('promptRemove.andOthers', { count: remaining });
     },
 
     // if the current route ends with the ID of the resource being deleted, whatever page this is wont be valid after successful deletion: navigate away
@@ -360,7 +354,7 @@ export default {
         <div class="mb-10">
           <template v-if="!hasCustomRemove">
             {{ t('promptRemove.attemptingToRemove', { type }) }} <span
-              v-clean-html="resourceNames(names, plusMore, t)"
+              v-clean-html="resourceNames(names, t)"
             />
           </template>
 
@@ -369,7 +363,7 @@ export default {
             v-if="hasCustomRemove"
             ref="customPrompt"
             v-model:value="toRemove"
-                        :close="close"
+            :close="close"
             :needs-confirm="needsConfirm"
             :value="toRemove"
             :names="names"
--- a/shell/components/PromptRemove.vue
+++ b/shell/components/PromptRemove.vue
@@ -363,6 +363,7 @@ export default {
             v-if="hasCustomRemove"
             ref="customPrompt"
             v-model:value="toRemove"
+            v-bind="$data"
             :close="close"
             :needs-confirm="needsConfirm"
             :value="toRemove"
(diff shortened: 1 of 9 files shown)


shell/components/PromptRemove.vue
```diff
@@ -368,7 +363,6 @@ export default {
             v-if="hasCustomRemove"
             ref="customPrompt"
             v-model:value="toRemove"
-            v-bind="_data"
```
torchiaf, line 371: I'm not sure where this variable comes from @rak-phillip

rak-phillip: I'm not 100% sure, this was added quite some time back to support a Harvester feature[^1], but it looks like shorthand or a private vue api to access the data property in templates:

- https://github.com/vuejs/core/blob/604d08760ee5b01772bc33cefd89ebdca9bb1b82/packages/runtime-core/src/compat/instance.ts#L166
- https://github.com/vuejs/vue/blob/9e88707940088cb1f4cd7dd210c9168a50dc347c/src/core/instance/state.ts#L63
- https://github.com/vuejs/vue/blob/9e88707940088cb1f4cd7dd210c9168a50dc347c/src/types/component.ts#L95

I advise against removing this line, doing so means that we are no longer binding properties from the data prop to the dynamically rendered component; there is a very high likelihood that this will break something. Either keep this line in, or find an approach to more explicitly bind the properties that we need.

[^1]: https://github.com/rancher/dashboard/commit/d96cf9c27b4e6d8a2d6b90000fbe955f5e36a41a#diff-0b5a35079524529e2ea0c2b7d649785e99121965d0541a9221ed4972c42d0ee7R306

rak-phillip: After taking some time to think about this a little more, I recall encountering this pattern in the past:

https://github.com/rancher/dashboard/commit/79850d131eb688fabc626f864a3d0c8290e55f9d#diff-5cf4eab2e7126559d21cb7359a67e030edc4332eafd9411598f9cdc048afa14cR425

The correct syntax in Vue3 should be `v-bind="$data"`. I suggest that we attempt this approach instead of removing `v-bind` altogether.

torchiaf: Thank you! I added the line with `v-bind="$data"`

shell/dialog/DeactivateDriverDialog.vue
```diff
@@ -31,15 +31,15 @@ export default {
   },
   computed: {
     formattedText() {
-      const namesSliced = this.drivers.map((obj) => obj.nameDisplay).slice(0, 5);
-      const remaining = this.drivers.length - namesSliced.length;
+      const namesSliced = this.drivers.map((obj) => obj.nameDisplay);
```
rak-phillip, line 34: nitpick: `namesSliced` is really a slice any more. I would update this variable name, but it's not really necessary.

rak-phillip (approved):
LGTM - tested each of the scenarios detailed in the attached issue and they are resolved.

Issue #13457, opened 2025-02-20
Refactor auth-config to clearly distinguish requirements for each provider


cnotv:
# Description

Refactor `auth-config` mixin and `auth` store to have separated logic for each providers, as well as containing logic in the same place.
Possibly convert to TS and add further unit tests as done for https://github.com/rancher/dashboard/issues/12477

Rewrite also `BASE_SCOPES` definitions after identifying all the use cases

PR #13514, opened 2025-02-26
Generic OIDC group scope is not sent in the request formed by the UI

### LINKED ISSUE #12477: Generic OIDC group scope is not sent in the request formed by the UI
Internal reference: SURE-9143
Reported in 2.9.2

Issue description:
Despite adding the "groups" scope in the OIDC config, the request that is generated is not including the "groups" scope.

Repro steps:
OIDC configured and checked the generated request.

Workaround:
None

Actual behavior:
The request does not contain the group scope despite being configured.

Expected behavior:
Expect the generated request to contain the group scope.

@mantis-toboggan-md was able to reproduce this issue and confirm it is a UI bug:

> The UI ignores configured scopes and always uses the values configured here https://github.com/rancher/dashboard/blob/master/shell/store/auth.js#L16. Likely, we need to update the redirectTo method in that file to fetch the genericoidc authconfig object and use its scope field to construct the redirect url, instead of that list of defaults.

cnotv:
### Summary
Fixes #12477

Included scopes in the Keycloack pop-up as indicated in the issue.

### Technical notes summary

- Replaced `addObjects()` with new `joinStringList()`, including unit tests
- Added tests to the `redirectTo`  action to generate redirect URL to auth provider pop up with all the involved cases
- Added tests to the `test` action to pass all the required options to compose the redirect URL
- Added tests to the `auth-config` mixin used in the view to submit the auth configuration
- Added tests to the utility function `concatStrings()`

**Side notes as report:**
It seems like `scope` was not added to the whole logic after passing the list from the UI. The function where it gets added has not been changed for 5 years and was not working, so nothing about this feature has been used before.
The scopes seem to be in a format `['a b c']` with a single item in the array, but a refactor has been pushed to a separate issue (#13457) since this is way out of scope.
In the store logic it's also misspelled to the plural as `scopes`, which may leads to confusion while reviewing.

As by reported issue.

- Auth with other non-SAML providers (Azure has E2E)

### Screenshot/Video

Video for this PR is not possible due to internal information, however a new page with the correct scope (as per test) is generated:

Screenshot 2025-02-26 at 10 28 56

### Checklist

- [x] The PR has a Milestone 

### DIFF
--- a/shell/mixins/__tests__/auth-config.test.ts
+++ b/shell/mixins/__tests__/auth-config.test.ts
@@ -1,34 +1,57 @@
 import { mount } from '@vue/test-utils';
 import authConfigMixin from '@shell/mixins/auth-config';
 
-describe('authConfigMixin', () => {
-  describe('on save', () => {
-    it('should not return error', () => {
-      const model = { authConfigName: 'whatever' };
-      const FakeComponent = {
-        render() {},
-        mixins:  [authConfigMixin],
-        methods: { applyHooks: jest.fn() },
-      };
-      const instance = mount(FakeComponent, {
-        data: () => ({
-          value: { configType: 'oidc' },
-          model
-        }),
-        global: {
-          mocks: {
-            $store: { dispatch: { 'auth/test': () => model } },
-            $route: {
-              params: { id: '123' },
-              query:  { mode: 'edit' },
-            },
-          }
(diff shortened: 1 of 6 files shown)


torchiaf (approved):
LGTM

PR #13519, opened 2025-02-26
fix misc issues for a11y

### LINKED ISSUE #12811: a11y: Fix keyboard navigation, labels/descriptions, roles and focus display for miscellaneous single occurrences
a11y issues: 223, 177, 166, 155, 145, 179

**Issue 223 Import Extension Catalog -  Keyboard focus traps**

Keyboard only users are not able to tab pass when the users navigate the page using 'Tab' key and focus gets trapped on combobox which resulted in keyboard-only users not being able to access the interactive elements after the form fields. Additionally, when users try to come out from mentioned combobox modal dialog gets closed.

image

Suggested remediation:

Ensure that the keyboard focus does not get trapped on any component and users can navigate the page content with ease in all the browsers. Modify the scripts associated with the dropdown list to ensure that keyboard trap can be avoided.

**Issue 177 Pod Security Admission: Creates - Role button not defined**

Role is not defined for the interactive image, making it difficult for screen reader users to access the functionality associated with it - issue is with the close button.

image

Suggested remediation:

Ensure that the interactive images is programmatically identified as buttons.
Apply the following changes:
• Code the mentioned interactive image using <button> element. Alternatively, use the role="button" and tabindex="0" attributes to define the button role.
• Modify the scripts to ensure that the image button is operable with keyboard and a mouse.
• Provide textual description such as "Close" for image button using "aria-label" attribute.

**Issue 166 Charts - Tab not programmatically determined**

The mentioned interactive elements are not programmatically determined as tab. Also, the number of tabs are not programmatically determined. On activating the interactive elements, the related content gets dynamically updated below. Although the dynamic change in the content is easier to understand for sighted users, it is not intuitive for screen reader users.

image

Suggested remediation:

Apply the following changes:
• Set the role="tablist" for the <div> tag containing the elements that behaves as a tab.
• Set the role="tab" for the <button> tags used for displaying "Browse" and "Featured". In addition, specify the ARIA attributes for tabs, such as "aria-controls" and "aria-selected". (The value of the aria-selected attribute needs to change on the basis of currently active tab.)
o Set aria-selected="true" for the active tab.
o Specify an "id" attribute for the <div> tags used to display the content of respective tab. Reference the value of the "id" attribute correctly via "aria-controls" attribute of each tab.
o Specify an 'id' attribute for the <button> tags used to display each tab. Reference the value of the "id" attribute correctly via "aria-labelledby" attribute of each tab panel.
• Set the role="tabpanel" for the <div> tags displaying the contents associated with "Browse" and "Featured" tabs.
o As best practice, provide tabindex="0" attribute on the container of tab panel.
• Ensure that the keyboard focus cycles between the tabs using arrow keys. The not selected tab should not receive the keyboard focus when navigated using "Tab" key.

More information can be found at https://www.w3.org/WAI/ARIA/apg/example-index/tabs/tabs-manual.html

**Issue 155 Delete Project - Missing form label or 'title' attribute**

The form field do not have the label associated with them resulting in screen reader users fail to understand the purpose of the form field.

image

Suggested remediation:

Apply the following changes:
• Replace the <div> tag containing the "Enter default below to confirm" text using <label> tag.
• Use "for" and "id" attributes to associate label with their respective input field. Ensure that the value of "for" attribute of the label exactly matches the value of "id" attribute of the input field.

**Issue 145 Kubectl Shell - Role and state not defined for interactive images**

The role and state are not programmatically defined for the interactive image making it difficult for screen reader users to access the state information and functionality associated with interactive image.

image

Suggested remediation:

Apply the following changes:
• Code the mentioned interactive elements using <button> tag. Alternatively, use role="button" and tabindex="0" attributes to define the button role.
• Modify the scripts to ensure that the functionality is accessible for all users.
• Set "aria-expanded" attribute to "false" for the image buttons when the buttons are in the collapsed state. Ensure that the value of "aria-expanded" attribute changes on user interaction via scripting.
• Provide textual description as "Resize" for the image button using "title" or "aria-label" attribute.

**Issue 179 Pod Security Admission: Create - Missing form label or 'title' attribute**

The form fields do not have the label associated with them and "title" attribute is not included in their source code either.
As a result, screen reader users will not understand the purpose of the form fields.

Pod Security Admission: Create
- Version (default: latest)
- Enter a comma separated list of usernames 
- Enter a comma separated list of runtimeClasses
- Enter a comma separated list of namespaces

image

Suggested remediation:

Apply the following changes:
• Provide unique and descriptive visual labels such as "Enforce - Version (default: latest)", "Audit - Version (default: latest)", "Usernames - Enter a comma separated list of usernames", "Namespaces - Enter a comma separated list of namespaces" and so on for the mentioned form fields using <label> tag.
• Use "for" and "id" attributes to associate labels with their respective form fields. Ensure that the value of "for" attribute of the label exactly matches the value of "id" attribute of the form field.

aalves08:
### Summary
Fixes #12811 

- Fixes a11y issue `223`, focus trap on `LabeledSelect` inside `Import Extension Catalog` modal by refactoring the mechanism to prevent `escape` key handler on modals based on other UI elements like `CodeMirror` and `LabeledSelect`. Architecture is now only based on preventing propagation of events 🪄 

Should be tested also the `Import YAML` modal since it features both `LabeledSelect` and `CodeMirror` as UI elements that have the need to prevent that `escape` key handler on `AppModal`

- a11y issue `177` fixed on a previous PR
- Fixes a11y issue `166` by improving aria-label's for the `ButtonGroup`
- Fixes a11y issue `155` and `179`  by allowing `LabeledInput` to pass an `aria-label` as a prop of `LabeledInput`. This covers the cases where `LabeledInput` is used without a label
- Fixes a11y issue `145` by adding the necessary a11y attributes and allowing keyboard interaction with arrow keys to resize the shell console window

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.vue
+++ b/pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.vue
@@ -111,7 +111,7 @@ export default defineComponent({
      * Optionally use this to comply with a11y IF there's no label
      * associated with the input
      */
-    accessibilityLabel: {
+    ariaLabel: {
       type:    String,
       default: ''
     }
@@ -373,7 +373,7 @@ export default defineComponent({
         :id="inputId"
         ref="value"
         v-bind="$attrs"
-        v-stripped-aria-label="!hasLabel && accessibilityLabel ? accessibilityLabel : undefined"
+        v-stripped-aria-label="!hasLabel && ariaLabel ? ariaLabel : undefined"
         :maxlength="_maxlength"
         :disabled="isDisabled"
         :value="value || ''"
@@ -388,7 +388,7 @@ export default defineComponent({
         v-else
         :id="inputId"
         ref="value"
-        v-stripped-aria-label="!hasLabel && accessibilityLabel ? accessibilityLabel : undefined"
+        v-stripped-aria-label="!hasLabel && ariaLabel ? ariaLabel : undefined"
         role="textbox"
         :class="{ 'no-label': !hasLabel }"
         v-bind="$attrs"
--- a/pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.vue
+++ b/pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.vue
@@ -105,6 +105,15 @@ export default defineComponent({
     class: {
       type:    String,
       default: ''
+    },
+
+    /**
+     * Optionally use this to comply with a11y IF there's no label
+     * associated with the input
+     */
+    accessibilityLabel: {
+      type:    String,
+      default: ''
     }
   },
 
@@ -364,6 +373,7 @@ export default defineComponent({
         :id="inputId"
         ref="value"
         v-bind="$attrs"
+        v-stripped-aria-label="!hasLabel && accessibilityLabel ? accessibilityLabel : undefined"
         :maxlength="_maxlength"
         :disabled="isDisabled"
         :value="value || ''"
@@ -378,6 +388,7 @@ export default defineComponent({
         v-else
         :id="inputId"
         ref="value"
+        v-stripped-aria-label="!hasLabel && accessibilityLabel ? accessibilityLabel : undefined"
         role="textbox"
         :class="{ 'no-label': !hasLabel }"
         v-bind="$attrs"
--- a/shell/components/CodeMirror.vue
+++ b/shell/components/CodeMirror.vue
@@ -123,12 +123,16 @@ export default {
   },
 
   async mounted() {
-    document.addEventListener('keyup', this.handleKeyPress);
+    const el = this.$refs.codeMirrorContainer;
+
+    el.addEventListener('keydown', this.handleKeyPress);
     this.codeMirrorContainerRef = this.$refs.codeMirrorContainer;
   },
 
   beforeUnmount() {
-    document.removeEventListener('keyup', this.handleKeyPress);
+    const el = this.$refs.codeMirrorContainer;
+
+    el.removeEventListener('keydown', this.handleKeyPress);
   },
 
   watch: {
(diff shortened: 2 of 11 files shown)


pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.vue
```diff
@@ -105,6 +105,15 @@ export default defineComponent({
...
       default: ''
+    },
+
+    /**
+     * Optionally use this to comply with a11y IF there's no label
+     * associated with the input
+     */
+    accessibilityLabel: {
+      type:    String,
+      default: ''
```
rak-phillip, line 116: nitpick: I'm not much of a fan of `accessibilityLabel` as a prop. I think that we can just name this `aria-label` because I believe that we would want to control where one is applied if a developer tries to supply one to the component.

shell/components/CodeMirror.vue
```diff
@@ -123,12 +123,16 @@ export default {
   },
 
   async mounted() {
-    document.addEventListener('keyup', this.handleKeyPress);
```
rak-phillip, line 126: Thanks for updating this 🙌. Removing listeners on the document will help prevent conflicts.

shell/components/nav/WindowManager/index.vue
```diff
@@ -293,6 +293,30 @@ export default {
...
+    },
+
+    resizeVertical(arrowUp) {
+      const resizeStep = 20;
+      const height = arrowUp ? this.height + resizeStep : this.height - resizeStep;
+
+      this.height = height;
+
+      this.setWmDimensions(height);
+    },
```
rak-phillip, line 305: It seems that I'm able to produce a scenario where the active terminal area gets clipped by dead space: 

image

I'm not able to produce this scenario in `master`

rak-phillip (changes requested):
I'm noting some odd behavior when resizing the terminal using keyboard controls.

shell/components/nav/WindowManager/index.vue
aalves08: Should be fixed mate

rak-phillip (commented):
I still see some visual glitches, but it's much more improved with this iteration:

https://github.com/user-attachments/assets/0f5f7684-4acd-4636-93b4-73e3f1d4c70d

image

rak-phillip (approved):
LGTM - Everything else in this PR works well and the visual inconsistency when resizing the shell is quite minor. I think that this can be addressed separately.

edit: I can force similar behavior in `master` when resizing the shell with the mouse. I'm thinking that this behavior is unrelated to the new functionality, it's just more prevalent on inspection.

PR #13609, opened 2025-03-05
working on issues reported in the a11y tests

### LINKED ISSUE #13833: a11y: Landmark role defined incorrectly
**a11y issue 46**

**Setup**
- Rancher version:
- Rancher UI Extensions:
- Browser type & version:

**Describe the bug**

Landmark role defined incorrectly

The `<main>` element is incorrectly defined twice resulting in a confusing experience for screen reader users. All the page regions should be appropriately defined.

**To Reproduce**

**Result**

**Expected Result**

All the page regions should be appropriately defined.

**Screenshots**

**Additional context**

aalves08:
### Summary
Fixes #13833

- update layouts to only render 1 `<main>` element and contain correct labels
- Fixes mostly the above mentioned issues from the accessibility report

Screenshot 2025-03-20 at 14 54 06

### Checklist
- [x] The PR has a Milestone 


shell/components/templates/blank.vue
```diff
@@ -9,7 +9,10 @@ export default {
 </script>
 
 <template>
-  <main class="main-layout">
+  <main
+    class="main-layout"
+    :aria-label="t('layouts.blank')"
```
rak-phillip, line 14: What is the benefit to adding an aria label to our `main` tags to describe the layout?

rak-phillip (commented):
Most of this looks good, I'm just unsure about adding aria labels to describe the layout to main. Why is is this required? Is there something more descriptive or beneficial that we can provide?

shell/components/templates/blank.vue
aalves08: Apparently it's a best practice, as per:

Screenshot 2025-03-27 at 11 52 42

https://github.com/dequelabs/axe-core/blob/develop/doc/rule-descriptions.md (search for `landmark-unique`)
https://dequeuniversity.com/rules/axe/4.10/landmark-unique?application=RuleDescription

rak-phillip: This seems to apply to multiple landmarks being present on the page[^1]. Is it the case that we have multiple `main` tags rendered?

[^1]: https://www.w3.org/WAI/ARIA/apg/patterns/landmarks/examples/main.html

aalves08: Given your last link, I think you might be right. Nevertheless, the accessibility report clearly states two different points:
`landmark-no-duplicate-main`, which was the case and the PR addresses that (There's no more than 1 `main` rendered)
`landmark-unique`, which is the adding of the `aria-label`

rak-phillip: I think that we can keep them. I think that you're reasoning to adding them is more than sufficient and addition of the labels can provide more context for screen reader users.

rak-phillip (approved):
LGTM

PR #13612, opened 2025-03-05
Use `Shift+Escape` for terminal and editor escape sequence 

### LINKED ISSUE #13601: ESCape to blur on terminal window causes issue with vi/kubectl edit
When using the the terminal window in a cluster, we now capture ESCape to blur focus.

When using the terminal, however, if editing a resource via `kubectl edit`, then `vi` is used and the escape key is important to the editing experience there - so using escape for blur causes issues.

**To reproduce:**

From the local cluster, open the kube terminal.
Type 'kubectl edit setting ui-pl`
Press `I` to enter insert mode
Press `ESCAPE` to exit insert mode 
The terminal window loses focus.

**Expected:**

Terminal window should not lose focus when escape is pressed.

**Suggestion:**

We require SHIFT+Escape in this case.

This is a slightly different to the code editor, but in this case, this is a terminal.

We can make this behaviour clear in the help text we show.

Maybe there's a better or more standard key combination for this case.

rak-phillip:
### Summary
The escape key conflicts with modal editing when working in VIM mode. Using `Shift+Escape` helps to prevent that conflict.

Fixes #13601 

- Use `Shift+Escape` to exit code mirror and shell focus trap

### Technical notes summary

I think there's an argument that can be made for making this key combo a configurable preference in the future.

- Code Mirror
  - Default Keybinds
  - Vim Keybinds
  - Emacs Keybinds
- Shell

- Code Mirror
- Shell

### Screenshot/Video

image

image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -2343,7 +2343,7 @@ resource:
     update: "Error updating {name}"
 
 codeMirror:
-  escapeText: Press Escape key to blur from editor
+  escapeText: Press Shift+Escape to blur from editor
   keymap:
     tooltip: Key mapping preference.
     indicatorToolip: "Key mapping: {name}"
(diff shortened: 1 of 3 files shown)


aalves08 (approved):
Gave it a run and works as advertised. Code changes are simple and easy to follow.

Thanks for replacing it on both kubectl shell and codemirror.

LGTM mate!

PR #13618, opened 2025-03-06
Use CSS flexbox for LDAP schema columns

### LINKED ISSUE #13616: LDAP Schema columns are out of order
There are two fields labeled "Search Attribute" in the Groups section. One of these fields should actually be in the Users section, and one should remain in the Groups section.

## Acceptance Criteria

- [ ] All user-related fields are under the "Users" column
- [ ] All group-related fields are under the "Groups" column

## Screenshots

Image

## Details

This regression was introduced in #13464

rak-phillip:
### Summary
This restores all form inputs for the Customize Schema section for the LDAP configuration form to their appropriate columns by refactoring this section to use CSS flexbox.

Fixes #13616

- Replace custom rows and columns for the Customize Schema section of the LDAP form with a CSS flexbox implementation

### Technical notes summary

🗒️ NOTE: if this refactor it too risky, we can revert to the previous implementation and manually shuffle around inputs in the rows/columns. I took on this effort because it was rather straightforward to implement and it improves the developer experience for managing these inputs moving forward.

PR #13464 introduced a change that regressed the layout of the Customize Schema section in the LDAP forms. I think that relying on CSS flexbox to layout the columns makes it much more clear that fields in these columns need to be grouped together. It wasn't clear that the first item in each row needed to be user-related and the second item needed to be group-related in the previous implementation.

This change will also reduce the need to manually "reflow" form elements to remove gaps in the columns when new inputs are added in the future.

We also get the benefit of updating the layout if the viewport width becomes too small to comfortably render the inputs side-by-side. 

- LDAP Auth Providers

- LDAP Auth Providers

#### Before (2.10)

image

image

#### After

image

image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/edit/auth/ldap/config.vue
+++ b/shell/edit/auth/ldap/config.vue
@@ -279,157 +279,125 @@ export default {
       color="info"
       label-key="authConfig.ldap.oktaSchema"
     />
-    <div class="row">
-      <div class="col span-6">
+    <div class="schema-container">
+      <div class="schema-column">
         <h4>{{ t('authConfig.ldap.users') }}</h4>
-      </div>
-      <div class="col span-6">
-        <h4>{{ t('authConfig.ldap.groups') }}</h4>
-      </div>
-    </div>
-    <div class="row mb-20">
-      <div class="col span-6">
         <LabeledInput
           v-model:value="model.userObjectClass"
           :mode="mode"
           :label="t('authConfig.ldap.objectClass')"
         />
-      </div>
-      <div class="col span-6">
-        <LabeledInput
-          v-model:value="model.groupObjectClass"
-          :mode="mode"
-          :label="t('authConfig.ldap.objectClass')"
-        />
-      </div>
-    </div>
-    <div class="row mb-20">
-      <div class="col span-6">
         <LabeledInput
           v-model:value="model.userNameAttribute"
           :mode="mode"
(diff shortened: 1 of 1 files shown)


shell/edit/auth/ldap/config.vue
```diff
@@ -279,157 +279,125 @@ export default {
...
+
+  .schema-column {
+    flex: 1;
+    display: flex;
+    flex-direction: column;
+    min-width: 16rem;
+
+    > :not(:first-child) {
+      margin-bottom: 20px;
+    }
```
rak-phillip, line 401: I'm using `margin-bottom` instead of gap so that I can preserve the spacing for the Users and Groups column headers. I'm open to alternative approaches.

shell/edit/auth/ldap/config.vue
```diff
@@ -279,157 +279,125 @@ export default {
...
+        </template>
       </div>
     </div>
   </div>
 </template>
+
+<style lang="scss" scoped>
+  .schema-container {
+    display: flex;
+    gap: 1.75%;
```
rak-phillip, line 389: `1.75%` isn't my preferred value, but it matches the spacing in the rest of the form.

codyrancher (approved):
I know we have quite a few places with the col span-6 pattern. It might be useful to extract what you did here into global styles or maybe components.

PR #13780, opened 2025-03-17
fix bug with Unit Input component not updating correctly

### LINKED ISSUE #13784: Bug with Unit Input component not updating correctly
**Setup**
- Rancher version:
- Rancher UI Extensions:
- Browser type & version:

**Describe the bug**
 
When using the arrow buttons on the component, the value doesn't update correctly/throws an error

**To Reproduce**

Cluster - Services - Create - Cluster IP - Session Affinity - ClientIP - then try to interact with the "arrow" buttons to increment/decrement the value

**Result**
value disappears. Warning is thrown in the console

**Expected Result**

value should update correctly

**Screenshots**

https://github.com/user-attachments/assets/88f8865e-b74e-4929-bd77-38430d5e0d00

**Additional context**

aalves08:
### Summary
Fixes #13784

- fix `UnitInput` component by reacting to `update:change`
- adjust unit tests
- bump shell version so that we can release a new version for a partner extension with this fix

### Screenshot/Video

**Before**

https://github.com/user-attachments/assets/699ce72a-9c5c-4b37-94de-61a0974ba566

**After**

https://github.com/user-attachments/assets/5169c2a2-17fb-4f58-b13e-3df5f8e80cbb

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/form/UnitInput.vue
+++ b/shell/components/form/UnitInput.vue
@@ -199,7 +199,6 @@ export default {
     },
 
     update(inputValue) {
-      console.warn(`UPDATE VAL ON UNIT INPUT raw`, inputValue);
       let out = inputValue === '' ? null : inputValue;
 
       if (this.positive && inputValue < 0) {
@@ -207,7 +206,7 @@ export default {
       }
 
       if (this.outputModifier) {
-        out = out === null ? null : `${ inputValue }${ this.unit }`;
+        out = out === null ? null : `${ parseInt(inputValue) }${ this.unit }`;
       } else if ( this.outputAs === 'string' ) {
         out = out === null ? '' : `${ inputValue }`;
       } else if (out) {
--- a/shell/components/form/UnitInput.vue
+++ b/shell/components/form/UnitInput.vue
@@ -199,6 +199,7 @@ export default {
     },
 
     update(inputValue) {
+      console.warn(`UPDATE VAL ON UNIT INPUT raw`, inputValue);
       let out = inputValue === '' ? null : inputValue;
 
       if (this.positive && inputValue < 0) {
@@ -235,7 +236,7 @@ export default {
     :required="required"
     :placeholder="placeholder"
     :hide-arrows="hideArrows"
-    @change="update($event.target.value)"
+    @update:value="update"
     @blur="update($event.target.value)"
   >
     <template #suffix>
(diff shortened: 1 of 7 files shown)


shell/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@rancher/shell",
-  "version": "3.0.2-rc.6",
+  "version": "3.0.2",
```
cnotv, line 3: I am not so sure about bumping a version with a fix, but I do not know what's the normal process for this either.

shell/components/form/UnitInput.vue
```diff
@@ -235,7 +235,7 @@ export default {
     :required="required"
     :placeholder="placeholder"
     :hide-arrows="hideArrows"
-    @change="update($event.target.value)"
+    @update:value="update"
```
cnotv, line 238: Have you tested this component with pages where it's consuming the emitter `update`?  In this modified version it seems to pass directly the whole event instead of the value, no?
Also is not changing the emitter from `update` to `change` affecting other components/views?

cnotv (commented):
Sorry if I do not test but I'm having issues setting up my instance

shell/package.json
aalves08: It's needed for a partner extension, as is described on the PR

shell/components/form/UnitInput.vue
aalves08: Imo, the problem lies with how `UnitInput` gets the value from it's child component `LabeledInput`. From my tests, what is passed upstream is the value that `UnitInput` parses. It can be a simple integer (in `number` or `string` format) or a string with units. I think that part looks good, tbh.
From searching the code there are 68 instances of `UnitInput` being used... I haven't checked them all, but we do have tests. I haven't found any problems with a few that I looked on but let's see what the tests pick up

cnotv: Yes I also double-checked in the `LabeledInput` and there's a check on that and it's just inconsistent how values are emitted, but the emission seems safe from your side.

```ts
      const value = typeof val === 'string' ? val : (val?.target as HTMLInputElement)?.value;
```

cnotv (dismissed):
LGTM

codyrancher (approved):

aalves08:
@codyrancher @cnotv @mantis-toboggan-md I went through all of the `UnitInput` instances on dashboard and found a few "broken" on the `Job` edit view. 

Last review 🙏

cnotv (approved):
LGTM

PR #13789, opened 2025-03-17
Validation support for RKE2 machine pool configuration

### LINKED ISSUE #13333: No validation for machine pool count on firefox or safari
Internal Reference: SURE-9757

Found by @jakefhyde in Rancher 2.10.1

Issue description:
The machine pool count form accepts non-numeric data when using firefox & safari.

Business impact:
Users can attempt to create clusters with negative machine pool counts, which will allow the cluster to be created but the machine pools to fail to provision. Users will also be able to click create for a count of "a", which causes the cluster to fail to be created.

Repro steps:
- Type anything into the count field

Workaround:
None

Actual behavior:
No validation occurs

Expected behavior:
Users are restricted from entering invalid data.

Image

momesgin:
### Summary
Fixes #13333 , #8760 

Added support for validation in machine pool configuration. For #8760 some items were already fixed, this PR completes fixing all the items mentioned there.

### Technical notes summary

- Added two validators, `poolName` for the name field ,and `poolQuantity` for the machine count field
- Validation error icon will be displayed, also the tooltips when the icon is hovered
- Validation errors will disable Create/Save button
- Pool name field won't accept empty value
- Machine count field won't accept negative values
- Machine count field won't accept empty or string values(string values could get entered in Safari and Firefox)
- Unit tests

### Screenshot/Video

https://github.com/user-attachments/assets/0a8c0d28-5024-4c99-9802-3963f680e2ee

### Checklist
- [x] The PR has a Milestone 


shell/edit/provisioning.cattle.io.cluster/tabs/MachinePool.vue
```diff
@@ -204,6 +228,26 @@ export default {
...
+    },
+
+    handlePoolName(val) {
+      this.updateValidation(MACHINE_POOL_VALIDATION.FIELDS.NAME, val === '');
+      this.$emit('valueChanged', 'pool.name', val);
+    },
+
+    handlePoolQuantity(val) {
+      this.updateValidation(MACHINE_POOL_VALIDATION.FIELDS.QUANTITY, val === '' || val < 0);
+      this.$emit('valueChanged', 'pool.quantity', val);
```
mantis-toboggan-md, line 250: I think that you may avoid these methods altogether by tweaking how you've implemented the form validation mixin in the component. Right now the mixin is only being used to compute and display errors in inputs; ideally, the same validator function used on the inputs would be used when determining whether or not the form overall (or in this case machine pool overall) is valid. This has the added benefit of being an easier pattern to maintain when validator functions themselves are complicated and subject to change.

form validation rulesets' `path` property is used to tell the mixin what on the component's `value` property to validate - eg if you set path to `pool.name` the mixin will run the rules you have defined against `value.pool.name`.


```
const FIELDS = {
  NAME:     'pool.name',
  QUANTITY: 'pool.quantity'
};
```

You should be able to simply watch the bool `fvFormIsValid` instead of your custom tracker, `validationErrors`, and drop the 3 methods I highlighted in this comment.

```
    fvFormIsValid: {
      handler(newValue) {
        this.$emit('validationChanged', newValue);
      },
      deep: true
    }
```

shell/utils/validators/machine-pool.ts
```diff
@@ -0,0 +1,32 @@
...
+];
+
+export const MACHINE_POOL_VALIDATION = {
+  FIELDS,
+  RULESETS
+};
+
+export const poolName = (ctx:any) => {
+  return (name: string) => {
+    return !!name ? undefined : ctx.t('validation.required', { key: 'Value' });
```
mantis-toboggan-md, line 24: Can this use the pool name label instead of "Value"? Same suggestion for the quantity validator

mantis-toboggan-md (changes requested):
I think this generally works but I have some suggestions to streamline the implementation of form validatio and better leverage existing functionality in the mixin. Let me know if you have questions about what I've written.

I also noticed the node role chips are not validating negative node counts and sort of contradict the input-level validation; should I file a separate issue for this?

Screenshot 2025-03-25 at 1 15 09 PM

shell/edit/provisioning.cattle.io.cluster/tabs/MachinePool.vue
momesgin: Thanks for your great feedback! I'm sure I've already tried `fvFormIsValid` but I might've forgotten about adding `deep: true`, will test it again and implement your suggestions. Regarding the chips, I think separating it from the validation would be better, I'll create the issue for it.

mantis-toboggan-md (approved):
LGTM

PR #13838, opened 2025-03-20
fix grouping of RadioGroup with meaningful title

### LINKED ISSUE #13837: a11y: Missing grouping of form controls. - RadioGroup
**Setup**
- Rancher version:
- Rancher UI Extensions:
- Browser type & version:

**Describe the bug**

**a11y issue 77**

The radio buttons are not programmatically grouped and associated with the instruction text resulting in screen reader users finding it difficult to understand the purpose of the radio buttons.

Main Content:
Login landing page
- Take me to the home page
- Take me to the area I last visited
- Take me to cluster
radio buttons 

Include the instruction "You can change where you land when you login" within `<legend>` element. Use `<fieldset>` and `<legend>` elements to group the instruction and radio buttons.

**To Reproduce**

**Result**

**Expected Result**

The screen reader should announce the instruction along with the label for the radio buttons.

**Screenshots**

**Additional context**

aalves08:
### Summary
Fixes #13837

- fix grouping of `RadioGroup` with meaningful title/description

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/LandingPagePreference.vue
+++ b/shell/components/LandingPagePreference.vue
@@ -102,12 +102,7 @@ export default {
 </script>
 
 <template>
-  <fieldset>
-    <legend>
-      <p class="set-landing-leadin">
-        {{ t('landing.landingPrefs.body') }}
-      </p>
-    </legend>
+  <div>
     <RadioGroup
       id="login-route"
       :value="afterLoginRoute"
@@ -115,6 +110,11 @@ export default {
       :options="routeRadioOptions"
       @update:value="updateLoginRoute"
     >
+      <template #label>
+        <p class="set-landing-leadin">
+          {{ t('landing.landingPrefs.body') }}
+        </p>
+      </template>
       <template #2="{option}">
         <div class="custom-page">
           <RadioButton
@@ -137,7 +137,7 @@ export default {
         </div>
       </template>
     </RadioGroup>
-  </fieldset>
+  </div>
 </template>
 
 <style lang="scss" scoped>
--- a/shell/components/LandingPagePreference.vue
+++ b/shell/components/LandingPagePreference.vue
@@ -102,10 +102,12 @@ export default {
 </script>
 
 <template>
-  <div>
-    <p class="set-landing-leadin">
-      {{ t('landing.landingPrefs.body') }}
-    </p>
+  <fieldset>
+    <legend>
+      <p class="set-landing-leadin">
+        {{ t('landing.landingPrefs.body') }}
+      </p>
+    </legend>
     <RadioGroup
       id="login-route"
       :value="afterLoginRoute"
@@ -135,7 +137,7 @@ export default {
         </div>
       </template>
     </RadioGroup>
-  </div>
+  </fieldset>
 </template>
 
 <style lang="scss" scoped>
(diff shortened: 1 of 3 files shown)


shell/components/LandingPagePreference.vue
```diff
@@ -102,10 +102,12 @@ export default {
 </script>
 
 <template>
-  <div>
-    <p class="set-landing-leadin">
-      {{ t('landing.landingPrefs.body') }}
-    </p>
+  <fieldset>
```
rak-phillip, line 105: It looks like this change indents radio groups a little bit:

## Before

image

## After

image

shell/components/LandingPagePreference.vue
```diff
@@ -102,10 +102,12 @@ export default {
...
+    <legend>
+      <p class="set-landing-leadin">
+        {{ t('landing.landingPrefs.body') }}
+      </p>
+    </legend>
```
rak-phillip, line 110: Would it make sense to supply this legend to the label slot for RadioGroup? It seems like the design of our component should be able to handle this.

aalves08: The following screenshot shows a larger, repeated title, which is me adding the same title to the `label` prop of the component... Sizing is totally off in regards to the remaining title of the page

Screenshot 2025-03-27 at 12 17 15

I can only see 2 options:
- keep the hacky `fieldset` just for the `LandingPagePreference` component, which isn't the prettiest thing to do codewise but at least we keep the same styling
- create a prop-activated title styling just for this usecase in the `RadioGroup` component and pass the responsibility to the component, despite introducing an "edge-case" to the component code

Wdyt? 🤔

rak-phillip: I gave it a try with the following template

```html
<template>
  <RadioGroup
    id="login-route"
    :value="afterLoginRoute"
    name="login-route"
    :options="routeRadioOptions"
    @update:value="updateLoginRoute"
  >
    <template #label>
      <legend>
        <p class="set-landing-leadin">
          {{ t('landing.landingPrefs.body') }}
        </p>
      </legend>
    </template>
    <template #2="{option}">
      <div class="custom-page">
        <RadioButton
          :label="option.label"
          :val="false"
          :value="afterLoginRoute=== 'home' || afterLoginRoute === 'last-visited'"
          :v-bind="$attrs"
          :prevent-focus-on-radio-groups="true"
          @update:value="updateLoginRoute(null)"
        />
        <Select
          v-model:value="routeFromDropdown"
          :aria-label="t('landing.landingPrefs.ariaLabelTakeMeToCluster')"
          :searchable="true"
          :disabled="afterLoginRoute === 'home' || afterLoginRoute === 'last-visited'"
          :clearable="false"
          :options="routeDropdownOptions"
          class="custom-page-options"
        />
      </div>
    </template>
  </RadioGroup>
</template>
```

Which renders the following

image

rak-phillip: It's not as nice as I would like.. perhaps we can fix it if we ever have the opportunity to refactor the RadioGroup component

aalves08: Mate, your solution is much nicer.... I didn't pay any attention to the `label` slot on the `RadioGroup`. I've just tweaked your solution slightly so that we can meet the requirements. 

Give it another review whenever you have the chance 🙏  Thanks!

rak-phillip (approved):
LGTM

PR #13865, opened 2025-03-24
Auth Provider create/edit form can be saved with some mandatory fields in blank

### LINKED ISSUE #13771: Auth Provider create/edit form can be saved with some mandatory fields in blank
**Setup**
- Rancher version: 

Rancher | v2.11-94853c35bd1e02cf8aa114be4016bc3699288c6a-head
Dashboard | master https://github.com/rancher/dashboard/commit/9b894b44ecf1e2ff0fb479052253e3c35ffa0c30
Rancher CLI | v2.11.0-rc.1

**Describe the bug**
User can save the edit config form of OIDC Auth Provider with the mandatory fields Auth Endpoint and Scopes in blank

**To Reproduce**
1- From the burger menu bar, go to Users & Authentication then Auth Provider
2- Configure an OIDC Auth Provider - in this example, Keyclock(for further details, ping me internally) via the rancher dashboard UI
3- ~After complete setup, via UI, try to edit the Auth Provider config by filling the Client Secret, but removing all the Scopes and removing the Auth Endpoint~ Verified to happen also on creation, see new video attachment
4- Hit Save

**Result**
The Save button is clickable in the UI, allowing the user to save the config of Auth Providers missing these mandatory fields.
Also, when the Auth Endpoint is missing, after the Error pop-up is closed, the Save Button stays locked with a progress bar and status Saving even after we click in the Auth Provider and quit the edit form.

**Expected Result**
The Save button should be locked in the UI, if any mandatory fields are blank.

**Screenshots**

https://github.com/user-attachments/assets/5128bcc9-ca27-4749-a483-0d69483ec539

Image

Image

cnotv:
### Summary
Fixes #13771

- Updated validation to prevent the described scenario
  - Auth endpoint it's now required from the form
  - Basic scopes are now required from the view, as they are equally added programmatically in the store
- Updated tooltip for the scopes

### Technical notes summary

- Updated KeyValue Story while testing around a similar requirement
- Grouped existing tests
- Added test to ensure the save button is disabled (check the title, too much movement due grouping tests):
  - `given missing required basic scopes`
  - `given missing Auth endpoint URL`

### Screenshot/Video

https://github.com/user-attachments/assets/16229b9d-ca49-41eb-89b2-0746c18473a0

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -658,7 +658,7 @@ authConfig:
     scope:
       label: Scopes
       placeholder: openid
-      protip: The <code>openid</code> scope is required. If you are using custom scopes, ensure that it is included.
+      protip: The <code>openid</code>, <code>profile</code>, and <code>email</code> scopes are required and cannot be removed.
     cert:
       label: Certificate
       placeholder: Paste in the certificate, starting with -----BEGIN CERTIFICATE-----
--- a/shell/edit/auth/__tests__/oidc.test.ts
+++ b/shell/edit/auth/__tests__/oidc.test.ts
@@ -32,151 +32,178 @@ const mockModel = {
 };
 
 describe('oidc.vue', () => {
-  let wrapper: VueWrapper<any, any>;
-  const requiredSetup = () => ({
-    data() {
-      return {
-        isEnabling:     false,
-        editConfig:     false,
-        model:          { ...mockModel },
-        serverSetting:  null,
-        errors:         [],
-        originalModel:  null,
-        principals:     [],
-        authConfigName: 'oidc',
-      } as any; // any is necessary as in pre-existing tests we are including inherited mixins values
-    },
-    global: {
-      mocks: {
-        $fetchState: { pending: false },
-        $store:      {
-          getters: {
-            currentStore:              () => 'current_store',
-            'current_store/schemaFor': jest.fn(),
-            'current_store/all':       jest.fn(),
-            'i18n/t':                  (val: string) => val,
-            'i18n/exists':             jest.fn(),
-          },
-          dispatch: jest.fn()
-        },
-        $route:  { query: { AS: '' }, params: { id: 'oicd' } },
-        $router: { applyQuery: jest.fn() },
+  describe('given default valid values', () => {
+    let wrapper: VueWrapper<any, any>;
+    const requiredSetup = () => ({
+      data() {
+        return {
+          isEnabling:     false,
+          editConfig:     false,
+          model:          { ...mockModel },
+          serverSetting:  null,
+          errors:         [],
+          originalModel:  null,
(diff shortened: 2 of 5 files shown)


mosessond:
I see that this fixes the issue where the UI is misleading the user by making it seem that those scopes are not added, not added when they are silently added, but why not go the other way and just not add these extra scopes? My IDP is only configured with the openid scope, and this means that rancher can never integrate with it. 

Is there a technical reason all these scopes are needed?

cnotv:
> I see that this fixes the issue where the UI is misleading the user by making it seem that those scopes are not added, not added when they are silently added, but why not go the other way and just not add these extra scopes? My IDP is only configured with the openid scope, and this means that rancher can never integrate with it.
> 
> Is there a technical reason all these scopes are needed?

In the last 5 years these scopes have always been added there and it seems there are no reports of needing the opposite. I am making this just visible, as I expected it to be something required, but it seems now this is not based on what you said.
I'll discuss this internally to review the actual requirements.

mosessond:
I don't think that the generic OIDC provider has been available for that long, or at least we didn't know about it (though admittedly, we only recently moved to rancher 2.9, with an eye towards 2.11 now that we need to do a slew of upgrades for our clusters on both the k8s and rancher sides). Just to offer our perspective:

* Since our IdP supports mTLS authentication, ideally we don't have to provide a client secret at all, and the only fields needed are:
   * client ID
   * rancher url
   * issuer
   * auth endpoint
   * Private key
   * certificate (would include both the public key for the mTLS auth cert as well as the trust chain for the IdP)
* Alternatively, we could work with a client secret, certificate combination, where the certificate is just the trust chain for the IdP
* Lastly, and this is sometimes tricky, but we could also cases where the service mesh is adding TLS origination, and in that case we would not need the IdP certificate, we would set the connection to HTTP, and the mesh would do TLS origination and verify the token by checking it explicitly.

Historically we have run our own Keycloak instance with a plugin that federates to our IdP, but that process is painful, and we would like to switch to just using generic OIDC if possible. I am also will to help with PRs if desired, but would need a rough pointer to where in the code to look

rak-phillip:
@mosessond thanks for your feedback, improving the areas around generic OIDC will definitely help to make Rancher better. 

Can you open up a new issue that details everything mentioned above? This will help us analyze the requirements and consider any potential cross-boundary changes. It will also serve as a starting point for any contributions that you may wish to make as well 🙂

mosessond:
Issue created: https://github.com/rancher/dashboard/issues/13936

aalves08 (approved):
I've given this a test and issue seems to be fixed. Tests do seem to cover this scenario, so LGTM

rak-phillip (approved):
LGTM - I think it makes sense to get this in as we triage #13936

Issue #13843, opened 2025-03-21
Two projects with the same name mis-group the namespace


a110605:
**Setup**
- Rancher version: 2.9.3
- Rancher UI Extensions : N/A
- Browser type & version: Chrome

**Describe the bug**
Reported from SURE-9967 in harvester integrated rancher `Project/Namspaces` page.

Same issue happens in Rancher local -> Cluster ->  Project/Namspaces page.

Two projects with the same name mis-group the created namespace.

**To Reproduce**
1. Create a project named `test`
2. Create another project with the same name `test`
3. Create ns in first project : `ns1`
4. Create ns in second project: `ns2`

**Result**
ns1 and ns2 are both group to first test project.

**Expected Result**
ns1 and ns2 should group to different project, even their projects are all named test.

**Screenshots**

https://github.com/user-attachments/assets/a563548c-1aee-4650-bb42-43e8e9466c95

**Additional context**

a110605:
Current namespace is group by [project displayName](https://github.com/rancher/dashboard/blob/21ce896accdac134d88773a1d315d2d276b328d2/shell/models/namespace.js#L128).

If two projects with the same name `test`. The namespace will be grouped to first test project.

I think we should block user from creating the same project name via POST API `/v3/projects`.

If we allowed, it's hard to display the same project option when moving the namespace.

Hi @nwmac , how to do think ?

<img width="1379" alt="Image" src="https://github.com/user-attachments/assets/f8efb5b8-65f3-4969-943c-d038eded28c7" />

yonasberhe23:
Manual validation required. I think this is worth testing different types of clusters including Harvester.

a110605:
Validated on Rancher 2.12.0-alpha2 is good in project/namespace page ✅ 

<img width="1496" alt="Image" src="https://github.com/user-attachments/assets/27b439da-0da6-4da8-8ade-6e7206c0a145" />


But for imported harvester, we need to install 1.5.1-dev harvester-ui-extension from [v1.5-head](https://github.com/harvester/harvester-ui-extension/tree/v1.5-head) branch.

IsaSih:
Works as expected. Verified in both harvester and local cluster on 

[Rancher](https://github.com/rancher/rancher)	v2.11.2
[Dashboard](https://github.com/rancher/dashboard)	master c8b95b8
[Rancher CLI](https://github.com/rancher/cli)	v2.11.2.

[...]

PR #13872, opened 2025-03-25
Change namespace group by project Id.

a110605:
### Summary
Fixes # https://github.com/rancher/dashboard/issues/13843

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/ExplorerProjectsNamespaces.vue
+++ b/shell/components/ExplorerProjectsNamespaces.vue
@@ -11,7 +11,7 @@ import Masthead from '@shell/components/ResourceList/Masthead';
 import { mapPref, GROUP_RESOURCES, ALL_NAMESPACES, DEV } from '@shell/store/prefs';
 import MoveModal from '@shell/components/MoveModal';
 import ButtonMultiAction from '@shell/components/ButtonMultiAction.vue';
-
+import { escapeHtml } from '@shell/utils/string';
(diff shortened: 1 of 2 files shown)


cnotv (changes requested):
I can see that this logic is used in several other cases: https://github.com/search?q=repo%3Arancher%2Fdashboard+%22+groupByLabel%22&type=code
Should we not address the same issue everywhere at this point? It would create inconsistent logic across the project.
Maybe use a new method and call it `groupById` to avoid confusion also with the term?

The issue is labeled as `dev-automation` so we could pick up [the existing test to be implemented for the namespace model](https://github.com/rancher/dashboard/blob/master/shell/models/__tests__/namespace.test.ts#L87), as well as in the `ExplorerProjectsNamespaces` component. If it seems complex we could take a look together or I can commit the test for you.

a110605:
> I can see that this logic is used in several other cases: https://github.com/search?q=repo%3Arancher%2Fdashboard+%22+groupByLabel%22&type=code Should we not address the same issue everywhere at this point? It would create inconsistent logic across the project. Maybe use a new method and call it `groupById` to avoid confusion also with the term?
> 
> The issue is labeled as `dev-automation` so we could pick up [the existing test to be implemented for the namespace model](https://github.com/rancher/dashboard/blob/master/shell/models/__tests__/namespace.test.ts#L87), as well as in the `ExplorerProjectsNamespaces` component. If it seems complex we could take a look together or I can commit the test for you.

Thanks for review and comment. I will modify to groupById name and try to add unit test.

PR #13921, opened 2025-03-27
Preventing the re-render of cloud credentials when validation takes place

### LINKED ISSUE #13802: Cloud Creds: Secret input loses focus as footer button is enabled
**Setup**
- Rancher version: v2.11
- Rancher UI Extensions:
- Browser type & version:

**Describe the bug**
When creating/editing a cloud credential, the Secret/Access Token input field loses focus as the Create/Save button is enabled. Workaround is to refocus and continue typing secret. 
Order does matter - secret/access token needs to be filled last to reproduce this issue. 

**To Reproduce**
1. go to Cluster Management => Cloud Credentials => Create => Amazon
2. fill out form (secret last)
3. notice secret input loses focus as Create button is enabled

**Result**
Secret/Access Token input field loses focus as the Create/Save button is enabled.

**Expected Result**
Secret/Access Token input field does not lose focus as the Create/Save button is enabled.

**Screenshots**

https://github.com/user-attachments/assets/11b92bbb-0090-42f6-bc48-8c962c3d12e6

**Additional context**
Discovered while testing: https://github.com/rancher/dashboard/issues/13590

codyrancher:
### Summary
Fixes #13802

### Technical notes summary
This one is frustrating, it's absolutely not obvious what the root problem is. I found a fix through trial and error to narrow the change down.

For some reason if you visit the page after selecting the subtype if you then make the form valid it causes an event/state change to propagate all the to the root of the app and forces a re-render. 

The issue stopped once I removed the wrapping element but that broke other aspects because inherrited attributes were overwriting expected values for props. I then narrowed this down to `done-params` being present ultimately fixed the behavior.

Looking at all references to `done-params` doesn't explain the fix. And just adding other attributes in it's place doesn't either.

Cloud credentials which have validation. Azure, AWS, and S3 Compatible all had this issue when I tested.

Cloud credentials in general, creating and editing.

### Screenshot/Video
https://github.com/user-attachments/assets/a2c6f178-97ac-42a7-809c-6dbf0fe35b0f

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/edit/cloudcredential.vue
+++ b/shell/edit/cloudcredential.vue
@@ -276,6 +276,7 @@ export default {
     <Loading v-if="$fetchState.pending" />
     <CruResource
       v-else
+      :done-params="$attrs['done-params'] /* Without this, changes to the validationPassed prop end up propagating all the way to the root of the app and force a re-render when the input becomes valid. I haven't found a reasonable explanation for why this happens. */"
       :mode="mode"
       :validation-passed="validationPassed"
       :selected-subtype="value._type"
--- a/cypress/e2e/po/pages/cluster-manager/cloud-credentials-create-aws.po.ts
+++ b/cypress/e2e/po/pages/cluster-manager/cloud-credentials-create-aws.po.ts
@@ -1,0 +1,36 @@
+import PagePo from '@/cypress/e2e/po/pages/page.po';
+import LabeledInputPo from '@/cypress/e2e/po/components/labeled-input.po';
+
+export default class CloudCredentialsCreateAWSPagePo extends PagePo {
+  private static createPath(clusterId: string) {
+    return `/c/${ clusterId }/manager/cloudCredential/create?type=aws`;
+  }
+
+  static goTo(clusterId: string): Cypress.Chainable<Cypress.AUTWindow> {
+    return super.goTo(CloudCredentialsCreateAWSPagePo.createPath(clusterId));
+  }
+
+  constructor(private clusterId = '_') {
+    super(CloudCredentialsCreateAWSPagePo.createPath(clusterId));
+  }
+
+  accessKeyInput() {
+    return new LabeledInputPo('[data-testid="access-key"]');
+  }
+
+  secretKeyInput() {
+    return new LabeledInputPo('[data-testid="secret-key"]');
+  }
+
+  credentialNameInput() {
+    return new LabeledInputPo('[data-testid="name-ns-description-name"] .labeled-input input');
+  }
+
+  errorBanner() {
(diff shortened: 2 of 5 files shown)


shell/edit/cloudcredential.vue
```diff
@@ -272,47 +274,46 @@ export default {
 </script>
 
 <template>
-  <form class="filled-height">
```
codyrancher, line 275: Removing this to be consistent with all other uses of CruResource. I can put this back in if we'd like. It doesn't affect the footer which was the original purpose.

shell/edit/cloudcredential.vue
```diff
@@ -272,47 +274,46 @@ export default {
...
-    <Loading v-if="$fetchState.pending" />
-    <CruResource
-      v-else
+  <Loading v-if="$fetchState.pending" />
+  <CruResource
+    v-else
+    :done-params="$attrs['done-params'] /* Without this, changes to the validationPassed prop end up propagating all the way to the root of the app and force a re-render when the input becomes valid. I haven't found a reasonable explanation for why this happens. */"
```
codyrancher, line 280: The actual fix

codyrancher:
@rak-phillip I don't call this a proper fix but given the amount of time I've spent on this already and the given severity I thought it was appropriate to go with this change. If you're not comfortable with this fix let me know and maybe we can figure something else out.

shell/edit/cloudcredential.vue
rak-phillip: Would it be appropriate to do `v-bind="$attrs"` instead? It's hard to tell, but it seems like all the fallthrough attributes are intended for this node.

If not, I think that it would be more clear if we created a `doneParams` prop for this component. This way, we could have a more explicit contract for this component instead of relying on implicit attributes.

shell/edit/cloudcredential.vue
rak-phillip: I'm fine with keeping this out, the extra form tag appeared to be a little redundant in the dom tree.

rak-phillip (commented):
Tested the fix and it behaves as expected. I only have a few comments about data binding for `cloudcredential.vue`

shell/edit/cloudcredential.vue
rak-phillip: we had an out of band discussion and decided to keep the implementation as it is in this PR. There's too many conflicts and edge cases that complicate making a change.

rak-phillip (approved):
LGTM

codyrancher:
Updated based on our conversation.

rak-phillip (approved):

PR #14041, opened 2025-04-07
Allow cluster create and import locations to be overridden

### LINKED ISSUE #14012: Allow users to replace the cluster creation/import screens in an extensiuon
For some deployments of Rancher, we want to be able to skip the cluster creation selection screen and go directly to a creation screen for a provider.

We should add an extension point for this.

Image

nwmac:
### Summary
Fixes #14012

This PR adds configuration for the provisioning cluster resource to allow the `create` and `import` locations to be overridden by an extension.

This is then wired into the cluster list and home page views.

Extensions can change the locations - an example use case is where you might want a user to only be able to create or import a single cluster type - for example, adding this in a product file in an extension would only allow them to create/import EKS clusters:

```

export function init(plugin: any, store: any) {
  const {configureType } = plugin.DSL(store, 'manager');

  configureType('provisioning.cattle.io.cluster', {
    custom: {
      createLocation: (params: any) => {
        return {
          name:  'c-cluster-product-resource-create',
          query: {
            type: 'amazoneks'
          },
          params
        }
      },
      importLocation: (params: any) => {
        return {
          name:  'c-cluster-product-resource-create',
          query: {
            mode: 'import',
            type: 'amazoneks'
          },
          params
        }
      },
    }
  });
}
```

This is considered an advanced use case and experimental and will be documented once stable. 

Testing should focus on testing that the existing UI is not broken by this change.

Both on the home page and the cluster list page, check that clicking the 'Create' and 'Import' buttons takes the user to the default pages allowing them to chose the cluster type.

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/store/type-map.js
+++ b/shell/store/type-map.js
@@ -107,6 +107,7 @@
 //                               graphConfig: undefined   -- Use this to pass along the graph configuration
 //                               notFilterNamespace:  undefined -- Define namespaces that do not need to be filtered
 //                               localOnly: False -- Hide this type from the nav/search bar on downstream clusters
+//                               custom: any - Custom options for a given type
 //                           }
 // )
 // ignoreGroup(group):        Never show group or any types in it
@@ -524,6 +525,7 @@ export const getters = {
       depaginate:             false,
       customRoute:            undefined,
       resourceEditMasthead:   true,
+      custom:                 {},
     };
 
     return (schemaOrType, pagination) => {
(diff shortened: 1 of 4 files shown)


shell/store/type-map.js
```diff
@@ -524,6 +525,7 @@ export const getters = {
       depaginate:             false,
       customRoute:            undefined,
       resourceEditMasthead:   true,
+      custom:                 {},
```
aalves08, line 528: This needs to be added to `ConfigureTypeOptions`

aalves08 (changes requested):
Overall, works as advertised. Was able to override the creation of a cluster for Digital Ocean.

We should add `custom` to the `ConfigureTypeOptions` for Typescript support.

We should also document this feature on the extension docs, stating from which shell version it's available

nwmac:
@aalves08 I have updated and added to the type.

As per comment at top, I was proposing not documenting this for now and once we have finalised this API/approach, we can document. There may be changes we want to make, so I consider this experimental for now.

aalves08 (approved):
LGTM. Sorry, I didn't notice the comment about the documentation. That's fine by me 🙏

Issue #13939, opened 2025-03-31
Remove route logic from page inizialization (index and auth/verify)


cnotv:
# Description

Routing logic for login/logout handled in the files [`shell/pages/index.vue`](https://github.com/rancher/dashboard/blob/master/shell/pages/index.vue) and [`shell/pages/auth/verify.vue`](https://github.com/rancher/dashboard/blob/master/shell/pages/auth/verify.vue) should be handled in a route guard.

PR #14042, opened 2025-04-07
Can not redirect back to previous page after login

### LINKED ISSUE #4152: [rancher/ui] Can not redirect back to previous page after login
**Rancher Server Setup**
- Rancher version: v2.5.5

**Describe the bug**

With Github-OAuth, Rancher cannot redirect to previous page. 

**To Reproduce**
 
1. configure with Github-Oauth
2. logout
2. open a Rancher page. eg. https://rancher.example.com/p/c-zt94r:p-vdvp7/workloads
3. login by Github-OAuth

**Result**
Rancher can not redirect to previous page (https://rancher.example.com/p/c-zt94r:p-vdvp7/workloads) after login.

**Expected Result**

Rancher redirect to previous page automatically.

**Additional context**

I found that Rancher write current page url to localStorage before redirect to login page.
image

**BUT does not read it back after redirecting back from Github OAuth.**
image

**AND unset it after login**
image

cnotv:
### Summary
Fixes #4152

Restored redirection to last page with every auth provider.

### Technical notes summary

Last page storage was omitted due change in the logic, where the app is routed directly and last URL was lost.

- Added comments in markup and code
- Added note for tech debt issue to abstract logic
- Replaced router method call with logout handler
- Removed deprecated code in the store
- Adopted `pathname` over full path to comply with router requirements
- Added first tests for:
  - Header component
  - Index (empty) page

### Screenshot/Video

https://github.com/user-attachments/assets/e5997877-fe33-4fe4-a00c-e42be8808d16

### Checklist

- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/nav/Header.vue
+++ b/shell/components/nav/Header.vue
@@ -15,6 +15,7 @@ import { LOGGED_OUT, IS_SSO } from '@shell/config/query-params';
 import NamespaceFilter from './NamespaceFilter';
 import WorkspaceSwitcher from './WorkspaceSwitcher';
 import TopLevelMenu from './TopLevelMenu';
+import { BACK_TO } from '@shell/config/local-storage';
 
 import Jump from './Jump';
 import { allHash } from '@shell/utils/promise';
@@ -135,10 +136,6 @@ export default {
 
     isAuthLocalProvider() {
       return this.principal && this.principal.provider === 'local';
-    },
-
-    generateLogoutRoute() {
-      return this.isAuthLocalProvider ? { name: 'auth-logout', query: { [LOGGED_OUT]: true } } : { name: 'auth-logout', query: { [LOGGED_OUT]: true, [IS_SSO]: true } };
     },
 
     principal() {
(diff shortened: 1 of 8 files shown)


shell/components/nav/__tests__/Header.test.ts
```diff
@@ -0,0 +1,66 @@
...
+  });
+
+  it('pressing logout should store current path', async() => {
+    const pathname = '/whatever';
+    const localStorageSpy = jest.spyOn(Object.getPrototypeOf(localStorage), 'setItem');
+
+    Object.defineProperty(window, 'location', { value: new URL(`http://whatever${ pathname }`) } );
+
+    // await wrapper.find('[data-testid="nav_header_showUserMenu"]').trigger('click');
+    // await wrapper.find('[data-testid="nav_header_logout"]').trigger('click');
```
rak-phillip, line 61: nitpick: We can omit the commented out code

shell/components/nav/Header.vue
```diff
@@ -390,6 +387,31 @@ export default {
...
+     */
+    storeCurrentPath() {
+      const isLogin = this.$route.name === 'auth-login' || this.$route.path === '/login'; // Cover dashboard and case of log out from ember;
+      const isLogout = this.$route.name === 'auth-logout';
+
+      // Prevent loop to login/logout pages
+      if (!isLogin && !isLogout) {
+        window.localStorage.setItem(BACK_TO, window.location.pathname);
+      }
+    },
```
rak-phillip, line 403: Is it sufficient to store the current path in the header when logout is clicked? Do we need to consider cases where the session expires via `auth-user-session-ttl-minutes`?

shell/components/nav/__tests__/Header.test.ts
cnotv: yeah I left it because it should be that way, but I can remove it

shell/components/nav/Header.vue
cnotv: That's a good point, but I do not know if it was implemented before.
I would rather handle that when we refactor the whole logic and take it out from the components in https://github.com/rancher/dashboard/issues/13939
This issue touched already too much unnecessary parts.

cnotv: Wait why do we have a test `auth-user-session-ttl-minutes` which is not failing?

cnotv: Also is it actually working on `master`? 🤔

cnotv:
I am closing this PR as the functionality actually works as expected and I was just not aware of the option 🙃 

I will create a separated issue to integrate the tests in the current implementation.

PR #14046, opened 2025-04-07
Update IngressClass labeled select to use `update:value` instead of `selecting`

### LINKED ISSUE #14043: [BUG][2.11] Issue with Ingress Class Configuration Not Retaining Changes
**Rancher Server Setup**
- Rancher version: 2.11
- Proxy/Cert Details:

- Kubernetes version: 1.31.7
- Cluster Type (Local/Downstream): Local
 
**User Information**
- What is the role of the user logged in? Admin

**Describe the bug**
Changing the Ingress Class on an existing Ingress causes the previously configured Ingress Class to be deleted. Additionally, setting an Ingress Class for a new Ingress does not take effect.

**To Reproduce**
Service Discovery -> Ingresses -> Edit an Ingress -> Set an Ingress Class -> Save

**Result**

**Expected Result**
Ingress Class correctly saved after clicking the save button

**Screenshots**

Image

Image

**Additional context**

rak-phillip:
### Summary
This updates the IngressClass component so that it listens for the `update:value` event instead of `selecting`.

Fixes #14043 

- Update the IngressClass component so that it listens for the `update:value` event instead of `selecting`

### Technical notes summary

While investigating this, I noticed that value wasn't updating because `@selecting` gets emitted twice. It appears that first emitted event contains the expected data formatted properly, while the second does not, deleting the selected value altogether. I'm opting to use `update:value` because this event should explicitly emit only when a value is selected by the user. 

Ingress Class
- Cluster Explorer => Service Discovery => Ingresses

Ingress Class
- Cluster Explorer => Service Discovery => Ingresses

### Screenshot/Video

NA

### Checklist
- [ ] The PR has a Milestone 

### DIFF
--- a/shell/edit/networking.k8s.io.ingress/IngressClass.vue
+++ b/shell/edit/networking.k8s.io.ingress/IngressClass.vue
@@ -5,6 +5,7 @@ import { get, set, remove } from '@shell/utils/object';
 
 export default {
   components: { LabeledSelect },
+  emits:      ['update:value'],
   props:      {
     value: {
       type:    Object,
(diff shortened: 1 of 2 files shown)


rak-phillip:
@codyrancher It appears that we started emitting `selecting` along with `update:value` in f52d9cf26a10d9f466e6737de6afc114733322fd. I think this is the source of the double-emit of the `selecting` event, do you recall why this change was needed at the time?

PR #14054, opened 2025-04-08
Remove deleted props from tooltips

rak-phillip:
### Summary
This change removes warnings that were printed to the console when hovering over primary navigation items. The `popperOptions` prop was removed from floating-vue[^1] and hasn't done anything since we migrated to Vue3.

[^1]: https://floating-vue.starpad.dev/migration/migration-from-v3#removed-props

- Remove deleted `popperOptions` prop from the return value of `getTooltipConfig()`

### Technical notes summary

- The `popperOptions` prop was removed from floating-vue[^1] and hasn't done anything since we migrated to Vue3

- Primary navigation

- Primary navigation

#### Before

image

#### After

image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/nav/TopLevelMenu.vue
+++ b/shell/components/nav/TopLevelMenu.vue
@@ -430,8 +430,7 @@ export default {
 
       return {
         content,
-        placement:     'right',
-        popperOptions: { modifiers: { preventOverflow: { enabled: false }, hide: { enabled: false } } },
+        placement: 'right',
         popperClass
       };
     },


aalves08 (approved):
I gave this a try with the closed an open sidebar and looks the same as in a `head` version of Rancher (2.11), so LGTM. Just out of curiosity, how do we now pass options to a specific tooltip? 🤔

rak-phillip:
> I gave this a try with the closed an open sidebar and looks the same as in a head version of Rancher (2.11), so LGTM. Just out of curiosity, how do we now pass options to a specific tooltip? 🤔

That's a good question. Since popper options has been removed, it looks like the old modifiers have been replaced with props. At least that's the case for `preventOverflow`[^1].

`hide` appears to be tied to the `popperJs` option from the original `v-tooltip` implementation. This might not apply to floating-vue anymore.

@aalves08 shall we attempt to add `preventOverflow` back instead of removing it entirely?

[^1]: https://floating-vue.starpad.dev/api/#preventoverflow

aalves08:
> > I gave this a try with the closed an open sidebar and looks the same as in a head version of Rancher (2.11), so LGTM. Just out of curiosity, how do we now pass options to a specific tooltip? 🤔
> 
> That's a good question. Since popper options has been removed, it looks like the old modifiers have been replaced with props. At least that's the case for `preventOverflow`[1](#user-content-fn-1-84ca1ed2c3071c64dca011daf4b69947).
> 
> `hide` appears to be tied to the `popperJs` option from the original `v-tooltip` implementation. This might not apply to floating-vue anymore.
> 
> @aalves08 shall we attempt to add `preventOverflow` back instead of removing it entirely?
> 
> ## Footnotes
> 1. https://floating-vue.starpad.dev/api/#preventoverflow [↩](#user-content-fnref-1-84ca1ed2c3071c64dca011daf4b69947)

I remember that I did some work on the tooltips around styling and implemented it in a couple of places in the UI, don't remember exactly where.... My point is that I had to use some of their configuration to achieve the desired goal. It would definitely be good to add/fix those configuration points/properties because they were put there for a reason, just don't know exactly why... 😝 

Nevertheless, regarding your changes, I didn't notice anything strange, so they all seem to be ✅

rak-phillip:
I'm going to merge this PR. I'm not in favor of adding code to address a scenario that doesn't have any open regressions and without knowing 100% why we need the configuration.

PR #14081, opened 2025-04-09
Enter edit mode if updating roles fails during user creation

### LINKED ISSUE #13539: Create User: Improve error handling
**Setup**
- Rancher version: 2.7.6, 2.9.6, 2.9.8, 2.10.1
- Rancher UI Extensions:
- Browser type & version:
- Internal reference: SURE-9866

**Describe the bug**
When creating a user with more rights than is currently possible (e.g. a restricted-admin trying to create a full-admin user), an error banner related to RBAC issues appears on the same page. The user was not created so another attempt is made to modify the roles and try creating again.

What actually happens though is the user was actually created in the background, albeit with no associated role. This means that even if you try to modify the permission checkboxes on the current page and create again, a conflict error will appear due to a user with this name already existing.

The guess is that the Rancher UI creates the user, then the global role bindings, and when this second step fails, the former is not rolled back.

Workaround:
Refresh the page or leave the page and return

Image

Image

Upon analysis from @richard-cox:
> The user shouldn't be offered the ability to create a user with escalated permissions. This is protected by the backend, but the UI still offers the option. The UI should in theory be able to tell for the provided global roles if they're creating anything above them (user-base --> standard-user --> restricted-admin --> admin), however this wouldn't cover custom roles (like the replacement to restricted-admin). I would recommend we don't change that part

> The UI does create the user first and then create the bindings. We just need to track that the create user request succeeds, so any further saves result in the edit flow (the same component handles both create and edit user actions).

The RBAC team chimed in:
> Creating a user and assigning a Global Role are 2 different operations. The UI makes the experience nicer by doing them both for you in a single page, but what we're seeing is the 1st operation succeeds and then the 2nd operation fails. Maybe the UI could delete the user if the 2nd operation fails? If that's not too difficult that would probably be sufficient.

rak-phillip:
### Summary
This emits  an `update:mode` event that sets to mode to "edit" and refreshes all resource-related variables when the `updateRoles()` method fails during user creation. This provides an opportunity for users to correct mistakes while still staying in the "Create User" context.

Fixes #13539 

- Switch mode to "edit" if the `updateRoles()` method fails during user creation
- Add a user readable error message if an escalation error is raised

### Technical notes summary

An alternate approach that explored navigating to an edit route was explored, but this required re-rendering the entire form and would eliminate the error message; additional state to persist the error messages would be required and I found that requirement to be more complex than the approach taken here.

User Creation: See details in the associated issue.

User Creation: Ensuring that the form still functions properly after an error is raised during user creation.

### Screenshot/Video

image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/cypress/e2e/tests/pages/users-and-auth/users.spec.ts
+++ b/cypress/e2e/tests/pages/users-and-auth/users.spec.ts
@@ -543,6 +543,8 @@ describe('Users', { tags: ['@usersAndAuths', '@adminUser'] }, () => {
       usersPo.waitForPage();
       usersPo.list().create();
 
+      const runTimestamp = +new Date();
+      const runPrefix = `e2e-test-${ runTimestamp }`;
       const adminUsername = `${ runPrefix }-admin-user`;
       const adminPassword = 'admin-password';
 
--- a/cypress/e2e/tests/pages/users-and-auth/users.spec.ts
+++ b/cypress/e2e/tests/pages/users-and-auth/users.spec.ts
@@ -6,6 +6,8 @@ import * as path from 'path';
 import { generateUsersDataSmall } from '@/cypress/e2e/blueprints/users/users-get';
 import BurgerMenuPo from '@/cypress/e2e/po/side-bars/burger-side-menu.po';
 import SortableTablePo from '@/cypress/e2e/po/components/sortable-table.po';
+import HomePagePo from '@/cypress/e2e/po/pages/home.po';
+import FixedBannerPo from '@/cypress/e2e/po/components/fixed-banner.po';
 
 const usersPo = new UsersPo('_');
 const userCreate = usersPo.createEdit();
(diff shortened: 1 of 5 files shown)


shell/assets/translations/en-us.yaml
```diff
@@ -5134,6 +5134,8 @@ rbac:
         label: Login Access
       clustertemplaterevisions-create:
         label: Create RKE Template Revisions
+  errors:
+    escalation: You do not have sufficient permissions to grant Global Permissions above your own.
```
rak-phillip, line 5138: I'm open to suggestions for improving this error message. This is what the original error looks like

```
admission webhook "rancher.cattle.io.globalrolebindings.management.cattle.io" denied the request: errors due to escalation: user "u-q46zm" (groups=["system:authenticated" "system:cattle:authenticated"]) is attempting to grant RBAC permissions not currently held:
{APIGroups:["*"], Resources:["*"], Verbs:["*"]}
{NonResourceURLs:["*"], Verbs:["*"]}
```

eva-vashkevich: I feel like you can just say:
"You cannot grant permissions above your own. Please verify your permissions." or something along those lines. There are no permissions that would be sufficient to grant someone permissions above your own

rak-phillip: > There are no permissions that would be sufficient to grant someone permissions above your own

Right, that's logical in hindsight.. Thanks for the feedback 🙌

shell/components/ResourceDetail/index.vue
```diff
@@ -414,6 +403,50 @@ export default {
...
+     * @param {string} payload.mode - The mode to set.
+     * @param {Object} payload.user - The user object containing user-specific
+     * information.
+     * @param {string} payload.resource - The resource string to use for
+     * initialization.
+     */
+    setMode({ mode, user, resource }) {
+      this.mode = mode;
+      this.configureResource(user, resource);
+    }
```
rak-phillip, line 449: I'm not 100% satisfied with this approach, but it satisfies the constraint of allowing a user to continue making modifications if roles fail to assign during user creation for any reason. I'm open to alternative approaches..

rak-phillip:
@yonasberhe23 can you provide feedback on the e2e test for this PR?

cypress/e2e/tests/pages/users-and-auth/users.spec.ts
```diff
@@ -496,4 +498,62 @@ describe('Users', { tags: ['@usersAndAuths', '@adminUser'] }, () => {
...
   });
+
+  describe('Create admin user with standard user', () => {
+    before(() => {
+      cy.login();
+    });
+
+    it('User creation should complete after admin user fails to create for Standard user with Manage Users Role', () => {
+      // create standard user
+      usersPo.goTo();
```
yonasberhe23, line 509: I think its good practice to add `usersPo.waitForPage();` after navigation

cypress/e2e/tests/pages/users-and-auth/users.spec.ts
```diff
@@ -496,4 +498,62 @@ describe('Users', { tags: ['@usersAndAuths', '@adminUser'] }, () => {
...
+
+      // logout admin
+      cy.logout();
+
+      // login as standard user
+      cy.login(standardUsername, standardPassword);
+
+      // navigate to the users page and attempt to create an admin user
+      HomePagePo.goToAndWaitForGet();
+      usersPo.goTo();
```
yonasberhe23, line 531: add `usersPo.waitForPage();`

yonasberhe23 (changes requested):
Just two comments for the e2e test. Otherwise looks great!

codyrancher (approved):
Verified it works. 

I did notice that upon error it will change the global permissions back to a standard user. I think it could cause some confusion if the user goes back to check why the error was received.

rak-phillip:
> Verified it works.
>
> I did notice that upon error it will change the global permissions back to a standard user. I think it could cause some confusion if the user goes back to check why the error was received.

@codyrancher I fixed this behavior in 2fffa89ecda28885b83ef100471f9d9ef1961ce4. It turns out that the `userId` watcher was overriding values when we switched to the "edit" mode. I'm opting to override the watcher in this scenario to prevent the update.

yonasberhe23 (approved):

codyrancher (approved):

rak-phillip:
All e2e tests pass, looking for a final review.

codyrancher (approved):

yonasberhe23 (approved):

PR #14093, opened 2025-04-10
Improve Banners settings screen reader output

### LINKED ISSUE #14092: a11y: Identical label for form fields - `banners` settings area in Settings
**Setup**
- Rancher version:
- Rancher UI Extensions:
- Browser type & version:

**Describe the bug**

Identical label for form fields - currently the screen reader announces every UI element without providing the context of the area (ex: text for header banner is announced as `text`, same for footer banner) so the user will have much more difficulties navigating through the UI.

**To Reproduce**

**Result**
Identical label is specified for the form fields resulting in screen reader users might find it difficult to understand the purpose of the form fields effectively.

**Expected Result**

Screen reader should announce unique and descriptive label for checkboxes and radio buttons.

**Screenshots**

**Additional context**

aalves08:
### Summary
Fixes #14092 

- Fixes the settings area for `banners` in order to improve context of screen reader output for the UI elements involved
- Adds unit tests related to `a11y`

### Technical notes summary

- Some of these fixes might differ in their pattern because, imo, our HTML markup is bloating some of our components, making it much more difficult to get the same screen readers outputs as one would get by browsing https://www.w3.org/WAI/ARIA/apg/patterns/ various examples with the screen reader

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/pkg/rancher-components/src/components/Form/Radio/RadioGroup.vue
+++ b/pkg/rancher-components/src/components/Form/Radio/RadioGroup.vue
@@ -108,9 +108,9 @@ export default defineComponent({
   emits: ['update:value'],
 
   data() {
-    return { 
-      currFocusedElem: undefined as undefined | EventTarget | null,
-      radioOptionsIdPrefix: `radio-option-${generateRandomAlphaString(12)}-`
+    return {
+      currFocusedElem:      undefined as undefined | EventTarget | null,
+      radioOptionsIdPrefix: `radio-option-${ generateRandomAlphaString(12) }-`
     };
   },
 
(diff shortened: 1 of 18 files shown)


pkg/rancher-components/src/components/Form/Radio/RadioGroup.vue
```diff
@@ -199,42 +237,42 @@ export default defineComponent({
 </script>
 
 <template>
-  <fieldset>
+  <div>
```
rak-phillip, line 240: Why are we removing the fieldset here?

rak-phillip (commented):
Looks good for the most part. Just one question about the radio group change.

pkg/rancher-components/src/components/Form/Radio/RadioGroup.vue
aalves08: We were "bloating" the HTML when all we required was the proper wiring and passing of those attributes.

rak-phillip (approved):
LGTM

PR #14095, opened 2025-04-10
Replace the done param fix and changed how we render the credential dynamic components

### LINKED ISSUE #13802: Cloud Creds: Secret input loses focus as footer button is enabled

codyrancher:
### Summary

Fixes #13802

I reverted the `done-param` changes, I still have no idea why those functioned. 

I was able to trace what was causing the re-renders further and it seems to be caused by re-computing the component that we wanted to render even if that component didn't change. We now manually compute these values which prevents the re-render from occurring when the form input becomes valid.

Cloud credentials create pages, hosted providers create credentials  section, node provisioned cluster create credentials section.

Anywhere with creating/selecting credentials.

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/edit/cloudcredential.vue
+++ b/shell/edit/cloudcredential.vue
@@ -100,6 +100,30 @@ export default {
   computed: {
     rke2Enabled: mapFeature(RKE2_FEATURE),
 
+    hasCustomCloudCredentialComponent() {
+      const driverName = this.driverName;
+
+      return this.$store.getters['type-map/hasCustomCloudCredentialComponent'](driverName);
+    },
+
+    cloudCredentialComponent() {
+      const driverName = this.driverName;
+
+      return this.$store.getters['type-map/importCloudCredential'](driverName);
+    },
+
+    genericCloudCredentialComponent() {
+      return this.$store.getters['type-map/importCloudCredential']('generic');
+    },
+
+    cloudComponent() {
+      if (this.hasCustomCloudCredentialComponent) {
+        return this.cloudCredentialComponent;
+      }
+
+      return this.genericCloudCredentialComponent;
+    },
+
     validationPassed() {
       return this.credCustomComponentValidation && this.nameRequiredValidation;
     },
@@ -110,14 +134,6 @@ export default {
 
     driverName() {
       return this.value?.provider;
-    },
-
-    cloudComponent() {
-      if (this.$store.getters['type-map/hasCustomCloudCredentialComponent'](this.driverName)) {
-        return this.$store.getters['type-map/importCloudCredential'](this.driverName);
-      }
-
-      return this.$store.getters['type-map/importCloudCredential']('generic');
     },
 
     // array of id, label, description, initials for type selection step
@@ -194,7 +210,6 @@ export default {
   },
 
   methods: {
-
     createValidationChanged(passed) {
       this.credCustomComponentValidation = passed;
     },
@@ -276,7 +291,6 @@ export default {
     <Loading v-if="$fetchState.pending" />
     <CruResource
       v-else
-      :done-params="$attrs['done-params'] /* Without this, changes to the validationPassed prop end up propagating all the way to the root of the app and force a re-render when the input becomes valid. I haven't found a reasonable explanation for why this happens. */"
       :mode="mode"
       :validation-passed="validationPassed"
       :selected-subtype="value._type"
--- a/shell/edit/provisioning.cattle.io.cluster/SelectCredential.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/SelectCredential.vue
@@ -81,6 +81,30 @@ export default {
   },
 
   computed: {
+    hasCustomCloudCredentialComponent() {
+      const driverName = this.driverName;
+
+      return this.$store.getters['type-map/hasCustomCloudCredentialComponent'](driverName);
+    },
+
+    cloudCredentialComponent() {
+      const driverName = this.driverName;
+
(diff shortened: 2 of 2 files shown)


shell/edit/cloudcredential.vue
```diff
@@ -112,14 +119,6 @@ export default {
       return this.value?.provider;
     },
 
-    cloudComponent() {
```
codyrancher, line 115: I'm moving the components from computed properties to data properties that need to be manually assigned.

It looks as though the re-render was being caused when these properties got recomputed even if the resulting value didn't change.

shell/edit/cloudcredential.vue
```diff
@@ -80,20 +80,27 @@ export default {
...
+      cloudComponent:                null
     };
   },
 
   watch: {
     'value._name'(newValue) {
       this.nameRequiredValidation = newValue?.length > 0;
+    },
+
+    driverName() {
```
codyrancher, line 102: Driver name isn't always available by the time data/fetch are invoked so I needed to add a watcher.

shell/edit/cloudcredential.vue
rak-phillip: This is a great observation, I hadn't considered that there might be an issue with the design of this computed prop. 

This piqued my interest, so I decided to investigate what would happen if we adhered to the principle of "keep computed props as simple as possible" to see if this affects the outcome - it appears that it does. 

I decided to split the getters out into individual computed properties and the issue appears to be resolved with this approach as well:

```javascript
    hasCustomCloudCredentialComponent() {
      const driverName = this.driverName;

      return this.$store.getters['type-map/hasCustomCloudCredentialComponent'](driverName);
    },

    cloudCredentialComponent() {
      const driverName = this.driverName;

      return this.$store.getters['type-map/importCloudCredential'](driverName);
    },

    genericCloudCredentialComponent() {
      return this.$store.getters['type-map/importCloudCredential']('generic');
    },

    cloudComponent() {
      if (this.hasCustomCloudCredentialComponent) {
        return this.cloudCredentialComponent;
      }

      return this.genericCloudCredentialComponent;
    },
```

This style of access in more in-line with how vuex getters are typically accessed in a Vue component. My current hypothesis is that our style of accessing getters can produce scenarios that invalidate a computed prop's cache.

codyrancher: I'll switch it over to this.

shell/edit/cloudcredential.vue
```diff
@@ -80,20 +80,27 @@ export default {
...
   },
 
   data() {
     return {
       credCustomComponentValidation: false,
       nameRequiredValidation:        false,
       nodeDrivers:                   null,
-      kontainerDrivers:              null
+      kontainerDrivers:              null,
+      cloudComponent:                null
```
rak-phillip, line 93: Storing `cloudComponent` in data produces the following warning:

```
[Vue warn]: Vue received a Component that was made a reactive object. This can lead to unnecessary performance overhead and should be avoided by marking the component with `markRaw` or using `shallowRef` instead of `ref`. 
Component that was made reactive:  
Object { name: "AsyncComponentWrapper", __asyncLoader: load(), __asyncHydrate: __asyncHydrate(el, instance, hydrate), __asyncResolved: Getter, setup: setup()
 }
 
  at <CruResource key=1 mode="create" validation-passed=false  ... > 
  at <CruCloudCredential key=4 ref="comp" value= 
Object { type: "cloudcredential", __clone: true, metadata: {…}, _name: "", annotations: {…}, amazonec2credentialConfig: {}, _type: "provisioning.cattle.io/cloud-credential", … }
  ... > 
  at <AsyncComponentWrapper key=4 ref="comp" value= 
Object { type: "cloudcredential", __clone: true, metadata: {…}, _name: "", annotations: {…}, amazonec2credentialConfig: {}, _type: "provisioning.cattle.io/cloud-credential", … }
  ... > 
  at <Index resource-override="cloudcredential" parent-route-override="c-cluster-manager-cloudCredential" class="outlet" > 
  at <CloudCredentialCreate class="outlet" onVnodeUnmounted=fn<onVnodeUnmounted> ref=Ref< 
 > > 
  at <RouterView key="/c/_/manager/cloudCredential/create" class="outlet" > 
 > > 
  at <RouterView> 
  at <App>
``

This can be resolved via `toRaw` or `shallowRef` when assigning a value to `cloudComponent`.

rak-phillip (changes requested):
LGTM - this behaves much more reliably than before. We just need to resolve the new warning as a result of storing components in data.

shell/edit/cloudcredential.vue
codyrancher: I should've caught that. I was just looking at another instance of this last week.

Issue #14115, opened 2025-04-11
Fleet get Bundle resources from BundleDeployment


torchiaf:
Fleet issue: https://github.com/rancher/fleet/issues/3482

Fleet will remove the `bundle.status.resourceKey` field to fix related performance issues.
The UI uses this field to calculate the Bundle's resources statuses in Bundle's details page.

**Describe the solution you'd like**

- We can use the `bundledeployment.status.resources` instead.
- Add a `cluster` column to specify which cluster the resources are deploy on.

**Describe alternatives you've considered**

**Additional context**

yonasberhe23:
- The Resources table no longer relies on status.resourceKey. It now uses status.resources from /v1/fleet.cattle.io.bundledeployments/....
- The new source (BundleDeployment.status.resources) is correctly populated on:
- - Git Repos -> Git Repo Detail page Resources Tab
- - Advanced -> Bundles -> Bundle Detail page (Resources)
- Bundle resources are now accurately listed per downstream cluster using BundleDeployments.
- Resource statuses and GVK/cluster metadata are correctly displayed.

Tested on: Rancher v2.12-9d8a81e5882a1c435fd4b5f68428e59dd69d73d1-head / Dashboard master `f206b54`

PR #14124, opened 2025-04-14
Use BundleDeployment status for showing Bundle resources list

aruiz14:
### Summary
Refers to #14115

The Fleet `Bundle`s detail page shows a list of all resources deployed to downstream clusters from a single `Bundle`. However, since it takes the list from `status.resourceKey`, it only contains the GVK and key of the resource, but not the downstream cluster. Also, in order to obtain the state of every resource, the `status.summary`'s `nonReadyResources` only contains limited information, and makes it hard to trace it back to the actual object in that state.

The main goal of this PR is to use a different source of information for the Bundle resources list, to use the BundleDeployments directly, as they contain all the information needed, instead of just a summary provided by the Fleet controller.

### Technical notes summary

We'll filter BundleDeployments created from a specific Bundle using their well-known labels, which will give us access to a concrete list of resources and the cluster they belong to. By using this and not the `Bundle`'s `.status.resourceKeys`, that will remove the only usage of that field, allowing Fleet to deprecate it. This field could get many entries for large setups, contributing to the already-big payload size of Bundles.

Also, I'm dropping the `FleetBundleResources` Vue Component in favor of `FleetResources`, which is already used by the `GitRepo` details page, for better reusability of code and a consistent UX.

In order to be able to reuse the `FleetResources` component, I had to do a small refactoring. This is now used by:
 - GitRepo details > "Resources" tab.
 - Bundle details page.

`BundleDeployments` will need to be fetched before being able to display Bundle details, adding some latency on first load (similar to GitRepos).

### Screenshot/Video

<details>
  <summary>Bundle details:</summary>
Before:

Screenshot 2025-04-15 at 10 11 25

After:

Screenshot 2025-04-15 at 10 10 18
</details>

<details>
  <summary>GitRepo details - Resources tab:</summary>
Before:

Screenshot 2025-04-15 at 10 12 16
After:

Screenshot 2025-04-15 at 10 11 56
</details>

### Checklist
- [X] The PR has a Milestone 


torchiaf (approved):
LGTM, some comments:

- This will introduce a small delay when users navigate to a Budle's details page for the first time, until the UI finishes to load the BundleDeployments.
- Not related to the current change: the resource list in GitRepo details page is different from the one in the Bundle's details page as non-Active Bundle's resources are missing from GitRepo's resources list in the GitRepo's YAML. This requires some UX analysis.

PR #13532, opened 2025-02-27
Fleet: switch workspace when navigating to GitRepo details page

torchiaf:
### Summary
Fixes #13531

### Checklist
- [ ] The PR is linked to an issue and the linked issue has a Milestone, or no issue is needed
- [ ] The PR has a Milestone 
- [ ] The PR template has been filled out
- [ ] The PR has been self reviewed 
- [ ] The PR has a reviewer assigned
- [ ] The PR has automated tests or clear instructions for manual tests and the linked issue has appropriate QA labels, or tests are not needed
- [ ] The PR has reviewed with UX and tested in light and dark mode, or there are no UX changes

PR #13703, opened 2025-03-11
Refactor form components so data prop no longer contains init logic

rak-phillip:
### Summary
This refactors form components that contain initialization logic in their data props so that they use patterns that are more idiomatic to Vue.

Fixes #13702 

### Technical notes summary

🗒️ NOTE: I recommend reviewing this PR commit-by-commit to reduce the complexity of reviewing some larger changes. 

🗒️ NOTE: This is a large PR that touches many forms throughout Dashboard. I am open to splitting this into smaller PRs if we find that validating and testing these changes takes too much effort.

Some form components require a larger refactor than others because of watcher usage. Watchers will trigger when watched values are updated in the created hook; this is undesirable because this modifies component behavior on initial render, potentially introducing regressions in forms. Refactoring these components to use setup() has the advantage of allowing us to retain a similar data initialization pattern to what we previously used before. I kept these refactors limited to components that utilize watchers on data props that require initialization in order to reduce the amount of complexity involved in this refactor. I think that the tradeoff is worth it in order solve the issue while maintaining existing behavior.  

### Areas or cases that should be tested

In effect, this touches all of our forms. Components like `NameNsDescription.vue` are widely used throughout Dashboard. Focus on core form functionality and pay attention to larger refactors with components:

- Probe.vue
- ValueFromResource.vue
- NameNsDescription.vue
- ArrayList.vue

### Areas which could experience regressions

Forms have different modes: Create, Edit, View. these should not regress as a result of this change.

### Screenshot/Video

NA

### Checklist
- [x] The PR is linked to an issue and the linked issue has a Milestone, or no issue is needed
- [x] The PR has a Milestone 
- [x] The PR template has been filled out
- [x] The PR has been self reviewed 
- [x] The PR has a reviewer assigned
- [x] The PR has automated tests or clear instructions for manual tests and the linked issue has appropriate QA labels, or tests are not needed
- [x] The PR has reviewed with UX and tested in light and dark mode, or there are no UX changes

PR #14160, opened 2025-04-18
Fleet, implement Dashboard

### LINKED ISSUE #10339: Fleet "continous delivery dashboard" should show Clusters/Bundles/Resources in error at "mouse over"
Internal Reference: SURE-5804

In the fleet dashboard, when a Cluster,Bundle or Resource has errors it would be nice if the mouse over would show those which are in the error state.

### LINKED ISSUE #13172: Fleet Epic: Dashboard UX improvements

- We want to revisits the Fleet dashboard and improve the UX part of it.
- The Dashboard should show Fleet resources in a structured way.
- We should improve error handling and fix performance issues.

- [x] https://github.com/rancher/dashboard/issues/10339
- [x] https://github.com/rancher/dashboard/issues/10580

**Additional context**

### LINKED ISSUE #13433: Fleet: Tooltips refactoring

Some of the badges in the Dashboard columns show tooltip messages on mouse over, to explain the status of Fleet resources:

Image

For some combinations of columns/status, the tooltip is missing, e.g. Cluster Ready/Ready Status:

Image

We want to add  missing tooltips and improve the message of  the existent ones

**Additional context**

torchiaf:
### Summary
Fixes #13172
Fixes #13532
Fixes #13433
Fixes #10339

We want to re-build the Fleet dashboard. The main goal is to improve its usability based on user feedback.

### Fixing 13172 , 13433 ,  10339 :
- A main panel is visible for each Fleet workspace. Each panel should be collapsed at first access.
- Panels can be expanded by clicking on arrow button on the right side or directly by clicking of the status badge filters or related chart section.
- The view mode button on top right corner of expanded panels allows to select one of [Table,Cards] mode.
- Each panel has 3 Summary panels on top of workspace panel: [GitRepos, Clusters, ClusterGroups]. 
- Only GitRepos and Clusters have chart doughnut.
- GitRepos cards and table are visible in the workspace body panel when expanded.
- GitRepos are grouped by their Fleet status
- Fleet status are grouped in the chart panel on top, by color: [Green, Blue, Yellow, Red]
- Each color is represented by a Chart's section and a selectable badge in the Chart panel.
- It is possible to have 2 resource statuses with different labels but same colors. For example: ['Modified', 'Not Ready'] statuses are both Yellow (warning), so cards having 1 of those statuses are in different card groups in the body but can be selected by clicking on color group (Yellow badge) or Chart yellow section. See the video below.
- The [status filters, badges, modeView, expanded] features are saved by the UI in the browser storage and restored when navigating to the dashboard page. Saved settings are invalidated when updating Rancher version. See `shell/mixins/preset.js`
- It's possible to click on cards to open the slide-in details page.
- Keyboard navigation is implement for all dashboard elements and slide-in panel.

### Fixing 13532 :
- When navigating from dashboard to GitRepos, the correct workspace is automatically selected

- UX has been reviewed by me and @edenhernandez-suse during implementation phase.

### Technical notes summary

- `shell/components/fleet/dashboard/ResourceCard.vue` component depends on https://github.com/rancher/dashboard/pull/14330 . We squashed the changes from the PR into this [porting](https://github.com/rancher/dashboard/pull/14160/commits/d948b7677fb21846510786c0a78fe116cd3a739a) so that we can go ahead with the review until the original PR is merged on master branch.
- The code contains some references to the upcoming HelmOps [features](https://github.com/rancher/dashboard/pull/13886) . Those were to simplify the work of implementing the UX and testing the dashboard with HelmOps enabled. They can be left out from current review.
- More e2e tests to be implemented after the first round of review.

- Fleet dashboard page
  - Functionalities and links to the GitRepo pages. 

### Screenshot/Video

[Screencast from 2025-06-04 16-38-35.webm](https://github.com/user-attachments/assets/462a1a3c-b904-4649-9383-e62735d6b13e)

### Checklist
- [x] The PR has a Milestone 


package.json
```diff
@@ -120,6 +121,7 @@
     "ufo": "0.7.11",
     "unfetch": "4.2.0",
     "url-parse": "1.5.10",
+    "vue-chart-3": "3.1.8",
```
rak-phillip, line 124: note: `vue-chart-3` appears to be in maintenance mode[^1]. I think that we should reconsider adding it as a dependency to Dashboard:

> I'm joining [vue-chartjs](https://github.com/apertureless/vue-chartjs) as a core maintainer so this package will be kept just for those who already use it. I'm encouraging new users to use vue-chartjs instead

[^1]: https://github.com/victorgarciaesgi/vue-chart-3?tab=readme-ov-file#im-joining-vue-chartjs-as-a-core-maintainer-so-this-package-will-be-kept-just-for-those-who-already-use-it-im-encouraging-new-users-to-use-vue-chartjs-instead

torchiaf: `vue-chart-3` is a temporary solution, we need to use a canvas + chart.js directly or enhance our Doughnut chart to support multiple colors: `shell/components/graph/Circle.vue` in the final version.

shell/components/SlideInPanelManager.vue
```diff
@@ -23,8 +28,44 @@ const panelTop = computed(() => {
 const panelHeight = computed(() => `calc(100vh - ${ panelTop?.value })`);
 const panelWidth = computed(() => currentProps?.value?.width || '33%');
 const panelRight = computed(() => (isOpen?.value ? '0' : `-${ panelWidth?.value }`));
+const panelZIndex = computed(() => `${ (isOpen?.value ? 1 : 2) * (currentProps?.value?.zIndex ?? 1000) }`);
 
-const panelTitle = computed(() => currentProps?.value?.title || 'Details');
+const showHeader = computed(() => currentProps?.value?.showHeader ?? true);
+const panelTitle = showHeader.value ? computed(() => currentProps?.value?.title || 'Details') : null;
+
+watch(
```
aalves08, line 36: I gave this a test and works correctly. Might need some tweaking in light of my previous comment where I advise to keep the focus trap always on (no flag needed).

On `PromptModal` we trigger automatically the focus trap on `AppModal` (always true) and I pass a the "variable to watch" that is dependent on wether the component is rendered or not https://github.com/rancher/dashboard/blob/master/shell/components/PromptModal.vue#L125

I think it might work for you here as well

aalves08 (commented):
Left a few comments.

Let me know if you need a hand tweaking the focus trap to always be "on" with the slide in manager.

Also, I would advise to create a unit like https://github.com/rancher/dashboard/blob/master/shell/components/__tests__/PromptModal.test.ts to assert that this and future slide in panels using this component will work just fine

richard-cox (commented):
@torchiaf can you sync with @oboc-sts and @momesgin?

The cards used in the fleet dashboard should be the new generic item card that's Mo's going to create to support the new helm chart view

shell/components/fleet/dashboard/ResourceDetails.vue
```diff
@@ -0,0 +1,178 @@
...
+          class="ml-5 state-icon"
+          :class="statePanel.icon"
+          :style="{ color: statePanel.color }"
+        />
+      </div>
+      <i
+        class="icon icon-close"
+        data-testid="slide-in-close"
+        :aria-label="'slide-in-close'"
+        :trigger-focus-trap="true"
```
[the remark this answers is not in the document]
torchiaf: This is a typo, I'll remove it.

shell/components/SlideInPanelManager.vue
torchiaf: Yeah, the watcher is a good idea. I'll remove the `triggerFocusTrap` as well so that it will be always enabled for the slide-in. Thanks

shell/components/fleet/dashboard/ResourceCard.vue
```diff
@@ -0,0 +1,175 @@
...
+    :header="{
+      title: { text: value.nameDisplay },
+      image: {},
+      statuses,
+    }"
+    :content="{}"
+    :value="value"
+    @click="select"
+    @keydown.enter.stop.prevent="select"
+    @keydown.space.stop.prevent="$router.push(value.detailLocation)"
```
rak-phillip: It's mot possible to activate the dropdown menu for each card with the keyboard. We need to add the `self` modifier so that child items can still receive keyboard input. 

```suggestion
    @keydown.self.enter.stop.prevent="select"
    @keydown.self.space.stop.prevent="$router.push(value.detailLocation)"
```

shell/pages/c/_cluster/fleet/index.vue
```diff
@@ -284,223 +425,510 @@ export default {
...
               />
-            </template>
-
-            <template #cell:target="{row}">
-              {{ row.targetInfo.modeDisplay }}
-            </template>
-          </ResourceTable>
-        </template>
-      </CollapsibleCard>
+            </div>
```
rak-phillip: Is there anything that would prevent us from using the button component? We can get keyboard behaviors and accessibility options out of the box without the needing to manually add all of this to the icon:

```js
<RcButton 
  :aria-label="t(`${ isWorkspaceCollapsed[workspace.id] ? 'expand': 'collapse' }-${ workspace.id }`)"
  @click="toggleCard(workspace.id)"
>
  <i
    :class="{
      ['icon icon-lg icon-chevron-right']: isWorkspaceCollapsed[workspace.id],
      ['icon icon-lg icon-chevron-down']: !isWorkspaceCollapsed[workspace.id],
    }"
  />
</RcButton>
```

shell/pages/c/_cluster/fleet/index.vue
```diff
@@ -284,223 +425,510 @@ export default {
...
+                role="button"
+                tabindex="0"
+                :aria-label="t(`${ isWorkspaceCollapsed[workspace.id] ? 'expand': 'collapse' }-${ workspace.id }`)"
+                :class="{
+                  ['icon icon-lg icon-chevron-right']: isWorkspaceCollapsed[workspace.id],
+                  ['icon icon-lg icon-chevron-down']: !isWorkspaceCollapsed[workspace.id],
+                }"
+                @click="toggleCard(workspace.id)"
+                @keydown.space.enter.stop.prevent="toggleCard(workspace.id)"
               />
```
rak-phillip: The focus ring has an extra white border in dark mode:

image

shell/components/fleet/dashboard/ResourcePanel.vue
```diff
@@ -0,0 +1,357 @@
...
+    <div class="details">
+      <div
+        class="description"
+        :data-testid="'description'"
+      >
+        <span class="count">{{ count }}</span>
+        <span class="label">{{ typeLabel }}</span>
+      </div>
+      <div class="states">
+        <BadgeState
```
rak-phillip, line 259: The focus ring for the badges has a white border in dark mode

image

shell/pages/c/_cluster/fleet/index.vue
```diff
@@ -284,223 +425,510 @@ export default {
...
+            >
+              <div
+                v-if="cardResources[workspace.id][state.stateDisplay]?.length"
+                class="card-panel"
+              >
+                <div
+                  role="button"
+                  tabindex="0"
+                  class="title"
+                  :aria-label="t(`${ isStateCollapsed[workspace.id]?.[state.stateDisplay] ? 'expand': 'collapse' }-${ state.stateDisplay }`)"
```
rak-phillip: This aria-label evaluates to `aria-label="%collapse-Error%"` for the first error under fleet-default

image

The translation key is either wrong or missing.

shell/pages/c/_cluster/fleet/index.vue
```diff
@@ -284,223 +425,510 @@ export default {
...
+            <div
+              v-for="(state, j) in applicationStates[workspace.id]"
+              :key="j"
+              :data-testid="`state-panel-${ state.stateDisplay }`"
+            >
+              <div
+                v-if="cardResources[workspace.id][state.stateDisplay]?.length"
+                class="card-panel"
+              >
+                <div
```
rak-phillip, line 610: The focus ring has a white border in dark mode

image

shell/pages/c/_cluster/fleet/index.vue
```diff
@@ -284,223 +425,510 @@ export default {
...
+                      <ResourceCard
+                        v-if="y < (cardsCount[workspace.id]?.[state.stateDisplay] || CARDS_MIN)"
+                        role="button"
+                        tabindex="0"
+                        :aria-label="`aria-label-card-${ item.id }`"
+                        :data-testid="`resource-card-${ item.id }`"
+                        :value="item"
+                        :state-panel="state.statePanel"
+                        @click="showResourceDetails(item, state.statePanel, workspace, `${ item.id }-${ y }`)"
+                      />
```
rak-phillip: The focus ring has a white border in dark mode

image

shell/pages/c/_cluster/fleet/index.vue
```diff
@@ -284,223 +425,510 @@ export default {
...
+                      :class="{
+                        ['selected']: selectedCard === `${ item.id }-${ y }` && isOpenSlideInPanel
+                      }"
+                      :data-testid="`card-${ item.id }`"
+                    >
+                      <ResourceCard
+                        v-if="y < (cardsCount[workspace.id]?.[state.stateDisplay] || CARDS_MIN)"
+                        role="button"
+                        tabindex="0"
+                        :aria-label="`aria-label-card-${ item.id }`"
```
rak-phillip: This needs a descriptive label that's easily relayed via assistive technologies.

shell/components/cards/RcItemCard.vue
```diff
@@ -0,0 +1,391 @@
...
+                  v-clean-tooltip="labelText(status.tooltip)"
+                  :class="['icon', status.icon, status.color]"
+                  :style="{color: status.customColor}"
+                  :data-testid="`item-card-header-status-${i}`"
+                />
+              </div>
+            </div>
+
+            <template v-if="$slots['item-card-actions']">
+              <div class="item-card-header-action-menu no-card-click">
```
rak-phillip: What is the `no-card-click` class for?

rak-phillip: I see, it's to support some functionality for the Card component in #14330 and is unrelated to this PR.

rak-phillip (commented):
This is quite a large PR, so I will need to split my review into chunks. I left notes on based on my experience while testing functionality. I'll follow-up with an additional review that focuses on reviewing implementation.

shell/pages/c/_cluster/fleet/index.vue
torchiaf: Replacing the icon with RcButton fixes the issue :thumbsup:

shell/components/fleet/dashboard/ResourcePanel.vue
torchiaf: We need to involve UX for this, for now I'm changing to default `primary` color

shell/pages/c/_cluster/fleet/index.vue
torchiaf: I'm removing the translation and the expand/collpase chunk, they are not required.

shell/components/cards/RcItemCard.vue
torchiaf: Yeah, this will be not part of this PR after rebase

torchiaf:
> This is quite a large PR, so I will need to split my review into chunks. I left notes on based on my experience while testing functionality. I'll follow-up with an additional review that focuses on reviewing implementation.

I've fixed the accessibility issues (focus color and aria-labels) and replaced icon buttons with RcButton component.

torchiaf:
I pushed new commits, 74de9f5b2a2a0d1e3895ee5df51d770ae470b006 and 6bf4466029b292f0da89a489f42a71948995823a to fix potential issues.

shell/components/fleet/FleetRepos.vue
```diff
@@ -95,6 +108,10 @@ export default {
         AGE
       ];
     },
+
+    _showIntro() {
```
rak-phillip: nit: I think that the underscore used in this function name is to avoid conflict with the `showIntro` prop. Consider if naming this `shouldShowIntro` or something similar will help better communicate the intent of this computed prop.

shell/components/fleet/FleetResources.vue
```diff
@@ -49,50 +59,67 @@ export default {
...
+    <template
+      v-for="(_, slot) of $slots"
+      v-slot:[slot]="scope"
+      :key="slot"
+    >
+      <slot
+        :name="slot"
+        v-bind="scope"
+      />
+    </template>
```
rak-phillip, line 124: nit: I know that we do this in several areas throughout Dashboard, but consider adding a comment that this is used to pass down templates provided by the caller. I think that this is helpful for future developers who might encounter this snippet.

shell/components/fleet/FleetSummary.vue
```diff
@@ -67,6 +26,13 @@ export default {
     },
   },
 
+  data() {
+    return {
+      bundlesDefaultStates:   FleetUtils.getBundlesDefaultState(this.$store.getters['i18n/withFallback'], this.stateKey),
+      resourcesDefaultStates: FleetUtils.getResourcesDefaultState(this.$store.getters['i18n/withFallback'], this.stateKey),
+    };
+  },
```
rak-phillip: This data prop contains some potentially complex logic. Consider what might happen in the event that calls to `getBundlesDefaultState()` or `getResourcesDefaultState()` fails. See #11807, #13669, & #13703 for more details and methods for avoiding these failures. 

If it's absolutely necessary to keep this logic in the data prop, consider wrapping this logic in a `try..catch` block and specifying defaults in the event of a failure.

shell/components/fleet/dashboard/Empty.vue
```diff
@@ -0,0 +1,77 @@
...
+    };
+  }
+};
+</script>
+<template>
+  <div
+    class="fleet-empty-dashboard"
+    :data-testid="'fleet-empty-dashboard'"
+  >
+    <i class="icon-fleet mb-30" />
```
rak-phillip, line 34: I think that we can add `aria-hidden="true"` to icons so that they won't be picked up be screen readers. 

FYI @aalves08

shell/components/fleet/dashboard/Empty.vue
```diff
@@ -0,0 +1,77 @@
...
+    <h1>{{ t('fleet.dashboard.welcome') }}</h1>
+    <p class="mb-30">
+      <span>{{ t('fleet.dashboard.gitOpsScale') }}</span>
+      <a
+        :href="t('fleet.dashboard.learnMoreLink')"
+        target="_blank"
+        rel="noopener noreferrer nofollow"
+      >
+        {{ t('fleet.dashboard.learnMore') }} <i class="icon icon-external-link" />
+      </a>
```
rak-phillip, line 44: I think that we should add some hidden text so that screen readers can announce that the link will open in a new tab. Otherwise, opening a new tab can be disorienting to users who rely on assistive tech. 

We can do something like this:

```html
  <span class="sr-only">{{ t('generic.opensInNewTab') }}</span>
```

FYI @aalves08

shell/components/fleet/dashboard/ResourceCard.vue
```diff
@@ -0,0 +1,175 @@
+<script>
```
rak-phillip: I think that we should be using typescript for new components at this point.

shell/components/fleet/dashboard/ResourceCard.vue
```diff
@@ -0,0 +1,175 @@
...
+    margin: 1px;
+
+    :deep() .item-card-body {
+      width: -webkit-fill-available;
+    }
+  }
+
+  :deep() .item-card-header-title {
+    margin-left: 8px;
+  }
```
rak-phillip: nit: Ideally, we should only require deep selectors in unsupported edge cases. Consider syncing with @oboc-sts & @momesgin to see if there are any enhancements that can be made at this point to support all our card use-cases.

shell/components/fleet/dashboard/ResourceCardSummary.vue
```diff
@@ -0,0 +1,189 @@
...
+        v-else
+        class="count"
+      >
+        <div class="label">
+          <span class="large">{{ summary.partial }}</span>
+          <span
+            v-if="summary.partial !== summary.total && summary.total !== 0"
+            class="label-secondary"
+          >/{{ summary.total }}</span>
+          &nbsp;
```
rak-phillip: nit: can we use CSS to provide a gap instead of using `&nbsp`?

shell/components/fleet/dashboard/ResourceDetails.vue
```diff
@@ -0,0 +1,182 @@
...
+          {{ value.id }}
+        </router-link>
+        <i
+          v-if="statePanel.id !== 'success'"
+          class="ml-5 state-icon"
+          :class="statePanel.icon"
+          :style="{ color: statePanel.color }"
+        />
+      </div>
+      <i
```
rak-phillip: Consider if a button works better than a clickable icon. Using a button would cut down on the amount of accessibility concerns in this area.

shell/components/fleet/dashboard/ResourcePanel.vue
```diff
@@ -0,0 +1,365 @@
...
+      type:    Object,
+      default: () => ({})
+    },
+  },
+
+  data() {
+    return {
+      chartId:      `${ this.workspace }-${ this.type }`,
+      chart:        null,
+      chartOptions: {
```
rak-phillip: Consider moving `chartOptions` outside of the data prop. Based on the usage in the `mounted()` hook, it doesn't look like this needs to be made reactive at all. The addition of methods in the data is also a pattern that we want to avoid.

rak-phillip (commented):
This is my second round of comments. I have reviewed up through `shell/components/fleet/dashboard/ResourcePanel.vue` 🙂

shell/components/fleet/dashboard/ResourceCard.vue
torchiaf: Fixed by adding the margin to the icon

shell/components/fleet/dashboard/ResourcePanel.vue
torchiaf: Moved in mounted()

torchiaf:
Thanks @rak-phillip I pushed new commits and left some comments.

- 26cc81ca926e6e3ac9bd0014f855606f0947914b adds typescript in new Fleet components.
- 147ab3eec4d8a801388865617de8cbc3d2fa75f4 fixes remaining issues from second review round.

shell/components/fleet/FleetResources.vue
cnotv: Comments in general are always welcomed :D

shell/components/fleet/FleetSummary.vue
cnotv: The util itself should contain the catch tbh, as it should be a pure function and always return the same value, which in case of error is an empty list. This is also a matter of clean code, so if you use the utils you do not have to think about other issues.

shell/components/fleet/dashboard/Empty.vue
cnotv: Adding aria as reason for this component https://github.com/rancher/dashboard/issues/7588

shell/components/fleet/dashboard/ResourceCard.vue
cnotv: I would say that `deep` should be used exclusively in case of vendors and even in such case it should be a wrapper component.

pkg/rancher-components/src/components/BadgeState/BadgeState.vue
```diff
@@ -64,7 +64,9 @@ export default defineComponent({
       v-if="icon"
       class="icon"
       :class="{[icon]: true, 'mr-5': !!msg}"
-    />{{ msg }}
+    />
+    <span class="msg">{{ msg }}</span>
```
cnotv, line 68: What is class `msg` for? Also what should be the use case of the new slot? I do not see anything in the video

shell/assets/styles/themes/_dark.scss
```diff
@@ -226,7 +228,29 @@
   --badge-state-disabled-bg    : #{$medium};
   --badge-state-disabled-border: #{$lighter};
 
+  --click-badge-error          : #{$lightest};
```
cnotv, line 231: is this element now an interactive? It was initially mentioned to be static and if used as interactive element just in case of a wrapper element: https://rancher.github.io/storybook/?path=/docs/stories-components-badgestate--docs

shell/components/SlideInPanelManager.vue
```diff
@@ -23,8 +28,55 @@ const panelTop = computed(() => {
...
+const panelZIndex = computed(() => `${ (isOpen?.value ? 1 : 2) * (currentProps?.value?.zIndex ?? 1000) }`);
+
+const showHeader = computed(() => currentProps?.value?.showHeader ?? true);
+const panelTitle = showHeader.value ? computed(() => currentProps?.value?.title || 'Details') : null;
+
+watch(
+  () => currentProps?.value?.triggerFocusTrap,
+  (neu) => {
+    if (neu) {
+      const opts = {
```
cnotv, line 43: Could we please add some comments about what are we doing?
Should this logic not be part of the composable? It seems rather out of scope here, plus I am sure it's a reused logic.

shell/components/SlideInPanelManager.vue
```diff
@@ -23,8 +28,55 @@ const panelTop = computed(() => {
...
+        ...DEFAULT_FOCUS_TRAP_OPTS,
+        setReturnFocus: () => {
+          const returnFocusSelector = currentProps?.value?.returnFocusSelector;
+
+          if (returnFocusSelector && !document.querySelector(returnFocusSelector)) {
+            console.warn('SlideInPanelManager: cannot find elem with "returnFocusSelector", returning focus to main view'); // eslint-disable-line no-console
+
+            return '.dashboard-root';
```
cnotv, line 54: Use of classes should not be used and rather rely on data. 
Classes are for styling and CSS not JS logic. If we are ever going to touch the markup, there will be no hint about not touching it.

shell/components/SlideInPanelManager.vue
```diff
@@ -23,8 +28,55 @@ const panelTop = computed(() => {
...
+  () => currentProps?.value?.triggerFocusTrap,
+  (neu) => {
+    if (neu) {
+      const opts = {
+        ...DEFAULT_FOCUS_TRAP_OPTS,
+        setReturnFocus: () => {
+          const returnFocusSelector = currentProps?.value?.returnFocusSelector;
+
+          if (returnFocusSelector && !document.querySelector(returnFocusSelector)) {
+            console.warn('SlideInPanelManager: cannot find elem with "returnFocusSelector", returning focus to main view'); // eslint-disable-line no-console
```
cnotv, line 52: Is this something we should inform the users about?

shell/components/__tests__/RcItemCard.test.ts
```diff
@@ -0,0 +1,153 @@
+import { mount } from '@vue/test-utils';
+import RcItemCard from '@shell/components/cards/RcItemCard.vue';
+
+const id = 'test';
+
+const baseProps = {
```
cnotv: Keep in mind these props could be mutated and make the test unreliable.

shell/components/cards/RcItemCard.vue
```diff
@@ -0,0 +1,391 @@
+<script setup lang="ts">
```
cnotv: Could we kindly create a Story for each new component? It would be very helpful to debug, highlight features and maintain it.
I may help you if it takes more than 30 minutes.

shell/components/fleet/dashboard/ResourceCard.vue
```diff
@@ -0,0 +1,180 @@
...
+        id:          state.id,
+        icon:        state.icon,
+        color:       '',
+        customColor: state.color,
+        tooltip:     {},
+        handleClick: () => {},
+      }];
+    },
+
+    resourceCounts() {
```
cnotv, line 76: Maybe a little summary of which resources we will consider?

shell/components/fleet/dashboard/ResourcePanel.vue
```diff
@@ -0,0 +1,376 @@
...
+  data(): DataType {
+    return {
+      chartId: `${ this.workspace }-${ this.type }`,
+      chart:   null,
+    };
+  },
+
+  mounted() {
+    if (this.showChart) {
+      const container = document.getElementById(`${ this.chartId }-container`);
```
cnotv, line 73: Could we not use a `ref` for reactivity?

torchiaf:
@rak-phillip rebased after RcItemCard merging

shell/components/fleet/dashboard/ResourceCard.vue
```diff
@@ -0,0 +1,180 @@
...
+        return this.value.nameDisplay;
+      }
+
+      return null;
+    }
+  },
+
+  methods: {
+    select(value: PointerEvent) {
+      const elem = value?.srcElement as HTMLElement;
```
rak-phillip: `srcElement` has been marked deprecated. Let's avoid using this, if possible.

https://developer.mozilla.org/en-US/docs/Web/API/Event/srcElement

shell/components/fleet/dashboard/ResourceCardSummary.vue
```diff
@@ -0,0 +1,195 @@
...
+  },
+
+  computed: {
+    counts() {
+      return this.value.status?.resourceCounts || {};
+    },
+
+    countsPerCluster() {
+      return ((this.value.targetClusters || []) as { id: string }[]).map(({ id }) => this.value.status?.perClusterResourceCounts?.[id] || { desiredReady: 0 });
+    },
```
rak-phillip, line 34: nit: Quite a few of these computed props have a return type of `any`; it looks like the root cause can be attributed to now knowing the type of `value`.

shell/components/fleet/dashboard/ResourceDetails.vue
```diff
@@ -0,0 +1,185 @@
...
+      clusterId: ''
+    };
+  },
+
+  mounted() {
+    this.clusterId = this.clusters[0]?.value || '';
+  },
+
+  computed: {
+    clusters() {
```
rak-phillip, line 49: nit: This has a similar problem, we don't know the type of `value` so the return type is inferred as `any`.

shell/mixins/preset.js
```diff
@@ -0,0 +1,95 @@
...
+import { NORMAN } from '@shell/config/types';
+
+export default {
+  data() {
+    const principal = this.$store.getters['rancher/byId'](NORMAN.PRINCIPAL, this.$store.getters['auth/principalId']) || {};
+
+    return {
+      _init:   [],
+      presets: {},
+      user:    principal.loginName,
```
rak-phillip: Including `user` in the data prop implies that we expect this value to be mutated. Would this fit better as a computed prop?

shell/mixins/preset.js
rak-phillip, line 1: The only usage for `preset` appears to be in `shell/pages/c/_cluster/fleet/index.vue`, does this need to be a mixin?

I think it's best that we try to avoid mixin usage where it's not needed[^1]

> No Longer Recommended
>
> In Vue 2, mixins were the primary mechanism for creating reusable chunks of component logic. While mixins continue to be supported in Vue 3, [Composable functions using Composition API](https://vuejs.org/guide/reusability/composables.html) is now the preferred approach for code reuse between components.

[^1]: https://vuejs.org/api/options-composition#mixins

shell/mixins/preset.js
```diff
@@ -0,0 +1,95 @@
...
+          try {
+            const presets = JSON.stringify(this.presets);
+
+            window.localStorage.setItem(this.presetKey, presets);
+          } catch (error) {
+            console.warn(`Preset: save presets failed, invalid presets [${ this.presetKey }]`); // eslint-disable-line no-console
+          }
+        },
+        { deep: true }
+      );
```
rak-phillip, line 97: I'm unsure if this usage of `watch` inside of a method is something that we want to be doing. I expect to see watchers declared in `setup()` or `<script setup>`.

shell/pages/c/_cluster/fleet/index.vue
```diff
@@ -71,211 +95,330 @@ export default {
...
-      admissableAreas: ['clusters', 'bundles', 'resources'],
-      headers:         [
-        {
-          name:          'name',
-          labelKey:      'tableHeaders.repoName',
-          value:         'nameDisplay',
-          sort:          ['nameSort'],
-          formatter:     'LinkDetail',
-          canBeVariable: true,
+      repoSchema:  this.$store.getters['management/schemaFor'](FLEET.GIT_REPO),
```
rak-phillip: `repoSchema` is better served as a computed prop.

rak-phillip (changes requested):
Thanks for the updates. I was able to make it through all files this time around 🙂.

I have a few comments additional comments.

shell/pages/c/_cluster/fleet/index.vue
rak-phillip: This looks much better now, but I can still see the original border. I think that we need to offset the focus ring to cover up this border, similar to how labeled inputs are handled 

image

shell/components/fleet/dashboard/ResourceCardSummary.vue
torchiaf: Yes, we will need to define a type for GitRepos and use it as type for `value` prop, so that all the computed properties will automatically get types from there. I would leave this as is for now and improve typescript definitions in next releases.

shell/components/fleet/dashboard/ResourceDetails.vue
torchiaf: see above comment

shell/mixins/preset.js
torchiaf: The aim was to make this mechanism reusable in other components in the future, maybe adding the possibility to define different storage (browser storage/preferences APIs)

shell/mixins/preset.js
torchiaf: The reason for putting it in `preset()` method is that we want to watch the preset variable only when `preset` method is called in the component where we are importing the mixin.

For instance, in the Fleet dashboard, we want to 'preset' the 'typeFilter' only after getting its default value based on the value of workspaces:
```
  watch: {
    workspaces(neu) {
      if (neu) {
        neu?.forEach((ws) => {
          this.isWorkspaceCollapsed[ws.id] = neu.length > 1;

          this.isStateCollapsed[ws.id] = { Active: true };

          this.typeFilter[ws.id] = {
            [FLEET.GIT_REPO]: true,
            [FLEET.HELM_OP]:  true,
          };

          this.stateFilter[ws.id] = {};
        });

        this.preset('isWorkspaceCollapsed', 'object');
        this.preset('isStateCollapsed', 'object');
        this.preset('typeFilter', 'object');
        this.preset('stateFilter', 'object');
      }
    }
  }
  ```
 This way, the first time a user access to the dashboard, we assing a default value to presets before they are saved by `preset.js` watch.
 This would not be possible if we assign the watcher in the `preset.js` setup()

pkg/rancher-components/src/components/BadgeState/BadgeState.vue
torchiaf: `msg` is the label of the badge:

image

However, in current PR we added only the <slot name="content-right" />

shell/assets/styles/themes/_dark.scss
torchiaf: It's interactive in the dashboard. It would be better to define a wrapper for it but i would do this outside of this PR.

shell/components/SlideInPanelManager.vue
torchiaf: Done

shell/components/__tests__/RcItemCard.test.ts
torchiaf: This was part of RcItemCard feature.

shell/components/cards/RcItemCard.vue
torchiaf: This was part of RcItemCard feature.

shell/components/fleet/dashboard/ResourcePanel.vue
torchiaf: I think this should not be reactive, the charts to show in the panel are hardcoded by design (we show charts only for `GitRepos` and `Clusters` panel)

torchiaf:
I fixed the issues from last round of reviews. @richard-cox the RcItemCard has been merged and we are now use it in the dashboard.

rak-phillip (approved):
LGTM

cnotv (approved):
LGTM

PR #14172, opened 2025-04-22
Replace container logs actions with new dropdown menu component

### LINKED ISSUE #14157: a11y: Replace v-dropdown in Container Logs
Replace the `v-dropdown` implementation used in Container logs with the new dropdown menu component.

## Proposed Changes

- Replace the custom v-dropdown implementation used in container logs with the new dropdown menu component
- Ensure that all functionality currently present in the custom v-dropdown implementation is present in the replacement
- Ensure that the replacement is keyboard accessible

## Acceptance Criteria

- [ ] Custom v-dropdown is replaced with the new dropdown menu component
- [ ] Previously available actions are all represented in the new implementation
- [ ] Implementation is keyboard accessible

## Details

The current custom `v-dropdown` implementation in Container Logs is inconsistent with other dropdown menu implementations used throughout Dashboard. Using the new dropdown menu component provides a more consistent user experience.

rak-phillip:
### Summary
This replaces the custom container actions dropdown menu with the new dropdown menu component.

Fixes #14157

- Create a new dropdown menu item checkbox component
- Create a new dropdown menu item select component
- Abstract duplicated dropdown menu item logic into new composable, `useDropdownItem`
- Create a dedicated component for comtainer logs actions
- Replace container logs actions implementation with new component

### Technical notes summary

I opted to keep the design as close as possible to the existing implementation. Unfortunately, this means that we are utilizing a non-standard pattern with the Labeled Select in the menu, but it keeps the design consistent with what we had before. I attempted to create a sub-menu abstraction that I had 90% working, but I was taking too much time to finalize the components; the approach that I took in this PR was much faster to implement. 

We can also take an alternative approach to implement the range select as a group of radio buttons, this would better adhere to design patterns prescribed by the APG. I would like to sync with @edenhernandez-suse so that we can settle on a preferred approach.

Container logs

Container logs

#### Before

image

image

#### After

image

image

image

image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/pkg/rancher-components/src/components/RcDropdown/RcDropdownItemCheckbox.vue
+++ b/pkg/rancher-components/src/components/RcDropdown/RcDropdownItemCheckbox.vue
@@ -1,0 +1,68 @@
+<script setup lang="ts">
+/**
+ * An item for a dropdown menu. Used in conjunction with RcDropdown.
+ */
+import { Checkbox as RcCheckbox } from '@components/Form/Checkbox';
+import { useDropdownItem } from '@components/RcDropdown/useDropdownItem';
+
+const props = defineProps({ modelValue: Boolean, disabled: Boolean });
+const emits = defineEmits(['click']);
+
+const { handleKeydown, handleActivate } = useDropdownItem();
+
+const handleClick = () => {
+  if (props.disabled) {
+    return;
+  }
+
+  emits('click', !props.modelValue);
+};
+</script>
+
+<template>
+  <div
+    ref="dropdownMenuItem"
+    dropdown-menu-item
+    tabindex="-1"
+    role="menuitemcheckbox"
+    :disabled="disabled || null"
+    :aria-disabled="disabled || false"
+    @click.stop="handleClick"
+    @keydown.enter.space="handleActivate"
+    @keydown.up.down.prevent.stop="handleKeydown"
+  >
+    <rc-checkbox :value="modelValue">
+      <template #label>
+        <slot name="default">
+          <!--Empty slot content-->
+        </slot>
+      </template>
+    </rc-checkbox>
+  </div>
+</template>
+
+<style lang="scss" scoped>
+  [dropdown-menu-item] {
+    display: flex;
+    gap: 8px;
+    align-items: center;
+    padding: 9px 8px;
+    margin: 0 9px;
+    border-radius: 4px;
+
+    &:hover {
+      cursor: pointer;
+      background-color: var(--dropdown-hover-bg);
+    }
+    &:focus-visible, &:focus {
+      @include focus-outline;
+      outline-offset: 0;
+    }
+    &[disabled] {
+      color: var(--disabled-text);
+      &:hover {
+        cursor: not-allowed;
+      }
+    }
+  }
+</style>
--- a/shell/mixins/vue-select-overrides.js
+++ b/shell/mixins/vue-select-overrides.js
@@ -1,8 +1,5 @@
 
 export default {
-  data() {
-    return { overridesMixinPreventDoubleTriggerKeysOpen: false };
-  },
   methods: {
     mappedKeys(map, vm) {
       // Defaults found at - https://github.com/sagalbot/vue-select/blob/master/src/components/Select.vue#L947
@@ -63,7 +60,6 @@ export default {
           if (vm.closeOnSelect) {
             // this ties in to the Select component implementation
             // so that the enter key handler doesn't open the dropdown again
-            this.overridesMixinPreventDoubleTriggerKeysOpen = false;
             vm.open = false;
             vm.typeAheadPointer = -1;
           }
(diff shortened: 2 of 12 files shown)


shell/mixins/vue-select-overrides.js
```diff
@@ -63,7 +63,7 @@ export default {
           if (vm.closeOnSelect) {
             // this ties in to the Select component implementation
             // so that the enter key handler doesn't open the dropdown again
-            this.overridesMixinPreventDoubleTriggerKeysOpen = true;
+            this.overridesMixinPreventDoubleTriggerKeysOpen = false;
```
rak-phillip, line 66: @aalves08 I don't 100% understand the intent of `overridesMixinPreventDoubleTriggerKeysOpen`, do we have scenarios that demonstrate the issue that it solves? I'm finding that this variable interferes with multiple opening and closing attempts with the dropdown menu item implementation.

pkg/rancher-components/src/components/RcDropdown/RcDropdownItemCheckbox.vue
```diff
@@ -0,0 +1,138 @@
...
+        :aria-describedby="uidLabel"
+      >
+      <label
+        :id="uidLabel"
+        class="rc-checkbox-label"
+      >
+        <slot name="default">
+          <!--Empty slot content-->
+        </slot>
+      </label>
```
rak-phillip, line 54: @aalves08 We are using a custom checkbox implementation here because the existing Checkbox component is not centered properly. I attempted to update the styles to resolve this issue, but I still couldn't get it quite right. There's something about our checkbox component that I'm missing.. 

I'm open to recommendations on how we can fix the checkbox component. Otherwise, if we like this, we might have a path forward to simplify the existing checkbox implementation. 

@edenhernandez-suse

rak-phillip:
@edenhernandez-suse I would like a design review for these changes 🙂

edenhernandez-suse (approved):

shell/mixins/vue-select-overrides.js
aalves08: If you go to `preferences` and try to to interact with that first labeled select (language one), open it and try to select an option with the `enter` key. This interaction on the dropdown should select that option and close it.

Now switch the `overridesMixinPreventDoubleTriggerKeysOpen` to `false` and try performing the same action with the `enter` key. The dropdown closes and opens it right away... this is because it collides with the `keydown.enter` event in `Select` and `LabeledSelect` (`focusSearch` method).


Uploading Screen Recording 2025-04-28 at 10.15.35.mov…



At the time that was the only way I could handle this problem. I've just noticed that these events in both areas are `keydown`, so it appears that if you add `e.stopPropagation()` after that first `if` condition on `mappedKeys` for the `enter` handler (13) it seems to work. 

Double-check this in `Select` and `LabeledSelect` and, if possible, just remove this `overridesMixinPreventDoubleTriggerKeysOpen` from our codebase. I don't remember any other reason why I added this...

pkg/rancher-components/src/components/RcDropdown/RcDropdownItemCheckbox.vue
aalves08: Is it that bad @rak-phillip ? What's the visual difference here?

aalves08 (commented):
Left a couple of comments

pkg/rancher-components/src/components/RcDropdown/RcDropdownItemCheckbox.vue
rak-phillip: It's not horrible, but the existing checkbox component is decidedly off-center:

image

I'm willing to attempt another go-around at trying to fix this.. although, I'm still not sure what the issue is.

shell/mixins/vue-select-overrides.js
rak-phillip: > At the time that was the only way I could handle this problem. I've just noticed that these events in both areas are keydown, so it appears that if you add e.stopPropagation() after that first if condition on mappedKeys for the enter handler (13) it seems to work.

Indeed, I replaced this with `@keydown.enter.stop` in 2927468bc173483b5ac782bf447309e65e40ac79. What do you think?

aalves08: LGTM mate, from what I could test. Thanks! 🙏

aalves08 (dismissed):
Thanks for that fix on the `overridesMixinPreventDoubleTriggerKeysOpen` part @rak-phillip .

Tbh, I am not fond of us having a custom checkbox implementation in the `DropdownItemCheckbox`. Looking at the screenshots, I would have definitely lived with that vs introducing new code.

But I don't think it's a blocker and we can revisit it later. Would you mind creating an issue for that? I think the issue is to either replace Checkbox with your simplified version OR just use the Checkbox component we already have. Wdyt?

rak-phillip:
> Tbh, I am not fond of us having a custom checkbox implementation in the DropdownItemCheckbox. Looking at the screenshots, I would have definitely lived with that vs introducing new code.

I reverted the commit for now and will raise a separate PR to address issues in the checkbox component. This way, we can keep the conversation focused on issues with the design of that component without the need for duplication in this PR.

Ready for another review.

aalves08 (approved):
LGTM

PR #14186, opened 2025-04-24
Update AppModal, PromptModal and focusTrap

aalves08:
### Summary
Fixes #14221

- update `AppModal`, `PromptModal` and `focusTrap` to use only the watcher based focus trap and wire in required props for prompt modals
- add unit test for currently existing prompt modals display assertion
- minor tweaks to current prompt modals so that they work fine on unit tests

### Technical notes summary

I've had to create a new `PromptModal` prop called `fixUnitTestTrapOptions` so that we are able to run these unit tests without the `focusTrap` directive complaining that it cannot find a target node to focus on, as per https://github.com/focus-trap/tabbable#common-options + https://github.com/focus-trap/tabbable#testing-in-jsdom

It does not have any impact whatsoever on the `AppModal` because it's off by default. Manual tests on all modals changed by this PR reveal zero problems with the added bonus of them having a working focus trap by default.

- check `DrainNode` modal -> `node` list view table action
- check `AddCustomBadgeDialog` modal -> Cluster explorer OR cluster provisioning edit view (badge config)
- check `ScalePoolDownDialog` modal ->provisioning cluster detail view (machine pools)
- check `ForceMachineRemove` modal -> `cluster.x-k8s.io.machine` list view table action

Double-check other modals if you feel it's necessary (check promptmodal unit test)

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/AppModal.vue
+++ b/shell/components/AppModal.vue
@@ -86,32 +86,20 @@ export default defineComponent({
       default: DEFAULT_ITERABLE_NODE_SELECTOR,
     },
     /**
-     * trigger focus trap - but watcher based
-     */
-    triggerFocusTrapWatcherBased: {
-      type:    Boolean,
-      default: false,
-    },
-    /**
      * watcher-based focus trap variable to watch
      */
     focusTrapWatcherBasedVariable: {
       type:    Boolean,
       default: undefined,
     },
-    fakeWatchVar: {
+    /**
+     * prop used to immediately trigger the focus trap when a proper watch variable is not required
+     * will default to this prop IF there's no "focusTrapWatcherBasedVariable" being passed
+     * prop used only in setup (if you need a proper watcher variable, pass it via "focusTrapWatcherBasedVariable")
+     */
+    autoTriggerFocusTrapWatcher: {
       type:    Boolean,
       default: true,
-    },
-    /**
-     * property need to manipulate focus-trap lib for unit tests
-     * as per https://github.com/focus-trap/tabbable#common-options
-     * which has to be "none" in order for them to work
-     * manual test on Rancher Dashboard shows all modals work fine without this
-     */
-    fixUnitTestTrapOptions: {
-      type:    Boolean,
-      default: false,
     }
   },
   computed: {
@@ -139,13 +127,6 @@ export default defineComponent({
         width: this.modalWidth,
         ...this.stylesPropToObj,
       };
-    },
-    watcherVariableForFocusTrap(): boolean {
-      if (this.focusTrapWatcherBasedVariable !== undefined) {
-        return this.focusTrapWatcherBasedVariable;
-      }
-
-      return true;
     }
   },
   setup(props) {
@@ -170,12 +151,7 @@ export default defineComponent({
         };
       }
 
-      // needed as a workaround/fix for PromptModal unit tests
-      if (props.fixUnitTestTrapOptions) {
-        opts.tabbableOptions = { displayCheck: 'none' };
-      }
-
-      useWatcherBasedSetupFocusTrapWithDestroyIncluded(() => props.focusTrapWatcherBasedVariable !== undefined ? props.focusTrapWatcherBasedVariable : props.fakeWatchVar, '#modal-container-element', opts, true);
+      useWatcherBasedSetupFocusTrapWithDestroyIncluded(() => props.focusTrapWatcherBasedVariable ?? props.autoTriggerFocusTrapWatcher, '#modal-container-element', opts, true);
     }
   },
   mounted() {
--- a/shell/components/AppModal.vue
+++ b/shell/components/AppModal.vue
@@ -2,7 +2,6 @@
 import { defineComponent } from 'vue';
 import {
   DEFAULT_FOCUS_TRAP_OPTS,
-  useBasicSetupFocusTrap,
   getFirstFocusableElement,
   useWatcherBasedSetupFocusTrapWithDestroyIncluded
 } from '@shell/composables/focusTrap';
@@ -98,8 +97,22 @@ export default defineComponent({
      */
     focusTrapWatcherBasedVariable: {
       type:    Boolean,
-      default: false,
+      default: undefined,
     },
+    fakeWatchVar: {
+      type:    Boolean,
+      default: true,
+    },
+    /**
+     * property need to manipulate focus-trap lib for unit tests
+     * as per https://github.com/focus-trap/tabbable#common-options
+     * which has to be "none" in order for them to work
+     * manual test on Rancher Dashboard shows all modals work fine without this
+     */
+    fixUnitTestTrapOptions: {
+      type:    Boolean,
+      default: false,
+    }
   },
   computed: {
     modalWidth(): string {
@@ -127,6 +140,13 @@ export default defineComponent({
         ...this.stylesPropToObj,
       };
     },
+    watcherVariableForFocusTrap(): boolean {
+      if (this.focusTrapWatcherBasedVariable !== undefined) {
+        return this.focusTrapWatcherBasedVariable;
+      }
+
+      return true;
+    }
   },
   setup(props) {
     if (props.triggerFocusTrap) {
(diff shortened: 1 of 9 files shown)


shell/dialog/AddCustomBadgeDialog.vue
```diff
@@ -213,10 +213,10 @@ export default {
   <Card
     class="prompt-badge"
     :show-highlight-border="false"
-    :trigger-focus-trap="true"
   >
     <template #title>
       <h4 class="text-default-text">
+        ALEX
```
rak-phillip, line 219: Unfortunately, we'll need to remove the ALEX header 🙂

shell/components/__tests__/PromptModal.test.ts
```diff
@@ -0,0 +1,116 @@
...
+    ['ScalePoolDownDialog', ScalePoolDownDialog]
+  ])('prompt Modal should render modal %p', (modalName, component) => {
+    document.body.innerHTML = '<div id="modals"></div>';
+    const wrapper = mount(PromptModal,
+      {
+        attachTo: document.body,
+        data() {
+          return {
+            opened:                 true, // this controls modal content visibility
+            fixUnitTestTrapOptions: true // promptModal flag to allow unit tests to run on focus trap
```
rak-phillip, line 100: We already mock the focus trap implementation for unit tests in `shell/components/form/__tests__/SSHKnownHosts.test.ts` and `shell/components/form/SSHKnownHosts/__tests__/KnownHostsEditDialog.test.ts`. Can we also mock the focus trap in this test and remove the `fixUnitTestTrapOptions` prop? I think it's best if we avoid allowing test code to drive our component design.

shell/components/AppModal.vue
```diff
@@ -98,8 +97,22 @@ export default defineComponent({
...
-      default: false,
+      default: undefined,
+    },
+    fakeWatchVar: {
+      type:    Boolean,
+      default: true,
     },
+    /**
+     * property need to manipulate focus-trap lib for unit tests
+     * as per https://github.com/focus-trap/tabbable#common-options
```
rak-phillip, line 108: The link provided also links to documentation on how we can better mock for unit tests[^1]

> You can globally overwrite the diplayCheck property by including this file in your __mocks__ folder:
> 
> ```js
> // __mocks__/tabbable.js
> 
> const lib = jest.requireActual('tabbable');
> 
> const tabbable = {
>    ...lib,
>    tabbable: (node, options) => lib.tabbable(node, { ...options, displayCheck: 'none' }),
>    focusable: (node, options) => lib.focusable(node, { ...options, displayCheck: 'none' }),
>    isFocusable: (node, options) => lib.isFocusable(node, { ...options, displayCheck: 'none' }),
>    isTabbable: (node, options) => lib.isTabbable(node, { ...options, displayCheck: 'none' }),
> };
> 
> module.exports = tabbable;
> ```

I think that we should reconsider introducing code changes that accommodate unit tests in production code and encourage the usage of mocks for unit tests.

[^1]: https://github.com/focus-trap/tabbable#testing-in-jsdom

shell/components/AppModal.vue
```diff
@@ -98,8 +97,22 @@ export default defineComponent({
      */
     focusTrapWatcherBasedVariable: {
       type:    Boolean,
-      default: false,
+      default: undefined,
+    },
+    fakeWatchVar: {
```
rak-phillip, line 102: What is the purpose of `fakeWatchVar`? It isn't immediately clear to me how this is used. Can we rename this prop so that its purpose is more clear?

shell/components/AppModal.vue
```diff
@@ -150,19 +170,12 @@ export default defineComponent({
...
-  },
-  created() {
-    // This usecase is to cover the PromptModal scenario where it's renders a generic component inside.
-    // Due to the architecture of the PromptModal it needs to be handled with a watcher based focus trap
-    // but with a dedicated unmount hook
-    if (this.triggerFocusTrapWatcherBased) {
-      const opts:any = DEFAULT_FOCUS_TRAP_OPTS;
 
-      useWatcherBasedSetupFocusTrapWithDestroyIncluded(() => this.focusTrapWatcherBasedVariable, '#modal-container-element', opts, true);
+      useWatcherBasedSetupFocusTrapWithDestroyIncluded(() => props.focusTrapWatcherBasedVariable !== undefined ? props.focusTrapWatcherBasedVariable : props.fakeWatchVar, '#modal-container-element', opts, true);
```
rak-phillip, line 178: Can this be simplified to the following?

```suggestion
      useWatcherBasedSetupFocusTrapWithDestroyIncluded(        props.focusTrapWatcherBasedVariable ?? props.fakeWatchVar,, '#modal-container-element', opts, true);
```

In which scenarios would the `focusTrapWatcherBasedVariable` be omitted but the `fakeWatchVar` is provided?

shell/components/AppModal.vue
```diff
@@ -127,6 +140,13 @@ export default defineComponent({
         ...this.stylesPropToObj,
       };
     },
+    watcherVariableForFocusTrap(): boolean {
```
rak-phillip, line 143: This computed prop looks to be unused.

rak-phillip (changes requested):

shell/components/__tests__/PromptModal.test.ts
aalves08: Your next comment seems to be more in line of what we want (mocking the `tabbable` library).

Imho, I don't think the correct way is to mock the focus trap, because we want to test it properly, even it's quirks. Otherwise we are kind of defeating the purpose of this. I really want to make "bullet-proof" that we are actually rendering the content we want inside of a `PromptModal` at all times.

I know you dislike this prop pattern... I'll make an effort to look into mocking the `tabbable` library and see where that leads us. 🙏

shell/components/AppModal.vue
aalves08: As per previous comment, I'll have a look at this mock 🙏

shell/components/AppModal.vue
aalves08: Ah, yes... The nullish coalescing operator is a much better fit here, you're right. It does need the function though. 

That's needed because of why the `watch` works in the composition API https://vuejs.org/guide/essentials/watchers.html check the getter function usage

shell/components/AppModal.vue
aalves08: You're right. Old stuff from previous iterations. Nice catch mate 🙏

shell/dialog/AddCustomBadgeDialog.vue
aalves08: Wanted to leave my mark on the codebase 🤣 🤣 🤣

shell/components/AppModal.vue
```diff
@@ -150,19 +151,7 @@ export default defineComponent({
...
-  },
-  created() {
-    // This usecase is to cover the PromptModal scenario where it's renders a generic component inside.
-    // Due to the architecture of the PromptModal it needs to be handled with a watcher based focus trap
-    // but with a dedicated unmount hook
-    if (this.triggerFocusTrapWatcherBased) {
-      const opts:any = DEFAULT_FOCUS_TRAP_OPTS;
-
-      useWatcherBasedSetupFocusTrapWithDestroyIncluded(() => this.focusTrapWatcherBasedVariable, '#modal-container-element', opts, true);
+      useWatcherBasedSetupFocusTrapWithDestroyIncluded(() => props.focusTrapWatcherBasedVariable ?? props.autoTriggerFocusTrapWatcher, '#modal-container-element', opts, true);
```
rak-phillip, line 154: Thanks for updating this, but I think I'm still missing a detail about how these props are expected to work with each other. Here's what I see:

1. `focusTrapWatcherBasedVariable` is of type `boolean?` with a default value of `undefined`
2. `autoTriggerFocusTrapWatcher` is of type `boolean` with a default value of `true`
3. When `focusTrapWatcherBasedVariable` is `null` or `undefined`, we use the value of `autoTriggerFocusTrapWatcher` (defaults to `true`)

With the above being true, can this behavior be controlled with a single prop via `focusTrapWatcherBasedVariable ?? true`? What is the purpose of having two props in this scenario?

rak-phillip (commented):
This looks good to me. I just have one question about props usage and then we can merge.

shell/components/AppModal.vue
aalves08: The `focusTrap` composable method has been reduced to a watcher-based trigger, as per https://github.com/rancher/dashboard/blob/master/shell/composables/focusTrap.ts#L53

Afaik we can't just pass a "value" to the watcher. Only variables, hence why we can't just pass `true` https://vuejs.org/guide/essentials/watchers.html

rak-phillip (approved):
LGTM

PR #14207, opened 2025-04-25
Handle UI plugins with chart version and appVersion not matching

### LINKED ISSUE #14204: Shell: Chart version is expected to match AppVersion when determining extension upgrade availability
**Setup**
- Rancher version: 2.11
- Rancher UI Extensions: 3.0.4

**Describe the bug**
When the UI [determines availability of upgrade for the extension](https://github.com/rancher/dashboard/blob/master/shell/pages/c/_cluster/uiplugins/index.vue#L360-L362) it is comparing UIPlugin version with available charts versions. However, when the UIPlugin CR is created, [it's version is populated from chart.appVersion](https://github.com/rancher/dashboard/blob/master/shell/scripts/extension/helm/charts/ui-plugin-server/templates/cr.yaml#L10).

This can result in incorrect upgrade availability identification in case when chart version and chart appVersion don't match.

**To Reproduce**
Publish an extension with version e.g. 1.3.1 and then manually change the generated chart version to different version (e.g. 302.0.0+up1.3.1). This is for example used in downstream Edge extensions charts.

**Result**
As a result when the extension is installed, it immediately suggests an upgrade to a newer version 1.3.1 -> 3.0.2.0.0+up1.3.1. Because UIPlugin version is 1.3.1 (from chart.appVersion) but the newly available version is identified 3.0.2.0.0+up1.3.1 (from chart.version). 

**Expected Result**
The extension is not set as upgradable because UIPlugin version is the same as chart appVersion.

**Screenshots**

**Additional context**

jtomasek:
This change adds handling for UIPlugins which are installed from charts where chart version and chart appVersion don't match.

- Determine plugin upgradability by comparing UIPlugin version with chart appVersion with a fallback to chart version if appVersion is not available
- Update PluginInfoPanel to handle plugin version not matching chart version by checking appVersion first
- Update InstallDialog to handle plugin version not matching chart version by checking appVersion first

I've tested the PR with KubeVirt dashboard extension from the following Helm chart repository:
Type: git,
URL: https://github.com/jtomasek/dashboard-extensions,
Branch: ghp-test

### Summary
Fixes #14204

### Checklist
- [ ] The PR has a Milestone 


aalves08:
@jtomasek this PR needs a rebase from `master` since the failing e2e test has been fixed there 🙏

shell/pages/c/_cluster/uiplugins/InstallDialog.vue
```diff
@@ -82,25 +82,31 @@ export default {
   },
 
   methods: {
+    getChartVersion(pluginVersion) {
```
aalves08, line 85: can we add this function to https://github.com/rancher/dashboard/blob/master/shell/utils/uiplugins.ts and update it's usage on this PR?

aalves08 (changes requested):
Left a specific comment @jtomasek 

Also, could you be clear on how I can reproduce this issue with the repo+branch you've mentioned on the PR? Where in the published extensions did you make the manual change?

shell/pages/c/_cluster/uiplugins/InstallDialog.vue
jtomasek: Done

jtomasek:
> Also, could you be clear on how I can reproduce this issue with the repo+branch you've mentioned on the PR? Where in the published extensions did you make the manual change?

This is the change I did on the repository to reproduce the problem: https://github.com/jtomasek/dashboard-extensions/commit/ab1829d92c242f0992667808220bcc7549195795

I am setting the chart version to be different from appVersion. Which matches the convention Edge team uses for releasing charts. In such case, when an extension is installed the UIPlugin version is populated form appVersion and when UI runs the logic to determine if an upgrade is available it finds a chart version which does not match the UIPlugin version and suggests an upgrade. This PR fixes the problem by checking the UIPlugin version against Chart appVersion (or Chart version if appVersion is not set).

aalves08:
@jtomasek I ran your branch and the `update` disappears, which I think it's fine to fit your usecase. Let's wait for @nwmac feedback as well.
Unless we manually change the `appVersion`, I think `appVersion` is always equal to `version`...

With this logic change you implemented, shouldn't the card also read `999.0.0+up1.3.1`? There's a difference between what's displayed in the slidein panel when comparing with the extension card.

Screenshot 2025-04-29 at 09 32 58

jtomasek:
> With this logic change you implemented, shouldn't the card also read `999.0.0+up1.3.1`? There's a difference between what's displayed in the slidein panel when comparing with the extension card.

Yeah, so if the extension is not installed, we are showing helm chart versions in both extension box and extension panel. If the extension is installed, the UIPlugin exists and in that case in the extension card we are showing UIPlugin version which equals Helm chart appVersion. In the extension panel we are always showing available Helm chart versions. That is why we are getting a mismatch (visible in your screenshot). Should we always show Helm chart appVersion (if present) in the extension box and extension panel?

aalves08:
> > With this logic change you implemented, shouldn't the card also read `999.0.0+up1.3.1`? There's a difference between what's displayed in the slidein panel when comparing with the extension card.
> 
> Yeah, so if the extension is not installed, we are showing helm chart versions in both extension box and extension panel. If the extension is installed, the UIPlugin exists and in that case in the extension card we are showing [UIPlugin version which equals Helm chart appVersion](https://github.com/rancher/dashboard/blob/master/shell/scripts/extension/helm/charts/ui-plugin-server/templates/cr.yaml#L10). In the extension panel we are always showing available Helm chart versions. That is why we are getting a mismatch (visible in your screenshot). Should we always show Helm chart appVersion (if present) in the extension box and extension panel?

I think that the versions must match in all places where we convey that information to the user. If you look at the card when it's `available` for install, it says that the most up-to-date version to be installed is `999.0.0+up1.3.1`:

Screenshot 2025-04-29 at 14 32 06

If you look at the card in the `installed` state, it's reading `1.3.1`, as per:

Screenshot 2025-04-29 at 09 32 58

These are all taken from running your branch locally

jtomasek:
I agree, the question I am raising is whether to show an appVersion (the actual version of the extension, in my branch equal to `1.3.1`) or a helm chart version (the version of the helm chart which installs the extension, in my branch equal to `999.0.0+up1.3.1`).

My take is that it makes much more sense to show an extension version (= appVersion) because that is the information which the user is interested in when deciding to install the extension.

aalves08:
> I agree, the question I am raising is whether to show an appVersion (the actual version of the extension, in my branch equal to `1.3.1`) or a helm chart version (the version of the helm chart which installs the extension, in my branch equal to `999.0.0+up1.3.1`).
> 
> My take is that it makes much more sense to show an extension version (= appVersion) because that is the information which the user is interested in when deciding to install the extension.

I would say just go for the same approach: `appVersion` falling back to `version`. 98% of the usecases they'll both be the same, so no harm no foul. It fits your usecase, so everything evens out in the end

jtomasek:
Screenshot 2025-07-04 at 17 10 16

aalves08 (dismissed):
LGTM

aalves08:
@jtomasek could you add some tests for this, please?

FYI @richard-cox

PR #14209, opened 2025-04-28
App card component

### LINKED ISSUE #14198: [Apps/Marketplace] - Create New Card Component for Apps
[Confluence page](https://confluence.suse.com/display/CU/Charts+catalogue)

### LINKED ISSUE #14244: [Apps & Marketplace] - Create App Card Component
As shown in the screenshot the action menu(will be used on Tools and Extensions pages) and the last updated status(e.g. 5 Days Ago) are out of scope for this task.

**Screenshot**

Image

momesgin:
### Summary
Fixes #14244 

### Checklist
- [ ] The PR has a Milestone 


momesgin:
@oboc-sts @richard-cox some updates/questions:

1. Except for the "Deprecated" status, I couldn't find specific icon names. I temporarily used svg tags from Figma directly, but ideally, we should update our icon library to avoid having noisy code.
2. Since I couldn't find any explicit dark mode designs, I made some decisions myself, for example, around the logos I added a white background(like the old design) to ensure proper contrast.
3. Some badges like "Experimental", "Linux Only", etc are not covered in the new design, I was wondering if it's a good idea to display them as a tag or a category?
4. For now, the latest available version is displayed, until we figure out that we can have the info for the installed version reliably
5. I removed the colored circles next to the repo names in the repos dropdown

oboc-sts:
Hey Mo. Replying over mobile - from my side:

1. Yep, we should update the icon lib - I know I was looking to reuse some
icons, but they are not in the lib OR consistent between them;

2. I didn't use new colors / elements so the dark more should reflect the
current alternatives wr're using. Use your best judgement and we'll tweak
it a bit after I get back if that's the case;

3. In the design, they are tags (not categories) - look at the filter and
the different cards;

4. Oky;

5. I can't 'visualize' the repos dropdown right now, but I'm pretty sure we
don't use one anymore in the new design...

Aaaand that's about it, hope it helps :) I'm on the road again today.

momesgin:
@richard-cox I think we might be able rely on the chart's `catalog.cattle.io/ui-source-repo` annotation (which holds the repo name the chart was installed from) to display the installed and upgradable statuses. This annotation is always added when the app is installed through the UI, and since the Apps page only shows apps from the user's repositories, it should be reliable.

There's one problem though, if a user installs an app, then removes and later re-adds the repository under a different name, we won't be able to match the app to the repo anymore. In that case, users would need to reinstall the app for it to be recognized correctly.

richard-cox:
@momesgin there's a set of functions that link an app to chart and version (see shell/models/catalog.cattle.io.app.js). those need to be flipped to link chart to app. We need to take into account other parts from that as well, like not offering upgrade on charts that link to fleet owned apps. 

We touched on the limitations of those in the previous call linking charts to apps, it's a best effort and not applicable in a lot of scenarios.

momesgin:
> @momesgin there's a set of functions that link an app to chart and version (see shell/models/catalog.cattle.io.app.js). those need to be flipped to link chart to app. We need to take into account other parts from that as well, like not offering upgrade on charts that link to fleet owned apps.
> 
> We touched on the limitations of those in the previous call linking charts to apps, it's a best effort and not applicable in a lot of scenarios.

Initially I wanted to use those functions like `matchingChart` but I thought if your app has the annotation that I mentioned, you wouldn't need to be worried about some criteria like the fleet stuff, since I've never seen fleet apps being included in repos, but I'm probably wrong, I'll look further into that

shell/components/cards/AppCard.vue
```diff
@@ -0,0 +1,499 @@
+<script setup>
```
richard-cox, line 1: Think the idea was for this to be a `rancher-component`.


In terms of name we were going to go with three generic cards that serve three purposes. Check with @oboc-sts about those purposes and come up with a nice name for this one. Feel free to discuss here or in slack if it helps.

shell/components/cards/AppCard.vue
```diff
@@ -0,0 +1,499 @@
...
+  },
+  logo: {
+    type:    String,
+    default: ''
+  },
+  logoAltText: {
+    type:    String,
+    default: 'logo'
+  },
+  featured: {
```
richard-cox, line 23: In order to make this generic we should remove hard coded references to helm app specific things like feature, deprecated, etc. The card should then accept configuration to populate those areas instead. Maybe an array for each that's a collection of 'things' like pills, links, etc. It's probably worth defining an object that caters for the known possibilities. For example

- id - unique id that we could use for testid
- label: { key - l10n key, text - text instead of key }
- link: { href, route}
- pill: { color: aaa } 

Consider the above from a ui extension developer's perspective and what structure would be easiest for them to understand / easily implement.

shell/components/cards/AppCard.vue
```diff
@@ -0,0 +1,499 @@
...
+  if (e.target.closest('.no-card-click')) {
+    return;
+  }
+
+  emit('card-click');
+}
+
+</script>
+
+<template>
```
richard-cox, line 72: once nearer to completion double check with @aalves08 about what updates this should have to be accessible (arias, tab order, etc)

richard-cox (changes requested):

momesgin:
closing this PR in favor of [this PR](https://github.com/rancher/dashboard/pull/14330) for more clarity.

PR #14214, opened 2025-04-29
Update `namespaceFilter` accessibility features

### LINKED ISSUE #12779: a11y: Fix keyboard navigation, aria-label, role and focus display issues on the namespace chooser control
a11y Issue: 34, 24, 25, 27, 29, 32, 42

- [x] **Issue 34**: The clear button on the namespace picker does not show keyboard focus and does not response to enter/space to activate.

- [x] **Issue 25**: Need to add `role-button` and `aria-label` to the clear button.
- [x] **Issue 29**: Role is not defined for the interactive element, making it difficult for screen reader users to access the functionality associated with it.

image

- [x] **Issue 32**: The form field do not have the label associated with them resulting in screen reader users fail to understand the purpose of the form field.

- [x] **Issue 24**: - Only User Namespaces input field - Unclear functionality of input field -  As soon as user starts entering data in mentioned input field, the content is automatically updated below resulting in screen reader users unable to understand the dynamic update happening on the page.

- [x] **Issue 27**: Only User Namespaces input field - while filtering namespaces, "no matching options" Status message not defined programmatically - On typing the data in the input field the status message gets dynamically added below resulting in screen reader users not aware about the message being displayed on the page.

Screen reader should announce the dynamically added status message for their users.

Use ARIA live region to make sure that the dynamic update is announced for screen reader users. This can be achieved by either adding a live region role="status" or  aria-live="polite" and aria-atomic="true" attributes to the container used to display the message. Make sure the live region is present in the page source by default.

- [x] **Issue 42**: Namespace filter in header - interactive element receive the keyboard focus resulting in keyboard-only and screen reader users will have to navigate through the expanded content even when not required. The mentioned content must be available only when the Namespace filter interactive element is activated. 

Filter box:
image

rak-phillip:
### Summary
This updates the namespace filter to make is more keyboard accessible. This also updates aria attributes so that different elements and behaviors of the namespace filter can be better detected by screen readers.

Fixes #12779

- Fix roles for namespace filter options
- Replace namespace filter icons with buttons
- Fix RcButton classes
- Update namespace filter roles and keyboard behavior
- Update aria attributes

Most notably, the keyboard behavior has been updated so that all elements are accessible via keyboard:

- Namespace filter is focused
   - Down/Enter opens the namespace filter and focuses on the filter input
- Namespace filter is opened
  - Tab will focus the next element
  - Down will enter the list of options
  - Tabbing away from final element in the dropdown will close the dropdown and focus the next element in tab order
  - Escape will close the dropdown and return focus to the namespace filter
- Namespace filter items have focus
  - Keyboard up/down will focus the next/previous element
  - Focus will loop to top/bottom when moving past last/first element
  - Tab will close the dropdown and focus next element in tab order
  - Escape will close the dropdown and return focus to the namespace filter
  - Shift+Tab will focus previous element in dropdown

### Technical notes summary

I explored potential options like using `vue-select` or `RcDropdown` to perform a larger refactor and get better out of the box accessibility support. Each approach had major downsides that would either alter the design of the namespace filter component greatly or require a great deal of customization to retain the current design and behavior. Because of this, I opted to apply fixes to the existing component. 

It appears that this namespace filter most closely follows the combobox pattern[^1], but not entirely. I did my best to apply the guidelines so that the namespace filter follows the correct patterns.

[^1]: https://www.w3.org/WAI/ARIA/apg/patterns/combobox/

- Namespace filter

- Namespace filter

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/pkg/rancher-components/src/components/RcButton/RcButton.vue
+++ b/pkg/rancher-components/src/components/RcButton/RcButton.vue
@@ -51,7 +51,7 @@ defineExpose({ focus });
   <button
     ref="RcFocusTarget"
     role="button"
-    :class="{ ...buttonClass, ...($attrs.class || { }) }"
+    :class="{ ...buttonClass }"
   >
     <slot name="before">
       <!-- Empty Content -->
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -3945,7 +3945,7 @@ namespaceFilter:
       other {{total} items selected}
       }
   input: Filter namespace options
-  buttons:
+  button:
     clear: Remove applied namespace filters
     clearFilter: Clear namespace options filter 
 
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -3944,6 +3944,10 @@ namespaceFilter:
       one {1 item selected}
       other {{total} items selected}
       }
+  input: Filter namespace options
+  buttons:
+    clear: Remove applied namespace filters
+    clearFilter: Clear namespace options filter 
 
 namespaceList:
   selectLabel: Namespace
--- a/shell/components/nav/NamespaceFilter.vue
+++ b/shell/components/nav/NamespaceFilter.vue
@@ -800,12 +800,17 @@ export default {
       data-testid="namespaces-menu"
     >
       <div class="ns-controls">
-        <div class="ns-input">
+        <div
+          class="ns-input"
+          role="status"
+          aria-live="polite"
+        >
           <input
             ref="filterInput"
             v-model="filter"
             tabindex="0"
             class="ns-filter-input"
+            :aria-label="t('namespaceFilter.input')"
             @click="focusFilter"
             @keydown="inputKeyHandler($event)"
           >
@@ -814,6 +819,7 @@ export default {
             small
             ghost
             class="ns-filter-clear"
+            :aria-label="t('namespaceFilter.button.clearFilter')"
             @click="clearFilter"
             @keydown.enter.stop="clearFilter"
           >
@@ -836,6 +842,7 @@ export default {
           small
           ghost
           class="ns-clear"
+          :aria-label="t('namespaceFilter.button.clear')"
           @click="clear()"
           @keydown.enter.stop="clear()"
         >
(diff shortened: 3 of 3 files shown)


aalves08 (changes requested):
Gave this a review and test and here's some of the stuff I've picked up:

- can’t open dropdown with `space` key
- Imo, we should focus the whole bottom “group” where we can select multiple options… I think that clicking on tab should focus immediately on the first item of that bottom group (check video)
- can’t keyboard nav into “chips” to dismiss applied selections (check screenshot)
- clear filter input box button as a problem with the translation on `aria-label` (check screenshot)
missing `aria-expanded` on on item with role combobox
missing `aria-activedescendant`
missing `aria-selected`

https://www.w3.org/WAI/ARIA/apg/patterns/combobox/

**"tabbing into group"**

https://github.com/user-attachments/assets/76a848bd-0771-4271-9abf-94c2a74f5d7e

**"chips"**
Screenshot 2025-04-30 at 11 30 22

**"aria-label issue"**
Screenshot 2025-04-30 at 11 35 10

rak-phillip:
@aalves08 Thanks for the review! The following should be fixed now:

- [X] can’t open dropdown with `space` key
- [X] Imo, we should focus the whole bottom “group” where we can select multiple options… I think that clicking on tab should focus immediately on the first item of that bottom group (check video)
- [X] can’t keyboard nav into “chips” to dismiss applied selections (check screenshot)
- [X] clear filter input box button as a problem with the translation on `aria-label` (check screenshot)
- [X] missing `aria-expanded` on on item with role combobox
- [X] missing `aria-activedescendant`
- [X] missing `aria-selected`

I kept the `shift+tab` behavior the same as before, where we focus on the container. Alternatively, we can consider removing the container as a focusable element altogether and just allow tab/down from the input/button to focus the options.

aalves08 (changes requested):
@rak-phillip thanks for the fixes, mate!

One last item missing: When you click `space` to open the dropdown it also does a "space" on the filter input, therefore hiding a lot of options on the dropdown because it's essentially filtering for " ". It needs fixing, sorry.

https://github.com/user-attachments/assets/5ea993ee-fb2a-495d-8a21-bb099570acef


With this last one I think the keyboard nav will be as good as it can get 💪

rak-phillip:
> One last item missing: When you click space to open the dropdown it also does a "space" on the filter input, therefore hiding a lot of options on the dropdown because it's essentially filtering for " ". It needs fixing, sorry.

@aalves08 good catch. This has been resolved in 8a94e57c515f3f7961cdeb9e03ab08eee62c8082

aalves08 (approved):
All fixed. LGTM mate!

PR #14320, opened 2025-05-09
fix problem with extra space and scroll on settings banner page

### LINKED ISSUE #14314: Global Settings: Banners: White space appears at bottom of Banners Page
**Setup**
- Rancher version: v2.12

**Describe the bug**
White space is visible at the bottom of the Banners page, even when there is no additional content to scroll to.

**To Reproduce**
1. go to Global Settings -> Banners
2. scroll to bottom of page

**Result**
Excessive white space.
**Expected Result**
No white space.

**Screenshots**

Image

**Additional context**
Discovered while testing https://github.com/rancher/dashboard/issues/14252

aalves08:
### Summary
Fixes #14314 

- fix problem with extra space and scroll on settings banner page

### Technical notes summary

Couldn't really find the root cause, but the overflow on that CSS rule doesn't do anything apparently (it's a grid well defined in size) and it's causing that extra scroll to happen. Only seems to happen on chromium based browsers

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/pkg/kubectl-explain/components/SlideInPanel.vue
+++ b/pkg/kubectl-explain/components/SlideInPanel.vue
@@ -35,11 +35,20 @@ export default {
     };
   },
 
-  created() {
-    useWatcherBasedSetupFocusTrapWithDestroyIncluded(() => this.isOpen, '.slide-in', {
-      escapeDeactivates: false,
-      allowOutsideClick: true
-    });
+  watch: {
+    // better to trigger focus-trap based on "isOpen" rather than using the "created" hook
+    // because re-opening an already loaded resource data would not run the "created" hook
+    // and therefore not triggering the focus trap
+    isOpen(neu, old) {
+      if (neu && neu !== old) {
+        useWatcherBasedSetupFocusTrapWithDestroyIncluded(() => this.isOpen, '[data-testid="slide-in-panel-resource-explain"]', {
+          escapeDeactivates: false,
+          allowOutsideClick: true,
+          // putting the initial focus on the first element that is not conditionally displayed
+          initialFocus:      '[data-testid="slide-in-panel-close-resource-explain"]'
+        });
+      }
+    }
   },
 
   computed: {
@@ -199,7 +208,7 @@ export default {
       class="slide-in"
       :class="{ 'slide-in-open': isOpen }"
       :style="{ width, right, top, height }"
-      data-testid="slide-in-panel"
+      data-testid="slide-in-panel-resource-explain"
     >
       <div
         ref="resizer"
@@ -267,7 +276,7 @@ export default {
             role="button"
             :aria-label="t('kubectl-explain.scrollToTop')"
             class="icon icon-close"
-            data-testid="slide-in-panel-close"
+            data-testid="slide-in-panel-close-resource-explain"
             tabindex="0"
             @click="close()"
             @keydown.space.enter.stop.prevent="close()"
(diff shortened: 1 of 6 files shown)


rak-phillip (dismissed):
LGTM - I tested this in both Chrome and Firefox and scrolling still behaves as expected across several areas of Dashboard.

I have no means to reproduce the original issue, so I think that we should get a final signoff from @yonasberhe23.

yonasberhe23:
I tested the fix in dev mode. The original issue no longer reproduces when scrolling the page however, I was able to reproduce the issue by tabbing through the page instead. @aalves08 Are you seeing this as well?

rak-phillip:
I can see elements from the window manager receiving focus when tabbing, that might be root cause of this issue. It looks like `div.wm.drag-end` and `div.resizer.resizer-y` in `shell/components/nav/WindowManager/index.vue` are able to receive keyboard focus even though they are not visible on the page.

aalves08:
@rak-phillip even if we remove everything from the `default` template except the `router-view`, there's still problems with that extra space. 

Removing the `overflow-y` like the original fix seems to be the only thing that fixes it 🤷 

You were right with the `WindowManager`, so now is "hidden" under a `v-if`.

I also noticed that the `SlideInManager` close button was also receiving focus because of `tabindex=0`. I just did a condition to not focus on it if it's not open. Also removed a weird `trigger-focus-trap` that Jordon had on that close button. That doesn't do anything and is just prone to problems.

https://github.com/user-attachments/assets/84597794-c693-4f21-bfb0-e35aead604ad

Such a weird bug this one... Ready for re-review @rak-phillip @yonasberhe23

rak-phillip (dismissed):
LGTM - this might require a rebase on master to resolve failing e2e tests.

yonasberhe23 (approved):

aalves08:
@rak-phillip @yonasberhe23 I had to address some flaky e2e tests on `describe-resource` and a broken e2e on the kubectl explain. Up for final review 🙏

yonasberhe23 (approved):
LGTM!

pkg/kubectl-explain/components/SlideInPanel.vue
```diff
@@ -35,11 +35,20 @@ export default {
...
+    isOpen(neu, old) {
+      if (neu && neu !== old) {
+        useWatcherBasedSetupFocusTrapWithDestroyIncluded(() => this.isOpen, '[data-testid="slide-in-panel-resource-explain"]', {
+          escapeDeactivates: false,
+          allowOutsideClick: true,
+          // putting the initial focus on the first element that is not conditionally displayed
+          initialFocus:      '[data-testid="slide-in-panel-close-resource-explain"]'
+        });
+      }
+    }
```
rak-phillip, line 51: I get the following error when I navigate to Local Cluster => Nodes => Click on Describe Resource:

```
Uncaught runtime errors:
ERROR
Your focus-trap must have at least one container with at least one tabbable node in it at all times
updateTabbableNodes@https://127.0.0.1:8005/js/chunk-vendors.js:48797:13
activate@https://127.0.0.1:8005/js/chunk-vendors.js:49251:9
./shell/composables/focusTrap.ts/useWatcherBasedSetupFocusTrapWithDestroyIncluded/</</<@https://127.0.0.1:8005/js/index.js:39228:39
```

I think that it has something to do with this change. This only happens when I first render the panel; subsequent renders will not throw the error.

rak-phillip (changes requested):
I'm encountering a runtime error. I think that we need to investigate it and resolve before merging.

pkg/kubectl-explain/components/SlideInPanel.vue
rak-phillip: Perhaps.. The issue must've been on my end. I attempted a repro in a fresh environment and I don't see the behavior any more.

rak-phillip (approved):

PR #14367, opened 2025-05-15
Update Rancher theme

rak-phillip:
### Summary
This updates the default theme for Rancher Dashboard.

Fixes #

- Set default theme to "modern"
- Preserve original theme as "classic"
- Update SUSE and CSP themes for modern styles

This affects all dashboard themes. Brands can be modified by navigating to Global Settings and setting `ui-brand` to one of the values:

- modern (default theme, empty string)
- classic
- suse
- csp

Light and Dark themes need review for each brand.

Same as above - buttons, inputs, tabs, navigation, etc...

##### Modern

image

##### Classic

image

##### SUSE

image

##### CSP

image

##### Modern

image

##### Classic

image

##### SUSE

image

##### CSP

image

### Checklist
- [x] The PR has a Milestone 


shell/assets/styles/themes/_dark.scss
```diff
@@ -29,19 +29,84 @@
...
-  $contrasted-dark: $lightest !default;
+  $contrasted-dark: $grey-0 !default;
   $contrasted-light: $darkest !default;
 
   $drag-over: #DCDEE7;
+  $body-bg: #2F2F42;
+
+  /*********************/
+  
+  --body__bg: #{$body-bg};
```
rak-phillip, line 40: We had an out of band discussion on the introduction of new css variables like `--body__bg`. Widespread usage of variables like this could cause backwards compatibility issues with extensions that are built with these new variables in place, as they will not exist in previous versions of Rancher Dashboard.

shell/assets/styles/themes/_light.scss
```diff
@@ -44,25 +54,60 @@ $drag-over: #DCDEE7;
...
+  --on-active: #{$grey-0};
+  --active--hover: #{lighten($primary-110, 5%)};
+  --category--active: #{$primary-15};
+  --category--active--hover: #{$grey-20};
+
+  --toggle-OFF__bg: #{$grey-5};
+  --toggle-OFF__color: #{$grey-50};
+  --toggle-ON__bg: var(--active);
+  --toggle-ON__color: var(--on-active);
+  --toggle-OFF--hover: #{$grey-10};
```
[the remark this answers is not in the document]
edenhernandez-suse: That's is BEM (Block, Element, Modifier), a rather old CSS naming methodology, so things after `__` are elements/properties and things after `--` are modifiers/states. I used them because it helps me write CSS and understand variables, but I know it's not the norm in our code and I probably we shouldn't switch to it if that's going to be a problem.

shell/assets/styles/themes/_csp.scss
```diff
@@ -4,9 +4,9 @@
   $selected: rgba($primary, .5);
 
   --primary                    : #{$primary};
-  --primary-text               : #fff;
+  --on-primary               : #fff;
```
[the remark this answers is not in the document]
edenhernandez-suse: In modern Design Systems, design tokens for "content on a primary element" are commonly described as `on-primary` instead of `primary-text`, `primary-icon`, etc. Literally, anything without its own colour on top of a primary element.

As with the rest, it's my preferred way to write variables (and what we use in Figma), but it's you guys, the devs, who should decide whatever you prefer to use.

aalves08:
My main concern here is the variables names that change and even some new variables that come in... 

I've locally built two versions of Kubewarden:
- one with the current shell it employs  (let's call it `OLD SHELL`)
- another one with shell based on this PR (let's call it `NEW SHELL`)

Some other concepts: 
- `OLD SYSTEM` means any Rancher without this new theming
- `NEW SYSTEM` means Rancher with this new theming

I then did developer load into different Rancher systems.

Here are some comparisons I did:

**image 1 - old-shell vs old-system**
1-old-shell-old-system

**image 2 - new-shell vs old-system**
1-new-shell-old-system

There's impact on the radio buttons, which can barely be seen when active

**image 3 - old-shell vs old-system**
2-old-shell-old-system

**image 4 - new-shell vs old-system**
2-new-shell-old-system

Grouped buttons for table grouping look weird

`Observability` extension (with it's old shell) running against this PR:

_problem icon sidebar (this particular has a CSS hack which I've raised awareness to the developers to remove it and fix the svg)_
hover-observability-icon-weird

_Weird hover effect on button that impacts badges_
https://github.com/user-attachments/assets/aaf9f756-0f57-4660-9148-3cb40d5b3207


Besides these types of problems we might find with extensions, there's a few areas to have some tweaks:

_linux badge is so light that is hard to read_
light-yellow-barely-visible-charts

_icon local cluster bad on dark mode_
bad-dark-mode-icon-local-cluster

_describe icon on header is wrong_
describe-icon-header-wrong

Overall, I think that I thought the impacts were worse on extensions, but I am really keen on testing this PR after we've consolidated the CSS variable names to have the same names.

shell/assets/styles/themes/_csp.scss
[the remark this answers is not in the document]
richard-cox: this is another one where unfortunately i think we need to stick to the existing convention

richard-cox:
Neil made a good comment, we should try to keep the original (community only?) styles and provide a way to return to them if needed

rak-phillip:
@richard-cox @edenhernandez-suse @aalves08 I think that this is ready for more feedback. I started adding fallbacks for variables in the base "brand". If we want to use the new styles, we should set the brand to "modern"

image

This still still a WIP and I expect to run into regressions with the base brand. Please take not of all inconsistencies that you might find along the way.

shell/store/prefs.js
```diff
@@ -205,6 +205,16 @@ export const getters = {
     return theme;
   },
 
+  brand: (_state, _getters, _rootState, rootGetters) => {
+    const brandSetting = rootGetters['management/byId'](MANAGEMENT.SETTING, SETTING.BRAND);
+
+    if (brandSetting && brandSetting.value && brandSetting.value !== '') {
```
richard-cox: if `brandSetting.value` is `''` it'll never get to the third condition

think this can be shortened to 

```
brandSetting?.value !== ''
```

edenhernandez-suse:
I've seen a bug with the Home SVG icon on hover:
 
image

The colour applied is `on-tertiary` and it should be `on-active`:

image

EDITED: I initially said it should be `on-focus`

edenhernandez-suse:
The `User retention settings` button/link is broken:

image
image

edenhernandez-suse:
The button group used inside Helm charts installation wizards use weird classes and look kinda broken as well. The inactive one is using the `disabled` class, they have `bg-primary` which was never updated and, on hover, they take the new community primary hover (although I'm in the Prime theme).

Maybe we should have a separate bug for this component, but `bg-primary` should not be used and, if we want to keep it for backwards compatibility, it should be adapted to the theme primary (in this case, it should be green).

image

HOVER
image

edenhernandez-suse:
First-level nav items get the wrong hover background colour when active:

Simply active:
image

Active on hover:
image

rak-phillip:
> i applied the suse brand however got csp

@richard-cox can you explain how you confirmed that you got csp instead of SUSE? I'm not able to reproduce this behavior.

richard-cox:
@rak-phillip i updated the confluence page (tagged changes with `#2`), suse brand was showing csp colours in dark mode.

let me know when this is good for another pass

richard-cox:
@rak-phillip i've gone through the confluence page and updated with resolved issues, a couple of new. worth a rebase to get the new resource detail / config UI

shell/assets/styles/themes/_csp.scss
```diff
@@ -1,25 +1,43 @@
 .csp {
-  $primary: hsl(151, 59%, 46%);
-  $info:  mix($primary, $secondary, 50%);
-  $selected: rgba($primary, .5);
+  $primary  : hsl(151, 59%, 46%);
+  $info     : mix($primary, $secondary, 50%);
+  $selected : rgba($primary, .5);
```
richard-cox, line 5: at some point we should re-org this folder

- themes - light|dark
- brands - csp|modern|suse (root, but not default, brand being classic)

shell/assets/styles/base/_variables-classic.scss
```diff
@@ -0,0 +1,85 @@
+.classic {
```
richard-cox, line 1: are these all lift and shifted from `_light.scss`?

shell/assets/styles/base/_basic.scss
```diff
@@ -23,6 +23,10 @@ BODY {
   }
 }
 
+HEADER {
```
richard-cox, line 26: is this needed?

In therms of hyerachy it's 
1. body / dashboard-body
   - has same background style 
3. dashboard-root
4. dashboard-content
5. header

pkg/rancher-components/src/components/BadgeState/BadgeState.vue
```diff
@@ -74,27 +74,30 @@ export default defineComponent({
...
     display: inline-flex;
-    padding: 2px 10px;
+    padding: 2px 8px;
     border: 1px solid transparent;
     border-radius: 20px;
+    height: 20px;
 
     &.bg-info {
-      border-color: var(--info);
+      color: var(--on-info-banner);
```
richard-cox, line 83: i got lost on this. in components we 

- introduce new colours with no fallback on an old colour
- when vars have been renamed however we do fallback on old an colour (which now does not exist)

should we either always fallback, or it's not needed in both places?

shell/assets/styles/global/_button.scss
```diff
@@ -21,23 +22,27 @@ button,
   user-select: none;
   border: 0;
   padding: $btn-padding;
-  border-radius: var(--border-radius);
+  border-radius: $btn-radius;
```
richard-cox, line 25: given `border-radius: $btn-radius;`, is this needed?

shell/components/BrandImage.vue
```diff
@@ -78,9 +78,13 @@ export default {
       }
     },
 
+    isDarkTheme() {
```
richard-cox, line 81: there's no `dark-modern` now

shell/components/IconOrSvg.vue
```diff
@@ -24,8 +24,8 @@ const cssCache = {};
 
 const colors = {
   header: {
-    color: '--header-btn-text',
-    hover: '--header-btn-text-hover'
+    color: '--on-tertiary',
```
richard-cox: fallbacks needed?

richard-cox (changes requested):
- storybook needs validating

pkg/rancher-components/src/components/BadgeState/BadgeState.vue
rak-phillip: We will need to define a fallback for new colors where one wasn't defined before. FYI @edenhernandez-suse

shell/assets/styles/base/_variables-classic.scss
rak-phillip: Correct - I introduced this so that we can share classic colors between light and dark themes, we can't implicitly fallback to these variables without `_light.scss` as the default anymore.

shell/assets/styles/themes/_csp.scss
rak-phillip: Much agreed. I think there are many opportunities to refactor how we handle themes and styles across the board.

shell/components/BrandImage.vue
rak-phillip: Correct. Good catch.

shell/components/IconOrSvg.vue
rak-phillip: This requires some effort.. I introduced a change to attempt this in the component.

rak-phillip:
@nwmac @richard-cox I will close out this PR and work on isolating the changes. 

FYI @edenhernandez-suse we might need to sync to ensure that we don't unintentionally regress in certain areas as we move this forward.

nwmac:
Thanks @rak-phillip - let's start with the core color changes

PR #14396, opened 2025-05-27
Don't show agent status when API calls fail

### LINKED ISSUE #14394: Cattle Agent and Fleet Agent status boxes still shown when the user does not have permission
Internal Reference: SURE-10211

This is a follow-up to https://github.com/rancher/dashboard/issues/12259

There are still cases where we show the Fleet and/or Cattle agent status cards on the cluster dashboard home page, e.g.

Image

This occurs when the API calls return an error response - e.g. 403.

The current code does handle this by setting the agent info to `null` - but this is not then taken into account when rendering the page.

The UI should NOT show these status boxes if the backend API calls return errors.

nwmac:
### Summary
Fixes #14394

This PR fixes the issue above.

The existing code does handle the case where the backend API returns an error code, but does not then use this correctly.

This PR is a relatively small fix, which ensures we hide the boxes when we get errors from the APIs, since these indicate that the user does not have permission. 

Automated tests are extended to cover the gap - I believe they are sufficient to cover the use cases given the code paths that we have.

### Screenshot/Video

Before:

image

After:

image

### Checklist
- [x] The PR has a Milestone 


aalves08 (approved):
LGTM

PR #14419, opened 2025-05-29
Add Notification Center

### LINKED ISSUE #14007: [UI EPIC] Add notification centre to UI
For management tracking purposes.

nwmac:
### Summary
Fixes #14007 

This PR adds support for the new Notification Center UI that is shown in the top-right of the UI.

Currently this is only wired into 2 places:

- Release notes - the notification to read the release notes for a release notes has been moved from the home page banner into the notification center
- Growls - all growls will add a corresponding notification to the notification center

### Technical notes summary

To ensure backwards compatibility with UI Extensions, we leave the existing `growl` store in place and add a new `notification` store. This should be the considered the go-forward API that developers should use and the `growl` store should be considered deprecated.

Note that for now, developers should use the notification store directly - we will add an API for this later as part of introducing APIs for extensions and developers.

Of note:

- e2e tests for release notes have been updated to cover the use of the notification center
- `GrowlManager` has been updated to fix a bug where one or more growls could get stuck on screen. The button that would show to `Clear all` has been removed, since growls will appear in the notification center. The growl notifications now also show below the header, rather than at the top of the page, so that they do not overlap on the new notification center icon
- Growl store - We will only show 5 growls at most - since growls are also in the notification center, if 5 are on screen and a new one appears, the oldest will be removed - they are always available in the notification center.
- Growl store - Growls now send notifications. A new action is added, `notification`, since notifications will also show growls for `success`, `warning` and `error` growls (not info and others). This ensures we do not get any circular issues.
- Growl store - errors do not override the timeout to be shown indefinitely - they will use the 5 second default and auto disappear like the others, since the user will be able to access them via the notification center
- Growl store - closing a growl that was creating from a notification will mark the associated notification as read
- Notification store - pretty straight-forward. Notifications are persisted in local storage in the browser. We keep at most 50 notifications and 14 days of notifications
- The notification center UI uses the new `RcDropDown` with custom items. A lot of effort has been put into making sure this is accessible.
- The notification center uses a focus trap for each item, so when the user clicks, they tab between the read/unread button and the primary and secondary buttons (if they are present) - the component manages this directly as the focus trap util did not seem to work for this use case
- Release notes will not check at startup if it needs to add a notification. It checks the user preference and uses a notification with a well-known identifier to ensure we only re-add the release notes notification for a release when necessary.

e2e tests cover release notes.

We may want to do some manual testing of the UI and the keyboard navigation.

### Screenshot/Video

Example with some test notifications:

image

Notifications all clear:

image

### Checklist
- [x] The PR has a Milestone 


shell/components/nav/NotificationCenter/index.vue
```diff
@@ -0,0 +1,160 @@
...
+  if (index === 1) {
+    setTimeout(() => {
+      scroller?.value?.scrollTo(0, 0);
+    }, DELAY);
+  } else if (index === count - 1) {
+    setTimeout(() => {
+      scroller?.value?.scrollTo(0, 2000000);
+    }, DELAY);
+  }
+};
```
rak-phillip: I think that we can remove (most) of the workaround if we simply add a `.prevent` handler to each of the `@keydown.up.down` events. It appears that the offset that causes the an item to not fully scroll into view is the same as one tick when pressing down on the keyboard. 

We may need to utilize `scrollPanel` for items that are partially rendered outside of the container, but I think that we can simplify this function and remove the delay if we prevent the default up/down key behavior.

rak-phillip: I gave it a quick test with the following `scrollPanel()` implementation

```
const scrollPanel = (index: number, count: number) => {
};
```

These are the results in Chrome


https://github.com/user-attachments/assets/89ef4727-46ad-408c-81bd-ad28b2665293

shell/components/nav/NotificationCenter/Notification.vue
```diff
@@ -0,0 +1,446 @@
...
+import { DropdownContext, defaultContext } from '@components/RcDropdown/types';
+
+const CLASSES = {
+  [NotificationLevel.Announcement]: 'icon-notify-announcement text-info',
+  [NotificationLevel.Error]:        'icon-notify-error text-error',
+  [NotificationLevel.Info]:         'icon-notify-info text-info',
+  [NotificationLevel.Task]:         'icon-notify-busy text-info',
+  [NotificationLevel.Warning]:      'icon-notify-warning text-warning',
+  [NotificationLevel.Success]:      'icon-notify-tick text-success',
+};
```
rak-phillip, line 18: nit: an enum might fit better than an object for this use case.

shell/components/nav/NotificationCenter/Notification.vue
```diff
@@ -0,0 +1,446 @@
...
+
+const emits = defineEmits(['didFocus']);
+
+const props = defineProps({ item: { type: Object as PropType<StoredNotification>, required: true } });
```
rak-phillip: nit: we can drop the need for `PropType` if we use type-based declaration

```suggestion
const props = defineProps<{item: StoredNotification}>();
```

shell/components/nav/NotificationCenter/index.vue
```diff
@@ -0,0 +1,160 @@
...
+    :aria-label="t('nav.notifications.label')"
+    @update:open="open"
+  >
+    <rc-dropdown-trigger
+      tertiary
+      data-testid="notifications-center"
+      :aria-label="t('nav.notifications.button.label')"
+    >
+      <div class="level-indicator">
+        <i class="icon icon-lg icon-notify-bell" />
```
rak-phillip, line 87: hmm.. I think I'm missing some icons? When all notification are read, I see an empty void in the header

image

shell/components/nav/NotificationCenter/Notification.vue
```diff
@@ -0,0 +1,446 @@
...
+          </div>
+        </div>
+        <div
+          v-if="item.primaryAction || item.secondaryAction"
+          class="notification-actions"
+        >
+          <button
+            v-if="item.primaryAction"
+            ref="primaryActionButton"
+            tab-index="1"
```
rak-phillip: Avoid manually manipulating the tab index for individual elements. The structure of our markup should suffice in ensuring that the tab order is correct.

shell/components/nav/NotificationCenter/Notification.vue
```diff
@@ -0,0 +1,446 @@
...
+            role="button"
+            class="btn btn-sm role-primary"
+            @keydown.enter.space.stop="action(item.primaryAction)"
+            @keydown.tab.stop="innerFocusNext($event)"
+            @keydown.escape.stop="exitFocusTrap"
+            @click.stop="action(item.primaryAction)"
+          >
+            {{ item.primaryAction.label }}
+          </button>
+          <button
```
rak-phillip, line 299: optional: We do have the `RcButton` component that should handle things like roles and styling for us.

shell/components/nav/NotificationCenter/Notification.vue
```diff
@@ -0,0 +1,446 @@
...
+
+// Exit inner focus - this is the state when the notification has a focus trap and you are tabbing between
+// the controls in the notification - this function is typically called when the user presses ESCape
+// and we want to exit this focus trap
+const exitFocusTrap = (e: MouseEvent | KeyboardEvent) => {
+  // Return focus to the outer notification div
+  dropdownMenuItem.value?.focus();
+
+  e.stopPropagation();
+  e.preventDefault();
```
rak-phillip, line 141: optional: we can remove the need for `e.stopPropagation()` and `e.preventDefault()` if we add the `.stop` and `.prevent` modifiers to the associated events.

shell/components/nav/NotificationCenter/NotificationHeader.vue
```diff
@@ -0,0 +1,112 @@
...
+</script>
+
+<template>
+  <div
+    dropdown-menu-item
+    tabindex="-1"
+    role="menuitem"
+    class="notification-header"
+    @click.stop
+    @keydown.up.down.stop="handleKeydown"
```
rak-phillip: Adding `.prevent` should stop the default scroll behavior:

```suggestion
    @keydown.up.down.stop.prevent="handleKeydown"
```

shell/components/nav/NotificationCenter/NotificationHeader.vue
```diff
@@ -0,0 +1,112 @@
...
+        {{ t('notificationCenter.title') }}
+      </div>
+      <div v-if="unreadCount !== 0">
+        <a
+          ref="markAllReadButton"
+          role="button"
+          tabindex="-1"
+          href="#"
+          data-testid="notifications-center-markall-read"
+          @keydown.up.down.stop="handleKeydown"
```
rak-phillip: Adding `.prevent` should stop the default scroll behavior:

```suggestion
    @keydown.up.down.stop.prevent="handleKeydown"
```

shell/components/nav/NotificationCenter/Notification.vue
```diff
@@ -0,0 +1,446 @@
...
+
+<template>
+  <div
+    ref="dropdownMenuItem"
+    dropdown-menu-item
+    tabindex="-1"
+    role="menuitem"
+    data-testid="notifications-center-item"
+    :aria-label="t('notificationCenter.ariaLabel', { title: item.title })"
+    @keydown.up.down.stop="handleKeydown"
```
rak-phillip: Adding `.prevent` should stop the default scroll behavior:

```suggestion
    @keydown.up.down.stop.prevent="handleKeydown"
```

shell/components/nav/NotificationCenter/NotificationHeader.vue
```diff
@@ -0,0 +1,112 @@
...
+    @click.stop
+    @keydown.up.down.stop="handleKeydown"
+    @focus="gotFocus"
+  >
+    <div class="notification-header">
+      <div class="notification-title">
+        {{ t('notificationCenter.title') }}
+      </div>
+      <div v-if="unreadCount !== 0">
+        <a
```
rak-phillip, line 68: I think it's best for maintainability if we could use an actual button instead of a link that masquerades as a button. The `RcButton` component contains both `link` and `ghost` styles that should provide us with buttons that are minimally styled. Hopefully, that can fit our needs for this use case.

shell/store/notifications.ts
```diff
@@ -0,0 +1,272 @@
...
+// Notifications store state
+interface NotificationsStore {
+  notifications: StoredNotification[],
+}
+
+function persist(state: NotificationsStore) {
+  window.localStorage.setItem(LOCAL_STORAGE_KEY, JSON.stringify(state.notifications));
+}
+
+export const state = function(a: any): NotificationsStore {
```
rak-phillip: `a: any` is unused. I think that we need a better name for `a` if it is needed.

shell/store/notifications.ts
```diff
@@ -0,0 +1,272 @@
+import { randomStr } from '@shell/utils/string';
+
+const LOCAL_STORAGE_KEY = '#notifications';
+
+// Expire in seconds (14 days)
+const EXPIRY = 14 * 24 * 60 * 60;
```
rak-phillip: Can we use `dayjs` to make this easier to read and maintain?

shell/store/notifications.ts
```diff
@@ -0,0 +1,272 @@
...
+interface NotificationsStore {
+  notifications: StoredNotification[],
+}
+
+function persist(state: NotificationsStore) {
+  window.localStorage.setItem(LOCAL_STORAGE_KEY, JSON.stringify(state.notifications));
+}
+
+export const state = function(a: any): NotificationsStore {
+  let notifications = [];
```
rak-phillip: This implicitly has a type of `any[]`. Explicitly adding a type will assist with future maintainability:

```suggestion
  let notifications: StoredNotification[] = [];
```

shell/store/notifications.ts
```diff
@@ -0,0 +1,272 @@
...
+      const diff = (now.getTime() - created.getTime()) / 1000; // Diff in seconds
+
+      return diff < EXPIRY;
+    } catch (e) {}
+
+    return true;
+  });
+
+  return { notifications };
+};
```
rak-phillip: Similar to a data prop in a vue component, a store's state should avoid complex logic to focus on declaring reactive state. Would this logic initialization logic better fit in the `init()` action?

rak-phillip (changes requested):
What's here looks really good - It's good to see that the RcDropdown component can be extended to support additional use cases.

The notification dot looks a little wrong in dark mode, almost as if slightly squared and has a few white pixels bleeding through at the border? I think it's more pronounced when it's in the checked state:

image

image

I'm not sure if I like having `escape` be the only hatch to exit the focus trap. I feel that being able to press up/down on they keyboard to exit the trap and focus the next item would feel more natural. I think that we should let others play around with the existing implementation and collect feedback on how they feel about the focus trap interaction.

shell/components/nav/NotificationCenter/index.vue
nwmac: Yeah - sorry, I had not updated the icon pack - I have done this now.

shell/components/nav/NotificationCenter/index.vue
nwmac: Okay, I've done this - there is still a slight issue with the first notification, but I don't think that's worth having the work-around for, so I have taken it out.

shell/components/nav/NotificationCenter/Notification.vue
nwmac: Gonna leave that for another time

shell/store/notifications.ts
nwmac: Removed. Good spot

shell/store/notifications.ts
nwmac: We don't currently bring in the `duration` plugin for `dayjs`, which I think is what we'd need here. I had a look, but it was more change than I expected - can we do this later?

shell/store/notifications.ts
nwmac: Done

shell/components/nav/NotificationCenter/Notification.vue
nwmac: Thanks, updated.

shell/store/notifications.ts
nwmac: Good idea - done

shell/components/nav/NotificationCenter/Notification.vue
nwmac: Removed.

nwmac:
Added a fix for the dark mode border color issue

shell/components/nav/NotificationCenter/NotificationHeader.vue
nwmac: I tried this, but we don't quite have the styling on RcButton for this case - see last comment, I have created a follow-on issue to address this and a couple of other things from this PR.

nwmac:
@rak-phillip I think I've addressed most of your feedback - I did create a follow-on issue for a couple of things - https://github.com/rancher/dashboard/issues/14512 - I'd like to get this in, so it can be wired in to the registration work.

shell/components/nav/NotificationCenter/Notification.vue
```diff
@@ -0,0 +1,446 @@
...
+    dropdown-menu-item
+    tabindex="-1"
+    role="menuitem"
+    data-testid="notifications-center-item"
+    :aria-label="t('notificationCenter.ariaLabel', { title: item.title })"
+    @keydown.up.down.stop.prevent="handleKeydown"
+    @focus.stop="gotFocus"
+    @click.stop
+    @keydown.enter.space.stop="enterFocusTrap"
+  >
```
rak-phillip: We can listen on `@focusin` and add a method to invoke `scrollIntoView()`[^1] on the notification items. I think this solves all of the scrolling problems that we've encountered.

```html
  <div
    ref="dropdownMenuItem"
    dropdown-menu-item
    tabindex="-1"
    role="menuitem"
    data-testid="notifications-center-item"
    :aria-label="t('notificationCenter.ariaLabel', { title: item.title })"
    @keydown.up.down.stop.prevent="handleKeydown"
    @focusin="scrollIntoView"
    @focus.stop="gotFocus"
    @click.stop
    @keydown.enter.space.stop="enterFocusTrap"
  >
```

```js
const scrollIntoView = (event: Event) => {
  const target = event.target;

  if (target instanceof HTMLElement) {
    target?.scrollIntoView({
      behavior: 'smooth',
      block:    'center',
      inline:   'nearest',
    });
  }
};
```

[^1]: https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

rak-phillip: This is the result of the change above, what do you think?

https://github.com/user-attachments/assets/c3e0cd05-2b80-49e6-93f5-0facfc416eb8

rak-phillip (commented):
I think that this is looking good. I left a comment for a potential improvement of scrolling behavior.

The tab behavior is also less than ideal when attempting to tab away while the notification center is open. We should focus on the next element in the tab order, but it appears that an element in the notification center receives focus before the menu is closed. This isn't too disruptive, but it leaves us in a confusing state. You can see what I'm talking about if you compare the tab behavior of the page actions menu when it is open and compare it to the notification center.

Everything else looks good at this point.

docs/developer/notification-center.md
```diff
@@ -0,0 +1,95 @@
...
+
+The Notification Center adds a 'bell' icon in the top-right of the Rancher Manager UI which supports a history of application notifications. When there are unread notifications, then bell icon shows a red dot, for example:
+
+![Example Notification](./images/notification-headerbar.png)
+
+> The existing `growl` store is deprecated - notifications added to the notification center of certain notification levels will automatically show growls.
+
+## Adding a Notification
+
+> Note: Types are defined in `shell/notifications.ts`
```
richard-cox: shell/store/notifications.ts, though strictly speaking they should be in `shell/types`?

shell/store/notifications.ts
```diff
@@ -0,0 +1,287 @@
+import { randomStr } from '@shell/utils/string';
+
+const LOCAL_STORAGE_KEY = '#notifications';
+
+// Expire in seconds (14 days)
+const EXPIRY = 14 * 24 * 60 * 60;
+
+// Maximum number of notifications that will be kept
```
richard-cox: These and comments below should be jsdoc notation.

shell/store/notifications.ts
```diff
@@ -0,0 +1,287 @@
...
+const MAX_NOTIFICATIONS = 50;
+
+/**
```
richard-cox, line 8: even though there's a gap, the jsdoc here is applied to `NotificationAction` below

shell/utils/release-notes.ts
```diff
@@ -0,0 +1,47 @@
...
+      }
+    }
+  }
+
+  // Also use pref, so if we expired it, we don't add it back
+  if (!found && lastReadVersion !== rancherVersion) {
+    // Register notification
+    const notification = {
+      id,
+      level:      2,
```
richard-cox: NotificationLevel.Info

shell/utils/release-notes.ts
```diff
@@ -0,0 +1,47 @@
...
+      title:      t('landing.whatsNew.title', { version: rancherVersion }),
+      message:    t('landing.whatsNew.message'),
+      preference: {
+        key:   READ_WHATS_NEW,
+        value: rancherVersion
+      },
+      primaryAction: {
```
richard-cox, line 40: Improvement for the future, and to fix a bug we had before, we should tailor this in airgap world

docs/developer/notification-center.md
richard-cox: moving them to types would avoid extensions et al importing `NotificationLevel` from /store as well

shell/store/notifications.ts
```diff
@@ -0,0 +1,287 @@
+import { randomStr } from '@shell/utils/string';
+
+const LOCAL_STORAGE_KEY = '#notifications';
```
richard-cox: should this have a `#` in front?

shell/components/nav/NotificationCenter/Notification.vue
```diff
@@ -0,0 +1,446 @@
...
+          </div>
+          <div class="progress-percent text-muted">
+            {{ item.progress }}%
+          </div>
+        </div>
+        <div
+          v-if="item.primaryAction || item.secondaryAction"
+          class="notification-actions"
+        >
+          <button
```
richard-cox, line 287: haven't checked the design... but primary comes before secondary. in prompts we have the actionable primary button on the right and cancel like button on the left. guessing here primary and secondary could be anything, but it felt a little jarring seeing the bright blue button on

image

shell/store/notifications.ts
```diff
@@ -0,0 +1,287 @@
...
+    state.notifications.forEach((notification) => {
+      if (!notification.read) {
+        notification.read = true;
+      }
+    });
+
+    persist(state);
+  },
+
+  update(state: NotificationsStore, notification: Notification) {
```
richard-cox:
```suggestion
  update(state: NotificationsStore, notification: Partial<Notification>) {
```

docs/developer/notification-center.md
```diff
@@ -0,0 +1,95 @@
...
+|level|Notification level|
+|title|Title to be displayed in the notification|
+|message|Message to be shown in the notification (optional)|
+|progress|Progress (0-100) for notifications of type `Task` (optional)|
+|primaryAction|Primary action to be shown in the notification (optional)|
+|secondaryAction|Secondary to be shown in the notification (optional)|
+|id|Unique id (optional)|
+|preference|User Preference tied to the notification (optional) (the preference will be updated when the notification is marked read)|
+
+The `level` specifies the importance of the notification and determines the icon that is shown in the notification:
```
richard-cox: Worth adding that levels are represented via the `NotificationLevel` type, details below

docs/developer/notification-center.md
```diff
@@ -0,0 +1,95 @@
...
+|Info|Information notificatoon|
+|Success|Notification that something has completed successfully|
+|Warning|Notification of a warning|
+|Error|Notification of an error|
+
+Example notification of type `Announcement`:
+
+![Example Notification](./images/notification-example.png)
+
+Note that the `id` field is optional - if not specified, the notification will be assigned a unique identifier. Using a known identifier is only really required if you wish to be able to look up a notification by a known identifier OR you wish to be able yo update a notification after being added.
```
richard-cox:
```suggestion
Note that the `id` field is optional - if not specified, the notification will be assigned a unique identifier. Using a known identifier is only really required if you wish to be able to look up a notification by a known identifier OR you wish to be able to update a notification after being added.
```

richard-cox: i hit this whilst playing around. i added a notification on the home page with an id and then two tabs were auto-reloaded. end result was two notification each with the same id and content. could the add action switch to update if the id already exists or error if the same id was previously added?

shell/store/index.js
```diff
@@ -881,6 +882,11 @@ export const actions = {
       setBrand(brand);
     }
 
+    // Add the notification for the release notes
+    if (isRancher) {
+      await addReleaseNotesNotification(dispatch, getters);
```
richard-cox, line 887: will isRancher skip harvester (and epinio) standalone modes? noticed it checks for the upstream cluster for the project schema... which they might have but avoid them in the namespace list

docs/developer/notification-center.md
```diff
@@ -0,0 +1,95 @@
+# Notification Center API
```
richard-cox, line 1: do we need to call out that developers shouldn't send sensitive information (id's, token content, etc) given we use local storage?

shell/components/nav/NotificationCenter/Notification.vue
```diff
@@ -0,0 +1,446 @@
+<script setup lang="ts">
```
richard-cox, line 1: comment only. i'll add this to the 'what needs improving' retro point and pick up the action. we should come up with a best practise for structuring composition components. this would avoid a wall of disjointed variables and function calls. someone somewhere out there must have come up with a good approach that we can adopt/improve

shell/components/nav/NotificationCenter/index.vue
```diff
@@ -0,0 +1,137 @@
...
+            {{ t('notificationCenter.caughtUp', {}, true) }}
+          </div>
+        </div>
+        <div
+          ref="scroller"
+          class="scroll-container"
+        >
+          <template
+            v-for="(a, index) in allNotifications"
+            :key="a.title"
```
richard-cox:
```suggestion
            :key="a.id"
```

though this could be duped given add with id x 2 = two notifications

shell/components/nav/NotificationCenter/index.vue
```diff
@@ -0,0 +1,137 @@
+<script setup lang="ts">
```
richard-cox, line 1: the ocd part of me really wanted to clear my notifications, though there's no way to do this atm. was it out of scope for 2.12 or not scoped at all?

shell/store/growl.js
```diff
@@ -50,11 +63,29 @@ export const actions = {
...
+
+    // If the growl has a notification, mark it as read if the user dismisses the growl
+    if (growl.notification) {
+      await dispatch('notifications/markRead', growl.notification, { root: true });
+    }
+  },
+
+  async success({ commit, dispatch }, data) {
+    // Send a notification for the growl
+    const notification = await dispatch('notifications/fromGrowl', {
```
richard-cox, line 79: we might want to re-asses some of the existing growl levels

for instance, we show a growl in the cluster management cluster list when the user clicks 'copy kubeconfig to clipboard' (so there's some kind of visual feedback, done also for snapshot creation). if i click that three times i have to close each growl or go to notifications to read all or each to remove the unread notification dot. should that growl be at info level?

alternatively this is working as intended? i could click on copy kubeconfig, miss the growl because i read slowly but can then go to the notification to check text

nwmac: No, I think the growl there is un-necessary. Let me have a look.

shell/components/nav/NotificationCenter/index.vue
nwmac: Yeah, I know. Discussed with UX and said it was okay as is.

I was looking at adding something in the User Preferences - maybe I can do that as part of the follow-on ticket.

shell/components/nav/NotificationCenter/Notification.vue
nwmac: Nice. Thank you Phil. I have updated the PR with this change.

docs/developer/notification-center.md
nwmac: Moved the types

shell/utils/release-notes.ts
nwmac: Done

shell/utils/release-notes.ts
nwmac: Yes, we should

docs/developer/notification-center.md
nwmac: I have added a comment

shell/components/nav/NotificationCenter/Notification.vue
nwmac: Would love this - I can't say I love the composition API

shell/components/nav/NotificationCenter/index.vue
nwmac: Good spot. Fixed

shell/store/notifications.ts
nwmac: Done

shell/store/notifications.ts
nwmac: Done

shell/store/growl.js
nwmac: I've tuned those down to be `info` growls, rather than `success` - we can revisit, but this stops them also appearing in the notification center.

shell/store/notifications.ts
nwmac: Yeah, It doesn't matter - I thought i'd use that. I can remove it you like.

shell/components/nav/NotificationCenter/Notification.vue
nwmac: That is the design! Can raise with UX

shell/store/index.js
nwmac: That is what I am hoping

docs/developer/notification-center.md
nwmac: Done

docs/developer/notification-center.md
nwmac: Done

nwmac: Added a check for an existing notification with the same id

shell/store/notifications.ts
nwmac: Done

shell/components/nav/NotificationCenter/Notification.vue
nwmac: Will add to the follow-on work

shell/components/nav/NotificationCenter/NotificationHeader.vue
nwmac: Added

shell/components/nav/NotificationCenter/Notification.vue
nwmac: Added

shell/components/nav/NotificationCenter/NotificationHeader.vue
nwmac: Added

shell/components/nav/NotificationCenter/Notification.vue
nwmac: Will leave for now

nwmac:
@richard-cox Updated typo in doc as per our discussion and also switched the local storage key to be keyed off of the user id and moved the initialisation into `isLoggedIn`, so that we initialise notifications once we know what the user id is.

docs/developer/notification-center.md
```diff
@@ -0,0 +1,97 @@
...
+
+> &#x26a0;&#xfe0f; Documentation in this directory is intended for internal use only. Any information contained here is unsupported.
+
+> For now, developers can use the Notification Centre via the VueX store. In the future, we will add an API for this, that will be fully supported as part of the UI Extensions API.
```
rak-phillip, line 5: nit: we switch between "Centre" and "Center". We should stick to one for consistency and use that throughout 🙂

rak-phillip (approved):
LGTM

I found an edge case where the notification header receives focus if the list is empty. I think that this is an issue with the Dropdown Menu implementation not respecting `tabindex=-1`.

image

nwmac:
@rak-phillip @richard-cox Richard asked me to encrypt the notification data in local storage - now updated for this - this required a bit of change, but hopefully clear what I have done.

docs/developer/notification-center.md
nwmac: Indeed - I try to use the US spelling, but my better conscience is struggling with it

PR #14440, opened 2025-06-02
Add description formatter

a110605:
### Summary
Fixes https://github.com/harvester/harvester/issues/6869
SURE-5788

### Technical notes summary
- Add description formatter to avoid too long description column.
- Add description column in project/namespace table

### Screenshot/Video
Before 

Screenshot 2025-06-02 at 3 42 38 PM

After 

Screenshot 2025-06-02 at 3 52 44 PM

Add description column 
Screenshot 2025-06-02 at 3 58 22 PM

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/formatter/Description.vue
+++ b/shell/components/formatter/Description.vue
@@ -1,24 +1,19 @@
 <script>
 export default {
   props: {
-    row: {
-      type:     Object,
+    value: {
+      type:     String,
       required: true,
-    },
-  },
-  computed: {
-    descriptionText() {
-      return this.row?.description || '';
-    },
+    }
   },
 };
 </script>
 <template>
   <span
-    v-clean-tooltip="{content: descriptionText, skidding: -30, placement: 'top-start', popperClass: ['description-tooltip']}"
+    v-clean-tooltip="{content: value, placement: 'auto', popperClass: ['description-tooltip']}"
     class="description-text"
   >
-    {{ descriptionText }}
+    {{ value }}
   </span>
 </template>
 <style lang="scss">
--- a/shell/components/formatter/Description.vue
+++ b/shell/components/formatter/Description.vue
@@ -1,0 +1,35 @@
+<script>
+export default {
+  props: {
+    row: {
+      type:     Object,
+      required: true,
+    },
+  },
+  computed: {
+    descriptionText() {
+      return this.row?.description || '';
+    },
+  },
+};
+</script>
+<template>
+  <span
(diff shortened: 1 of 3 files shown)


a110605:
@nwmac, do I need to open an issue under `rancher/dashboard` to fix gh-project-integration workflow failed ?

or we can modify `.github/workflows/scripts/pr-gh-project.js` to include harvester/harvester issues.

richard-cox:
I've added the label `QA/None`, which skips the requirement for a linked dashboard issue. QA should be carried out externally as part of the harvester projects issue management

shell/components/formatter/Description.vue
```diff
@@ -0,0 +1,35 @@
+<script>
+export default {
+  props: {
+    row: {
```
richard-cox: this can be replaced with the below prop, `computed` deleted, and then `descriptionText` references replaced with `value`

```
    value: {
      type:     String,
      required: true,
    }
```

shell/components/formatter/Description.vue
```diff
@@ -0,0 +1,35 @@
...
+  computed: {
+    descriptionText() {
+      return this.row?.description || '';
+    },
+  },
+};
+</script>
+<template>
+  <span
+    v-clean-tooltip="{content: descriptionText, skidding: -30, placement: 'top-start', popperClass: ['description-tooltip']}"
```
richard-cox: cluster management --> cloud credentials. 

note the triangle misaligned with main tooltip area and also the where it points to

image

image

richard-cox (changes requested):
At first i though it would look a lot better underneath the name itself (neater, allows for more width for name due to removal of column, will apply everywhere), however then users cannot sort/filter by the description. Think sorting/filtering might be really important, so lets keep as is

image

Screenshot_20250602_161704

shell/components/formatter/Description.vue
richard-cox: note - most cloud credentials are validated before being created (making it harder to create test entries), however `S3-Compatible` is not (can be used to test the table).

a110605: It's hard to calucate the triangle position based on text length. 
Change placement to `auto` to let component self-decide how to render the tooltip.


https://github.com/user-attachments/assets/7c02a260-670d-4bc2-a7f3-a1693bc88e0e

richard-cox (approved):

PR #14460, opened 2025-06-04
Add a button to quickly install helm charts outside the apps and marketplace

### LINKED ISSUE #13105: Add a generic component to install specific charts outside the apps and marketplace context
Our UI extensions often require a helm chart installed alongside them: elemental, kubewarden, capi, virtual clusters. It would be nice if, upon installing the UI extension, users that didn't have the required app were notified of that and offered an easy way to install it from whatever page they're currently on, without navigating into the cluster explorer->apps and marketplace. 

A generic modal component that extensions can use to install specific helm apps with minimal configuration. Users should be able to pick which cluster to install the app in, without necessarily being in the cluster explorer context: ~eg, from the cluster management, I would like to be able to install the k3k chart into any downstream cluster~ eg, from the CAPI turtles landing page, I would like to be able to install the turtles chart.

**Additional context**
Kubewarden already has something like this -- we should look at making the code there more generic, as it needs to be used across extensions.

mantis-toboggan-md:
### Summary
Fixes #13105 

This PR adds a component to install charts outside the apps and marketplace with minimal configuration. 

Test using https://github.com/rancher/capi-ui-extension/pull/169

none

### Screenshot/Video

https://github.com/user-attachments/assets/748bdf28-80e7-4dff-8f6c-f18696f17662

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -1141,6 +1141,7 @@ catalog:
           error: Error installing {chart}
       viewLogs: View Installation Logs
       customizeInstall: Customize Install Options
+      waitForLogs: Installation initialized; waiting for logs...
     chart: Chart
     warning:
       managed:
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -1144,7 +1144,7 @@ catalog:
         installChart:
           action: Install {chart}
           waiting: Installing {chart}...
-          success: '{chart} installation  started'
+          success: '{chart} installation started'
           error: Error installing {chart}
         waitForLogs:
           action: View installation logs
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -1122,6 +1122,25 @@ catalog:
       goToUpgrade: Edit/Upgrade
     appReadmeMissing: This chart doesn't have any additional chart information.
     appReadmeTitle: Chart Information (Helm README)
+    button:
+      stages:
+        addRepo:
+          action: Add the {repo} repository
+          waiting: Adding {repo} repository...
+          success: Repository added
+          error: Error adding {repo} repository
+        loadCharts:
+          action:  Fetch {repo} repository chart(s)
+          waiting: Fetching {repo} repository chart(s)...
+          success: '{chart} found'
+          error: '{chart} chart not found - Try again?'
+        installChart:
+          action: Install {chart}
+          waiting: Installing {chart}...
+          success: '{chart} installation  initialized'
+          error: Error installing {chart}
+      viewLogs: View Installation Logs
+      customizeInstall: Customize Install Options
     chart: Chart
     warning:
       managed:
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -1123,6 +1123,13 @@ catalog:
     appReadmeMissing: This chart doesn't have any additional chart information.
     appReadmeTitle: Chart Information (Helm README)
     button:
+      alreadyInstalled: '{chart} version {version} is already installed.'
+      viewDetails: View chart details
+      customizeInstall: Customize Install Options
+      viewLogs: View Installation Logs
+      waitForLogs: Installation initialized; waiting for logs...
+      canNotCreateRepos: It looks like you don't have permission to add repositories and install helm charts. 
+      canNotInstallApps: The {repo} repository has been added, but you don't have permission to install charts from it.
       stages:
         addRepo:
           action: Add the {repo} repository
@@ -1137,11 +1144,16 @@ catalog:
         installChart:
           action: Install {chart}
           waiting: Installing {chart}...
-          success: '{chart} installation  initialized'
+          success: '{chart} installation  started'
           error: Error installing {chart}
-      viewLogs: View Installation Logs
-      customizeInstall: Customize Install Options
-      waitForLogs: Installation initialized; waiting for logs...
+        waitForLogs:
+          action: View installation logs
+          waiting: Waiting for installation logs...
+          error: Unable to load installation logs
+      errors:
+        fetchChart: Error retrieving chart from repository
+        alreadyInstalled: Chart already installed
+        timeoutWaiting: Timed out waiting for logs
     chart: Chart
     warning:
       managed:
--- a/shell/components/InstallHelmCharts.vue
+++ b/shell/components/InstallHelmCharts.vue
@@ -12,7 +12,7 @@ import AsyncButton from '@shell/components/AsyncButton';
 import { isPrerelease } from '@shell/utils/version';
 import { nextTick } from 'vue';
 import { set } from '@shell/utils/object';
-import { defaultCmdOpts } from '@shell/pages/c/_cluster/apps/charts/install.vue';
+import { defaultCmdOpts as defaultOpts } from '@shell/pages/c/_cluster/apps/charts/install.vue';
 
 /**
  * Assumptions made by this component
@@ -23,6 +23,8 @@ import { defaultCmdOpts } from '@shell/pages/c/_cluster/apps/charts/install.vue'
  *  - if the app is already installed
  *  - if the user has permission to install apps
  */
+
+const defaultCmdOpts = { ...defaultOpts, timeout: '600s' };
 
 const MAX_TRIES = 5;
 const RETRY_WAIT = 500;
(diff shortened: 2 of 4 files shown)


richard-cox:
This concept of bootstrapping is used in kubewarden, elemental and harvester (that one though is just to install the extension). Is the plan to convert those over to use this one? 

Worth running the video by UX as well.

mantis-toboggan-md:
> This concept of bootstrapping is used in kubewarden, elemental and harvester (that one though is just to install the extension). Is the plan to convert those over to use this one?
> 
> Worth running the video by UX as well.

Can't believe I forgot to include them here- tagging in UX now


Yes, the plan is to have a shared component to use across extensions though I'm not 100% certain on the harvester use case

shell/components/InstallHelmCharts.vue
```diff
@@ -0,0 +1,637 @@
...
+    }
+
+  },
+
+}
+;
+</script>
+
+<template>
+  <Loading v-if="$fetchState.pending" />
```
eva-vashkevich, line 589: This components is not imported

shell/components/InstallHelmCharts.vue
```diff
@@ -0,0 +1,637 @@
...
+      this.operation = await this.$store.dispatch(`${ this.store }/find`, {
+        type: CATALOG.OPERATION,
+        id:   operationId
+      });
+
+      try {
+        await this.operation.waitForLink('logs');
+        this.logsReady = true;
+      } catch (e) {
+        // The wait times out eventually, move on...
```
eva-vashkevich, line 431: I think we need to display an error here. Sometimes, if I try to install Turtles with a user that doesn't have sufficient permissions ie standard user, it just gets stuck with 'Installation initialized; waiting for logs...' forever. Other times, it does show me the logs that contain an error. I am not quite able to tell what leads to one outcome vs the other (other than luck and timing)

eva-vashkevich (changes requested):
I think we should hide this page if user doesn't have sufficient permissions if possible. I don't think relying on consuming components to do it is sufficient. This component should be able to behave intuitively independently from how it is consumed. Otherwise, you won't be able to install the charts and we do not show it in any way, and clicking on 'Customize Install Options' just takes you to Home page.

shell/assets/translations/en-us.yaml
```diff
@@ -1046,6 +1046,26 @@ catalog:
...
+          error: Error adding {repo} repository
+        loadCharts:
+          action:  Fetch {repo} repository chart(s)
+          waiting: Fetching {repo} repository chart(s)...
+          success: '{chart} found'
+          error: '{chart} chart not found - Try again?'
+        installChart:
+          action: Install {chart}
+          waiting: Installing {chart}...
+          success: '{chart} installation  initialized'
```
eva-vashkevich: I am being extra nit picky, but there is an extra space in this sentence

shell/assets/translations/en-us.yaml
```diff
@@ -1046,6 +1046,26 @@ catalog:
...
+          error: Error installing {chart}
+      viewLogs: View Installation Logs
+      customizeInstall: Customize Install Options
+      waitForLogs: Installation initialized; waiting for logs...
```
eva-vashkevich: This might be extra extra nit picky and more of a preference, but it is better to use simpler language when possible in case our users aren't native speakers. IMO, "Stared the installation" or "Installation started" is better (but it is up to you)

shell/components/InstallHelmCharts.vue
```diff
@@ -0,0 +1,637 @@
...
+          // some errors here are expected
+          // we only show error state when while block finishes and still no chart found
+        }
+      }
+
+      this.stages.loadCharts.loading = false;
+
+      // tried all our tries and the chart still isn't found - tell users something has gone wrong
+      if (!this.chart) {
+        this.btnCb(false);
```
eva-vashkevich, line 313: Should we also show an error saying that we couldn't find the chart?

shell/components/InstallHelmCharts.vue
```diff
@@ -0,0 +1,637 @@
...
+    },
+
+    async installChart() {
+      this.stages.configureChart.errors = '';
+      this.stages.installChart.loading = true;
+
+      if (this.stages.installChart.done) {
+        this.stages.installChart.loading = false;
+
+        this.stages.installChart.error = 'Chart already installed';
```
eva-vashkevich: we should use translation

shell/components/InstallHelmCharts.vue
mantis-toboggan-md: Screenshot 2025-06-10 at 10 44 42 AM

Yes. I've updated the component to show an error if the chart can't be found. You can test this by modifying the capi-ui-extension example to use an invalid chart name prop, eg   `chart-name="rancher-turtles-aaaa"`

Note that it takes time for this error to show up, because we try reloading repositories a few times before giving up.

shell/components/InstallHelmCharts.vue
mantis-toboggan-md: I've added a few permissions checks before the user ever reaches this stage. Now the component checks:
- if the user can create repositories
- if the user can see apps (I don't think checking create permissions on this schema makes sense as we're technically not creating app resources directly)
- if the user has the 'install' action on this specific repo

By default standard users can see cluster repositories but can't install charts in the local cluster. So they'll pass the first of the above checks but then see a message saying they dont have permission to install this chart.  This message is contained in a slot, so parent components can choose to render nothing at all if that makes more sense in some particular context.
Screenshot 2025-06-10 at 10 41 23 AM

shell/assets/translations/en-us.yaml
mantis-toboggan-md: That's fair. I changed initialized to started.

mantis-toboggan-md:
I've updated the component to improve error handling and permission checks. I moved the wait for logs/view logs actions into the same asyncbutton the installation process uses to make the component look more cohesive. I've also refactored the code used to determine which label and phase to show in the async button so it is hopefully a bit easier to follow.


I rewrote how this component displays errors a bit so it emits growls instead of showing banners - I think this will make the component easier to drop into a variety of pages. The component still tracks error messages locally and has a slot that'll be rendered when there are errors so the parent component can display them if needed.

https://github.com/user-attachments/assets/72bdebd5-e415-49d7-812a-f3659c84b699

shell/components/InstallHelmCharts.vue
```diff
@@ -0,0 +1,661 @@
...
+      }
+
+      // server url and project ids are used in global values
+      try {
+        this.serverUrlSetting = await this.$store.dispatch(`${ this.store }/find`, {
+          type: MANAGEMENT.SETTING,
+          id:   SETTING.SERVER_URL,
+        });
+      } catch (e) {
+        console.error('Unable to fetch `server-url` setting: ', e); // eslint-disable-line no-console
```
eva-vashkevich: Can this be a growl instead?

shell/components/InstallHelmCharts.vue
```diff
@@ -0,0 +1,661 @@
...
+      WAIT:        'waitForLogs'
+    };
+
+    return {
+      throttledRefreshCharts: null,
+
+      // check the cluster repo schema
+      canCreateRepos:   false,
+      // once the target repo is created/found, we'll check that the user can perform the install action with that specific repo
+      // canInstallChart:  true,
```
eva-vashkevich: was this line left intentionally?

mantis-toboggan-md: No; I made this a computed prop and overlooked removing the comment

shell/components/InstallHelmCharts.vue
mantis-toboggan-md: yes. I also made it use the error object that is caught instead of an un-localized string

shell/components/InstallHelmCharts.vue
```diff
@@ -0,0 +1,660 @@
...
+
+        res = await this.targetRepo.doAction('install', this.installCmd);
+      } catch (err) {
+        this.btnCb(false);
+        this.errors.push(err);
+        this.doingButtonAction = false;
+      }
+
+      this.installOperationName = res.operationName;
+      this.installOperationNamespace = res.operationNamespace;
```
eva-vashkevich, line 407: if error occurs and res is undefined, this throws a console error

shell/components/InstallHelmCharts.vue
```diff
@@ -0,0 +1,660 @@
...
+
+        this.doingButtonAction = false;
+      } else {
+        this.btnCb(true);
+      }
+    },
+
+    // once a chart is located fetch installation info from the repo
+    async fetchVersionInfo() {
+      if (!this.doingButtonAction) {
```
eva-vashkevich: Should we also just return if !this.chart ?

shell/components/InstallHelmCharts.vue
```diff
@@ -0,0 +1,660 @@
...
+
+          this.errors.push(e);
+          this.doingButtonAction = false;
+
+          this.btnCb(false);
+
+          return;
+        }
+
+        await nextTick();
```
eva-vashkevich: why do we need to wait for nextTick here?

shell/components/InstallHelmCharts.vue
eva-vashkevich: Also, why do we need to return if !this.doingButtonAction ? Could you please add a comment explaining this?

shell/components/InstallHelmCharts.vue
mantis-toboggan-md: We don't anymore - good catch. Before refactoring following review comments this was 

```
        this.stages.addRepo.loading = false;
        this.stages.addRepo.done = true;
        this.stages.loadCharts.loading = true;
        this.stages.loadCharts.error = '';
        await nextTick();
```

`nextTick` here allowed computed properties dependent on `this.stages` recalculate before proceeding. Which was needed because the component was badly written and doing way too much in one computed property ostensibly used to label a button.

shell/components/InstallHelmCharts.vue
mantis-toboggan-md: You know what I don't think this if block makes sense either and I think I must have left it in by mistake when I was refactoring the button stage logic. I removed it. Thanks for catching that.

We don't need to check that `this.chart` is defined here because this function is now _only_ called by a watcher on `chart` 

```
 chart: {
      handler(neu) {
        if (neu) {
          this.fetchVersionInfo();
        }
      },
      immediate: true
    },
```

PR #14463, opened 2025-06-04
Separated creating new namespace from createNamespace check

### LINKED ISSUE #14462: Add a way to create a namespace without having to use "Add namespace" in NameNsNamespace

Currently, we do not have a standardized way to create a new namespace while creating other resources
Needed for [85](https://github.com/rancher/capi-ui-extension/issues/85)

Add a way to work around createNamespace state if needed

**Additional context**

eva-vashkevich:
### Summary
Fixes #14462 

Separated new namespace creation logic from a check to see if CruResource should create a new namespace.

### Technical notes summary

CAPI providers usually have pre-defined namespaces they are supposed to be created at. Those namespaces usually do not exist if provider has not been installed before. Without this change, it is impossible to pre-set the namespace we want to create and we would have to rely on users to know the correct namespace name and create it using 'Create New Namespace' functionality of the NameNsDescription control, which would set createNamespace state to true. We want to avoid having to rely on the user to do this, however, there was no other way to overwrite the value of createNamespace check. With this change, other components can call createNamespaceIfNeeded method directly, sidestepping createNamespace check if needed.

Make sure that NameNsDescription functionality hasn't changed:
- You can create a new namespace if 'Create a New Namespace' is selected. 
- For components that do not allow in-line creation of a new namespace, no new namespace is created

This could technically affect any component that utilizes CruResource

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/CruResource.vue
+++ b/shell/components/CruResource.vue
@@ -371,8 +371,6 @@ export default {
     },
 
     async clickSave(buttonDone) {
-      let errorOccurred = false;
-
       if (this.createNamespace) {
         try {
           await this.createNamespaceIfNeeded();
@@ -381,16 +379,15 @@ export default {
         // show any applicable errors if the namespace is
         // invalid.
           this.$emit('error', exceptionToErrorsArray(err.message));
-          errorOccurred = true;
           buttonDone(false);
+
+          return;
         }
       }
 
       // If the attempt to create the new namespace
       // was successful or no ns needs to be created, save the resource.
-      if (!errorOccurred) {
-        this.$emit('finish', buttonDone);
-      }
+      this.$emit('finish', buttonDone);
     },
 
     save() {
--- a/shell/components/CruResource.vue
+++ b/shell/components/CruResource.vue
@@ -371,19 +371,21 @@ export default {
     },
 
     async clickSave(buttonDone) {
-      try {
-        await this.createNamespaceIfNeeded();
-
-        // If the attempt to create the new namespace
-        // is successful, save the resource.
-        this.$emit('finish', buttonDone);
-      } catch (err) {
+      if (this.createNamespace) {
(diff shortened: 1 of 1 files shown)


shell/components/CruResource.vue
```diff
@@ -371,19 +371,21 @@ export default {
...
-        this.$emit('error', exceptionToErrorsArray(err.message));
-        buttonDone(false);
+          this.$emit('error', exceptionToErrorsArray(err.message));
+          buttonDone(false);
+        }
       }
+
+      // If the attempt to create the new namespace
+      // was successful or no ns needs to be created, save the resource.
+      this.$emit('finish', buttonDone);
```
momesgin, line 390: I might be wrong, but I think this will be called even if there's an error which is different from previous behaviour

momesgin: I just tested
image

eva-vashkevich: Good catch, I did not think about this and assumed buttonDone would prevent further action. Should be fixed now

shell/components/CruResource.vue
```diff
@@ -371,18 +371,25 @@ export default {
...
+          await this.createNamespaceIfNeeded();
+        } catch (err) {
         // After the attempt to create the namespace,
         // show any applicable errors if the namespace is
         // invalid.
-        this.$emit('error', exceptionToErrorsArray(err.message));
-        buttonDone(false);
+          this.$emit('error', exceptionToErrorsArray(err.message));
+          errorOccurred = true;
+          buttonDone(false);
```
momesgin, line 382: wouldn't it be better if we exit(`return`) early here instead of creating a new variable `errorOccurred`?

momesgin (approved):
LGTM

PR #14491, opened 2025-06-11
Add the card components necessary for the new detail page

codyrancher:
### Summary
Implementation of the card components described in `@confluence /display/CU/Resource+page+redesign` ask for a link if you'd like it.

This change will not affect any rendering currently in master

Here's an example of the usage:
https://github.com/rancher/dashboard/pull/14210/files#diff-7d0976f24eefa02ccce988873ffb94ec5ceec0ae290b89c7ef187890be99489f

Adding new components for the detail page redesign.

### Technical notes summary
I'm breaking up the changes into smaller easier to review chunks. This means that none of this code will have an affect on rendering in master until later PRs are merged.

None, there's unit tests for the cards.

None should

### Screenshot/Video
ResourceUsageCard:
image

PodsCard and StateCards:
image

### Checklist
- [x] The PR has a Milestone 


jest.setup.js
```diff
@@ -64,19 +64,14 @@ beforeAll(() => {
     }))
   });
 });
-jest.mock('@shell/composables/useI18n', () => ({ useI18n: () => (key) => key }));
+
```
codyrancher: This was getting overriden below.

jest.setup.js
```diff
@@ -64,19 +64,14 @@ beforeAll(() => {
...
-// jest.mock('@shell/composables/useI18n', () => ({ useI18n: () => (key) => key }));
 jest.mock('@shell/composables/useI18n', () => {
   return {
     useI18n() {
-      return {
-        t(key) {
-          return key;
-        }
-      };
+      return { t: (key, options) => `${ key }${ options ? `-${ JSON.stringify(options) }` : '' }` };
```
codyrancher, line 75: There were times we wanted to also see the arguments for the string parameters,

shell/assets/styles/base/_spacing.scss
```diff
@@ -0,0 +1,29 @@
+// Text indent, margins, and paddings from 4-40px in 4px increments
```
codyrancher, line 1: We have new spacing requirements. I added in these helper classes which closely reflect what we already have but with new values.

shell/assets/translations/en-us.yaml
```diff
@@ -8221,6 +8225,39 @@ errors:
   notFound:
     withUrl: '{msg}: {url}'
     withoutUrl: '{msg}'
-
-
-
+    
+component:
```
codyrancher, line 8257: Some of these aren't currently used in this PR but it's easier to push all of them now instead of breaking it up.

shell/components/Resource/Detail/Card/ResourceUsageCard/composable.ts
```diff
@@ -0,0 +1,51 @@
+import { Props } from '@shell/components/Resource/Detail/Card/ResourceUsageCard/index.vue';
```
codyrancher, line 1: These are placeholders, I'll be adding the implementation once I work on the cluster dashboard page.

shell/components/Resource/Detail/Card/StateCard/composables.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  const rows = useDefaultResources(remainingPairs);
+
+  return {
+    title: 'Resources',
+    rows:  rows.value
+  };
+}
+
+export function useDefaultWorkloadInsightsCardProps(): StateCardProps {
```
codyrancher, line 89: Just a placeholder. I still need to implement this for the insights card.

jest.config.js
```diff
@@ -44,6 +44,7 @@ module.exports = {
   snapshotSerializers: ['<rootDir>/node_modules/jest-serializer-vue'],
 
   // Coverage
+  coverageProvider:    'v8',
```
codyrancher, line 47: Without this we can't run coverage and have the two `<script>` tags in one file. Without this jest will complain about having a duplicate identifier.

shell/components/Resource/Detail/Card/Scaler.vue
```diff
@@ -0,0 +1,80 @@
...
+
+<template>
+  <div class="scaler">
+    <button
+      class="decrease"
+      :disabled="!!props.min && (props.value <= props.min)"
+      @click="() => emit('decrease', props.value - 1)"
+    >
+      <i class="icon icon-sm icon-minus" />
+    </button>
```
rak-phillip: We need an accessible label for these buttons so that screen readers communicate their function.

shell/assets/styles/base/_spacing.scss
```diff
@@ -0,0 +1,29 @@
...
+@each $keyword, $property in $spacing-property-map {
+    .#{$keyword}-0 {
+        #{$property}: 0 !important;
+    }
+
+    @for $size from 1 through 10 {
+        $val: $size * 4;
+
+        .#{$keyword}-#{$size} {
+            #{$property}: $val * 1px !important;
```
rak-phillip: nit: What is the significance of using `!important` for the spacing? `!important` is _usually_ not needed and the same effect can be achieved by increasing specificity.

shell/components/Resource/Detail/Card/PodsCard/composable.ts
```diff
@@ -0,0 +1,32 @@
...
+          err
+        },
+        { root: true });
+      }
+    };
+  };
+
+  const workloadNameValue = toValue(workloadName);
+  const scaleUp = computed(() => scale(() => scaleUpFn(), workloadNameValue, 'up'));
+  const scaleDown = computed(() => scale(() => scaleDownFn(), workloadNameValue, 'down'));
```
rak-phillip: What do we gain out of `scaleUp` and `scaleDown` being computed properties? It looks like both of these would be better suited as methods, without the need for caching.

shell/components/Resource/Detail/Card/PodsCard/index.vue
```diff
@@ -0,0 +1,108 @@
...
+import VerticalGap from '@shell/components/Resource/Detail/Card/VerticalGap.vue';
+import StatusBar from '@shell/components/Resource/Detail/StatusBar.vue';
+import StatusRow from '@shell/components/Resource/Detail/StatusRow.vue';
+import { useI18n } from '@shell/composables/useI18n';
+import { StateColor } from '@shell/utils/style';
+import { computed } from 'vue';
+import { useStore } from 'vuex';
+
+export interface Props {
+  pods?: any[];
```
rak-phillip, line 13: Can we specify a better type than `any[]` for `pods`?

shell/components/Resource/Detail/StatusRow.vue
```diff
@@ -0,0 +1,63 @@
...
+} = defineProps<Props>();
+</script>
+
+<template>
+  <div class="status-row">
+    <div
+      class="indicator"
+      :style="{backgroundColor: stateColorCssVar(color)}"
+    >
+&nbsp;
```
rak-phillip: Can we define our spacing via css instead of using `&nbsp`?

shell/components/Resource/Detail/Card/StateCard/composables.ts
```diff
@@ -0,0 +1,111 @@
+import { extractCounts, Props as ResourceRowProps } from '@shell/components/Resource/Detail/ResourceRow.vue';
+import { useI18n } from '@shell/composables/useI18n';
+import { INGRESS, SERVICE } from '@shell/config/types';
+import { getHighestAlertColor } from '@shell/utils/style';
+import { computed, Ref, toValue } from 'vue';
+import { useStore } from 'vuex';
+import { Props as StateCardProps } from '@shell/components/Resource/Detail/Card/StateCard/index.vue';
+
+export function usesResourceCardRow(label: string, resources: any[], to?: any): ResourceRowProps {
```
rak-phillip: The `uses` naming is inconsistent with the rest of the composables used throughout this file

```suggestion
export function useResourceCardRow(label: string, resources: any[], to?: any): ResourceRowProps {
```

shell/components/Resource/Detail/Card/StateCard/composables.ts
```diff
@@ -0,0 +1,111 @@
...
+    color:  resources.length ? getHighestAlertColor(colors) : undefined,
+    counts: resources.length ? extractCounts(states) : undefined,
+    to
+  };
+}
+
+export interface Pairs {
+    label: string;
+    to?: any;
+    resources: any[];
```
rak-phillip, line 25: I think that we should strive to provide more robust types than `any` when possible.

shell/components/Resource/Detail/Card/StateCard/composables.ts
```diff
@@ -0,0 +1,111 @@
...
+}
+
+export function useDefaultResources(pairs: Ref<Pairs[]>) {
+  const pairsValue = toValue(pairs);
+  const rows = computed(() => pairsValue.map(({ label, resources, to }) => usesResourceCardRow(label, resources, to)));
+
+  return rows;
+}
+
+export function useResourcePair(label: string, type: string, resources?: any[]) {
```
rak-phillip, line 35: Many of these composables can benefit from doc strings to detail usage. It's not immediately clear to me what a `ResourcePair` is, but it looks like we're preparing data to be used with a router.

shell/components/Resource/Detail/PercentageBar.vue
```diff
@@ -0,0 +1,45 @@
...
+const usedStyle = computed(() => ({
+  backgroundColor: props.percentageColor,
+  width:           `${ props.percent }%`
+}));
+</script>
+
+<template>
+  <div
+    class="percentage-bar"
+    :style="baseStyle"
```
rak-phillip: optional: we can consider using `v-bind` in CSS instead of binding to inline styles[^1].

[^1]: https://vuejs.org/api/sfc-css-features.html#v-bind-in-css

shell/components/Resource/Detail/PercentageBar.vue
```diff
@@ -0,0 +1,45 @@
...
+  >
+    <div
+      class="used"
+      :style="usedStyle"
+    >
+&nbsp;
```
rak-phillip: nit: can we replace `&nbsp` usage with css?

shell/components/Resource/Detail/ResourceRow.vue
```diff
@@ -0,0 +1,119 @@
...
+
+  labels.forEach((l: string) => {
+    accumulator[l] = accumulator[l] || 0;
+    accumulator[l]++;
+  });
+
+  const counts: Count[] = Object.entries(accumulator).map(([label, count]) => ({ label, count }));
+
+  return sortBy(counts, 'label');
+}
```
rak-phillip, line 30: optional: have we considered the tradeoffs in moving functions that we export into composables? I'm unsure of what the overall design implication will be for making such a change, but it feels more natural (imo) for components to import from composables as opposed to the inverse. I'm basing this question off the fact that `shell/components/Resource/Detail/Card/StateCard/composables.ts` is the only usage of `extractCounts` that I see in this PR.

shell/components/Resource/Detail/ResourceRow.vue
```diff
@@ -0,0 +1,119 @@
...
+        <span
+          v-if="color"
+          class="dot"
+          :style="{backgroundColor: stateColorCssVar(color)}"
+        >
+&nbsp;
+        </span>
+        <span
+          v-for="(count, i) in counts"
+          :key="i"
```
rak-phillip: We generally want to avoid using the index as a key in `v-for`, but I think this looks safe; the content appears to be static.

shell/assets/styles/base/_spacing.scss
codyrancher: This was directly lifted from https://github.com/rancher/dashboard/blob/master/shell/assets/styles/base/_helpers.scss#L9-L30


But you're right, `!important` really shouldn't be needed. I'll remove it from mine but leave the original in until we migrate everything to the new spacing.

shell/components/Resource/Detail/Card/PodsCard/index.vue
codyrancher: As far as I know all of our models are JS without interfaces so I don't think there's a reasonable alternative unless that changes. I think that's out of scope for this change though.

shell/components/Resource/Detail/StatusRow.vue
codyrancher: We do actually, we have the `.indicator` class defined below.

I think this may have been a span before and I added the content so the element wouldn't collapse. 

I'll go ahead and remove it.

shell/components/Resource/Detail/PercentageBar.vue
codyrancher: done.

shell/components/Resource/Detail/ResourceRow.vue
codyrancher: This one has been on a journey, so here's some details:

Ultimately I would like as many components as possible to be stateless (exception of translations) and fully configurable through props. I would also like pages to be relatively easy to build without implicitly passing things down.

To achieve that I originally created functions which would return default props that we could pass to the components. This turned out to not be reactive so I went ahead and made composables with refs and computed props.

To me, the example I linked, seems intuitive. 
image

All data is flowing in one direction, everything but the page is stateless and it's still compact compared to how we  previously defined the pages.

I'm totally happy to change this if there's a more idiomatic way to achieve the same goal. Do you have

shell/components/Resource/Detail/ResourceRow.vue
codyrancher: We can just go with the count.label for the key. I'll switch it up.

shell/components/Resource/Detail/Card/PodsCard/composable.ts
codyrancher: Yeah, you're right. I've switched them to methods. I just got too computed happy after some reactivity issue chasing.

shell/components/Resource/Detail/Card/StateCard/composables.ts
codyrancher: Fixed. I did it on my end so I could do all the changes locally.

shell/components/Resource/Detail/Card/StateCard/composables.ts
codyrancher: I think I've taken care of all instances except for resources.

shell/components/Resource/Detail/Card/StateCard/composables.ts
codyrancher: I agree. I'd like to take care of this later though. These are still in flux and based on your other comment we may want to move away from composables.

shell/components/Resource/Detail/PercentageBar.vue
codyrancher: I didn't know about that feature. Thanks, I've updated it. I know there's some other style props in this pr but I don't think it works in the other locations.

shell/components/Resource/Detail/Card/Scaler.vue
codyrancher: Added
image

shell/components/Resource/Detail/Card/StateCard/composables.ts
rak-phillip: > These are still in flux and based on your other comment we may want to move away from composables.

My intent was to encourage moving more logic into composables 🙂. But I agree, we can postpone docs for the time being if the design is subject to change.

codyrancher: Ah, I must've misunderstood. Well I'm definitely in favor of more composables.

pkg/aks/components/Config.vue
```diff
@@ -881,6 +881,7 @@ export default defineComponent({
         :show-tabs-add-remove="mode !== 'view'"
         :rules="fvGetAndReportPathRules('vmSize')"
         class="mb-20"
+        :use-hash="false"
```
rak-phillip: How do the tabbed components for aks/eks/gke fit into this PR?

shell/components/Resource/Detail/ResourceRow.vue
rak-phillip: My vote is that we leave it for now. There's no issues caused by the current implementation and we can refactor if we find that the relationship between composables and components can be improved. 

My intuition tells me that the intended relationship is:

- Components import composables
- Composables import composables

While it's possible for composables to import components, this feels like a bit of an anti-pattern to me. My concern is that another component might want to import functionality specified in a composable and we can end up with an inadvertent transitive dependency to the original component imported by the composable.

rak-phillip:
@momesgin can you take a look at this as well? I want to make sure that there isn't overlap with the recent card work that you've done.

pkg/aks/components/Config.vue
codyrancher: It doesn't. I believe these got picked up from a rebase trying to fix merge conflicts.

https://github.com/rancher/dashboard/commit/ff397e327e1a70542c9f1784b6be69a07d20b7d1

I'm honestly a little confused because my commit doesn't show these locally.

codyrancher: This should now be resolved

shell/components/Resource/Detail/Card/ResourceUsageCard/composable.ts
```diff
@@ -0,0 +1,51 @@
...
+import { useI18n } from '@shell/composables/useI18n';
+import { computed } from 'vue';
+import { useStore } from 'vuex';
+
+export const useCpuUsageCardDefaultProps = (resource: any): Props => {
+  const store = useStore();
+  const i18n = useI18n(store);
+
+  //   const resourceValue = toValue(resource);
```
momesgin, line 10: testing leftover?

shell/components/Resource/Detail/Card/ResourceUsageCard/composable.ts
```diff
@@ -0,0 +1,51 @@
...
+      available: 1.93
+    };
+  }).value;
+};
+
+export const useMemoryUsageCardDefaultProps = (resource: any): Props => {
+  const store = useStore();
+  const i18n = useI18n(store);
+
+  //   const resourceValue = toValue(resource);
```
momesgin, line 25: same here

shell/components/Resource/Detail/Card/ResourceUsageCard/composable.ts
```diff
@@ -0,0 +1,51 @@
...
+      unit:               'GiB',
+    };
+  }).value;
+};
+
+export const usePodUsageCardDefaultProps = (resource: any): Props => {
+  const store = useStore();
+  const i18n = useI18n(store);
+
+  //   const resourceValue = toValue(resource);
```
momesgin, line 42: same

shell/components/Resource/Detail/StatusBar.vue
```diff
@@ -0,0 +1,59 @@
...
+    flex-direction: row;
+    justify-content: center;
+
+    column-gap: 2px;
+    height: 21px;
+
+    .segment {
+        height: 4px;
+
+        &:first-of-type {
```
momesgin, line 48: nice! didn't know about it

shell/components/Resource/Detail/Card/ResourceUsageCard/composable.ts
codyrancher: I commented above

image

momesgin: sorry, I missed that comment

PR #14505, opened 2025-06-12
Fix actions support in RcItemCard

### LINKED ISSUE #14497: RcItemCard: using the prop "actions" (ActionMenuShell custom-actions) is not transmitting events correctly
**Setup**
- Rancher version: 2.12-head
- Rancher UI Extensions:
- Browser type & version:

**Describe the bug**

When using the prop "actions" (ActionMenuShell custom-actions) is not transmitting events correctly. With this we can't use the embedded menu actions properly. Workaround is to use the slot and define there the same `ActionMenuShell`, which works as expected.

Also, the TS type for `actions` is wrong: should be an array of actions and not a simple object https://github.com/rancher/dashboard/blob/master/pkg/rancher-components/src/components/RcItemCard/RcItemCard.vue#L83

**To Reproduce**

- Define some table actions, like:
```
[
   {
          action: 'regenerateSecret',
          label:  this.t('oidcclient.regenerate')
    }
]
```

- it's expected that `RcItemCard` would emit an event called `regenerateSecret`, but it not being propagated up to it's parent 

**Result**
- No event it's propagated to the parent container from `ActionMenuShell`, therefore rendering the embedded `ActionMenuShell` useless

**Expected Result**

- Event should propagated to the parent container from `ActionMenuShell` so that we can use the embedded menu

**Screenshots**

**Additional context**

momesgin:
### Summary
Fixes #14497 

### Technical notes summary

- Replaced `div` with `rc-item-card-action` inside `RcItemCard.vue`
- Unit tests

QA can be omitted for this issue since the first real usage of this change will be on the Cluster Tools page which will test this change too.

Fleet Dashboard uses RcItemCard, make sure nothing is broken on the cards visually or functionally 

### Screenshot/Video

n/a
### Checklist
- [x] The PR has a Milestone 


momesgin:
@rak-phillip @aalves08 this was waiting for the master branch to be available after 2.12 release, it's ready for review now 🙏

pkg/rancher-components/src/components/RcItemCard/RcItemCard.vue
```diff
@@ -79,8 +80,24 @@ interface RcItemCardProps {
...
+   *     {
+   *       action: 'focusSearch',
+   *       label: t('catalog.charts.search'),
+   *       enabled: true
+   *     }
+   *   ]"
+   *   @focusSearch="focusSearch"
+   * />
+   */
+   actions?: DropdownOption[];
```
rak-phillip: nit: it looks like actions has an extra space in front of it. I think this it noteworthy because it looks like the spacing rules aren't being applied within interfaces in vue components. Ultimately, this is unrelated to this PR.

pkg/rancher-components/src/components/RcItemCard/RcItemCard.vue
```diff
@@ -370,7 +404,14 @@ $image-medium-box-width: 48px;
...
     &-action-menu {
-      margin-left: 12px;
+      margin-left: 8px;
+
+      &:deep(.v-popper) {
+        .icon-actions {
+          color: var(--body-text);
+          font-size: 16px;
+        }
+      }
```
rak-phillip: Is this change related to this PR? The associated issue and PR description doesn't mention icon styling.

We generally want to avoid using deep selectors where possible - their usage is typically an indicator that we are missing some sort of feature or functionality in our components.

pkg/rancher-components/src/components/RcItemCard/RcItemCard.vue
momesgin: I didn't see that!

pkg/rancher-components/src/components/RcItemCard/RcItemCard.vue
momesgin: In my opinion it should be fine to include this styling since it's solely related to the actions icon, however, I'm gonna add this to the ticket too for a better history record. 

Unfortunately using `deep` selector is needed here.

rak-phillip: Can you elaborate on why the deep selector is needed here? Deep selectors can be prone to breaking if the underlying implementation changes, and they can make styles more difficult to maintain over time. They should generally be used as a last resort for providing overrides when other alternatives are not feasible.

To reiterate, deep selector usage is typically an indicator that we are missing some sort of feature or functionality in our components. 

This change leads me to believe that one of the following scenarios might be true:

1. The default style for the icon action may be incorrect and needs to be updated to maintain consistency
2. The underlying `ActionMenuShell`/`RcDropdownMenu` implementation is missing some interface for customizing the icon/button in a controlled way

momesgin: You're right, I hadn’t considered updating the `ActionMenuShell` or `RcDropdownMenu` directly. My initial thinking was that since deep selectors are used in various parts of the codebase already, handling this individually as tech debt might be acceptable in the short term. I'll revert this part for now so I can address it separately.

momesgin: done, appreciate the detailed feedback and your emphasis on doing things the right way

rak-phillip (approved):
LGTM

PR #14557, opened 2025-06-23
14144 s3 endpoint field in cluster configuration missing input validation

### LINKED ISSUE #14144: S3 endpoint field in cluster configuration missing input validation
**Setup**
- Rancher version: 2.9.2
- Rancher UI Extensions:
- Browser type & version:

Internal ref: SURE-10082

**Describe the bug**

When a cluster has etcd snapshot backups configured with S3, the `endpoint` field accepts any input. It seems that using input that includes https protocol in the URL breaks the snapshot backup saving procedure silently. There is no UI error, but backups are not being saved. Only the rke2-server log shows the error.

**To Reproduce**

- Create a guest cluster in Harvester
- Configure etcd with "Backup snapshots to S3"
- Use endpoint in format "https://..."

**NOTE: The area in question is the cluster provisioning/edit interface**

**Result**

When protocol is used in S3 endpoint of etcd backup there is no error and snapshots are not saved to S3

**Expected Result**

When the format of S3 endpoint in etcd backup configuration is not correct, there should be an error. Correct format should be documented/suggester by UI

**Screenshots**

**Additional context**

marytlf:
### Summary
The core requirement was to prevent users from entering "http://" or "https://" at the beginning of the S3 endpoint URL in a form, and consequently, disable the main "Save" button of the cluster creation/edit form. Additionally, a tooltip should appear next to the endpoint field to inform the user about this specific error.

Fixes #14144 

Check if value typed on endpoint form is http:// or https://, in case of positive, return an error and disable the save button.
The save button was kept enable in view mode, the fix was to disable it in case the mode is view.

Rancher new cluster`s provisioning with etcd/s3 endpoint information
Browser: Opera 119.0.5497.88
Steps to reproduce:
**Scenario 1)**
- Create new cluster (custom)
- Tab etcd > Save Backup to s3 (enable) > set values on endpoint field (https://localhost:9000 or http://localhost)

**Scenario 2)**
- Create new cluster (custom)
- Tab etcd > Save Backup to s3 (enable) > set values on endpoint field (localhost:9000)
- Save
 

### Screenshot/Video
[Screencast From 2025-06-23 14-06-54.webm](https://github.com/user-attachments/assets/1b049186-8769-477c-8f02-49e5aca8ca3c)

### Checklist


richard-cox:
Note that the issue is currently targeting 2.13, and 2.12 priorities need to come first

shell/edit/provisioning.cattle.io.cluster/rke2.vue
```diff
@@ -266,13 +281,38 @@ export default {
...
-
+      labelForAddon,
+      s3ConfigValue:                            { ...initialS3Config },
     };
   },
 
   computed: {
     ...mapGetters({ features: 'features/get' }),
+
+    s3ConfigComponent() {
```
richard-cox: this logic should be encapsulated inside the `S3Config` component and it's valid state can be `emit`ed out

- logic specific to inside a component stays inside the component
- the interface between parent and child components is clear, rather than parent reaching into child and using internal properties

shell/edit/provisioning.cattle.io.cluster/tabs/etcd/S3Config.vue
```diff
@@ -12,6 +16,7 @@ export default {
     Checkbox,
     SelectOrCreateAuthSecret,
   },
+  mixins: [CreateEditView, FormValidation],
```
richard-cox: the CreateEditView mixin is intended for root resources create/edit/view components rather than a sub component

shell/edit/provisioning.cattle.io.cluster/tabs/etcd/S3Config.vue
```diff
@@ -63,22 +72,68 @@ export default {
...
+
 };
 </script>
 
 <template>
   <div>
     <SelectOrCreateAuthSecret
       v-model:value="config.cloudCredentialName"
+      :mode="mode"
+      :disable="isView"
```
richard-cox: `:disable="isView"` on this and below shouldn't be needed with `:mode="mode"`

shell/edit/provisioning.cattle.io.cluster/tabs/etcd/S3Config.vue
```diff
@@ -123,9 +181,17 @@ export default {
         <LabeledInput
```
richard-cox, line 143: the validation should be wired directly into this component, which handles the error icon when validation has not been met. this would also remove a lot of the custom watch stuff above

shell/utils/validators/setting.js
```diff
@@ -4,6 +4,14 @@ export const isServerUrl = (value) => value === 'server-url';
 
 export const isHttps = (value) => value.toLowerCase().startsWith('https://');
 
+export const isHttpsOrHttp = (value) => {
+  if (typeof value !== 'string') {
+    return false; // Or throw new Error('Input is not a string');
+  }
+
+  return value.toLowerCase().startsWith('https://') || value.toLowerCase().startsWith('http://');
```
richard-cox: generally it's better to reduce duplication

```
  const lowerCaseValue = value.toLowerCase();
  
  return lowerCaseValue.startsWith('https://') || lowerCaseValue.startsWith('http://');
```

richard-cox (changes requested):

marytlf:
Hi @richard-cox
I made the changes you asked, when you have time can you take a look, please?
thank you

richard-cox:
@marytlf thanks, it's on the stack

richard-cox:
Apologies, i haven't managed to get back to this one, moving to 2.13.0

shell/edit/provisioning.cattle.io.cluster/rke2.vue
```diff
@@ -273,13 +288,28 @@ export default {
...
-
+      labelForAddon,
+      s3ConfigValue:                            { ...initialS3Config },
     };
   },
 
   computed: {
     ...mapGetters({ features: 'features/get' }),
+
+    isS3EndpointTrulyValid() {
```
richard-cox: See https://github.com/rancher/dashboard/pull/14557#discussion_r2168577814

This still does S3Config specific validation outside of the S3Config component

shell/edit/provisioning.cattle.io.cluster/rke2.vue
```diff
@@ -2424,7 +2461,19 @@ export default {
               @update:value="$emit('input', $event)"
               @s3-backup-changed="handleS3BackupChanged"
               @config-etcd-expose-metrics-changed="handleConfigEtcdExposeMetricsChanged"
-            />
+            >
+              <template #s3-config>
```
richard-cox: this isn't used, there's no `<slot name="s3-config">` or `$slot` management in the Etcd component

shell/utils/validators/formRules/index.ts
```diff
@@ -164,6 +164,8 @@ export default function(
 
   const https: Validator = (val: string) => val && !isHttps(val) ? t('validation.setting.serverUrl.https') : undefined;
 
+  const httpsOrHttp: Validator = (val: string) => val && !isHttpsOrHttp(val) ? t('validation.setting.serverUrl.https') : undefined;
```
richard-cox: this should have a new l10n (httpsOrHttp)

shell/edit/provisioning.cattle.io.cluster/rke2.vue
```diff
@@ -146,7 +149,13 @@ export default {
     providerConfig: {
       type:    Object,
       default: () => null
-    }
+    },
+
+    s3EndpointHasError: {
```
richard-cox: this should be a `data` property (it's internal), rather than a component property (nothing will pass this in to the component)

shell/edit/provisioning.cattle.io.cluster/tabs/etcd/S3Config.vue
```diff
@@ -63,22 +70,66 @@ export default {
 
       return {};
     },
+
+    isEndpointInvalid() {
```
richard-cox: nothing uses this

richard-cox (changes requested):

shell/edit/provisioning.cattle.io.cluster/rke2.vue
richard-cox: superseded by https://github.com/rancher/dashboard/pull/14557#discussion_r2235877563 (problem remains)

shell/utils/validators/setting.js
```diff
@@ -4,6 +4,16 @@ export const isServerUrl = (value) => value === 'server-url';
 
 export const isHttps = (value) => value.toLowerCase().startsWith('https://');
 
+export const isHttpsOrHttp = (value) => {
+  if (typeof value !== 'string') {
+    return false; // Or throw new Error('Input is not a string');
```
eva-vashkevich: we should remove this comment

shell/edit/provisioning.cattle.io.cluster/rke2.vue
```diff
@@ -2424,6 +2441,7 @@ export default {
               @update:value="$emit('input', $event)"
               @s3-backup-changed="handleS3BackupChanged"
               @config-etcd-expose-metrics-changed="handleConfigEtcdExposeMetricsChanged"
+              @s3-config-validation-changed="handleS3ConfigValidation"
```
eva-vashkevich: we should rename this to something like 
etcd-validation-changed in case we add more validation to this component i the future. rke2 shouldn't care which part of etcd is invalid, just that the component is invalid

shell/edit/provisioning.cattle.io.cluster/rke2.vue
```diff
@@ -176,6 +179,10 @@ export default {
       this.value.spec.rkeConfig = {};
     }
 
+    if (!this.value.spec.rkeConfig.etcd) {
+      this.value.spec.rkeConfig.etcd = {};
```
eva-vashkevich: This changes behavior of [L1017-1028](https://github.com/rancher/dashboard/blob/ff9c93eb24e57f08b2d91d07286bb591ea4f1384/shell/edit/provisioning.cattle.io.cluster/rke2.vue#L1017) is that intentional?

shell/edit/provisioning.cattle.io.cluster/rke2.vue
```diff
@@ -122,6 +123,7 @@ export default {
     Advanced,
     ClusterAppearance,
     AddOnAdditionalManifest,
+    S3Config,
```
eva-vashkevich: I don't think this is used anywhere?

shell/edit/provisioning.cattle.io.cluster/rke2.vue
```diff
@@ -273,13 +280,14 @@ export default {
       isAuthenticated:                          this.provider !== GOOGLE || this.mode === _EDIT,
       projectId:                                null,
       REGISTRIES_TAB_NAME,
-      labelForAddon
-
+      labelForAddon,
+      s3ConfigValid:                            true,
```
eva-vashkevich: let's rename this to etcdConfigValid

shell/edit/provisioning.cattle.io.cluster/tabs/etcd/S3Config.vue
```diff
@@ -70,15 +78,51 @@ export default {
       const out = { ...this.config };
 
       this.$emit('update:value', out);
+
+      this.validateEndpoint(this.config.endpoint);
+    },
+    isHttpsOrHttp,
```
eva-vashkevich: this is not needed

shell/edit/provisioning.cattle.io.cluster/tabs/etcd/S3Config.vue
```diff
@@ -70,15 +78,51 @@ export default {
...
+
+    validateEndpoint(value) {
```
eva-vashkevich: this function basically goes around our validation mechanisms. I would recommend utilizing FormValidation mixin instead https://github.com/rancher/dashboard/blob/68d2249d917965537fcad6645db0cb2a9a885ba1/docusaurus/docs/internal/code-base-works/forms-and-validation.md?plain=1#L66
Personally, I think https://github.com/rancher/dashboard/blob/master/shell/cloud-credential/aws.vue demonstrates it quite well

eva-vashkevich (changes requested):

torchiaf:
@marytlf @eva-vashkevich  The 'Save Backups to S3' section is now visible only when 'Automatic Backups' are enabled. Is this a wanted change?

image

image

eva-vashkevich:
Hi @torchiaf I see "Save Backups to S3" only visible when "Automatic backups" are enabled on master too. Can you please elaborate on what you meant?

shell/utils/validators/setting.js
```diff
@@ -1,9 +1,14 @@
 import isUrl from 'is-url';
 
+// Note that these function cover specific use cases and you need to make sure it works for your use case before using them.
+// ie they would consider empty values as valid, not all endpoint formatting is enforced
+
 export const isServerUrl = (value) => value === 'server-url';
 
 export const isHttps = (value) => value.toLowerCase().startsWith('https://');
 
+export const isAWSStyleEndpoint = (value) => (/^(?!https?:\/\/)(?:[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?\.)+[a-z]{2,}(?::\d{1,5})?(?:\/\S*)?$/).test(value);
```
richard-cox: The other validators are pretty straight so don't need jsdocs, this one though definitely does.

Would also vote this is a little bit more specific, like isAWSS3Endpoint (given other aws urls may need the protocol?)

There are other protocols (like ftp, udp), are they likely to be supplied as bucket endpoints (specifically ftp)?

shell/assets/translations/en-us.yaml
```diff
@@ -6902,6 +6915,7 @@ validation:
       localhost: If the Server URL is internal to the Rancher server (e.g. localhost) the downstream clusters may not be able to communicate with Rancher.
       trailingForwardSlash: Server URL should not have a trailing forward slash.
       url: Server URL must be an URL.
+      awsStyleEndpoint: Endpoint URL has to be a valid domain-based endpoint and cannot start with a protocol (https://, http://)
```
richard-cox: for the regex as is at the moment this would be `(i.e. https:// and http://)` (as these are the only ones excluded), if it's expanded to check for others it would become `(e.g https:// or http://)`

shell/utils/validators/setting.js
eva-vashkevich: I feel like isAWSS3Endpoint is a bit misleading, because I am not actually checking if it is s3 (I think those are supposed to have s3 in them as well, but I wasn't sure how mandatory that is), just that it's domain without protocol. I've googled a bit more and it sounds like no protocol should be provided, so I will exclude ftp and udp too, though they seem to cause less confusion among users

eva-vashkevich:
@richard-cox Naturally, we also weren't clearing our s3 fields if Automatic backups are disabled, so I fixed that too
I've improved the regex a bit as well, please let me know if you think it is too restrictive

richard-cox (approved):

PR #14740, opened 2025-07-08
Update value watcher for ArrayList to preserve focus on input

### LINKED ISSUE #14692: `ArrayList` items lose the focus while typing.
**Setup**
- Rancher version: 2.12-head
- Rancher UI Extensions:
- Browser type & version:

**Describe the bug**

**To Reproduce**

ArrayList elements lose the focus while typing.

This is a regression from https://github.com/rancher/dashboard/pull/14568

https://github.com/user-attachments/assets/7c5754a4-ca7b-410a-ba19-37994402df69

- Go to Fleet -> Create App Bundles - GitRepo
- Go to step 2
- Add a new path and type something 

**Result**

**Expected Result**

**Screenshots**

**Additional context**

rak-phillip:
### Summary
The previous implementation attempted to preserve the id when newVal matched oldVal. This means that a new id would be generated when editing an existing value, causing the component to re-render and lose focus on the input.

Fixes #14692 

### Technical notes summary

This new approach preserves the existing id by locating the modified index and updating the value in place.

This solution comes with a tradeoff - add or deleting values will cause a new ID to be generated for each member of the ArrayList. This will force a re-render of the entire list and can become expensive when the ArrayList becomes arbitrarily large. Attempting to preserve the ids can become complicated on deletion when multiple members of the ArrayList have the same value.

The attached issue contains a good example for testing. Other instances of ArrayList should be retested.

Instances of ArrayList used throughout Dashboard.

### Screenshot/Video

NA

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/form/ArrayList.vue
+++ b/shell/components/form/ArrayList.vue
@@ -164,21 +164,29 @@ export default {
       { deep: true }
     );
 
+    /**
+     * Watching props.value helps to maintain internal state when slots are used
+     */
     watch(
       () => props.value,
-      (newVal) => {
+      (newVal, oldVal) => {
         lastUpdateWasFromValue.value = true;
-        const newRows = (newVal || []).map((value) => {
-          const existingRow = rows.value.find((row) => row.value === value);
 
-          if (existingRow) {
-            return { value, id: existingRow.id };
-          } else {
-            return { value, id: randomStr() };
-          }
+        // find member of the array that has been modified
+        const modifiedIdx = newVal.findIndex((current, idx) => {
+          return current !== oldVal[idx];
         });
 
-        rows.value = newRows;
+        // a member has been added or removed
+        if (newVal.length !== oldVal.length || modifiedIdx < 0) {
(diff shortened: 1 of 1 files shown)


shell/components/form/ArrayList.vue
```diff
@@ -164,21 +164,29 @@ export default {
...
-          const existingRow = rows.value.find((row) => row.value === value);
 
-          if (existingRow) {
-            return { value, id: existingRow.id };
-          } else {
-            return { value, id: randomStr() };
-          }
+        // find member of the array that has been modified
+        const modifiedIdx = newVal.findIndex((current, idx) => {
+          return current !== oldVal[idx];
```
codyrancher, line 177: By any chance do we need to do a deep equality check? I'm pretty sure the array can have objects instead of primitives

rak-phillip:
Closing this PR without merging. Opting to revert the original change - there are too many edge cases in ArrayList and its usage where the added ids are causing unexpected behavior.

PR #14765, opened 2025-07-10
Backport/2.11.4/13539 create user

### LINKED ISSUE #13539: Create User: Improve error handling

### LINKED ISSUE #14608: [backport v2.11.4] Create User: Improve error handling
This is a backport issue for https://github.com/rancher/dashboard/issues/13539, automatically created via [GitHub Actions workflow](https://github.com/rancher/dashboard/actions/runs/15914095149) initiated by @gaktive

Original issue body:

**Setup**
- Rancher version: 2.7.6, 2.9.6, 2.9.8, 2.10.1
- Rancher UI Extensions:
- Browser type & version:
- Internal reference: SURE-9866

**Describe the bug**
When creating a user with more rights than is currently possible (e.g. a restricted-admin trying to create a full-admin user), an error banner related to RBAC issues appears on the same page. The user was not created so another attempt is made to modify the roles and try creating again.

What actually happens though is the user was actually created in the background, albeit with no associated role. This means that even if you try to modify the permission checkboxes on the current page and create again, a conflict error will appear due to a user with this name already existing.

The guess is that the Rancher UI creates the user, then the global role bindings, and when this second step fails, the former is not rolled back.

Workaround:
Refresh the page or leave the page and return

Image

Image

Upon analysis from @richard-cox:
> The user shouldn't be offered the ability to create a user with escalated permissions. This is protected by the backend, but the UI still offers the option. The UI should in theory be able to tell for the provided global roles if they're creating anything above them (user-base --> standard-user --> restricted-admin --> admin), however this wouldn't cover custom roles (like the replacement to restricted-admin). I would recommend we don't change that part

> The UI does create the user first and then create the bindings. We just need to track that the create user request succeeds, so any further saves result in the edit flow (the same component handles both create and edit user actions).

The RBAC team chimed in:
> Creating a user and assigning a Global Role are 2 different operations. The UI makes the experience nicer by doing them both for you in a single page, but what we're seeing is the 1st operation succeeds and then the 2nd operation fails. Maybe the UI could delete the user if the 2nd operation fails? If that's not too difficult that would probably be sufficient.

rak-phillip:
Fixes #14608

This is a backport of PR #14081

### Summary
This emits  an `update:mode` event that sets to mode to "edit" and refreshes all resource-related variables when the `updateRoles()` method fails during user creation. This provides an opportunity for users to correct mistakes while still staying in the "Create User" context.

Fixes #13539 

- Switch mode to "edit" if the `updateRoles()` method fails during user creation
- Add a user readable error message if an escalation error is raised

### Technical notes summary

An alternate approach that explored navigating to an edit route was explored, but this required re-rendering the entire form and would eliminate the error message; additional state to persist the error messages would be required and I found that requirement to be more complex than the approach taken here.

User Creation: See details in the associated issue.

User Creation: Ensuring that the form still functions properly after an error is raised during user creation.

### Screenshot/Video

image

### Checklist
- [x] The PR has a Milestone 


shell/assets/translations/en-us.yaml
```diff
@@ -318,6 +318,15 @@ suffix:
...
 ##############################
+layouts:
+  home: home layout
+  plain: plain layout
+  default: default layout
+  blank: blank layout
+  unauthenticated: unauthenticated layout
+  logout: logout layout
+  verify: verify layout
+
```
rak-phillip: hmm... it looks like the cherrypick pulled in some additional translations along the way[^1]. I think that these need to be resolved before merging.

[^1]: https://github.com/rancher/dashboard/pull/14081/files#diff-2f4699c41daff9d49b0ef303edbd950cc74e4b6baaf7f5cf46d30fa2fb1a6220

eva-vashkevich (approved):
lgtm

PR #14784, opened 2025-07-11
Remove complex init logic from ingress data props

### LINKED ISSUE #14777: Data Exceptions: networking.k8s.io.ingress
Refactor ingress components to replace data props that contain complex init logic.

## Acceptance criteria

The following components are refactored so that the `data` prop no longer contains logic and only specifies component state: 

- [ ] `shell/edit/networking.k8s.io.ingress/Certificate.vue`
- [ ] `shell/edit/networking.k8s.io.ingress/DefaultBackend.vue`
- [ ] `shell/edit/networking.k8s.io.ingress/IngressClass.vue`
- [ ] `shell/edit/networking.k8s.io.ingress/Rule.vue`
- [ ] `shell/edit/networking.k8s.io.ingress/RulePath.vue`

rak-phillip:
### Summary
This refactors form components that contain initialization logic in their data props so that they use patterns that are more idiomatic to Vue.

Fixes #14777

- Refactor data props in 
   - shell/edit/networking.k8s.io.ingress/Certificate.vue
   - shell/edit/networking.k8s.io.ingress/DefaultBackend.vue
   - shell/edit/networking.k8s.io.ingress/IngressClass.vue
   - shell/edit/networking.k8s.io.ingress/Rule.vue
   - shell/edit/networking.k8s.io.ingress/RulePath.vue

### Technical notes summary

The most complex changes in this PR involve shifting initialization logic outside of `data()` and into the `created()` hook. We will want to validate that the initialization order remains correct in cases where the `created()` hook already contains logic.

Assert that CRUD operations behave as expected for

- shell/edit/networking.k8s.io.ingress/Certificate.vue
- shell/edit/networking.k8s.io.ingress/DefaultBackend.vue
- shell/edit/networking.k8s.io.ingress/IngressClass.vue
- shell/edit/networking.k8s.io.ingress/Rule.vue
- shell/edit/networking.k8s.io.ingress/RulePath.vue

Same as the areas that should be tested.

### Screenshot/Video

NA

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/edit/networking.k8s.io.ingress/Certificate.vue
+++ b/shell/edit/networking.k8s.io.ingress/Certificate.vue
@@ -27,7 +27,7 @@ export default {
   data() {
     return {
       hosts:      this.value?.hosts ?? [''],
-      secretName: this.value?.secretName ?? DEFAULT_CERT_VALUE,
+      secretName: this.value?.secretName === undefined ? DEFAULT_CERT_VALUE : this.value?.secretName,
       secretVal:  this.value?.secretName ?? DEFAULT_CERT_VALUE,
     };
   },
--- a/shell/edit/networking.k8s.io.ingress/Certificate.vue
+++ b/shell/edit/networking.k8s.io.ingress/Certificate.vue
@@ -25,17 +25,14 @@ export default {
     }
   },
   data() {
-    const defaultCert = {
-      label: this.t('ingress.certificates.defaultCertLabel'),
-      value: DEFAULT_CERT_VALUE,
-    };
-    const { hosts = [''], secretName = defaultCert.value } = this.value;
-
     return {
-      defaultCert,
-      hosts,
-      secretName,
-      secretVal: this.value.secretName ?? DEFAULT_CERT_VALUE,
+      defaultCert: {
+        label: this.t('ingress.certificates.defaultCertLabel'),
+        value: DEFAULT_CERT_VALUE,
+      },
+      hosts:      this.value?.hosts ?? [''],
+      secretName: this.value?.secretName ?? DEFAULT_CERT_VALUE,
+      secretVal:  this.value?.secretName ?? DEFAULT_CERT_VALUE,
     };
   },
   watch: {
--- a/shell/edit/networking.k8s.io.ingress/Certificate.vue
+++ b/shell/edit/networking.k8s.io.ingress/Certificate.vue
@@ -26,10 +26,6 @@ export default {
   },
   data() {
     return {
-      defaultCert: {
-        label: this.t('ingress.certificates.defaultCertLabel'),
-        value: DEFAULT_CERT_VALUE,
-      },
       hosts:      this.value?.hosts ?? [''],
       secretName: this.value?.secretName ?? DEFAULT_CERT_VALUE,
       secretVal:  this.value?.secretName ?? DEFAULT_CERT_VALUE,
@@ -44,7 +40,12 @@ export default {
   },
   computed: {
     certsWithDefault() {
-      return [this.defaultCert, ...this.certs.map((c) => ({ label: c, value: c }))];
+      const defaultCert = {
+        label: this.t('ingress.certificates.defaultCertLabel'),
+        value: DEFAULT_CERT_VALUE,
+      };
+
+      return [defaultCert, ...this.certs.map((c) => ({ label: c, value: c }))];
     },
     certificateStatus() {
       const isValueAnOption = !this.secretName || this.certsWithDefault.find((cert) => this.secretName === cert.value);
--- a/shell/edit/networking.k8s.io.ingress/Rule.vue
+++ b/shell/edit/networking.k8s.io.ingress/Rule.vue
@@ -32,12 +32,9 @@ export default {
     }
   },
   data() {
-    const { host = '', http = {} } = this.value;
-    const { paths = [{ id: random32(1) }] } = http;
-
     return {
-      host,
-      paths,
+      host:     this.value?.host ?? '',
+      paths:    this.value?.http?.paths ?? [{ id: random32(1) }],
       ruleMode: this.value.asDefault ? 'asDefault' : 'setHost',
     };
   },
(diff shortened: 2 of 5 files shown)


rak-phillip:
@eva-vashkevich this is out of draft and ready for review.

shell/edit/networking.k8s.io.ingress/Certificate.vue
```diff
@@ -25,17 +25,14 @@ export default {
...
-    };
-    const { hosts = [''], secretName = defaultCert.value } = this.value;
-
     return {
-      defaultCert,
-      hosts,
-      secretName,
-      secretVal: this.value.secretName ?? DEFAULT_CERT_VALUE,
+      defaultCert: {
+        label: this.t('ingress.certificates.defaultCertLabel'),
```
eva-vashkevich: is it okay to have this.t inside data? should we move defaultCert to a computed property?

eva-vashkevich:
I probably should have reviewed this sooner, but what is the benefit of moving these to created instead of turning them into computed properties? The values are just coming from this.value and this feels unnecessary? I feel like using computed properties would allow us to simplify updating them as well?

shell/edit/networking.k8s.io.ingress/Certificate.vue
rak-phillip: After reviewing the implementation, it should be fine - the risk of throwing an unhandled exception is extremely low. But, you make an accurate assessment, there'e no reason for `defaultCert` to be a data prop and it does not benefit from having reactivity attached. I've moved `defaultCert` into the `certsWithDefault()` computed.

rak-phillip:
> what is the benefit of moving these to created instead of turning them into computed properties? The values are just coming from this.value and this feels unnecessary? I feel like using computed properties would allow us to simplify updating them as well?

My primary motivation for using lifecycle hooks is to keep the complexity of the refactor to a minimum. The `created()` hook closely mirrors the same initialization logic that was originally implemented in the `data` prop. I agree that we can perform further refactoring to simplify the component implementation, but I'm of the opinion that the patterns applied in this PR lessen the risk and help us to more quickly resolve the widespread problem described in the associated epic: #11807

eva-vashkevich:
> My primary motivation for using lifecycle hooks is to keep the complexity of the refactor to a minimum. The `created()` hook closely mirrors the same initialization logic that was originally implemented in the `data` prop. I agree that we can perform further refactoring to simplify the component implementation, but I'm of the opinion that the patterns applied in this PR lessen the risk and help us to more quickly resolve the widespread problem described in the associated epic: #11807

That's fair. My main concern is that since we are already touching these components, we might as well make them more readable, since we don't know when (likely never) we will have time to refactor them more. So it will just carry some tech debt that can be easily avoided.

But I also see value in your approach and I think it comes down to our preferences. 
I'll let you make the call, just wanted to bring up the other perspective

rak-phillip:
>  I'll let you make the call, just wanted to bring up the other perspective

Thanks, your feedback is valid and warranted, and I hope that it doesn't appear like I'm discarding it without consideration. In a different scenario, I would gladly perform additional refactoring to address some of these problematic patterns in our components, but I think the risk isn't worth it for this change. It's my position that the best approach is to limit the scope of the refactor to reduce the surface area for regressions.

shell/edit/networking.k8s.io.ingress/Rule.vue
```diff
@@ -32,12 +32,9 @@ export default {
...
   },
   data() {
-    const { host = '', http = {} } = this.value;
-    const { paths = [{ id: random32(1) }] } = http;
-
     return {
-      host,
-      paths,
+      host:     this.value?.host ?? '',
+      paths:    this.value?.http?.paths ?? [{ id: random32(1) }],
```
eva-vashkevich, line 37: I am still a bit unsure why we are okay to do this here, but then in the file above we put it in created? since in both cases we are just accessing properties of value?
paths:    this.value?.http?.paths ?? [{ id: random32(1) }],

shell/edit/networking.k8s.io.ingress/RulePath.vue
```diff
@@ -35,23 +35,16 @@ export default {
...
-    set(this.value, 'pathType', this.value.pathType || pathTypes[0]);
-    set(this.value.backend, this.ingress.serviceNamePath, get(this.value.backend, this.ingress.serviceNamePath) || '');
-    set(this.value.backend, this.ingress.servicePortPath, get(this.value.backend, this.ingress.servicePortPath) || '');
-
-    const serviceName = get(this.value.backend, this.ingress.serviceNamePath);
-    const servicePort = get(this.value.backend, this.ingress.servicePortPath);
-
     return {
-      pathTypes, serviceName, servicePort, pathType: this.value.pathType, path: this.value.path
+      pathTypes: [
```
eva-vashkevich: Should this list be moved to a constant?

eva-vashkevich (commented):
I still have a few questions

shell/edit/networking.k8s.io.ingress/Rule.vue
rak-phillip: The `get()` implementation will explicitly `throw` if the path is null or undefined.. In hindsight, I think it's safe to leave the `get()` implementation in as-is in `shell/edit/networking.k8s.io.ingress/IngressClass.vue`. I perceived `random32()` to have less risk, that is why I thought it to be fine in data.

shell/edit/networking.k8s.io.ingress/RulePath.vue
rak-phillip: It could be, but then we will have to specify it as a non-reactive prop to allow it to be accessed in the template.. I made an update to this - lmk what you think.

eva-vashkevich: @rak-phillip what did you mean when you said you made an update to this? there haven't been any changes?

rak-phillip: hmm.. it looks like my commits didn't get pushed yesterday. @eva-vashkevich Can you check now?

eva-vashkevich (approved):
lgtm

PR #14820, opened 2025-07-16
Handle charts zero date

### LINKED ISSUE #14819: Broken date for the apps from OCI repositories
**Setup**
- Rancher version:
- Rancher UI Extensions:
- Browser type & version:

**Describe the bug**

**To Reproduce**

1. Add an OCI repository e.g. `oci://registry-1.docker.io/bitnamicharts/redis`
2. Go to Apps/Charts page
3. Filter the result by only displaying apps from the repo you added
4. Check the date on the app card

**Result**

A broken date is displayed(e.g. `Dec 31, -0815`)

**Expected Result**

We should display `n/a` with a tooltip: `Last updated date is not available for this chart`.

**Screenshots**

Image

**Additional context**

### LINKED ISSUE #15124: Chart Detail Page - Update tooltip message
**To Reproduce**

Add an OCI repo like `oci://registry-1.docker.io/bitnamicharts/redis` then go to the detail page for that chart

**Result**
The tooltip says "The info is missing"

**Expected Result**
The tooltip should say "No date available for this chart version"

**Screenshots**

Image

momesgin:
### Summary
Fixes #14819 
Fixes #15124 

### Technical notes summary

- Check for `ZERO_TIME` in `shell/models/chart.js` when generating `cardContent`
- Updated `AppChartCardSubHeader.vue` to display a tooltip for when the date is missing
- Unit tests

Reproduce steps on the issue

Check if cards display the date properly from both oci and non-oci repositories
### Screenshot/Video

Charts page:
image
Chart detail page:
image

### Checklist
- [x] The PR has a Milestone 


shell/models/chart.js
```diff
@@ -121,18 +121,27 @@ export default class Chart extends SteveModel {
...
+          {
+            icon:        'icon-version-alt',
+            iconTooltip: { key: 'tableHeaders.version' },
+            label:       latestVersion.version
+          },
+          {
+            icon:        'icon-refresh-alt',
+            iconTooltip: { key: 'tableHeaders.lastUpdated' },
+            label:       hasZeroTime ? this.t('generic.na') : day(latestVersion.created).format('MMM D, YYYY'),
+            ...(hasZeroTime && { labelTooltip: this.t('generic.missingInfoMessage') })
```
nwmac: Won't that push `false` into the array? Does that just get ignored?

I'd probably go with an if statement and only add the sub header if hasZeroTime is true.

nwmac: Not sure who added `missingInfoMessage` - I think we should add a specific translation `Last updated date is not available for this chart`

nwmac (dismissed):

shell/models/chart.js
momesgin: > Won't that push false into the array? Does that just get ignored?

`...false` inside an object literal gets ignored, but I think for better readability I'll switch back to `if` statement 

> Not sure who added `missingInfoMessage` - I think we should add a specific translation `Last updated date is not available for this chart`

That was my idea to use a generic translation so it can be used in multiple places. I'll update the translation.

codyrancher (approved):

PR #14840, opened 2025-07-17
Resolving some z-index issues around tooltips and slide-ins

### LINKED ISSUE #14833: Tooltips in Resource Details are hidden behind the slide-in panel
**Setup**
- Rancher UI version: head-a0a36797a6
- Rancher UI Extensions:
- Browser type & version:

**Describe the bug**

Tooltips are hidden behind the slide-in panel of Resource Details.

**To Reproduce**

- Go to Explorer -> Deployments
- Click on a Deployment - name column to show the Deployment's details page
- Click on 'Show configurations' on top
- Go to General tab
- Hover your mouse on 'Pull Secrets' tooltip

**Result**

The Tooltip is behind the slide-in panel so not visible

Image

Fleet GitRepo example

Image

**Screenshots**

**Additional context**

codyrancher:
### Summary
Fixes #14833

Modals, tooltips and drawers

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/assets/styles/base/_variables.scss
+++ b/shell/assets/styles/base/_variables.scss
@@ -42,25 +42,31 @@ $z-indexes: (
 
   cruFooter: 19,
 
-  modalOverlay: 20,
-  modalContent: 21,
+  loadingMain: 51,
 
-  tooltip: 30,
+  slide-in: 52,
 
-  dropdownOverlay: 40,
-  dropdownContent: 41,
+  modalOverlay: 53,
+  modalContent: 54,
 
-  loadingMain: 51
+  dropdownOverlay: 55,
+  dropdownContent: 56,
+
+  tooltip: 57,
 );
 
 // Usage Example:
 // @media only screen and (min-width: map-get($breakpoints, '--viewport-*')) {
 // }
 $breakpoints: (
-  '--viewport-4':  480px, // Phone
-  '--viewport-7':  768px, // Tablet
-  '--viewport-9':  992px, // Laptop/Desktop
-  '--viewport-12': 1281px, // Desktop
+  '--viewport-4': 480px,
+  // Phone
+  '--viewport-7': 768px,
+  // Tablet
+  '--viewport-9': 992px,
+  // Laptop/Desktop
+  '--viewport-12': 1281px,
+  // Desktop
 );
 
 $font-size-h2: 21px;
--- a/pkg/kubectl-explain/components/SlideInPanel.vue
+++ b/pkg/kubectl-explain/components/SlideInPanel.vue
@@ -446,7 +446,6 @@ export default {
     flex-direction: column;
     position: fixed;
     top: 0;
-    z-index: 2000;
     width: $slidein-width;
     background-color: var(--body-bg);
     right: -$slidein-width;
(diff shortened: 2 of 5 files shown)


shell/assets/styles/base/_variables.scss
```diff
@@ -45,12 +45,14 @@ $z-indexes: (
...
 
-  tooltip: 30,
-
   dropdownOverlay: 40,
   dropdownContent: 41,
 
-  loadingMain: 51
+  loadingMain: 51,
+
+  slide-in: 52,
```
rak-phillip, line 47: Is it possible to have a slide-in that will trigger a modal? That's probably not the best UX, but it appears that modal will render beneath the slide-in in the event that this happens.

codyrancher: I don't think we have the intention of allowing that and it would therefore be a bug. At some point we may allow multiple slide-ins but that's a future issue.

@oboc-sts is there any intention to allow slideins (Drawers) to trigger modals?

oboc-sts: @codyrancher Nope. They (the elements inside drawers) can trigger the following overlays:
- popovers (when hovering / clicking an element - e.g. actions menu)
- another drawer, on the same level (chained drawers that allow the user to move backwards - e.g. view another resource's config)
- another drawer, on the 2nd level (on top of the existing one) - this is also the max levels (2) we have in mind for stacked drawers (not the case for now)
- toast messages (top-right corner of the viewport - e.g. a confirmation / error / etc that happens when the user does something)

The plan for the medium term is to move away from modals and replace them with drawers (of different sizes) or other (less intrusive) IxD patterns.

rak-phillip: We might disallow a pattern in our own development, but we can't control what extension developers do.. imo it's a good idea to take into consideration the hierarchy of these elements until we deprecate and remove modals.

rak-phillip: Ultimately, this concern is an edge case and I'm not going to push back on it too hard. As long as we don't subvert any expectations with modals potentially rendering underneath slide-ins, I think we are fine.

codyrancher: > We might disallow a pattern in our own development, but we can't control what extension developers do.. imo it's a good idea to take into consideration the hierarchy of these elements until we deprecate and remove modals.

I went ahead and moved modals above the slide-ins. I think maybe we'd actually prefer resolving this by having drawers close if a modal opens so that there isn't an issue with extensions but that's beyond the scope of what we want here.

rak-phillip (approved):
LGTM

nwmac:
I think this introduces a bug where some controls now overlap with the slidein - for example with the kubectl explain:

image

codyrancher:
> I think this introduces a bug where some controls now overlap with the slidein - for example with the kubectl explain:
> image

I'll take a look.

a110605:
Hi @codyrancher or @nwmac, thanks for the fixing.
Can we backport this fix to 2.12.1 ? 

https://github.com/rancher/dashboard/issues/15077#issuecomment-3164156199

codyrancher:
@a110605 we should've gotten this in but it looks like a backport was never made. Friday was the code cuttoff for 2.12.1 so I'll open a backport for 2.12.2.

Edit: Nevermind @a110605, I got permission to add it to 2.12.1.

PR #14853, opened 2025-07-18
Implements the StatusIndicator as described in our design system

codyrancher:
### Summary
Implements the StatusIndicator as described in our design system. I can provide a link to our figma. Trying to get some momentum with probably the easiest component to implement.
 
image

Just the new component. You can verify it in storybook.

It's a new feature which only adds new code so I don't anticipate any regressions.

### Screenshot/Video

https://github.com/user-attachments/assets/ea4bb0e5-a273-44c4-8058-30e24c8b2414

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/assets/styles/themes/_light.scss
+++ b/shell/assets/styles/themes/_light.scss
@@ -629,10 +629,10 @@ BODY, .theme-light {
 
 
   // These are meant to be the colors used within rancher-components. They will eventually be removed from shell once we have theming working in the components
-  --rc-info: #1F67DB;
-  --rc-success: #007032;
-  --rc-warning: #FFE47A;
-  --rc-error: #B13333;
-  --rc-unknown: #EDEFF3;
-  --rc-none: #DCDEE4;
+  --rc-info: #{$blue0001};
+  --rc-success: #{$green001};
+  --rc-warning: #{$yellow001};
+  --rc-error: #{$red001};
+  --rc-unknown: #{$gray001};
+  --rc-none: #{$gray002};
 }
--- a/shell/assets/styles/base/_color.scss
+++ b/shell/assets/styles/base/_color.scss
@@ -9,3 +9,10 @@
 .text-disabled {
   color: var(--disabled-text) !important;
 }
+
+$blue0001: #1F67DB;
+$green001: #007032;
+$yellow001: #FFE47A;
+$red001: #B13333;
+$gray001: #EDEFF3;
+$gray002: #DCDEE4;


shell/assets/styles/themes/_light.scss
```diff
@@ -625,4 +625,13 @@ BODY, .theme-light {
...
   --bubble-bg                : #{$lighter};
+
+
+  // These are meant to be the colors used within rancher-components. They will eventually be removed from shell once we have theming working in the components
+  --rc-info: #1F67DB;
+  --rc-success: #007032;
+  --rc-warning: #FFE47A;
+  --rc-error: #B13333;
+  --rc-unknown: #EDEFF3;
+  --rc-none: #DCDEE4;
```
codyrancher: These are the colors that are used in the DS but I've been told that they will change.

pkg/rancher-components/src/components/Pill/StatusIndicator/StatusIndicator.vue
codyrancher, line 1: I know some of our other components now have the Rc prefix. Do we want to carry that over here as well? I'm not sure what the original intention for the prefix was considering only some have it.

codyrancher:
@oboc-sts I just wanted to let you know this got started.

shell/assets/styles/themes/_light.scss
rak-phillip: nit: I would like to see a concrete set of color variables used across the design system and code. For example, instead of a hex value for `--rc-info`, we could have a `blue-100` (or whatever we choose to call it) that can be easily referenced wherever we define values for components. 

I think that utilizing an approach like this will help us to maintain consistency in the colors that we use across the board. 

FYI @oboc-sts

pkg/rancher-components/src/components/Pill/StatusIndicator/StatusIndicator.vue
rak-phillip: I think that we need to make a decision on a consistent naming convention if we intend to expand and formalize the component library. 

The intent of the Rancher Components (`Rc`) prefix was to:

1. Signal that base components are a part of the component library and not the core Rancher Dashboard product
2. Avoid single word component names, so that we prevent potential conflicts with existing and future HTML elements

shell/assets/styles/themes/_light.scss
codyrancher: Agreed. Just not sure when it's going to be ready
image

pkg/rancher-components/src/components/Pill/StatusIndicator/StatusIndicator.vue
codyrancher: Works for me. I'll update the name.

pkg/rancher-components/src/components/Pill/StatusIndicator/types.ts
```diff
@@ -0,0 +1,7 @@
+export type Shape = 'disc' | 'horizontal-bar' | 'vertical-bar';
+export type Status = 'info' | 'success' | 'warning' | 'error' | 'unknown' | 'none';
+
+export interface CardProps {
```
rak-phillip: `CardProps` doesn't appear to be the right name. What if we named this something like `StatusIndicatorProps`?

pkg/rancher-components/src/components/Pill/StatusIndicator/types.ts
```diff
@@ -0,0 +1,7 @@
+export type Shape = 'disc' | 'horizontal-bar' | 'vertical-bar';
+export type Status = 'info' | 'success' | 'warning' | 'error' | 'unknown' | 'none';
```
rak-phillip: nit: Other components like `BadgeState` and `Banner` use a `color` prop to specify similar behavior. 

I quite like the usage of `status`. My request is that we consider consistency moving forward so that we have predictable interfaces between components that specify similar behaviors - so, if `status` is preferred, I'd like to evaluate how we can update other components to utilize `status` as well.

codyrancher: Yeah, I waited until we had more than one to address that.

https://github.com/rancher/dashboard/pull/14889/files#diff-25eb8b299975e2777cc29d2fdadcc2877ea76e92a680a10844bedb19e15790ba

I'm not sure if the above is how we want to do it but it was on my mind.

pkg/rancher-components/src/components/Pill/StatusIndicator/types.ts
codyrancher: Correct. I'll fix that.

pkg/rancher-components/src/components/Pill/RcStatusIndicator/RcStatusIndicator.vue
```diff
@@ -0,0 +1,75 @@
+<script setup lang="ts">
+import { RcStatusIndicatorProps } from '@components/Pill/RcStatusIndicator/types';
+
+const props = defineProps<RcStatusIndicatorProps>();
```
rak-phillip, line 4: Should we specify defaults if no props are supplied, for example: `<status-indicator />`? I'm unsure if that's a valid use case.. without defaults, we seemingly render nothing.

rak-phillip (commented):
We need to consider color contrast for both light and dark themes:

## Light

image

## Dark

image

I'm also interested in example use-cases. Is this a visual only indicator or is this expected to be used in conjunction with some text to relay some status. If this is a visual-only indicator, how do we relay the status via screen readers?

pkg/rancher-components/src/components/Pill/RcStatusIndicator/RcStatusIndicator.vue
codyrancher: In my mind I don't think we want to provide defaults. At the very least the consumer needs to indicate what the shape is. 

I could see the default for status being `none` but my preference is to expect the consumer to specify both.

shell/assets/styles/themes/_light.scss
oboc-sts: > nit: I would like to see a concrete set of color variables used across the design system and code. For example, instead of a hex value for `--rc-info`, we could have a `blue-100` (or whatever we choose to call it) that can be easily referenced wherever we define values for components.
> 
> I think that utilizing an approach like this will help us to maintain consistency in the colors that we use across the board.
> 
> FYI @oboc-sts

This is indeed the plan @rak-phillip . There are 2 streams of work going on in parallel on our (design) side:
- @edenhernandez-suse is creating the variables for the colors (see below)
- I am focusing on the components

Since the variables were not ready yet when I created the components in the Rancher DS, the agreement (between Eden and myself) was to use the hex colors for now in a consistent way that will enable us to "search and replace all" with the variables when they are ready. Once applied, we'll republish the DS and all the components in use will automatically be updated.

Both streams of work take time (we're actually doing the work of a dedicated DesignOps team), especially the variables in the context of the new theming. But I'm confident we'll be ready with them by the time we release 2.13.

image

@codyrancher thanks for picking this up. Is this the pre-work needed to be able to implement the resource link popover?

image

codyrancher: @oboc-sts It's not pre-work. 

This is to start getting momentum and get a few of us aligned so we can get through more of the components.

codyrancher: @rak-phillip given the answer above, and in an effort to keep things moving. 

Would you like to maybe create a `colors.scss` which we can bind the temporary hex values to a name and then replace those vars everywhere we use them once it gets finalized? If you have another idea I'd be open to that as well.

rak-phillip: > Would you like to maybe create a colors.scss which we can bind the temporary hex values to a name and then replace those vars everywhere we use them once it gets finalized?

Yeah, I think that makes sense. With my original comment being a nit, I don't think it should block moving forward with this PR. I think we have a path forward for adding some structure around this area.

oboc-sts:
> We need to consider color contrast for both light and dark themes
> I'm also interested in example use-cases. Is this a visual only indicator or is this expected to be used in conjunction with some text to relay some status. If this is a visual-only indicator, how do we relay the status via screen readers?

These indicators will most likely be used in conjunction with the name of a resource and / or in a legend (these are the most common cases).  They are meant to represent the minified version of the status in different contexts where the full status badge either doesn't fit or it's not in focus. On hover, we will display a popover with additional information, including the status badge in full (including the status value).

On the contrast and accessibility issues, I had a chat with @edenhernandez-suse on this because the new theme colour scheme doesn't quite work out when applied at smaller sizes (the current status indicators you're working on) - long story short, I believe we should tweak those colors a bit or provide an alternative to be used for smaller elements (graph lines, status indicators, etc). 

Let's involve @edenhernandez-suse on this as well please.

rak-phillip (approved):
LGTM - colors and themes will need to be addressed in follow-up PRs. I think that my other accessibility concerns have been addressed via above discussions.

PR #14871, opened 2025-07-22
Reverted the color of the loading background while maintaining the new color for the rest of overlays

### LINKED ISSUE #14856: Switching loading backgrond color back to more muted color
The new loading overlays are a bit too jarring and we'd like to switch them back to what they were in 2.11.

Image

codyrancher:
### Summary
fixes #14856

The loading overlay and modal/drawer overlays

### Screenshot/Video

image

image

image

image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/CodeMirror.vue
+++ b/shell/components/CodeMirror.vue
@@ -444,7 +444,7 @@ export default {
         justify-content: center;
         border: 1px solid transparent;
         color: var(--darker);
-        background-color: var(--overlay-bg);
+        background-color: var(--subtle-overlay-bg);
         font-size: 12px;
 
         .close-indicator {
--- a/shell/assets/styles/themes/_dark.scss
+++ b/shell/assets/styles/themes/_dark.scss
@@ -95,6 +95,7 @@
 
   --modal-bg                   : #{$dark};
   --modal-border               : #{$medium};
+  --subtle-overlay-bg          : #{rgba($darkest, 0.75)};
   --overlay-bg                 : #{rgba($darkest, 0.75)};
   --shadow                     : #{rgba($darkest, 0.9)};
 
--- a/shell/assets/styles/themes/_light.scss
+++ b/shell/assets/styles/themes/_light.scss
@@ -407,6 +407,7 @@ BODY, .theme-light {
 
   --modal-bg                   : #{$lightest};
   --modal-border               : #{$dark};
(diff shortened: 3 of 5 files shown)


codyrancher:
@oboc-sts would you mind reviewing the screenshots to make sure they colors for each are what you intended? I just switched the loading background back to what it was previously but I can change to something else if preferred.

oboc-sts (dismissed):
Looking good @codyrancher 💪

eva-vashkevich:
Hi @oboc-sts should we do this for CodeMirror as well?
This is what we have now
Screenshot 2025-07-23 at 5 47 07 PM
This is what we had in 2.11
Screenshot 2025-07-23 at 5 48 49 PM

shell/components/CodeMirror.vue
```diff
@@ -301,7 +301,7 @@ export default {
   </div>
 </template>
 
-<style lang="scss">
+<style lang="scss" scoped>
```
codyrancher: Can you double check the styling is correct on your end?

I was seeing odd behavior where both the old and the new style was getting included on the same page and the old was overriding the new.

Scoping seemed to fix that but I saw it change once and I'm not sure if it's a problem with hot-reloading.

codyrancher:
>should we do this for CodeMirror as well? This is what we have now Screenshot 2025-07-23 at 5 47 07 PM This is what we had in 2.11 Screenshot 2025-07-23 at 5 48 49 PM

@eva-vashkevich  No, it was specifically for real overlays. Thanks for catching that and sorry for not digging as much as I should've. I switched that one and a couple others back as well.

I left a note about code mirror. I'd appreciate if you'd double check the fix there.

oboc-sts:
> Hi @oboc-sts should we do this for CodeMirror as well? 

@eva-vashkevich nope, but I see @codyrancher [already answered](https://github.com/rancher/dashboard/pull/14871#issuecomment-3111792384) - spot on (the change was indeed intended for the real overlays only)

shell/components/CodeMirror.vue
eva-vashkevich: Weird, I don't see an issue even if I remove scoped

codyrancher: Seems like the same is happening on my end after clearing node_modules. I removed the scoped.

eva-vashkevich (approved):
lgtm

PR #14906, opened 2025-07-24
Improve performance in apps/charts page by optimizing LazyImage component

### LINKED ISSUE #14905: [Apps/Charts] - Clicking on a chart takes too long to go to the chart detail page

On Apps/Charts page when you click on a chart card it takes too long to go to the chart detail page. The issue seems to be the `LazyImage` component trying to load its image even if it's not in the viewport which become problematic when the list is long. If you wait for the page to finish all the network calls for those images and then click on a chart it'll open the detail page much faster.

The `LazyImage` component should only load its image once it’s within the viewport.

Pagination could be an alternative, but it's is not implemented in the backend yet. However, even with that in place, we can still benefit from making the `LazyImage` component smarter by doing this.

**Additional context**

momesgin:
### Summary
Fixes #14905 

### Technical notes summary

- Utilized `IntersectionObserver` in `LazyImage` to make sure the images are loaded only when the component is in view.
- Unit tests

Check images get loaded properly in:
Go to Explorer => Apps/Charts and check the network tab, make sure you're not filtering any requests
Scroll down, new network requests should be made as you scroll down

Make sure the images are loaded properly in these pages:

- Explorer => Cluster  => Tools
- Extensions

### Screenshot/Video

https://github.com/user-attachments/assets/fec422c9-7308-4a78-a442-12b4f2b127f7

### Checklist
- [x] The PR has a Milestone 


richard-cox (dismissed):
Code looks good.

I looked at the performance going from Pods page to Charts page, pause, then scroll to bottom of charts list 
- both show similar high cpu whilst all cards are initially shown (no impact for additional watches). when adding the bitnami repo (going from 128 to 246 charts) there's less cpu impact on initial render
- loading images on demand shows more activity when scrolling but NOT high cpu

So all good there.

I think there's one part that needs to be fixed though, we now show the generic initial image all the time before the full image. This is a visual blip seen in places like the cluster provider type (cluster create) and extension (extensions tabs) icons. Best place is flicking between extension tabs in the extensions page. 

- now - white area --> initial / missing image icon --> actual icon
- before - white area --> actual icon

momesgin:
> Code looks good.
> 
> I looked at the performance going from Pods page to Charts page, pause, then scroll to bottom of charts list
> 
> * both show similar high cpu whilst all cards are initially shown (no impact for additional watches). when adding the bitnami repo (going from 128 to 246 charts) there's less cpu impact on initial render
> * loading images on demand shows more activity when scrolling but NOT high cpu
> 
> So all good there.
> 
> I think there's one part that needs to be fixed though, we now show the generic initial image all the time before the full image. This is a visual blip seen in places like the cluster provider type (cluster create) and extension (extensions tabs) icons. Best place is flicking between extension tabs in the extensions page.
> 
> * now - white area --> initial / missing image icon --> actual icon
> * before - white area --> actual icon

I've updated the code to have the same flow for showing the placeholder image

https://github.com/user-attachments/assets/c115611e-07d5-4312-b83c-42875e4bbaf7

shell/components/LazyImage.vue
```diff
@@ -17,32 +19,87 @@ export default {
     },
   },
 
+  data() {
+    return {
+      observer:    null,
+      intersected: false,
+      boundError:  null,
```
rak-phillip: question: Vue will add reactivity to all properties in `data()` - do any of these properties need reactivity attached?

I ask this question because I noticed that `boundError` was previously a non-reactive prop attached to the component instance. I don't think that this PR really changes the way we expect to interact with `boundError`, but I could be missing somthing.. I also wonder if we want `observer` to be reactive as well, it could be needless overhead for this use-case.

rak-phillip (commented):
I tested the change and it looks great. Previous comments look to be addressed. 

I have one additional question for us to consider before merging.

shell/components/LazyImage.vue
momesgin: thanks for catching this, it turned out I don't need them there at all

PR #14934, opened 2025-07-28
[12518] Fleet support Harvester clusters

### LINKED ISSUE #12518: Feature Request: Support harvester-baremetal-container-workload clusters
### Is your feature request related to a problem?

There does not appear to be any way to target Harvester clusters when the harvester-baremetal-container-workload feature flag is enabled. Harvester clusters are excluded from the "All clusters in this workspace" filter, "Advanced" filter, or cluster groups that contain Harvester clusters.

### Solution you'd like

When the harvester-baremetal-container-workload feature flag is enabled, don't filter Harvester clusters from the target.

### Alternatives you've considered

Using advanced targeting and cluster groups.

### Anything else?

_No response_

torchiaf:
### Summary
Fixes #12518

- Currently Harvester cannot be used for Fleet, so with the feature flag ON harvester-baremetal-container-workload it can now select them.
- Changes will be mostly on CREATE/EDIT flow to be able to select those options

When `harvester-baremetal-container-workload` flag is enabled, the UI should support Harvester clusters as targets in App Bundles (HelmOps/GitRepos), Workspaces and Clusters pages.

### Technical notes summary

- Important to notice that for cases where there are SELECTED CLUSTERS, it is not required to have the MATCH EXPRESSION to remove the other ones.
- The MATCH EXPRESSION is only useful on the case where you SELECT ALL, so we can worry for that MATCH EXPRESSION to exist only on the case where it is selecting all clusters.
- The HARVESTER CLUSTER on EDIT mode if you have it turned OFF will show on the SELECT as DISABLED, only the CLUSTER that has been previously selected, if removed it will disappear on the screen.
- There is currently a bug when the flag updates the cluster doesnt show.

- Create/Edit the AppBundle on Fleet
- Check for cases with the FLAG ON, should be able to select HARVESTER CLUSTER
- Check for cases with the FLAG OFF, should not be able to select HARVESTER CLUSTER
- Check for cases with the FLAG ON -> transition -> FLAG OFF, should be able to see the HARVESTER CLUSTER and remove it.

- CreateEdit AppBundle on fleet, Cluster Target page

### Screenshot/Video

- When the feature flag is turned off, it doesn’t show on all
Pasted Graphic

- Doesn’t show on the select
Pasted Graphic 1

- Can still select through Cluster Group [EXPECTED]
Pasted Graphic 2

With the flag turned off you can check the flag for already created AppBundles
Pasted Graphic 3

You can see on the select but it is disabled
Pasted Graphic 4

As soon as you remove the option cannot be added again.
Pasted Graphic 5

When creating with the FLAG ON, doesn’t have the MATCH SELECTORS
App Bundle Create

If Editing with FLAG ON, it doesn’t have the MatchSelector
App Bundle disabled-group-with-harv

### Checklist
- [x] The PR has a Milestone 


richard-cox:
@torchiaf i took a quick look at this to see if we could get into 2.13.0, however think it introduces more risk than i'm comfortable with. mainly around the fix for the `all in workgroup` option. 

previously
- the 'all' option always included a filter for `not harvester`

now
- we need to pass in an empty selector. if the selector is not provided or is an empty array it will not select any clusters

Angles i was looking at
- existing resources that restrict clusters by `not harvester` --> ff enabled --> how the form handles them
  - with update the filter remains until the user changes something
  - user sees 'select cluster' with nothing there
  - think this is fine
- gitrepos / helmops resources
- harvester clusters show in clusters list (and we don't show the warning model)

Can you take a look at this and confirm it's either too risky for 2.13.0 or ok to go in? If the later we'll need to get into 2.14.0 and 2.13.1

richard-cox:
@torchiaf now that 2.14 is on the radar can you take a look?

marcelofukumoto:
Main concern of these changes would be for the Harvester Cluster to disappear.
I did some tests and found out it disappear so the UI/UX is weird because the option disappear.
image

It feels the ticket as a whole would need some UX blessing.
The concerns would be for this specific case. for normal cases should be ok, the best I could think of is to have some way to know that those are harvester cluster.

So I think in the end the biggest problem is that turn on -> create app bundle -> turn off -> edit app bundle flow.
I think I can do some tweaks, but the way it is should be ok, at least it still showing on the screen.

I will check if we have some Rancher with both Harvester and normal Cluster.

shell/components/fleet/FleetClusterTargets/index.vue
```diff
@@ -170,6 +183,10 @@ export default {
         });
     },
 
+    isHarvesterClustersEnabled() {
```
richard-cox: this is a dupe of areHarvesterHostsVisible

shell/components/fleet/FleetClusterTargets/index.vue
```diff
@@ -276,14 +293,20 @@ export default {
 
     toTargets(): Target[] | undefined {
       switch (this.targetMode) {
-      case 'none':
+      case 'none': // No clusters
         return undefined;
-      case 'all':
+      case 'all': // All clusters in workspace
+        // return this.isHarvesterClustersEnabled ? this.targets : [excludeHarvesterRule]; TODO: RC
```
richard-cox: This was the part that i wasn't sure about. I think this was the new line that came in with recent changes... but conflicted with the lines below originally from this pr?

Should the comment go or the code below updated?

richard-cox (commented):
Couple of bugs
- enable ff, create a git repo bundle, open Resources --> Bundles. x bundle is hidden as it belows to a harvester cluster... message shown. Note that it doesn't show on refresh, though shows again after visiting Resources / Git Repos or Helm Ops pages
- same "not shown" banner shown for cluster registration tokens resource

One UX bug but i don't think related to this PR
- ff off, created a git repo bundle (with bad source repo) that targeted all clusters, enabled ff, editing bundle and set back to all clusters. detail page "no target clusters" banner. fixed bad source repo, detail page banner now not shown. so i think this is a bad message, the bundle does target clusters however none are ready?

shell/components/fleet/FleetClusterTargets/index.vue
marcelofukumoto: This would try to keep the targets from before, which would have the similar empty case.
On the below implementation it tries to clear up everything which represents ALL better.
It actually tries to recover from previous data. But overall I feel it is better to enforce it here.

If we use the code on top the excludeHarvesterRule would need to try to merge the this.targets to the exclude for the logic to make sense.

shell/components/fleet/FleetClusterTargets/index.vue
marcelofukumoto: Removed

marcelofukumoto:
> Couple of bugs
> 
> * enable ff, create a git repo bundle, open Resources --> Bundles. x bundle is hidden as it belows to a harvester cluster... message shown. Note that it doesn't show on refresh, though shows again after visiting Resources / Git Repos or Helm Ops pages
> * same "not shown" banner shown for cluster registration tokens resource
> 
> One UX bug but i don't think related to this PR
> 
> * ff off, created a git repo bundle (with bad source repo) that targeted all clusters, enabled ff, editing bundle and set back to all clusters. detail page "no target clusters" banner. fixed bad source repo, detail page banner now not shown. so i think this is a bad message, the bundle does target clusters however none are ready?

Comparing the GitRepo and the list, the difference is that the GitRepo and HelmOps list pages they fetch the Fleet Cluster which contains the required data to check if they are Harvester or not. Those are bug not added from this PR but I tried to fix those.
To make it better I merged the 2 FleetBundle with the bundle list page as well, because I was applying changes on both all the time.

shell/utils/fleet.ts
```diff
@@ -83,8 +90,11 @@ class Application {
       return target;
     });
 
-    // Check if targets contains only harvester rule after name normalizing
-    if (isEqual(normalized, [this.excludeHarvesterRule])) {
+    // Check if targets contains only harvester rule or no rule at all after name normalizing
+    // That means the ALL option has been selected previously
+    // one case for feature harvester-baremetal-container-workload ON
+    // and another for feature harvester-baremetal-container-workload OFF
+    if (isEqual(normalized, [this.includeAllWorkgroupRule]) || isEqual(normalized, [this.excludeHarvesterRule])) {
```
marcelofukumoto, line 97: I changed the logic here.
The problem of using the FEATURE FLAG is that the creation of the CLUSTER is based on the flag when it was created. So both 2 state means "ALL" for the cluster selection.

Ideally the information of what was the state of the flag when the BUNDLE is created could be inside the actual bundle so we can use it. But considering that both conditions represent the same "ALL" but in different flags, checking for one or another feels enough for the case. If there are more complexes check later in the future for the state of the flag, we would need something separately to check on that.

marcelofukumoto:
Applied changes to cluster registration token.

shell/components/fleet/FleetBundles.vue
```diff
@@ -18,14 +22,49 @@ export default {
...
+        cluster: {
+          inStoreType: 'management',
+          type:        FLEET.CLUSTER
+        },
+      }, this.$store);
+    } catch (e) {
+    }
+
+    // Only fetch all bundles if we're in list mode (no value prop)
+    if (!this.value && this.resource) {
```
richard-cox:
```suggestion
    if (this.isListMode) {
```

shell/components/fleet/FleetBundles.vue
```diff
@@ -18,14 +22,49 @@ export default {
...
+
+    useQueryParamsForSimpleFiltering: {
+      type:    Boolean,
+      default: false
+    }
   },
 
   async fetch() {
+    try {
+      await checkSchemasForFindAllHash({
```
richard-cox: these two could be huge calls, so need to be done together

We have some examples where we stick everything into a hash, wait for those promises to complete and then extract the properties we need


```
    const _hash = {
      hciClusters:  this.$store.dispatch(`${ inStore }/findAll`, { type: HCI.CLUSTER }),
      mgmtClusters: this.$store.dispatch(`${ inStore }/findAll`, { type: MANAGEMENT.CLUSTER }),
      catalogLoad:  this.$store.dispatch('catalog/load', { reset: true }),
    };

    const hash = await allHash(_hash);

    this.hciClusters = hash.hciClusters;
    this.mgmtClusters = hash.mgmtClusters;
```

here we can conditionally add the await this.$fetchType(this.resource);

richard-cox: i played around some more and not sure these are needed?

it looks like we already fetch all fleet clusters somewhere core in this product

richard-cox: i gave it a bash without this line (and the one in shell/list/fleet.cattle.io.clusterregistrationtoken.vue) and couldn't repro the `x bundles are hidden as they belong to Harvester clusters that can not be used with Continuous Delivery` banner bug

shell/utils/fleet.ts
richard-cox: Something is broken the other side now.

creating a bundle with FF on creates it with the targets.clusterSelector.matchExpression excluding the harvester cluster.

after enabling the FF and going to edit the bundle (which we'll now advise via release note) the radio button 'All Clusters in the workspace' is selected and we show all clusters except the harvester one. This is the semantic issue from before (we show "all" as selected but not all clusters are selected).

for this case the PR used to switch to "Manually selected" which showed the actual clusters the bundle would use (so no harvester cluster)

So i think if we stick with this approach every time we enter we should apply the 'all' matchExpression

- ff on and all selected? apply excludeHarvesterRule
- ff off and all selected? apply includeAllWorkgroupRule

we can always later on warn the user if we do change what's there

shell/components/fleet/FleetBundles.vue
marcelofukumoto: Really? I tested that the data from CLUSTERS were missing so it was necessary to fetch those to make sure the data that they are from "harvester" were needed. Let me do the test with and without and see it. But I am positive that I saw the bug without add this in.

marcelofukumoto: Here is the recording with the bug on the FleetBundles. What I am doing is acquiring same data that is required on gitrepos. Without the data from the Clusters you can't filter because you dont have the information if it is HARVESTER or not.

https://github.com/user-attachments/assets/ef3b07e6-dd41-42fe-9c09-f9714b3ac408

marcelofukumoto: Now with the added code to fetch the data

https://github.com/user-attachments/assets/12c2a33d-12cf-43e1-9ac5-29cd199408b8

shell/components/fleet/FleetBundles.vue
marcelofukumoto: Changed as suggested :)

shell/utils/fleet.ts
marcelofukumoto: > creating a bundle with FF on creates it with the targets.clusterSelector.matchExpression excluding the harvester cluster.

Start with the Flag OFF
I guess this here should be flag off since it says it will exclude.
All good: 
image

> after enabling the FF and going to edit the bundle (which we'll now advise via release note) the radio button 'All Clusters in the workspace' is selected and we show all clusters except the harvester one. This is the semantic issue from before (we show "all" as selected but not all clusters are selected).

Now enable the FF to ON
Editing it:
image
At first it will show the previously selected ones. Ok, I thought that was the correct, so my understanding was incorrect. I believe it was better to keep the screen as it is. Which is with the previous filter.
That would be the case for the VIEW mode as well.

---------

I tried to do something to "keep" the previous selected ones:

https://github.com/user-attachments/assets/25381863-023e-4587-83ec-864ee6bab78f

So it automatically select the ones that from previous state.
FF OFF Create -> FF ON Create means that there are now CLUSTER SELECTED < NEW ALL, so that means we change to 'clusters' and select the ones that are NOT HARVERSTER, so it keeps consistent to what it had before.

For the FF ON Create -> FF OFF Create, CLUSTER SELECTED > ALL, so that means we can keep ALL and show even the new ones.

----------

But now, let's go with the approach that seems like the correct one.
```
ff on and all selected? apply excludeHarvesterRule
ff off and all selected? apply includeAllWorkgroupRule
```

Are you sure on this one? Shouldn't be the opposite? 
ff on and all selected -> apply the includeAllWorkgroupRule
ff off and all selected -> apply the excludeHarvesterRule

Which is applying the NOW filter to the edit mode

---------

Not sure if that is correct. So let me summarize:
- My approach was to keep the screen intact (show all with the previous filter), which is weird since after changing the screen it changes it.
- Another approach that I showed before is to select the correct ones, for OFF -> ON is ok, but for ON -> OFF it can select all and keep the harvester ONE.
- Last approach is I think what I understood you want. Just set same as "CREATE" even if not exactly what was created before, it will show the NEW state. I think that will bug slightly on the "VIEW" mode. So I will keep the VIEW mode as the old filter.

https://github.com/user-attachments/assets/06b73cd9-9edf-490d-bc63-444cd4e004c4

So seems like this one is the correct one?

marcelofukumoto:
Retested the process again:

Happy Case
￼
App Bundle Create
Pasted Graphic 6

￼
If selected manually, even on OFF it will still show
But if removed you cannot it back again
￼
Pasted Graphic 8


——

￼
It will not show the option on all
Pasted Graphic 3
Pasted Graphic 5

Independently of the previous state, it will ignore and use based on the feature flag.

image
The only interesting part is that the Show Configuration will show it properly based on what is saved. So if the flag is off because it was created without the NotIn, it will respect that on the Show Configuration only.

----

The biggest changes at the end is that we made the EDIT mode to respect more the flag and show the proper data, makes it simpler since ignores completely what is there, it only worries about the option selected previously.

LGTM to test and move forward. Let me know if require a more in-depth discussion but we can check it on our meeting.

@richard-cox it is ok to review again.

richard-cox (approved):
I haven't done the normal meticulous review on the latest, but did give it a run through with it off, on and then also transitioning from off to on and it meets the requirement of `all clusters` showing what will be selected on save.

We need to release note the information that after toggling the feature they need to edit bundles created in the UI that target all clusters and ensure the new selection is as required

PR #14982, opened 2025-07-30
Fixed Harvester detail page through Cluster Management

### LINKED ISSUE #14285: Check if Harvester cluster import/edit uses Ember
As title says - we want to reduce Ember pages - need to check if Ember is still used for Harvester clusters
After discussion with Harvester team, we decided that editing Harvester cluster should not be allowed from Cluster management  page

### LINKED ISSUE #15005: 'Show configuration' component should not allow editing configuration/yaml for resources that do no allow it
**Setup**
- Rancher version: 2.12 and up
- Rancher UI Extensions: N/A
- Browser type & version: N/a

**Describe the bug**

We should not allow editing resources through 'Show configuration' if that action is not supported due to resource type or user's permissions

**To Reproduce**

Scenario 1:
1. Create a standard user and add it as a Cluster member to a downstream cluster
2. Login as that user
3. Go to Cluster Management and open action menu for the cluster. Note what actions are supported
4. Click 'Show Configuration' 

**Result**
Editing Configuration and Editing YAML are both supported

**Expected Result**

User should be able to view configuration and yaml but not edit them

Scenario 2:
1. Login as admin
2. Go to Cluster Management and open action menu. Note that you can edit configuration but only view YAML
3. Click 'Show Configuration' 

**Result**
Editing Configuration and Editing YAML are both supported

**Expected Result**

User should be able to view and edit configuration but only view yaml

**Screenshots**

**Additional context**

eva-vashkevich:
### Summary
Fixes #14285 
Fixes #15005  

Modified detail drawer to only allow editing if that action is enabled for the resource. This prevented loading Ember pages for Harvester cluster from management cluster detail page.

Added a way to see if cluster should be editable.
Separated check if resource can be edited through config vs through yaml.
Only enabled config tab for Harvester when it is accessed through Virtualization Manager

### Technical notes summary

Added a new canEdit property to the models that indicates that resource configuration should be editable.

For edit:
1. Create a standard user and add them as Cluster Member to a cluster.
2. Log in as that user
3. Cluster Management -> Check that 'Show Configuration' is enabled.
4. Verify that you can only view Config and YAML tabs but not edit them
5. Log in as admin
6. Cluster Management -> Check that 'Show Configuration' is enabled.
7. Verify that you can edit Config tab and view YAML tab
8. Go to local cluster -> workloads. Find one that allows editing YAML. Go to 'Show configuration'
9. Verify that you can view and edit both Config and YAML tabs
10.  Verify that same applies when 'Show configuration' is accessed through resource detail page

For Harvester:
Import Harvester cluster
1. Global settings -> feature flags -> "harvester-baremetal-container-workload" -> enable
5. Cluster Management -> Check that 'Show Configuration' is enabled, but only YAML tab is shown and you can only view it, but not save
11. Click on the Harvester cluster and got to the detail page. -> Check that 'Show Configuration' is enabled, but only YAML tab is shown and you can only view it, but not save

I've overwritten canEdit for Harvester resource, so that might introduce regressions, but I didn't see it being used in both dashboard and harvester ui extension.

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/Drawer/ResourceDetailDrawer/index.vue
+++ b/shell/components/Drawer/ResourceDetailDrawer/index.vue
@@ -50,6 +50,10 @@ const action = computed(() => {
     action
   };
 });
+
+const canEdit = computed(() => {
+  return props.resource.canEdit;
+});
 </script>
 <template>
   <Drawer
@@ -83,6 +87,7 @@ const action = computed(() => {
     </template>
     <template #additional-actions>
       <RcButton
+        v-if="canEdit"
         :primary="true"
         :aria-label="action.ariaLabel"
         @click="action.action"
--- a/shell/components/Drawer/ResourceDetailDrawer/index.vue
+++ b/shell/components/Drawer/ResourceDetailDrawer/index.vue
@@ -17,6 +17,8 @@ export interface Props {
 }
 </script>
 <script setup lang="ts">
+const editBttnDataTestId = 'save-configuration-bttn';
+const componentTestid = 'configuration-drawer-tabbed';
 const props = defineProps<Props>();
 const emit = defineEmits(['close']);
 const store = useStore();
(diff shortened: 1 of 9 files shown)


shell/components/Drawer/ResourceDetailDrawer/index.vue
```diff
@@ -50,6 +50,10 @@ const action = computed(() => {
     action
   };
 });
+
+const canEdit = computed(() => {
+  return props.resource.canEdit;
```
codyrancher: Edit: Based on the resource-class definition I think this needs to be canCustomEdit.

https://github.com/rancher/dashboard/blob/master/shell/plugins/dashboard-store/resource-class.js#L936-L941

eva-vashkevich: @codyrancher I think we might want to even check if edit is in the list of this._availableActions, in case it is disabled like we do for Harvester ie
https://github.com/eva-vashkevich/dashboard/blob/master/shell/models/provisioning.cattle.io.cluster.js#L163
what do you think?

codyrancher: I think we need to stick with canCustomEdit so we don't further muddy the source of truth when trying to figure out if a resource can be edited.

I think maybe doing:
`return props.resource.canCustomEdit && !props.resource.isRke1; // in 2.12 we removed access to the edit pages of rke1 resources to eliminate the use of ember pages`

That way when we remove rke1 from the code base this will get removed and we get to maintain the same source of truth elsewhere.

eva-vashkevich: I feel like this introduces too many conditions, because editing is disabled for Harvester instances as well, but not everywhere, so that will be messy, and now we need to keep track of that here as well

codyrancher: I don't love it but I think it will work so I'll approve it if that's what you'd like to do.

More of a rant than an attempt to sway you... It seems like harvester bucks too many conventions as is and should maybe conform by overriding canCustomEdit so we don't need to special case rancher code to accommodate it yet again.

Given that this will be backported I think your version will be safer when accounting for harvester too.

image

eva-vashkevich: Harvester already does necessary changes for case where Harvester instances are accessed through Harvester product. The complexity arises when we need to handle those instances outside of Harvester, because many of the generic functionalities are not supported for Harvester clusters or they should be done through Harvester

codyrancher (approved):

PR #14983, opened 2025-07-30
Chart readme page

### LINKED ISSUE #14919: Display chart readme separately in a new tab
On chart detail page, show a small secondary button in the top right corner of the chart readme container when the user hovers / focuses this container.

- on hover, show the following tooltip: "Open the readme in a new tab";
- on click, open the read me in a new tab (only the readme, nothing else).
- hide the button when the the container in no longer in focus.

Image

momesgin:
### Summary
Fixes #14919 

### Technical notes summary

- Added a new path `'/c/:cluster/apps/charts/chart-readme'` to render the chart readme separately
- Used the existing `standalone.vue` to render the page with no header, sidenav etc.
- Created a helper `getStandaloneReadmeUrl` to generate a url with the necessary parameters to be able to render the chart readme page without the need to fetch data or access the store
-  Utilized the session storage to pass the chart readme content to the new tab
- Tests 

Check the chart readme on these pages:

- chart detail page
- extensions page (click on an extension card)
- chart install page (click on "View chart info" button)

### Screenshot/Video

https://github.com/user-attachments/assets/49de0b7d-41c3-4cd4-bcb4-418309d5d07c

### Checklist
- [x] The PR has a Milestone 


shell/components/ChartReadme.vue
```diff
@@ -17,34 +18,74 @@ export default {
       default: true
     }
   },
+  created() {
+    // When rendered as a standalone page, the theme is passed as a query parameter and applied to the body.
+    if (this.isStandalone) {
+      const theme = this.$route.query.theme || 'light';
+
+      document.body.classList.add(`theme-${ theme }`);
```
codyrancher, line 26: I think it might make sense to create a separate template which includes the theme so you don't need to do this in the component.

shell/components/ChartReadme.vue
```diff
@@ -17,34 +18,74 @@ export default {
...
+
+      if (this.isStandalone) {
+        const { storageKey } = this.$route.query;
+
+        if (storageKey === CHART_README_STORAGE_KEY) {
+          const stored = sessionStorage.getItem(storageKey);
+
+          return stored ? JSON.parse(stored) : null;
+        }
+      }
```
codyrancher, line 58: I'm not sure I understand the motivation for using session storage. It seems like maybe you're doing this because the standalone page is unauthenticated.

I think the need for this could be removed if we made a new template as mentioned above.

codyrancher: It also means I couldn't share or save a link to the readme.

codyrancher (commented):
Reading through the original issue and this PR it's not obvious to me why we're using session storage here. 

It looks like it might be a work around, if so I think we may have some other options.

shell/components/ChartReadme.vue
oboc-sts: to double down on @codyrancher's comment above: sharing the link behind a readme is a must have.

oboc-sts:
@momesgin looks good imho - I can't yet evaluate the comment I've made [previously](https://github.com/rancher/dashboard/pull/14983#discussion_r2244621797)

As a secondary issue (simply because I don't know if it's part of this PR or not): the dashed divider (between the page header and the page content) needs to have a different border color when used in dark mode (I believe the equivalent is or should be `#4A4B52`?). 

image

oboc-sts (commented):
@momesgin looks good imho - I can't yet evaluate the comment I've made [previously](https://github.com/rancher/dashboard/pull/14983#discussion_r2244621797)

As a secondary issue (simply because I don't know if it's part of this PR or not): the dashed divider (between the page header and the page content) needs to have a different border color when used in dark mode (I believe the equivalent is or should be `#4A4B52`?). 

image

shell/components/ChartReadme.vue
momesgin: @codyrancher I wanted the chart readme page to load super fast by avoiding any data fetching. Initially I passed the whole readme content via the url using encoding/decoding, but I ran into browser url length limits for larger readme contents, so I switched to using session storage which solved the size issue, but I forgot about losing the shareability. Thanks for pointing that out, I'll look into creating a new template approach, I think we'll lose the super fast rendering but the ability to share the page is more important in this case.

nwmac:
@momesgin Is this for 2.13? Still draft

momesgin:
> @momesgin Is this for 2.13? Still draft

No, changed the milestone

PR #15001, opened 2025-07-31
[2.11] Migrate to staging registry for e2e tests

rak-phillip:
### Summary

This targets the staging registry for running e2e tests in the release-2.11 branch.

### Occurred changes and/or fixed issues

- target the staging registry for running e2e tests in the release-2.11 branch

### Technical notes summary

This is release maintenance. See #14364 for a reference to the change made for release-2.10

### Areas or cases that should be tested

e2e tests should pass in CI

### Areas which could experience regressions

e2e tests should pass in CI

### Screenshot/Video

NA

### Checklist
- [x] The PR is linked to an issue and the linked issue has a Milestone, or no issue is needed
- [x] The PR has a Milestone 
- [x] The PR template has been filled out
- [x] The PR has been self reviewed 
- [x] The PR has a reviewer assigned
- [x] The PR has automated tests or clear instructions for manual tests and the linked issue has appropriate QA labels, or tests are not needed
- [x] The PR has reviewed with UX and tested in light and dark mode, or there are no UX changes
- [x] The PR has been reviewed in terms of Accessibility

PR #15032, opened 2025-08-04
Update charts e2e test so that it doesn't rely on version

rak-phillip:
### Summary
This updates the `should call fetch when route query changes with valid parameters` test so that it navigates to the first active link in the list of versions.

The test does not need to rely on a specific version in order to assert its case. Relying on the specific version is prone to failure if the backend stops providing it for any reason.

- Remove specific version requirement from `should call fetch when route query changes with valid parameters` test

### Technical notes summary

This change has been introduced to the `release-2.11` and `release-2.12` branches[^1] [^2] - the intent is to keep the implementation as consistent as possible between all active development branches. 

[^1]: https://github.com/rancher/dashboard/pull/15014
[^2]: https://github.com/rancher/dashboard/pull/15001

Affected e2e test should pass.

NA

### Screenshot/Video

NA

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/cypress/e2e/po/pages/explorer/charts/chart.po.ts
+++ b/cypress/e2e/po/pages/explorer/charts/chart.po.ts
@@ -55,6 +55,10 @@ export class ChartPage extends PagePo {
     return this.self().find('[data-testid="chart-versions"]');
   }
 
+  versionLinks() {
+    return this.versions().find('[data-testid="chart-version-link"]');
+  }
+
   showMoreVersions() {
     return this.self().find('[data-testid="chart-show-more-versions"]');
   }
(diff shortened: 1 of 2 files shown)


aalves08 (commented):
It's not finding the element on the DOM @rak-phillip

rak-phillip:
> It's not finding the element on the DOM @rak-phillip

Indeed, it looks like the implementation changed in `master`. No problem here, all we need to do is change the selector - the rest of the change remains the same. 

@aalves08 this is ready for another review.

aalves08:
@rak-phillip so many tests failing 😱 😱 😱

rak-phillip:
> @rak-phillip so many tests failing 😱 😱 😱

@aalves08 those are all sorry cypress failures and not related to this change.. If you run the affected test yourself, you will see it pass.

rak-phillip:
@aalves08 all tests pass with a rerun. Can you review the change when you are ready?

cypress/e2e/po/pages/explorer/charts/chart.po.ts
```diff
@@ -55,6 +55,10 @@ export class ChartPage extends PagePo {
     return this.self().find('[data-testid="chart-versions"]');
   }
 
+  versionLinks() {
+    return this.self().find('.chart-body__info > :nth-child(1) a');
```
momesgin: I could be wrong but I thought we should only use testid attributes

rak-phillip: I updated the change so that it uses a combination of the existing `versions()` method along with the `chart-version-link` testid. Thanks for the feedback, I think this looks much better as a result.

PR #15076, opened 2025-08-06
Remove the concept of legacy/latest detail pages

codyrancher:
### Summary
This is one of 2-3 PRs to allow for detail pages to explicitly customize areas of the page. When finished I'll add support for #14977.

I only wanted to keep around the legacy pages in case we needed to disable the latest pages. Since things seem stable I'm removing legacy.

This is also one step to make it easier to fully customize more pages without doing full rewrites.

We only have one ResourceDetail implementation now. I moved the legacy definition back into index and allow models to define `fullDetailPageOverride` if there's a desire to override the default page.

Detail pages of configmap, secret and projectsecret as well as at least one other resource with a detail page.

See above

### Checklist
- [x] The PR has a Milestone 


shell/detail/configmap.vue
codyrancher, line 1: It looks like a large change but I really only moved  `shell/pages/explorer/resource/detail/configmap.vue` into this file.

Only difference being that we now receive the value from the parent.

shell/detail/secret.vue
codyrancher, line 1: It looks like a large change but I really only moved `shell/pages/explorer/resource/detail/secret.vue` into this file.

Only difference being that we now receive the value from the parent.

shell/components/ResourceDetail/index.vue
codyrancher, line 1: I moved `shell/components/ResourceDetail/legacy.vue` into here and added `isFullPageOverride` checks.

momesgin (dismissed):

momesgin (approved):

PR #15146, opened 2025-08-13
Implement the Tag component as defined by the DS system

codyrancher:
### Summary
image

Storybook component

### Screenshot/Video

Looks like story book is partly broken right now so I added the components to a page to test.

The dark theme follows the tags we already have in the detail page masthead even though it's not specified in the DS figma.

image

image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/pkg/rancher-components/src/components/Pill/RcTag/RcTag.vue
+++ b/pkg/rancher-components/src/components/Pill/RcTag/RcTag.vue
@@ -1,4 +1,5 @@
 <script setup lang="ts">
+import RcButton from '@components/RcButton/RcButton.vue';
 import { RcTagProps } from './types';
 
 const props = withDefaults(defineProps<RcTagProps>(), { disabled: false, showClose: false });
@@ -11,13 +12,14 @@ const emit = defineEmits(['close']);
     :class="{[props.type]: true, disabled: props.disabled}"
   >
     <slot name="default" />
-    <button
+    <RcButton
       v-if="props.showClose"
+      ghost
       :aria-label="props.closeAriaLabel"
       @click="emit('close')"
     >
       <i class="icon icon-close" />
-    </button>
+    </RcButton>
   </div>
 </template>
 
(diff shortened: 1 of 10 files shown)


codyrancher:
@rak-phillip the close button has been implemented. We also got a comment back in figma. Would you mind taking a look at this when you get a chance?

pkg/rancher-components/src/components/Pill/RcTag/RcTag.vue
```diff
@@ -0,0 +1,90 @@
...
+    :class="{[props.type]: true, disabled: props.disabled}"
+  >
+    <slot name="default" />
+    <button
+      v-if="props.showClose"
+      :aria-label="props.closeAriaLabel"
+      @click="emit('close')"
+    >
+      <i class="icon icon-close" />
+    </button>
```
rak-phillip: This button is missing focus styles when it receives keyboard focus. I suggest that we use `RcButton` so that we can get consistent button behavior out of the box. The ghost role should be able to satisfy the need for having a barebones button that we can style ourselves.

pkg/rancher-components/src/components/Pill/RcTag/RcTag.vue
```diff
@@ -0,0 +1,90 @@
...
+
+<template>
+  <div
+    class="rc-tag"
+    :class="{[props.type]: true, disabled: props.disabled}"
+  >
+    <slot name="default" />
+    <button
+      v-if="props.showClose"
+      :aria-label="props.closeAriaLabel"
```
rak-phillip, line 18: I'm considering if we should provide a default value for `aria-label` since this is an icon button. We can provide some context that's better than an icon with something simple like `Close Tag` and developers can override this to provide more context about the specific tag.

rak-phillip (commented):
Looks good - I have a few comments about the button.

pkg/rancher-components/src/components/Pill/RcTag/RcTag.vue
codyrancher: Done

pkg/rancher-components/src/components/Pill/RcTag/RcTag.vue
codyrancher: I omitted this change based on our offline conversation.

Ultimately we're not sure we want to add translations as a dependency for rancher components

PR #15292, opened 2025-09-02
Fix missing Fleet clusters updates after saving

### LINKED ISSUE #15284: [BUG] Fleet Cluster Labels are updated or removed from UI until Force Update used.
**Setup**
- Rancher version: v2.12-028474b31b5a8b01746f12b5ef88e722d377aa42-head
- Rancher UI Extensions:

**Describe the bug**

Apply cluster label, it is not shown immediately on UI, it takes too much time to reflect. At the end Force Update have to perform to get the label updated on UI.

This is happening while removing the cluster label.

**To Reproduce**

- Navigate to Continuous Delivery --> Clusters
- Edit the cluster and add label
- Wait for some time to reflect it on UI.

**Result**
Label change is not reflected on UI in time.

**Expected Result**

It should be reflect the cluster label on cluster page immediately, without using the Force Update.

**Screenshots**

**Additional context**

https://github.com/user-attachments/assets/07fcf872-12ed-4ca1-ab42-fdb47f4a8c38

torchiaf:
### Summary
Fixes #15284

Cached resource's re-watch was introduced in https://github.com/rancher/dashboard/pull/15011.
However, the `dispatch('watch')` call  triggers an `unwatch` with missing `type` and `id` params.

It seems we have a raise condition or a bug in the re-watch mechanism.

[Screencast from 2025-09-02 10-15-44.webm](https://github.com/user-attachments/assets/b705bba3-eac3-47fa-805a-7794031a7e56)

UPDATE @richard-cox :

When the `ui-sql-cache` flag (pagination) is `disabled`, the UI shows the updates, when enabled it doesn't.
Using a local UI instance on `master` branch, I compared the websocket incoming messages with the 2 flag conditions after adding a new label and hitting `save` in the edit form.
Left enabled, right disabled:

Screenshot from 2025-09-02 12-42-12

In the left side - no updates, flag enabled - there is a missing `resource.create` for `fleet.cattle.io.cluster` resource.
Pretty sure that this the root cause of the issue, since it's the only difference between the incoming messages.

### Technical notes summary

- We verified that incoming data from websocket are correct.
- The fix consists in adding a wait condition after saving the clusters, so that they receives the updates, then show up the clusters table.

1. Go to Fleet -> Clusters tab
2. Edit the `local` cluster, add a generic label
3. Verify that the cluster lists show the correct labels.

### Screenshot/Video

[Screencast from 2025-09-04 11-24-42.webm](https://github.com/user-attachments/assets/da1c2d5d-e2f9-47a3-974d-9ee20ffe187e)

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/edit/fleet.cattle.io.cluster.vue
+++ b/shell/edit/fleet.cattle.io.cluster.vue
@@ -61,6 +61,7 @@ export default {
 
         await this.normanCluster.save();
 
+        // Changes (such as labels or annotations fields) to normanCluster are reflected in the fleet cluster via Rancher services, so wait for that to occur
         await this.waitForFleetClusterLastRevision();
 
         this.done();
@@ -74,15 +75,15 @@ export default {
     async waitForFleetClusterLastRevision() {
       const inStore = this.$store.getters['currentProduct'].inStore;
 
+      const currRev = this.value?.metadata?.resourceVersion;
+
       try {
         return await this.value.waitForTestFn(() => {
-          const curr = this.value.metadata?.resourceVersion;
-          const nue = this.$store.getters[`${ inStore }/byId`](this.value.type, this.value.id)?.metadata?.resourceVersion;
+          const rev = this.$store.getters[`${ inStore }/byId`](this.value.type, this.value.id)?.metadata?.resourceVersion;
 
-          return curr && curr !== nue;
+          return currRev && currRev !== rev;
         }, `${ this.value.id } - wait for resourceVersion to change`, 1000, 200);
       } catch (e) {
-
       }
     }
   },
--- a/shell/edit/fleet.cattle.io.cluster.vue
+++ b/shell/edit/fleet.cattle.io.cluster.vue
@@ -56,9 +56,12 @@ export default {
     async save(buttonDone) {
       try {
         this.errors = [];
+
         await this.value.save();
 
         await this.normanCluster.save();
+
+        await this.waitForFleetClusterLastRevision();
 
         this.done();
         buttonDone(true);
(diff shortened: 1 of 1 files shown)


richard-cox:
The process looks good and the socket messages sent (watch/unwatch) and received (resource.change) are correct, however there's two bugs

1. user is on the cluster list page, watching all clusters --> user navs to edit/detail page, unwatch all clusters +  watch specific cluster --> user navs to list page, unwatches specific cluster + watches all clusters WITHOUT a revision
   - no revision means only new changes will be recieved
   - the local cache though is out of date and should send a revision to get updates to all resources from it's previous state
   - fix for this is `findAll` action `watch` call args basically nukes need fixing - `revision:  '',` needs removing
1. user is on the cluster list page, watching all clusters --> user navs to edit/detail page, unwatch all clusters +  watch specific cluster --> user navs to list page, resource.changes message received containing updated cluster, unwatches specific cluster which clears out the queue containing the pending resource.changes 
   - given we've just junked the change the cluster list never shows the updated resource
   - note we won't hit this pretty much anywhere else, 
     - the updates resource (in response of the save http request) is pushed to store, however the fleet cluster won't contain the updates labels because...
     - labels are applied a proxy way (read from fleet cluster, but set in mgmt cluster, which are synced back to fleet cluster)
     - if that loop wasn't there the updated labels would be added to the cache immediately. unfortunately i don't think the label stuff will change anytime soon
   - solution here could be a few different things
     - flush the queue for resources we've unwatched, though this could mean the cache gets polluted (if cache has been updated with resource set A and the watch being flushed contains a change from resource set B it will appear as if B resource is in A)
     - whilst in the fleet cluster save block until we've received the updated labels in the fleet cluster (could be a force fetch on the mgmt cluster)

Would vote we do number 2 block on save until the resource is correctly updated, given this is a weird use case and is less impactfull than option 1. We could do this by just blocking until the revision in the store version of fleet cluster changes from it's original value

richard-cox:
image

This is the kind of flow we'd expect

1. watch all fleet clusters
2. watch specific fleet cluster + unwatch all fleet clusters
3. receive resource.change with the new fleet cluster
4. unwatch specific fleet cluster + watch all fleet clusters

 (note 3 and 4 get jumbled , but it's roughly right)

I'm not quite sure where the resource.create came from in no-vai world, those should only be sent if resourceVersion is 0 (it provides the current state of all resources via resource.create messages) or one has just been created. There's a resource.error and resource.stop on that resource watch as well, so something seems off

torchiaf:
Alright, it sounds clear to me, thanks for the explanation.
Fix n.1 works, I adopted fix n.2 though, which I'm fine with but not sure if there are others of these cases through the UI.
Please take a look.

richard-cox (commented):
Code looks ok. Best if someone takes a functional review as well

cnotv (approved):
Screenshot 2025-09-04 at 12 29 51


LGTM

PR #15357, opened 2025-09-06
Fix inconsistent extension progress statuses

### LINKED ISSUE #15356: [Extensions Page] - extension flows and the slide-in info panel
**Extension Card:**

- We always show the latest available version in the subtitle of the card
- Show the timestamp in the subtitle, this will be the “last updated” timestamp
- If the card has the installed flag visible, we display the following tooltip on hover: “Installed  [version number]”

**Tabs:**

- We get rid of the “Updates” and the “All” tabs
- “Installed + <count badge>” = shows the installed extensions only
  - the tab is selected by default when you land on the extensions page unless...
  - if there are NO extensions installed, we do NOT show the tab --> the “Available” tab becomes the default tab
- “Available” = a tab containing all available extensions that can be installed
- “Built-In” = shows the built-in extensions only
- When refreshing the page the active tab should not be changed e.g. if you're viewing the built-in tab and refresh the page you should see the built-in tab again

**Slide-In Info Panel:**

In the extension slide-in, we add 2 new sections:
- Extension tags like "Built-in", "Experimental" or "Third-party" will be displayed on top
- “Actions”:
  - title = “Actions”; position = below “versions”, always visible;
  - In this section, we will display the available actions a user can take, depending on the current status (installed vs not installed), the installed version (if any) and selected version in the “VERSIONS” section (if it’s newer or older).
  - If there are no actions available (e.g. for the built-in extensions), we simply display “No actions available” in grey, under the section heading.
- We don't display the warning banners for "Built-in", "Experimental" or "Third-party" anymore

**Actions:**

We will have consistent actions in these three places:
1. Cards action menu(kebab menu)
2. Slide-in info panel
3. Install/Uninstall modal

Here are the scenarios:

- If extension is NOT installed, you'll have ONLY the install option
- If extension is installed:
  - The "Upgrade" option becomes available ONLY if there is a higher version to upgrade to
  - The "Downgrade" option becomes available ONLY if there is a lower version to downgrade to
  - The "Uninstall" option will always be available
- "No actions available" for built-in extensions

Please reach out to @momesgin for the detailed Figma design link

momesgin:
### Summary
Fixes #15356 

### Technical notes summary

- added new progress statuses to `shell/pages/c/_cluster/uiplugins/index.vue` for rollback and update.
- tests

### Screenshot/Video

https://github.com/user-attachments/assets/5cd5ed9a-7e8a-4cf9-890e-5d686d92d4a5

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -898,9 +898,9 @@ asyncButton:
     success: Redeployed
     waiting: Redeploying Workload
   rollback:
-    action: Roll Back
+    action: Rollback
     success: Rolled Back
-    waiting: Rolling Back Workload
+    waiting: Rolling Back&hellip;
   rotate:
     action: Rotate
     waiting: Rotating&hellip;
@@ -5105,6 +5105,8 @@ plugins:
     image: Image
     installing: Installing ...
     uninstalling: Uninstalling ...
+    updating: Updating ...
+    rollingBack: Rolling back ...
     menu: Extensions menu
     reloadRancher: Reload Rancher
   descriptions:
--- a/shell/dialog/InstallExtensionDialog.vue
+++ b/shell/dialog/InstallExtensionDialog.vue
@@ -29,9 +29,9 @@ export default {
       required: true
     },
     /**
-     * Modal mode
+     * The action to perform (install, update, rollback)
      */
-    mode: {
+    action: {
       type:     String,
       default:  '',
       required: true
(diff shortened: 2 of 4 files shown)


shell/assets/translations/en-us.yaml
```diff
@@ -898,9 +898,9 @@ asyncButton:
     success: Redeployed
     waiting: Redeploying Workload
   rollback:
-    action: Roll Back
+    action: Rollback
     success: Rolled Back
-    waiting: Rolling Back Workload
+    waiting: Rolling Back&hellip;
```
momesgin, line 903: I could add a new translation key but based on the usage of this translation key I found it to be safe to just change it, I think "Rolling back..." should be still informative enough.

shell/dialog/InstallExtensionDialog.vue
```diff
@@ -29,9 +29,9 @@ export default {
       required: true
     },
     /**
-     * Modal mode
+     * The action to perform (install, update, rollback)
      */
-    mode: {
+    action: {
```
momesgin, line 34: renamed `mode` to `action` so it's not mistaken with the edit/create/view modes we have in the app.

aalves08 (approved):
LGTM

PR #15420, opened 2025-09-11
Limit the scenarios that the Storybook workflow runs

### LINKED ISSUE #7590: Restrict Storybook CI workflow to meaningful paths
# Description

Prevent Storybook to run at every single change to the application, as not every single file is meaningful for a new publication. 

## Context

Interested [paths](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) should be defined in the [current workflow](https://github.com/rancher/dashboard/blob/9a53bfa09408a42ca93f28714c8d345491a21858/.github/workflows/storybook.yaml), including only:
- `@rancher/components`
- `shell` components
- assets

E.g.

```yml
on:
  pull_request:
    paths:
      - 'pkg/rancher-components/**/*'
      - 'shell/components/**/*'
      - 'shell/assets/**/*'
    branches:
      - master
```

codyrancher:
### Summary
Fixes #7590

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/.github/workflows/storybook.yaml
+++ b/.github/workflows/storybook.yaml
@@ -1,9 +1,19 @@
 name: storybook
 on:
   push:
+    paths:
+      - 'storybook/**/*'
+      - 'pkg/rancher-components/**/*'
+      - 'shell/components/**/*'
+      - 'shell/assets/**/*'
     branches:
       - master
   pull_request:
+    paths:
+      - 'storybook/**/*'
+      - 'pkg/rancher-components/**/*'
+      - 'shell/components/**/*'
+      - 'shell/assets/**/*'
     branches:
       - master
 jobs:
--- a/.github/workflows/storybook.yaml
+++ b/.github/workflows/storybook.yaml
@@ -6,6 +6,10 @@ on:
       - 'pkg/rancher-components/**/*'
       - 'shell/components/**/*'
       - 'shell/assets/**/*'
+      - 'shell/utils/**/*'
+      - 'shell/directives/**/*'
+      - 'shell/config/**/*'
+      - 'shell/composables/**/*'
     branches:
       - master
   pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/storybook.yaml
```diff
@@ -1,9 +1,19 @@
 name: storybook
 on:
   push:
+    paths:
```
codyrancher, line 4: I changed the paths from the original issue.  `storybook` covers `.storybook` and `stories`.

codyrancher:
@cnotv I think we'll need to remove the requirement for skipping this workflow to work. If so I think we'll need to talk with the team before merging.

.github/workflows/storybook.yaml
cnotv: yeah the architecture changed in some years 😅

cnotv:
> @cnotv I think we'll need to remove the requirement for skipping this workflow to work. If so I think we'll need to talk with the team before merging.

Yes sure it makes sense. Did you had a scenario in mind which may be tricky?

codyrancher:
> > @cnotv I think we'll need to remove the requirement for skipping this workflow to work. If so I think we'll need to talk with the team before merging.
> 
> Yes sure it makes sense. Did you had a scenario in mind which may be tricky?

No particular scenario. I'm just not sure if anyone will have concerns.

.github/workflows/storybook.yaml
```diff
@@ -1,9 +1,19 @@
...
+      - 'storybook/**/*'
+      - 'pkg/rancher-components/**/*'
+      - 'shell/components/**/*'
+      - 'shell/assets/**/*'
```
rak-phillip, line 8: Should utils and directives also be included? For example, I see imports for `'@shell/utils/validators/formRules/index'` & `'@shell/directives/clean-html`

codyrancher: Yeah, you're right. I found a couple more as well with this search.
image

This shell paths seem like a list of dependencies we need to remove rancher components.

rak-phillip: Hmm.. I think we'll want to follow up to see if it can make sense to break out utilities into their own module. I'm unsure how much appetite we might have for taking on such a task, but the ideal scenario is that Rancher Components never reaches directly into shell for functionality.

PR #15421, opened 2025-09-11
Add scoped css attribute to Rancher Components

### LINKED ISSUE #15422: Rancher Components missing scoped styles
Several components in Rancher Components are missing the scoped attribute on their style tag. 

## Acceptance Criteria

- [ ] Styles are scoped in the following Rancher Components
   - [ ] `Card.vue`
   - [ ] `Checkbox.vue`
   - [ ] `RadioButton.vue`
   - [ ] `RadioGroup.vue`
   - [ ] `LabeledTooltip.vue`
- [ ] Styles related to the above components don't regress

rak-phillip:
### Summary
This scopes css for all components in Rancher Components so that css will apply only to elements of the components in question. 

Fixes #15422

- Add scoped attribute to all components in Rancher Components

### Technical notes summary

Since css in these components were applied globally when the components were rendered, there's a chance that other components could regress as a result of adding the scoped attribute. The only area where I see the highest risk of regression is in `LabeledTooltip.vue`, where we apply styles to `v-popper` via `.v-popper__popper.v-popper--theme-tooltip {`. 

Checkboxes, radio buttons, radio groups, cards and labeled tooltips should all render properly. We need to confirm that other weren't inadvertently relying on one of these styles being active while any of the given components were rendered.. 

Checkboxes, radio buttons, radio groups, cards and labeled tooltips should all render properly.

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/pkg/rancher-components/src/components/Card/Card.vue
+++ b/pkg/rancher-components/src/components/Card/Card.vue
@@ -98,7 +98,7 @@ export default defineComponent({
   </div>
 </template>
 
-<style lang='scss'>
+<style lang='scss' scoped>
  .card-container {
   &.highlight-border {
     border-left: 5px solid var(--primary);
--- a/pkg/rancher-components/src/components/Form/Checkbox/Checkbox.vue
+++ b/pkg/rancher-components/src/components/Form/Checkbox/Checkbox.vue
@@ -360,7 +360,7 @@ export default defineComponent({
   </div>
 </template>
 
-<style lang='scss'>
+<style lang='scss' scoped>
 $fontColor: var(--input-label);
 
 .checkbox-outer-container {
--- a/pkg/rancher-components/src/components/Form/Radio/RadioButton.vue
+++ b/pkg/rancher-components/src/components/Form/Radio/RadioButton.vue
@@ -226,7 +226,7 @@ export default defineComponent({
   </label>
 </template>
 
-<style lang='scss'>
+<style lang='scss' scoped>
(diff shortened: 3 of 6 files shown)


codyrancher:
> The only area where I see the highest risk of regression is in LabeledTooltip.vue, where we apply styles to v-popper via .v-popper__popper.v-popper--theme-tooltip {.

@rak-phillip I didn't run into any issues when checking out labeledinput/select but shouldn't those selectors be within a :deep()? Is the css just not applying any noticeable styles?

rak-phillip:
> > The only area where I see the highest risk of regression is in LabeledTooltip.vue, where we apply styles to v-popper via .v-popper__popper.v-popper--theme-tooltip {.
> 
> @rak-phillip I didn't run into any issues when checking out labeledinput/select but shouldn't those selectors be within a :deep()? Is the css just not applying any noticeable styles?

@codyrancher Yeah, I had a hard time with this one as well. I don't see any noticeable styles applied to the popper.. I think that we can take one of the following approaches:

- Apply the `:deep()` selectors appropriately
- Apply this style in a global css file
- Remove this style entirely 

edit: I gave this some more thought and I think it makes sense to move this to the global css file. I don't see any reason why we wouldn't want to have this fix applied globally. Removing doesn't seem to be the correct choice because we could be missing some implementation that relies on this, and the deep selector just works to more tightly couple the component to the underlying implementation.

codyrancher (approved):

PR #15436, opened 2025-09-15
SCC: UI does not update the status in offline registration

### LINKED ISSUE #15272: SCC: UI does not update the status in offline registration
**Setup**
- Rancher version: 
Rancher Prime
[Rancher](https://github.com/rancher/rancher)	v2.12.1
[Dashboard](https://github.com/rancher/dashboard)	v2.12.1
[Rancher Prime CLI](https://github.com/rancher/cli)	v2.12.1

**Describe the bug**
By uploading the certificate when registering SCC offline, the UI hangs with an indeterminate loading progress bar. By refreshing the page, the status of the registration is unregistered, but by checking the Registrations page, it can be verified that the offline registration was successful. The UI, tho, is still allowing the user to perform both online and offline registration, which makes it possible to have 2 co-existing registrations, which is not allowed.

**To Reproduce**
1- Log in Rancher Prime with no active registration as admin
2- In Global settings, go to Registration
3- Select the offline tab
4- Follow the steps of the offline registration - (ping me @IsaSih if support is required on getting a SCC registration code as well as logging in the SCC account)
5- Observe how the UI behaves after uploading the certificate
6- Refresh the page if the Ui hangs in the loading progress bar state

PS: issue is more likely to occur after sucessfully registering and then deregistering. That is, trying to register a code for the second time in that environment.

**Result**
UI error message indicating timeout is not always consistent:
Image

Observe that even after refreshing the page, the status shows as Unregistered.

https://github.com/user-attachments/assets/1fa7a7d9-cf6c-4f29-be29-d274f384bbf9

**Expected Result**

the UI must always display the correct status of a registration, with no timeout errors, and no need to refresh the page. UI must disable both online and offline registration feature when a registration is already active. 

**Additional context**
Because the UI allows the user to register again when an active registration code is already in place, I was able to register 2 different codes from different organizations in the same instance

Image

cnotv:
### Summary
Fixes #15272

BE changed architecture without inform us:
`OfflineActivationDone` is now just `Done`

@mallardduck please next time let me know 🙏 

### Screenshot/Video

Screenshot 2025-09-15 at 16 11 21

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/pkg/rancher-prime/pages/registration.composable.test.ts
+++ b/pkg/rancher-prime/pages/registration.composable.test.ts
@@ -281,7 +281,7 @@ describe('registration composable', () => {
         links:    { view: '123' },
         status:   {
           activationStatus: { activated: true },
-          currentCondition: { type: 'OfflineActivationDone' },
+          currentCondition: { type: 'Done' },
         },
       }];
 
--- a/pkg/rancher-prime/pages/registration.composable.ts
+++ b/pkg/rancher-prime/pages/registration.composable.ts
@@ -324,7 +324,7 @@ export const usePrimeRegistration = (storeArg?: Store<any>) => {
     const isError = lastCondition.type === 'RegistrationActivated';
     const isError2 = lastCondition.type === 'RegistrationAnnounced';
     const isCompleteOnline = mode === 'online' && lastCondition.type === 'Done';
-    const isCompleteOffline = mode === 'offline' && lastCondition.type === 'OfflineActivationDone';
+    const isCompleteOffline = mode === 'offline' && lastCondition.type === 'Done';
 
     return isError || isError2 || isCompleteOnline || isCompleteOffline;
   };


mallardduck:
> BE changed architecture without inform us:
>`OfflineActivationDone` is now just `Done`


Nit - the BE didn't change in the manner claimed. Both conditions still exist and `OfflineActivationDone` indicates an offline activation is done meaning that the Cert was uploaded and validated. `Done` is simply the "overall done" which applies to both offline and offline modes universally.

The behaviour that had changed is that [SCC Operator added](https://github.com/rancher/scc-operator/pull/17) a new `.status.currentCondtion` to indicate the only condition that the UI should care about to fix UI having to use "magic number" based logic.

That in mind, the UI may still see transitional states where the operator is doing a series of condition updates during the process. So based on timing it may see the current condition as `OfflineActivationDone` and moments later it might change to `Done` when the operator fully completes the process.

cnotv:
> That in mind, the UI may still see transitional states where the operator is doing a series of condition updates during the process. So based on timing it may see the current condition as `OfflineActivationDone` and moments later it might change to `Done` when the operator fully completes the process.

If nothing else is different, there should not be anything specific. I may create a second issue where I may add any catch cases which is not handled in case of mismatching architecture from BE just in case.

mallardduck:
@jbiers  can keep me honest - since I had my time off so she may know more. From my understanding the behaviour here shouldn't have changed too much and would be explained by what I described. Which is essentially the "eventual consistency" nature of k8s in action.

cnotv:
> @jbiers can keep me honest - since I had my time off so she may know more. From my understanding the behaviour here shouldn't have changed too much and would be explained by what I described. Which is essentially the "eventual consistency" nature of k8s in action.

Then in such case we should cover all the fallback to at least hint to the right direction: if miss registration, miss a secret, mismatching conditions, etc.

PR #15446, opened 2025-09-15
Fixed adding a new machine pool to existing GCE instance

### LINKED ISSUE #15417: Null errors seen when adding a new machine pool to existing GCE cluster
**Setup**
- Rancher version: v2.12.0 / v2.12.1
- Rancher UI Extensions: n/a
- Browser type & version: Chrome, 140.0.7339.133 

**Describe the bug**
When provisioning a new GCE node driver cluster multiple machine pools can be configured, and the various dropdown options for GCP are populated as expected. However, null errors are shown when attempting to add a new machine pool after the cluster has been successfully provisioned, and the drop down boxes are not populated. 

**To Reproduce**
+ Provision a GCE cluster with any number of nodes. 
+ Wait for the cluster to become available 
+ Edit the cluster and attempt to add a new pool

**Result**
Three null errors are shown and the drop down boxes are not populated

**Expected Result**
New node pools can be configured after the cluster has been provisioned 

**Screenshots**
Image

**Additional context**

+ This use case seems to have been missed when validating https://github.com/rancher/dashboard/issues/14011
+ When editing an existing GCE cluster, all of the drop down options are populated for existing node pools
+ I don't see any additional network requests to Rancher in the browser inspector when attempting to create a new node pool

eva-vashkevich:
### Summary
Fixes #15417

Adding a new machine pool was being treated as edit and not creation, so I was not fetching and setting the required information.

Modified rke2.vue to assign projectId from existing pool to all new pools
Modified google.vue and GCEImage.vue files to differentiate between editing and creation better.

### Technical notes summary

Instead of re-authenticating with projectId and credentials like we do for GKE, I am taking projectId from an existing pool since we already confirmed that it is valid during provisioning, and we do not re-check credentials for any other provider on edit.

1. Enable Google GCE driver
2. Provision GCE instance. Make sure provisioning was successful
3. Edit -> Machine Pools -> Make sure existing pool was loaded successfully and you can modify it
4. Add new pool. No errors should be displayed and you should be able to add a new pool successfully

No functionality outside of GCE should be affected

### Screenshot/Video

https://github.com/user-attachments/assets/cf7f1728-586d-4636-b90c-692e783f8ba9

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/edit/provisioning.cattle.io.cluster/rke2.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/rke2.vue
@@ -221,6 +221,8 @@ export default {
   },
 
   data() {
+    const isGoogle = this.provider === GOOGLE;
+
     return {
       loadedOnce:                      false,
       lastIdx:                         0,
@@ -270,7 +272,8 @@ export default {
       clusterAgentDefaultPC:                    null,
       clusterAgentDefaultPDB:                   null,
       activeTab:                                null,
-      isAuthenticated:                          this.provider !== GOOGLE || this.mode === _EDIT,
+      isGoogle,
+      isAuthenticated:                          !isGoogle || this.mode === _EDIT,
       projectId:                                null,
       REGISTRIES_TAB_NAME,
       labelForAddon,
@@ -1029,6 +1032,9 @@ export default {
 
       if (!this.machinePools) {
         await this.initMachinePools(this.value.spec.rkeConfig.machinePools);
+        if (this.isEdit && this.isGoogle && this.machinePools?.length > 0 && this.machinePools[0]?.config?.project) {
+          this.projectId = this.machinePools[0]?.config?.project;
+        }
         if (this.mode === _CREATE && !this.machinePools.length) {
           await this.addMachinePool();
         }
@@ -2221,7 +2227,7 @@ export default {
     </div>
     <AccountAccess
       v-if="!isAuthenticated"
-      v-model:credential="credential"
+      v-model:credential="credentialId"
       v-model:project="projectId"
       v-model:is-authenticated="isAuthenticated"
       :mode="mode"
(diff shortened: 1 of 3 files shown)


shell/edit/provisioning.cattle.io.cluster/rke2.vue
```diff
@@ -2221,7 +2227,7 @@ export default {
     </div>
     <AccountAccess
       v-if="!isAuthenticated"
-      v-model:credential="credential"
+      v-model:credential="credentialId"
```
momesgin, line 2230: were we passing a wrong value without any side effects?

shell/machine-config/components/GCEImage.vue
```diff
@@ -111,8 +111,8 @@ export default {
 
       };
     },
-    isCreate() {
-      return this.mode === _CREATE;
+    isNewOrUnprovisioned() {
+      return this.mode === _CREATE || !this.value?.metadata?.uid;
```
momesgin: sorry if this doesn't makes sense, whenever we create a new machine pool we add a `create: true` to the new machine pool object inside `addMachinePool` method in `rke2.vue`, I was wondering if we could use it here too?

momesgin: a use case of it:
https://github.com/rancher/dashboard/blob/master/shell/edit/provisioning.cattle.io.cluster/rke2.vue#L2347

shell/edit/provisioning.cattle.io.cluster/rke2.vue
eva-vashkevich: yes, it would only cause an error if we were showing AccountAccess on edit as well. But we don't really have to do that, so I fixed it in case we do in the future.

PR #15464, opened 2025-09-17
Window Manager enhancements

### LINKED ISSUE #15505: Support secondary side window

Rancher UI currently supports the Console window to display K8s resources logs and Kubectl shell.
We want to add an additional side window to support the new Rancher AI chat extension. 

- Implement the new 'Secondary' window.
- It should be possible to inject custom components from the main UI + registered components from any Rancher UI extension.
- It should be possible to resize the window.
  - Default position is on the right side of the screen.
- When the new window is open, the primary window (Kubectl console) should be automatically moved to the bottom side and become fixed (drag functionality disabled), in order to avoid malfunctions and avoid to overwhelm the main layout.

Next iterations:

- We should consider to make the window movable to the left side.
- Primary and Secondary side windows should be the same components.
- The `wm` store should be improved to support both windows

### LINKED ISSUE #15527: Merge Primary and Secondary side windows in a single component

We are introducing a new side window https://github.com/rancher/dashboard/issues/15505

- We should merge Primary (Kubectl console) and Secondary windows in a single SideWindow component.
- We should merge `wm` and `wm/secondary` store and use it to handle the SideWindows
- The windows should support dragging by tab.

**Additional context**

torchiaf:
### Summary
Fixes #15505
Fixes #15527

### Technical notes summary

It adds the multiple side windows and tabs dragging support.

- Usage

```
  this.$store.dispatch('wm/open', {
    id:         `kubectl-console-id`,
    label:      'Kubectl',
    icon:       'terminal',
    component:  'KubectlShell',
    attrs:      {},
    position:   'right',
    layouts:    ['default', 'home'], // the layouts where we want to show the side windows 
    showHeader: true, // Header can be disabled
  }, { root: true });
  
  // tabs dragging and default pin can be locked by position 
  this.$store.commit('wm/setLockedPositions', ['right', 'bottom', 'left']);
 
```

### Screenshot/Video

[Screencast from 2025-10-06 09-58-04.webm](https://github.com/user-attachments/assets/d3a286b6-3d3a-4b8c-92ff-1d0bf07ab1ef)

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/nav/WindowManager/composables/useComponentsMount.ts
+++ b/shell/components/nav/WindowManager/composables/useComponentsMount.ts
@@ -8,6 +8,33 @@ const warn = (msg: string, ...args: any[]) => {
   console.warn(`[wm] ${ msg } ${ args?.reduce((acc, v) => `${ acc } '${ v }'`, '') }`); /* eslint-disable-line no-console */
 };
 
+/**
+ * This composable is responsible for loading the tabs body components.
+ *
+ * - It supports loading components from Rancher extensions as well as from the TypeMap (defined in shell/components/Window directory).
+ * - The component is cached after the first load to optimize performance.
+ *
+ * Loading a component from extension:
+ *
+ *   // Register the component in the extension index file:
+ *   plugin.register('component', 'TestComponent', defineAsyncComponent(() => import('./pages/TestComponent.vue')));
+ *
+ *   // Use the component in a tab by specifying its name and extensionId:
+ *   store.dispatch('wm/open', {
+ *     id:            PRODUCT_NAME,
+ *     extensionId:   PRODUCT_NAME,
+ *     label:         'Label',
+ *     component:     'TestComponent',
+ *     position:      'bottom',
+ *     layouts:       [
+ *       Layout.default,
+ *       Layout.home
+ *     ],
+ *     showHeader: false,
+ *   }, { root: true });
+ *
+ */
+
 export default () => {
   const store = useStore();
 
--- a/shell/components/nav/WindowManager/composables/useDimensionsHandler.ts
+++ b/shell/components/nav/WindowManager/composables/useDimensionsHandler.ts
@@ -5,6 +5,9 @@ import { BOTTOM, LEFT, RIGHT } from '@shell/utils/position';
 import { Position } from '@shell/types/window-manager';
 import { CSS_KEY } from '../constants';
 
+/**
+ * This composable is responsible for handling the dimensions of the window manager panels.
+ */
 export default (props: { position: Position }) => {
   const store = useStore();
 
--- a/shell/components/nav/WindowManager/composables/useDragHandler.ts
+++ b/shell/components/nav/WindowManager/composables/useDragHandler.ts
@@ -3,6 +3,16 @@ import { Position, Tab } from '@shell/types/window-manager';
 import { computed, ref } from 'vue';
 import { CENTER } from '@shell/utils/position';
 
+/**
+ * This composable is responsible for handling the drag-and-drop behavior of tabs within the window manager.
+ *
+ * dragOverPositionsActive: A reactive reference indicating whether a drag operation is currently over the position areas.
+ * pinArea: A reactive reference representing the current position area being hovered over during a drag operation.
+ *
+ * Both of these references are shared across all instances of this composable (used by Horizontal and Vertical Panels)
+ * to maintain consistent drag-and-drop state.
+ */
+
 const dragOverPositionsActive = ref(false);
 const pinArea = ref(CENTER as Position);
 
--- a/shell/store/wm.ts
+++ b/shell/store/wm.ts
@@ -3,6 +3,13 @@ import { Layout, Position, Tab } from '@shell/types/window-manager';
 import { addObject, removeObject } from '@shell/utils/array';
 import { BOTTOM, LEFT, RIGHT } from '@shell/utils/position';
 
+/**
+ * This module defines the Vuex store for the window manager, managing tabs, their positions,
+ * active states, panel dimensions, and user preferences.
+ *
+ * The store can be used to add, switch, and close tabs, as well as manage panel dimensions and locked positions.
+ * The store can be accessed also by Rancher extensions to integrate to handle the window manager.
+ */
 export interface State {
   tabs: Array<Tab>;
   active: Record<Position | string, string>;
@@ -50,6 +57,26 @@ export const getters = {
 };
 
 export const mutations = {
+  /**
+   * Adds a new tab to the window manager.
+   *
+   * Usage:
+   *
+   * store.dispatch('wm/open', {
+   *   id:            PRODUCT_NAME,
+   *   extensionId:   PRODUCT_NAME,
+   *   label:         'Label',
+   *   component:     'LabelComponent',
+   *   position:      'bottom',
+   *   layouts:       [
+   *     Layout.default,
+   *     Layout.home
+   *   ],
+   *   showHeader: false,
+   * }, { root: true });
+   *
+   * This will add a new tab with the specified properties to the window manager and set it as active.
+   */
   addTab(state: State, tab: Tab) {
     const existing = state.tabs.find((x) => x.id === tab.id);
 
@@ -80,6 +107,14 @@ export const mutations = {
     window.localStorage.setItem(STORAGE_KEY['pin'], tab.position);
   },
 
+  /**
+   * Switches a tab to a different position within the window manager.
+   *
+   * Usage:
+   * store.commit('wm/switchTab', { tabId: 'tab1', targetPosition: LEFT });
+   *
+   * This will move the tab with ID 'tab1' to the LEFT position and update the active tab accordingly.
+   */
   switchTab(state: State, { tabId, targetPosition }: { tabId: string, targetPosition: Position }) {
     const current = { ...(state.tabs.find((x) => x.id === tabId) || {}) };
 
@@ -165,11 +200,27 @@ export const mutations = {
     }
   },
 
+  /**
+   * Sets the user's preferred pin position for tabs in the window manager.
+   *
+   * Usage:
+   * store.commit('wm/setUserPin', LEFT);
+   *
+   * This will set the user's preferred pin position to LEFT and store it in local storage.
+   */
   setUserPin(state: State, pin: string) {
     state.userPin = pin;
     window.localStorage.setItem(STORAGE_KEY['pin'], pin);
   },
 
+  /**
+   * Sets the locked positions for tabs in the window manager.
+   *
+   * Usage:
+   * store.commit('wm/setLockedPositions', [LEFT, RIGHT]);
+   *
+   * This will lock tabs to the specified positions, preventing them from being moved elsewhere.
+   */
   setLockedPositions(state: State, positions: Position[]) {
     state.lockedPositions = positions;
   }
--- a/shell/components/nav/WindowManager/PinArea.vue
+++ b/shell/components/nav/WindowManager/PinArea.vue
@@ -4,6 +4,15 @@ import { BOTTOM, CENTER, LEFT, RIGHT } from '@shell/utils/position';
 import useDragHandler from './composables/useDragHandler';
 import { Z_INDEX } from './constants';
 
+/**
+ * This component is responsible for rendering the pin area used during tab dragging.
+ *
+ * Behavior:
+ * - Enable drag areas for each position (left, right, bottom, center) when dragging is active.
+ * - Highlights the area where the tab can be pinned.
+ * - Uses z-index variables to ensure proper layering of drag areas.
+ */
+
 const {
   dragOverPositionsActive, pin, pinArea, lockedPositions, onDragPositionOver
 } = useDragHandler();
--- a/shell/components/nav/WindowManager/composables/usePanelHandler.ts
+++ b/shell/components/nav/WindowManager/composables/usePanelHandler.ts
@@ -6,6 +6,9 @@ import useDimensionsHandler from '../composables/useDimensionsHandler';
 import useDragHandler from '../composables/useDragHandler';
 import useTabsHandler from '../composables/useTabsHandler';
 
+/**
+ * This composable is responsible for managing the state and behavior of the window manager panel.
+ */
 export default (props: { position: Position }) => {
   const store = useStore();
 
--- a/shell/components/nav/WindowManager/composables/useResizeHandler.ts
+++ b/shell/components/nav/WindowManager/composables/useResizeHandler.ts
@@ -3,6 +3,9 @@ import { boundingRect, LEFT, RIGHT, screenRect } from '@shell/utils/position';
 import { Position } from '@shell/types/window-manager';
 import { useStore } from 'vuex';
 
+/**
+ * This composable is responsible for handling the resize behavior of the window manager panels.
+ */
 export default (props: { position: Position, setDimensions: (args: { width?: number, height?: number }) => void }) => {
   const store = useStore();
 
--- a/shell/components/nav/WindowManager/composables/useTabsHandler.ts
+++ b/shell/components/nav/WindowManager/composables/useTabsHandler.ts
@@ -4,6 +4,9 @@ import { Tab } from '@shell/types/window-manager';
 
 const tabs = ref<{ tab: Tab, containerId: string }[]>([]);
 
+/**
+ * This composable is responsible for handling the tabs within the window manager.
+ */
 export default () => {
   const store = useStore();
 
--- a/shell/components/nav/WindowManager/index.vue
+++ b/shell/components/nav/WindowManager/index.vue
@@ -8,6 +8,21 @@ import useTabsHandler from './composables/useTabsHandler';
 import usePanelsHandler from '@shell/components/nav/WindowManager/composables/usePanelsHandler';
 import { Layout, Position } from '@shell/types/window-manager';
 
+/**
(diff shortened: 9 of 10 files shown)


shell/components/nav/SecondarySideWindow.vue
```diff
@@ -0,0 +1,291 @@
+<script lang="ts" setup>
```
torchiaf: Core changes are here. It adds the new side window

shell/config/store.js
```diff
@@ -66,7 +67,8 @@ let store = {};
       '../store/slideInPanel.ts',
       '../store/type-map.js',
       '../store/uiplugins.ts',
-      '../store/wm.js',
+      '../store/wm/index.js',
```
torchiaf: Core changes are here. It adds the new `wm/secondary` store.

shell/utils/dynamic-importer.js
```diff
@@ -130,7 +130,7 @@ export function resolveDetail(key) {
 }
 
 export function resolveWindowComponent(key) {
-  return require.resolve(`@shell/components/nav/WindowManager/${ key }`);
+  return require.resolve(`@shell/components/Console/${ key }`);
```
torchiaf: Decoupling Kubectl components from Primary window

shell/components/DraggableZone.vue
```diff
@@ -51,7 +51,13 @@ export default defineComponent({
     },
 
     onDragOver(event: DragEvent, zone: Zone) {
-      this.drag.zone = zone;
+      const isSecondaryOpen = this.$store.getters['wm/secondary/isOpen'];
```
rak-phillip: nit: Getters are reactive by nature and can be accessed via via `...mapGetters('wm/secondary', ['isOpen'])`.

shell/components/nav/SecondarySideWindow.vue
```diff
@@ -0,0 +1,291 @@
...
+          <span class="tab-label"> {{ tab.label }}</span>
+        </div>
+      </div>
+      <div
+        class="actions"
+      >
+        <RcButton
+          small
+          ghost
+          class=""
```
rak-phillip: question: Why do have an empty class?

shell/components/nav/SecondarySideWindow.vue
```diff
@@ -0,0 +1,291 @@
...
+
+const componentName = ref('');
+const component = ref<any>(null);
+
+const userPin = ref<'left' | 'right'>(RIGHT); // Support only right now
+const dragOffset = ref(0);
+
+const open = computed(() => store.getters['wm/secondary/isOpen']);
+const tabs = computed(() => store.getters['wm/secondary/tabs'] || []);
+const userWidth = computed(() => store.getters['wm/secondary/userWidth'] || window.localStorage.getItem('wm2-width') || '600px');
```
rak-phillip: Does it make more sense to read from local storage & set defaults in the store instead of in this component? Shifting the responsibility to the store makes `wm/secondary` the single source of truth for getting the width.

shell/components/nav/SecondarySideWindow.vue
```diff
@@ -0,0 +1,291 @@
...
+
+    loadComponent(tabs.value[0] || {});
+  }
+}
+
+function close() {
+  store.dispatch('wm/secondary/close');
+}
+
+function dragXStart(event: MouseEvent | TouchEvent) {
```
rak-phillip: This drag logic looks largely similar to what's in `shell/components/nav/PrimarySideWindow.vue`. Can this be made into a composable or utility function?

shell/components/nav/SecondarySideWindow.vue
```diff
@@ -0,0 +1,291 @@
...
+      </div>
+      <div
+        class="actions"
+      >
+        <RcButton
+          small
+          ghost
+          class=""
+          @click="close"
+          @keydown.enter.stop="close"
```
rak-phillip: Both enter and space should trigger a button.

shell/components/nav/SecondarySideWindow.vue
```diff
@@ -0,0 +1,291 @@
...
+
+    <div
+      ref="body"
+      class="body"
+    >
+      <component
+        :is="component"
+        v-if="component"
+        :tab="{}"
+        v-bind="{}"
```
rak-phillip: question: What is the purpose of `:tab="{}"` & `v-bind="{}"?

shell/components/nav/SecondarySideWindow.vue
```diff
@@ -0,0 +1,291 @@
...
+    class="wm2 secondary-side-window"
+  >
+    <div
+      ref="header"
+      class="header"
+    >
+      <div class="group">
+        <div
+          class="resizer resizer-x"
+          role="button"
```
rak-phillip: question: Is there any reason why this div can't be an button?

shell/components/nav/SecondarySideWindow.vue
```diff
@@ -0,0 +1,291 @@
...
+          tabindex="0"
+          :aria-label="'Resize Secondary Window'"
```
rak-phillip: I think this string constant needs to be added to `en-us.yaml`.

shell/components/nav/SecondarySideWindow.vue
torchiaf: This is a typo, fixed

shell/components/nav/SecondarySideWindow.vue
torchiaf: I moved the storage & default values here because we want to have a single `wm` store to handle both side windows in next iterations, so it makes sense to keep `wm/secondary` as much generic as possible.

shell/components/nav/SecondarySideWindow.vue
torchiaf: Absolutely, the goal in the next iteration is to use composable API or a single SideWindow component for both windows.
For now I would leave the code as it is for our demo, and fix this tech debt later though, I created a new ticket to tackle it https://github.com/rancher/dashboard/issues/15527

shell/components/nav/SecondarySideWindow.vue
torchiaf: Nothing, those are typos, fixed

shell/components/nav/SecondarySideWindow.vue
torchiaf: No, this is a copy-paste from PrimarySideWindow, as mentioned in the other comment, we need to merge the 2 components in 1 and make a bit of code-refactoring, so we can fix this later

shell/components/nav/Header.vue
```diff
@@ -393,7 +395,25 @@ export default {
...
+    },
+
+    async updateExtensionActionsEnabled() {
+      for (const [i, action] of this.extensionHeaderActions.entries()) {
+        this.extensionActionsEnabled[i] = await this.isActionEnabled(action);
+      }
+    },
+
+    async isActionEnabled(action) {
+      if (action.enabled === undefined) {
```
rak-phillip, line 407: question: What are the scenarios where `action.enabled === undefined`? Is it safe to default to true?

shell/components/nav/SecondarySideWindow.vue
```diff
@@ -0,0 +1,289 @@
...
+    break;
+  }
+
+  neu = Math.max(min, Math.min(neu, max));
+
+  throttle(() => {
+    store.commit('wm/secondary/setUserWidth', `${ neu }px`);
+    window.localStorage.setItem('wm2-width', `${ neu }px`);
+    document.documentElement.style.setProperty('--wm2-width', `${ neu }px`);
+  }, 250, { leading: true })();
```
rak-phillip: I don't think that this throttle is working - I don't see a difference between a throttled implementation and one that isn't. The root cause for this behavior is that a new throttle function is created on every `dragXMove` call, which defeats the purpose of the throttle in the first place. 

Ultimately, committing to the store and writing to local storage should be the expensive operations. We could probably refactor by:

1. moving the declaration of the `throttle`/`debounce` outside of `dragXMove`
2. only throttling/debouncing `store.commit` & `localStorage.setItem`

```js
const storeSize = debounce((width: string) => {
  store.commit('wm/secondary/setUserWidth', width);
  window.localStorage.setItem('wm2-width', width);
}, 500);

function dragXMove(event: MouseEvent | TouchEvent) {
  const screen = screenRect();
  const eventX = (event instanceof MouseEvent) ? event.screenX : (event as TouchEvent).touches[0].screenX;

  const min = 250;
  const max = Math.round(2 * screen.width / 5);
  let neu;

  switch (userPin.value) {
  case RIGHT:
    neu = screen.width - eventX + dragOffset.value;
    break;
  case LEFT:
    neu = eventX + dragOffset.value;
    break;
  }

  neu = Math.max(min, Math.min(neu, max));

  document.documentElement.style.setProperty('--wm2-width', `${ neu }px`);
  storeSize(`${ neu }px`);
}
```

shell/components/nav/SecondarySideWindow.vue
```diff
@@ -0,0 +1,289 @@
...
+
+    return;
+  }
+
+  // This should work at tab level, not global
+  // if (!!component.value || componentName.value === name) {
+  //   _warn(`component is already loaded`, name, extensionId);
+
+  //   return;
+  // }
```
rak-phillip: nit: remove commented code

rak-phillip (commented):
I left another round of comments. One thing to note with regards to accessibility - we will need to ensure that focus is set to an element within the secondary panel when it is opened.

nwmac:
I think we could do this with less change.

My thoughts on what we should do:

- Have the notion of Horizontal and Vertical panels, managed by a window manager.
- Horizontal panel is pinned to the bottom of the browser window, Vertical is pinned to the right-hand side
- The API for opening a window with the window manager takes an optional arg of which panel to open in - by default, it will open in the bottom, horizontal panel.
- (future) Windows can be moved between the horizontal and vertical panels


Not sure about renaming 'wm' to 'console' - I think the term 'console' is perhaps to specific.

shell/utils/dynamic-importer.js
nwmac: I'd suggest using `Window` rather than `Console` - this is the auto-import of Vue components that can be shown in the window manager, I don't think its specifically console.

shell/models/management.cattle.io.cluster.js
```diff
@@ -353,7 +353,7 @@ export default class MgmtCluster extends SteveModel {
   openShell() {
     this.$dispatch('wm/open', {
       id:        `kubectl-${ this.id }`,
-      label:     this.$rootGetters['i18n/t']('wm.kubectlShell.title', { name: this.nameDisplay }),
+      label:     this.$rootGetters['i18n/t']('console.kubectlShell.title', { name: this.nameDisplay }),
```
nwmac: I think keeping `wm` is fine - rather than `console`.

shell/components/templates/default.vue
```diff
@@ -256,8 +259,9 @@ export default {
         @dragstart="$refs.draggableZone.onDragStart($event)"
         @dragend="$refs.draggableZone.onDragEnd($event)"
       >
-        <WindowManager @draggable="draggable=$event" />
+        <PrimarySideWindow @draggable="draggable=$event" />
```
nwmac: I'd suggest naming `HorizontalWindowManager`

shell/components/templates/default.vue
```diff
@@ -256,8 +259,9 @@ export default {
...
       </div>
+      <SecondarySideWindow />
```
nwmac: I'd suggest naming `VerticalWindowManager`

nwmac:
Also, I'd be okay with disabling the window dragging for now (if that helps keep this simpler) and bringing it back when we re-factor to a cleaner single window manager.

shell/utils/dynamic-importer.js
torchiaf: Done

shell/components/DraggableZone.vue
torchiaf: Done

shell/components/nav/SecondarySideWindow.vue
torchiaf: Fixed

shell/components/nav/Header.vue
torchiaf: When `action.enabled` is not specified in the `addAction` call, it will be undefined.

```
  plugin.addAction(
    ActionLocation.HEADER,
    {},
    {
      tooltipKey: 'ai.action.openChat',
      shortcut: 'i',
      svg: require('./assets/chat-icon.svg'),
      // enabled: not specificed
```

The current implementation is 
```
const enabled = action.enabled ? action.enabled.apply(this, [this.ctx]) : true;
```
which enables the plugins action by default if the option is not specified, so here I wanted to keep the same logic.

shell/components/nav/SecondarySideWindow.vue
torchiaf: You right, `throttle` usage is wrong, I'll push your change, thanks

torchiaf:
@nwmac @rak-phillip I decided to re-implement the WindowManager functionality to fully support multiple windows and tabs dragging.
I also updated the description with new enhancements notes, please take a look.

pkg/rancher-components/src/components/RcDropdown/RcDropdownItemSelect.vue
```diff
@@ -55,7 +59,7 @@ const focusMenuItem = () => {
     <LabeledSelect
       ref="menuItemSelect"
       :value="modelValue"
-      :label="t('wm.containerLogs.range.label')"
+      :label="label"
```
rak-phillip, line 62: Thanks for fixing this! It should have been using a label prop from the outset 🙃

shell/components/nav/WindowManager/composables/useComponentsMount.ts
rak-phillip, line 1: Can we add a few quick comments to document the purpose and usage of this composable?

shell/components/nav/WindowManager/composables/useDimensionsHandler.ts
```diff
@@ -0,0 +1,102 @@
...
+
+      const windowHeight = window.innerHeight;
+      let height = panelHeight ? parseInt(panelHeight, 10) : 0;
+
+      if ( !height ) {
+        height = Math.round(windowHeight / 2);
+      }
+      height = Math.min(height, 3 * windowHeight / 4);
+
+      setDimensions({ height });
```
rak-phillip, line 30: Getters should be side-effect free. Both getters for `width` and `height` will commit a mutation. Perhaps, it would be better to initialize dimensions using the mounted hook or some other mechanism.

shell/components/nav/WindowManager/composables/useDragHandler.ts
```diff
@@ -0,0 +1,89 @@
+import { useStore } from 'vuex';
+import { Position, Tab } from '@shell/types/window-manager';
+import { computed, ref } from 'vue';
+import { CENTER } from '@shell/utils/position';
+
+const dragOverPositionsActive = ref(false);
+const pinArea = ref(CENTER as Position);
```
rak-phillip, line 17: Can we document why we want to specify these refs outside of main composable function? I'm assuming that we want to define these once and share across all instances of the composable, but I don't know why we want to do this.

shell/store/wm.ts
```diff
@@ -0,0 +1,193 @@
...
+
+  setPanelHeight(state: State, { position, height }: { position: string, height: number | null }) {
+    state.panelHeight[position] = height;
+    window.localStorage.setItem(STORAGE_KEY[BOTTOM], `${ height }`);
+
+    for (const tab of state.tabs) {
+      if (tab.position === position) {
+        tab.containerHeight = height;
+      }
+    }
```
rak-phillip, line 189: nit: It seems a little awkward to iterate over each tab and set the height when this could be driven by the parent container with css.. I assume that there's some legacy implementation detail that requires this?

shell/components/nav/WindowManager/composables/useComponentsMount.ts
torchiaf: Sure, I'm adding some descriptions and usage examples in the most important components and composables.

shell/components/nav/WindowManager/composables/useDimensionsHandler.ts
torchiaf: I agree, this is mostly legacy code and it needs to be cleaned up.
However, the logic there is a bit complex, so I would leave this kind of refactoring for later to avoid the risk of breaking the set-dimensions behavior, if you are fine with that.

shell/components/nav/WindowManager/composables/useDragHandler.ts
torchiaf: Done

shell/store/wm.ts
torchiaf: Yes, it's legacy code. Some of the components in `shell/components/Window`, which implement the Kubectl console, require to get the parent dimensions when loaded in the Window Manager's panels

https://github.com/torchiaf/dashboard/blob/7d17f8946d0b95b3b1c0323be30cc57a0a8d99ab/shell/components/nav/WindowManager/index.vue#L52-L53

shell/components/nav/WindowManager/composables/useDimensionsHandler.ts
rak-phillip: Yes, let's create a new issue so we can follow up and address in a separate PR. This will also help to manage the scope of this change as well.

PR #15465, opened 2025-09-17
Automation: Add support for multiple K8s versions in Jenkins pipeline

yonasberhe23:
### Summary
Fixes #https://github.com/rancher/qa-tasks/issues/1957

This PR adds support for testing multiple Kubernetes versions in the Jenkins pipeline, allowing testing across different Rancher/K8s version combinations.
The "Cluster Dashboard | can search by resource group" test was failing because it was running against K8s v1.26.13+k3s2 (which has `TokenReviews`) but expecting `SelfSubjectReviews` which is only available in newer K8s versions.

Instead of making the test more complex, this PR enables testing against the correct K8s versions for each Rancher image tag:

- Rancher v2.10 → K8s v1.31.11+k3s1
- Rancher v2.11 → K8s v1.32.7+k3s1
- Rancher v2.12 → K8s v1.33.4+k3s1 
- Rancher head → K8s v1.33.3+k3s1

PLEASE SEE JENKINS STAGING ENV FOR RESULTS:
- ui-automation-multi-recurring-job - run 42
- ui-automation-job-yonas - run 49-56

Screenshot 2025-09-18 at 12 21 35 PM

Dashboard:

- Enhanced logging to show which K8s/Image tag combinations are being tested
- Dynamically fetching K3s versions from KDM repo based on Rancher version
- Implemented paired testing approach where each Rancher version gets its compatible K3s version
- Removed dependency on K3S_KUBERNETES_VERSION

JJB:
- This PR relies on [**JJB PR**](https://github.com/rancherlabs/jenkins-job-builder/pull/860)

### Technical notes summary
**Benefits:**
- Tests run against the actual supported K8s versions for each Rancher release
- Eliminates version-related test failures

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/cypress/jenkins/Jenkinsfile_multi
+++ b/cypress/jenkins/Jenkinsfile_multi
@@ -1,8 +1,67 @@
 #!groovy
 
 def branch = "master"
-def test_tags = "${env.TEST_TAGS}".split(',')
+def test_tags_raw = "${env.TEST_TAGS}".split(',')
+def test_tags = []
+for (String tag : test_tags_raw) {
+    test_tags.add(tag.trim())
+}
 def all_cypress_tags = "${env.CYPRESS_TAGS}".split('\\|')
+
+// Function to get KDM branch for a Rancher version
+def getKdmBranch(rancherVersion, testTags) {
+    if (rancherVersion == "head") {
+        // Find the highest version from test tags and increment it
+        def latestVersion = ""
+        for (String tag : testTags) {
+            def versionMatch = tag =~ /v(\d+\.\d+)-head/
+            if (versionMatch) {
+                def currentVersion = versionMatch[0][1]
+                if (currentVersion > latestVersion) {
+                    latestVersion = currentVersion
+                }
+            }
+        }
+        // Increment minor version
+        if (latestVersion.isEmpty()) {
+            error("Cannot determine KDM branch for 'head' image. Please provide at least one version tag (e.g., v2.12-head) in TEST_TAGS.")
+        }
+        def parts = latestVersion.split('\\.')
+        def minor = (parts[1] as Integer) + 1
+        return "dev-v${parts[0]}.${minor}"
+    }
+    
+    // Extract version from image tags (e.g."v2.12-head")
+    def versionMatch = rancherVersion =~ /v(\d+\.\d+)-head/
+    if (versionMatch) {
+        def majorMinor = versionMatch[0][1]
+        return "dev-v${majorMinor}"
+    }
+    
+    error("Unable to determine KDM branch for Rancher version: ${rancherVersion}")
+}
+
+// Function to fetch latest K3s version from KDM
+def getLatestK3sVersion(rancherVersion, testTags) {
+    def kdmBranch = getKdmBranch(rancherVersion, testTags)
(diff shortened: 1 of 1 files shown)


izaac:
This will fail immediately after the KDM is updated.

Edit; We can check this offline, we may be able to make it more robust by querying the current kdm specification (a [json](https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.12/data/data.json) file)

yonasberhe23:
> This will fail immediately after the KDM is updated.
> 
> Edit; We can check this offline, we may be able to make it more robust by querying the current kdm specification (a [json](https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.12/data/data.json) file)

The K8s versions I'm using are stable versions (v1.33.4+k3s1, v1.33.3+k3s1, etc.) that won't suddenly become unsupported when KDM is updated. Correct me if I'm wrong, but KDM updates typically add new versions rather than remove existing stable ones.
These hardcoded version combinations that actually work in practice are more reliable than dynamically querying whatever KDM currently lists as "supported," but may not have been fully tested together.
If we need to update the versions later, it's a simple change to the JJB config.

izaac:
@yonasberhe23 That's correct. Release images are static, but "head" (development) images are built from kdm development branches that are subject to change before a final release. The last time I looked into that.

And we have to consider the different Rancher versions: 2.10, 2.11, 2.12, 2.13 that might or might not work with those officially.

yonasberhe23:
> @yonasberhe23 That's correct. Release images are static, but "head" (development) images are built from kdm development branches that are subject to change before a final release. The last time I looked into that.
> 
> And we have to consider the different Rancher versions: 2.10, 2.11, 2.12, 2.13 that might or might not work with those officially.

@izaac Ah, I see. That makes sense. I'll implement the dynamic approach to fetch K8s versions from the KDM JSON files. Thanks for catching that

izaac:
It's interesting the `2.12` branch has a newer k3s version than the `head`  (from the jenkins run logs). Is that accurate?

yonasberhe23:
> It's interesting the `2.12` branch has a newer k3s version than the `head` (from the jenkins run logs). Is that accurate?

Yes, if you check KDM (dev-v2.12 and dev-v2.13), you will see that 2.12 has a newer k3s version than 2.13

cypress/jenkins/Jenkinsfile_multi
```diff
@@ -32,11 +91,22 @@ node {
...
+            }
+            
             try {
                 stage('Run Tests') {
                     jobs = [:]
-                    for( String val : test_tags ) {
+                    for (int i = 0; i < test_tags.size(); i++) {
+                        String rancher_version = test_tags[i]
+                        String k8s_version = k8s_versions[i]
                         for ( String ct : all_cypress_tags ) {
```
izaac: I think we can make this less error prone by using Groovy's `eachForIndex`

```groovy
test_tags.eachWithIndex { String rancher_version, int i ->
    String k8s_version = k8s_versions[i]
    all_cypress_tags.each { String ct ->
        echo "RANCHER_TAG: ${rancher_version}, K8S_VERSION: ${k8s_version}, CYPRESS_TAGS: ${ct}"
        
          string(name: 'K3S_KUBERNETES_VERSION', value: k8s_version),
```

PR #15501, opened 2025-09-22
Update chart's downgrade action icon

### LINKED ISSUE #15207: [Apps & Marketplace] - Charts actions on detail/wizard/tools pages
**Chart detail page:**

- version on the header should be equivalent to the one on the right panel
- top right button should have the updated text

Image

Image

Image

Image

**Wizard page:**

We change the “mode” of the wizard (the text used before the steps, on submit buttons, etc)  to “Upgrade”, “Edit” or “Downgrade” in real time when the user changes the version in the version dropdown:

- if the selected version = a newer version than the current version, the “mode” = “Upgrade”
- if the selected version = the current version, the “mode” = “Edit”
- if the selected version = an older version than the current version, the “mode” = “Downgrade”

Image

Image

Image

Image

**Cluster tools page:**

- If it’s installed and there there is at least 1 version newer than the current version (= if it’s upgradeable). We auto-select the newest version in the version dropdown control of the wizard.
- If it’s installed. We auto-select the current version in the version dropdown control of the wizard.
- If it’s installed and there there is at least 1 version older than the current version. We auto-select the version prior to the current version in the version dropdown control of the wizard.

Image

momesgin:
### Summary
Fixes #15207 

### Technical notes summary

- Updated our icons package
- Updated icon name

### Screenshot/Video

image

### Checklist
- [x] The PR has a Milestone 


momesgin:
> LGTM. Doesn't the icon feel off centre? I understand why, but I'd suggest to bring up with Ovi if we want to keep it this way or if we should adjust the margins Screenshot 2025-09-23 at 12 07 44 PM

yeah it looks a tiny bit off to me too, we need to add more variants to our current button component, we only have small and standard variants at the moment, the alignment issue can be addressed as part of that task, I'll let Ovi know about it, thanks for catching it.

PR #15509, opened 2025-09-24
redoing PR without changes

aalves08:
### Summary
Fixes #

### Checklist
- [ ] The PR has a Milestone 


cypress/e2e/tests/pages/global-settings/settings.spec.ts
```diff
@@ -34,6 +34,100 @@ describe('Settings', { testIsolation: 'off' }, () => {
...
+    expect(newSettingsPage.inactivityModalCard().getCardBody().should('exist'));
+    expect(newSettingsPage.inactivityModalCard().getCardActions().should('exist'));
+
+    expect(newSettingsPage.inactivityModalCard().getCardTitle().should('contain', 'Session expiring'));
+    expect(newSettingsPage.inactivityModalCard().getCardBody().should('contain', 'Your session is about to expire due to inactivity. Any unsaved changes will be lost'));
+    expect(newSettingsPage.inactivityModalCard().getCardBody().should('contain', 'Click “Resume Session” to keep the session in this tab active or refresh the browser after the session has expired'));
+
+    // Clicking the resume button should close the modal and reset the timer
+    expect(newSettingsPage.inactivityModalCard().getCardActions().contains('Resume Session').click());
+    expect(newSettingsPage.inactivityModalCard().shouldNotExist());
```
richard-cox, line 125: as much as i hate waits, we should add one here to wait passed the original idle session timeout and then try to nav somewhere new (to kick off a new http request). if everything has worked (we've delayed the user being timed out idle) then it should succeed

cypress/e2e/tests/pages/global-settings/settings.spec.ts
```diff
@@ -34,6 +34,100 @@ describe('Settings', { testIsolation: 'off' }, () => {
...
+
+    // Reset
+    SettingsPagePo.navTo();
+    newSettingsPage.waitForPage();
+    newSettingsPage.editSettingsByLabel(sessionIdleSetting);
+
+    settingsEdit.waitForPage();
+    settingsEdit.title().contains(`Setting: ${ sessionIdleSetting }`).should('be.visible');
+    settingsEdit.useDefaultButton().click();
+    settingsEdit.saveAndWait(sessionIdleSetting).then(({ request, response }) => {
```
richard-cox, line 147: just to confirm, what happens now when the UI changes the idle setting? do they 

- still do nothing
- start all existing idle timeouts again
- apply the new time to existing idle timeouts, resetting them if applicable

shell/components/Inactivity.vue
```diff
@@ -3,76 +3,211 @@ import ModalWithCard from '@shell/components/ModalWithCard';
...
+        console.error('watcherData RAN neu!!!', neu);
+        console.error('watcherData RAN old!!!', old);
+        const currDate = Date.now();
+        const endDate = new Date(neu.userActivityExpiresAt || '0001-01-01 00:00:00 +0000 UTC').getTime();
 
-    // Total amount of time before the user's session is lost
-    const thresholdToSeconds = settings?.inactivity?.threshold * 60;
+        //   // this means that all conditions are met to start/stop the timers
+        if (endDate > currDate && neu?.sessionTokenName) {
+          console.error('watcherData passed gate 1!!!');
```
richard-cox: this gets triggered even when the feature is disabled. 

looking at network requests we're also fetching `/v1/ext.cattle.io.useractivities/token-tmhkz`

so there needs to be something high up the call stack to stop anything happening if idle and session timeouts match

shell/components/Inactivity.vue
```diff
@@ -3,76 +3,211 @@ import ModalWithCard from '@shell/components/ModalWithCard';
...
+    async initializeInactivityData() {
+      console.error('initializeInactivityData!!!');
+      const canListSettings = this.$store.getters[`management/canList`](MANAGEMENT.SETTING);
+      const canListUserAct = this.$store.getters[`management/canList`](EXT.USER_ACTIVITY);
+      const canListTokens = this.$store.getters[`rancher/canList`](NORMAN.TOKEN);
+
+      if (canListUserAct && canListTokens && canListSettings) {
+        // fetching all resources needed to start inactivity logic
+        const hash = {
+          userSessionTtlIdleSetting: this.$store.dispatch('management/find', { type: MANAGEMENT.SETTING, id: SETTING.AUTH_USER_SESSION_TTL_MINUTES }),
```
richard-cox:
```suggestion
          userSessionTtlIdleSetting: this.$store.dispatch('management/find', { type: MANAGEMENT.SETTING, id: SETTING.AUTH_USER_SESSION_IDLE_TTL_MINUTES }),
```

userSessionTtlIdleSetting and userSessionTtlSetting matching should in theory have disabled the feature. hopefully previous comment should fix it though

richard-cox: we should have all settings by this point, so these two can be getters, 

which means the only async bit is the tokens

shell/utils/inactivity.ts
```diff
@@ -0,0 +1,106 @@
...
+    const courtesyTimer = Math.min(courtesyTimerVal, 60 * 5); // Never show the modal more than 5 minutes
+
+    const courtesyCountdown = courtesyTimer;
+
+    // Amount of time before the user sees the inactivity warning
+    // Note - time before warning is shown + time warning is shown = settings threshold (total amount of time)
+    const showModalAfter = thresholdSeconds - courtesyTimer;
+
+    return {
+      enabled:          true,
```
richard-cox: is this the part that needs wiring in (if both timeouts are equal?), though it would need to be called earlier in watcherData if so

shell/utils/inactivity.ts
```diff
@@ -0,0 +1,106 @@
...
+  public storeSessionTokenName(tokenName: string): void {
+    this.sessionTokenName = tokenName;
+  }
+
+  public async checkUserActivityData(store: any, sessionTokenName: string): Promise<UserActivityResponse> {
+    try {
+      const updatedData = await store.dispatch('management/find', {
+        type: EXT.USER_ACTIVITY,
+        id:   sessionTokenName,
+        opt:  { force: true }
```
richard-cox: optional - could also put `watch: false` here. under the hood it's not much extra cpu, but makes it clear to people reading this that the resource isn't updating

shell/utils/inactivity.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  public getSessionTokenName(): string | undefined {
+    return this.sessionTokenName;
+  }
+
+  public storeSessionTokenName(tokenName: string): void {
+    this.sessionTokenName = tokenName;
+  }
+
+  public async checkUserActivityData(store: any, sessionTokenName: string): Promise<UserActivityResponse> {
```
richard-cox:
```suggestion
  public async getUserActivity(store: any, sessionTokenName: string): Promise<UserActivityResponse> {
```

shell/utils/inactivity.ts
```diff
@@ -0,0 +1,106 @@
...
+      });
+
+      return updatedData;
+    } catch (e: any) {
+      console.error(`Could not GET UserActivity for session token ${ sessionTokenName }`, e); // eslint-disable-line no-console
+      throw new Error(e);
+    }
+  }
+
+  public async updateUserActivityToken(store: any, sessionTokenName: string, seenAt: string): Promise<UserActivityResponse> {
```
richard-cox:
```suggestion
  public async updateUserActivity(store: any, sessionTokenName: string, seenAt: string): Promise<UserActivityResponse> {
```

shell/utils/inactivity.ts
```diff
@@ -0,0 +1,106 @@
...
+}
+
+export class Inactivity {
+  private sessionTokenName: string | undefined = undefined;
+
+  public getSessionTokenName(): string | undefined {
+    return this.sessionTokenName;
+  }
+
+  public storeSessionTokenName(tokenName: string): void {
```
richard-cox:
```suggestion
  public setSessionTokenName(tokenName: string): void {
```

shell/components/Inactivity.vue
richard-cox: optional: the token stuff could move into a nice helper `storeSessionTokenName` in inactivity

shell/components/Inactivity.vue
```diff
@@ -3,76 +3,211 @@ import ModalWithCard from '@shell/components/ModalWithCard';
...
+          userSessionTtlSetting:     this.$store.dispatch('management/find', { type: MANAGEMENT.SETTING, id: SETTING.AUTH_USER_SESSION_TTL_MINUTES }),
+          tokens:                    this.$store.dispatch('rancher/findAll', { type: NORMAN.TOKEN }),
```
richard-cox: this should have `opt:  { watch: false }` to improve performance for admins (who might have permissions to see a LOT)

shell/components/Inactivity.vue
```diff
@@ -3,76 +3,211 @@ import ModalWithCard from '@shell/components/ModalWithCard';
...
+        };
+
+        const res = await allHash(hash);
+
+        this.tokens = res.tokens;
+
+        // handle the fetching/storage of session token name
+        if (!this.sessionTokenName) {
+          const sessionToken = this.tokens.find((token) => {
```
richard-cox, line 133: we loop once through tokens here (called on mount), then also loop through tokens in a computed property sessionToken.

once we find the token here we should just be able to set it directly to a this.sessionToken value

shell/components/Inactivity.vue
```diff
@@ -3,76 +3,211 @@ import ModalWithCard from '@shell/components/ModalWithCard';
...
+
+          if (sessionToken?.name) {
+            console.error('SETTING TOKEN NAME', sessionToken.name);
+            this.sessionTokenName = sessionToken.name;
+            Inactivity.storeSessionTokenName(sessionToken.name);
+          }
+        }
+
+        // get the latest userActivity data so that get reactivity on all this logic
+        const userActivityData = await this.$store.dispatch('management/find', { type: EXT.USER_ACTIVITY, id: this.sessionTokenName });
```
richard-cox: to keep things neat this could use `checkUserActivityData` (just need to setup a prop for `force`). if you've done the update that function has become getUserActivity

shell/components/Inactivity.vue
```diff
@@ -85,91 +220,137 @@ export default {
...
-          this.unsubscribe();
           this.clearAllTimeouts();
+
+          console.error('LOGGING OUT USER FROM THE UI!!!!!!');
+
+          return this.$store.dispatch('auth/logout', { sessionIdle: true });
         } else {
+          // When we have X seconds to go we log out the user, we check again for activity on the backend flag
+          // it may have come from another tab in the same browser
+          if (now >= endTime - (BEFORE_LOGOUT_CHECK_DELAY_SECONDS * 1000) && !this.beforeLogoutCheckRan) {
```
richard-cox: i don't think we need to do this before the timer runs out, only just before the log out is about to happen (in `if` section above). this reduces the overlap where in another tab the user can, at the last moment, click resume session but still be logged out.

as it is at the moment we don't actually gain anything from doing it early.

shell/components/Inactivity.vue
```diff
@@ -85,91 +220,137 @@ export default {
...
+          }
+
           this.inactivityTimeoutId = setTimeout(checkInactivityTimer, 1000);
         }
       };
 
       checkInactivityTimer();
     },
+    async checkBackendInactivity() {
+      const userActivityData = await Inactivity.checkUserActivityData(this.$store, this.sessionTokenName);
```
richard-cox: when the user is logged out in another tab this throws an exception (because the session is gone, ui makes a request, no valid session found, backend returns error 401)

i've seen this cause the user to be logged out but shown the normal 'log in again to continue' message (after a few flashy like pages). 

in checkUserActivityData the `find` itself will check for a 401 and if found call `dispatch('auth/logout', opt.logoutOnError, { root: true });` before returned a rejected promise / throw error.

code in shell/plugins/steve/actions.js `request` --> `onError`.

what we could do is pass through `logoutOnError: false` to `find`, but we would then need to handle the exception.


 so we would need to catch `await Inactivity.checkUserActivityData` and 

- startCountdown - if err._status === 401 --> this.$store.dispatch('auth/logout', { sessionIdle: true })
- _trackInactivity - if err._status === 401 --> this.$store.dispatch('auth/logout', { sessionIdle: true })

so maybe given above we just need to update checkUserActivityData?


```
  public async checkUserActivityData(store: any, sessionTokenName: string): Promise<UserActivityResponse> {
    try {
      const updatedData = await store.dispatch('management/find', {
        type: EXT.USER_ACTIVITY,
        id:   sessionTokenName,
        opt:  { force: true, watch: false,  logoutOnError: false }
      });

      return updatedData;
    } catch (e: any) {
       if (e._status === 401) {
          return  this.$store.dispatch('auth/logout', { sessionIdle: true })
       }

      console.error(`Could not GET UserActivity for session token ${ sessionTokenName }`, e); // eslint-disable-line no-console
      throw new Error(e);
    }
  }

```

cypress/e2e/tests/pages/global-settings/settings.spec.ts
aalves08: The backend does update `UserActivity` `expiresAt` property when a user changes `auth-user-session-idle-ttl-minutes`. 
The backend does **not** reset (empty it) the `UserActivity` `expiresAt` when `auth-user-session-idle-ttl-minutes` === `auth-user-session-ttl-minutes`

shell/components/Inactivity.vue
aalves08: 🤦 fixed with next commit

shell/utils/inactivity.ts
aalves08: Fixed with next commit

shell/utils/inactivity.ts
aalves08: fixed w/ next commit

shell/utils/inactivity.ts
aalves08: fixed w/ next commit

shell/utils/inactivity.ts
aalves08: This is not needed. Feature is verified as enabled or not on the component watcher. fixed with next commit

shell/utils/inactivity.ts
aalves08: fixed with next commit

shell/components/Inactivity.vue
aalves08: This area of the code is already for the last countdown (when the modal is already being displayed). We need to check it **before** the backend really does throw the user out of the session (invalidate token) which might happen due to FE timers not being in absolute sync with the BE expiration date, hence why I do a check `BEFORE_LOGOUT_CHECK_DELAY_SECONDS`, which is now set to 5 seconds, before we force the FE logout.

Since this is a second buffer (when we parse the inactivity data, we give it a 3 seconds buffer already), we might be able to reduce this variable value but I would definitely advise keeping this in. Remember the bigger the `expiresAt` the larger will be the difference between BE and FE timers.

Maybe set `BEFORE_LOGOUT_CHECK_DELAY_SECONDS` to 2 seconds?

shell/components/Inactivity.vue
aalves08: fixed on next pr commit

shell/components/Inactivity.vue
aalves08: fixed on next pr commit

shell/components/Inactivity.vue
aalves08: Given the error I did ` userSessionTtlIdleSetting: this.$store.dispatch('management/find', { type: MANAGEMENT.SETTING, id: SETTING.AUTH_USER_SESSION_TTL_MINUTES })`, the getters is the only reason all of this worked. 😅 

clearing these dispatches (fixed on next pr commit)

shell/components/Inactivity.vue
aalves08: The mount part is used when we move between different layouts in the UI (we re-use what we had already fetched if it's available). That is taken into account in the `initializeInactivityData` with the `if (!this.sessionTokenName)` which get's skipped if it exists.

The part that doesn't make sense on the code is the computed... I've taken a look at all the code and the computed is not used anywhere so I've got ahead and removed it (probably bits of older refactors) 🙏 

gonna deem this as resolved. pr updated with next commit

shell/components/Inactivity.vue
aalves08: The watcher is reactive to a bunch of different variables:

```
userActivityExpiresAt
sessionTokenName
ttlIdleValue
ttlValue
```

If you look at these logs from the UI (page refresh with settings the same value):
Screenshot 2025-10-10 at 10 38 45

The watcher is reactive because of the update to `sessionTokenName` within `initializeInactivityData`, which causes that "gate 1" to log it's part but we are still missing the `userActivityData` from `initializeInactivityData` at this point... 

When that XHR request is done, the watcher is triggered again, but it doesn't allow for the timer to be initialised on the watcher because `neu.ttlIdleValue === neu.ttlValue`. 

If you look at the code, the only "get" to the `userActivity` is when the setting gets updated via a watcher:
```
async ttlIdleValue(neu) {
   await Inactivity.getUserActivity(this.$store, this.sessionTokenName);
},
```

we need this because we can't watch `userActivity` but we also need the whole logic to be reactive to a setting change and gather that new `expiresAt`.

I can't see anything wrong here 🤔

shell/components/Inactivity.vue
aalves08: We can do that 👍 . Fixed on the next pr commit

cypress/e2e/tests/pages/global-settings/settings.spec.ts
aalves08: added this new assertion to the test

richard-cox: From what i can tell we're on the settings page and the added navTo will just be a no-op (route does not change), importantly no new network requests are made.

Can the test instead navTo something like users and auth?

cypress/e2e/tests/pages/global-settings/settings.spec.ts
richard-cox: > The backend does not reset (empty it) the UserActivity expiresAt when auth-user-session-idle-ttl-minutes === auth-user-session-ttl-minutes

That feels like a bug, can you create something for the BE team?

shell/components/Inactivity.vue
richard-cox: If the session idle timeout equals or is greater than the session timeout we shouldn't need to do anything (apart from monitoring for the case where idle is less than session).

shell/components/Inactivity.vue
richard-cox: The point i'm making is that we can check that the backend has/has't timed out at the point we're just about to log them out. We don't need the buffer as it introduces a gap where the user can prevent the inactivity in Tab B but we still log them out in Tab A. 

If the clocks are out of sync, it's even better to check just before logging the user out for that new active session.

If we check before the ui timeout is zero and the user isn't logged out (active in another browser), we close the model early - no issue
If we check after the ui timeout is zero and the user isn't logged out (active in another browser), we close the model - no issue

If we check before the ui timeout is zero and the user isn't active in another tab, we wait till zero and log out - no issue
if we check after the ui timeout is zero and the user isn't active in another tab, we log out - no issue

If we check before the ui timeout is zero and the user isn't active in another tab.... but is then active in another tab... will still log them out - issue
If we check after the ui timeout is zero and the user isn't active in another tab.... but is then active in another tab... user is not logged out - no issue.

So i'm not sure there's any benefit of checking a few seconds before timeout is zero than just before we're about to log them out.

shell/utils/inactivity.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  public async updateUserActivity(store: any, sessionTokenName: string, seenAt: string): Promise<UserActivityResponse> {
+    const spec: SpecData = { tokenId: sessionTokenName };
+
+    if (seenAt) {
+      spec.seenAt = seenAt;
+    }
+
+    const updateUserActivity = await store.dispatch('management/create', {
```
richard-cox: This one got lost in the PR shuffle https://github.com/rancher/dashboard/pull/15137#discussion_r2318870225. Another issue required for BE.

The request to update the resource returns a http validation warning in the header of the response. 

This is normally corrected by I guess openapi spec in the crd, but this resource has not crd.

The rancher `schema` version `/v1/schema/management.cattle.io.userattribute` does not contain attributes.columns with spec or type in.

The rancher returned openapi spec `/v1/apiextensions.k8s.io.customresourcedefinition/useractivities.ext.cattle.io` does not container references to spec or type either

shell/assets/translations/en-us.yaml
```diff
@@ -5114,13 +5115,9 @@ podDisruptionBudget:
 
 inactivity:
   title: Session expiring
-  titleExpired: Session expired
   banner: Your session is about to expire due to inactivity. Any unsaved changes will be lost.
-  bannerExpired: Your session has expired in this tab due to inactivity.
   content: Click “Resume Session” to keep the session in this tab active or refresh the browser after the session has expired.
```
```suggestion
  content: Click “Resume Session” to keep the session in this browser active.
```

shell/components/Inactivity.vue
```diff
@@ -85,91 +200,137 @@ export default {
...
         if (now >= endTime) {
-          this.isInactive = true;
-          this.unsubscribe();
           this.clearAllTimeouts();
+
+          console.error('LOGGING OUT USER FROM THE UI!!!!!!');
+
+          return this.$store.dispatch('auth/logout', { sessionIdle: true });
         } else {
+          // When we have X seconds to go we log out the user, we check again for activity on the backend flag
```
richard-cox: Somewhere around here there's a bug

1. Tab A - open rancher
2. Tab A - set idle timeout to 5 minutes
3. Tab A - Refresh
4. Tab B - open rancher
5. wait until Tab A and Tab B show idle modal
6. Tab B - click resume
7. Tab A - does ping and get extended expiresAt
8. Tab A - does not remove the idle modal, just stays but bar on 100%

cypress/e2e/tests/pages/global-settings/settings.spec.ts
aalves08: Updated the test to, after modal is displayed, navigation is done to `Account & API keys`, which is a different template/layout. 

Will be updated once I push the changes to the PR.

FYI, unless I missed a crucial point of your comment, but the network requests have zero impact on user being active or not (backend doesn't update that, we don't send that update either... we only send the update on these cases:

```
document.addEventListener('mousemove', this.setUserAsActive);
document.addEventListener('mousedown', this.setUserAsActive);
document.addEventListener('keypress', this.setUserAsActive);
document.addEventListener('touchmove', this.setUserAsActive);
document.addEventListener('visibilitychange', this.setUserAsActive);
```

I think this resolves this thread 🙏

shell/components/Inactivity.vue
aalves08: Removed this logic and updated the code so that check is done right before we do the logout. Will be updated once I push the changes

shell/assets/translations/en-us.yaml
aalves08: updated once the code is pushed

shell/utils/inactivity.ts
aalves08: Issue got created on the backend board when request, as per https://github.com/rancher/rancher/issues/51933, but it's not a blocker and I don't think the BE team is going to fix this any time soon.

shell/components/Inactivity.vue
aalves08: Simplified watcher logic to be reactive to the following variables:

```
userActivityExpiresAt
sessionTokenName
isFeatureEnabled
```

where:

```
isFeatureEnabled = this.ttlIdleValue < this.ttlValue;
```

With this I cleaned up superfluous `if` conditions inside the watcher

shell/components/Inactivity.vue
aalves08: This is solved because on each watcher update we are now cleaning up any old timers and resetting variables to prepare for the new expiration date. That also closes any modals 

`this.stopInactivity()`

cypress/e2e/tests/pages/global-settings/settings.spec.ts
richard-cox: That's correct, however if our new process to extend the idle timeout failed then new http requests would fail with 401. That was the angle i was looking at.

I think we're good now

shell/components/Inactivity.vue
```diff
@@ -99,39 +99,37 @@ export default {
         const currDate = Date.now();
         const endDate = new Date(neu.userActivityExpiresAt || '0001-01-01 00:00:00 +0000 UTC').getTime();
 
-        //   // this means that all conditions are met to start/stop the timers
-        if (endDate > currDate && neu?.sessionTokenName) {
-          console.error('watcherData passed gate 1!!!');
-          console.error('neu.ttlIdleValue < neu.ttlValue', neu.ttlIdleValue < neu.ttlValue);
+        if (endDate > currDate && neu?.sessionTokenName && neu?.isFeatureEnabled) {
```
richard-cox, line 100: the latest change to use isFeatureEnabled to block functionality is good here, but it doesn't cover the `mounted` functionality which fetches all tokens + the useractivity object

had a play around and there's a nice fix. 

1. nuke `mounted`
2. add the following to the top of the watcherData watch handler

```
        if (!old?.isFeatureEnabled && neu?.isFeatureEnabled) {
          await this.initializeInactivityData();
        }
```

this kicks off the timer well if the user changes the settings from disabled --> enabled. if the user toggles enabled --> disabled --> enabled it kicks things off again ok as well

shell/components/Inactivity.vue
```diff
@@ -230,36 +228,33 @@ export default {
...
+      const checkCountdown = async() => {
         const now = Date.now();
 
         if (now >= endTime) {
           this.clearAllTimeouts();
+          const isUserActive = await this.checkBackendInactivity();
 
-          console.error('LOGGING OUT USER FROM THE UI!!!!!!');
+          if (!isUserActive) {
+            console.error('LOGGING OUT USER FROM THE UI!!!!!!');
```
richard-cox: this still happens even when the user presses `Resume`

image

richard-cox: Not sure it's connected. but startCoundown contains an async checkCountdown, but when called doesn't await it

shell/components/Inactivity.vue
aalves08: thank you for the suggested change. Will be updated once I push the code 🙏

shell/components/Inactivity.vue
aalves08: I found out where the problem was.... Always between the chair and the computer 😂  

Seriously, the problem wasn't the async nature of it but the fact that the computed `userActivityExpiresAt` was updated first than the check that `checkBackendInactivity` to compare if the `expiresAt` really got updated, making it `userActivityData?.status?.expiresAt === this.expiresAt`

shell/components/Inactivity.vue
```diff
@@ -262,6 +238,22 @@ export default {
 
       checkCountdown();
     },
+    async checkBackendInactivity(currExpiresAt) {
```
richard-cox, line 221: nit, if you're making functional changes (not just removing console logs) this can just be `async checkBackendInactivity()` again (it's called in two places with the same arg `this.checkBackendInactivity(this.expiresAt);`

richard-cox: no other functional changes, no need to tweak

richard-cox:
FWIW the angles i tested 

- single tab
  - log in - modal shows, times out, user logged out
  - refresh after logging in - modal shows, times out, user logged out
  - refresh and reduce the idle timeout - modal shows after new duration
  - refresh, wait a bit and reduce the idle timeout to a period that has already been exceeded - user is logged out
  - refresh and increase the idle timeout - model shows after new duration
  - refresh and set the idle timeout to the session timeout (disable feature) - model is never shown
- two tabs
  - refresh tab A and tab B - modal shows on both, times out, user logged out on both
  - refresh tab A and tab B - wait some time, interact with tab a, neither tab shows modal after previous idle time
  - refresh tab A and tab B - modal shows, resume session in tab A, user is not logged out of either

PR #15523, opened 2025-09-25
Implement UI context metadata

### LINKED ISSUE #15528: Create UI context metadata

We want to implement a mechanism to get visual and contextual information of the currently viewed page and make them accessible to the extensions and other UI functionalities.

- Implement a new `ui-context` store to save/expose the metadata into the UI.
- Implement a new `ui-context` directive to assign context getters to the UI elements.

**Additional context**

torchiaf:
### Summary
Fixes #15528

### Technical notes summary

It adds the new `ui-context` directive & store to handle the UI context metadata.

The directive will register the context on mount and unregister it before unmounting.
This means the context will be available in the store until the bound elements stay on screen.

Usage:
  ```
  <div v-ui-context="{ tag: 'example-tag', path: 'some.path.in.component', description: 'An example context' }"></div>
  or
  <div v-ui-context="{ tag: 'example-tag', value: 'Direct Value', description: 'An example context with direct value' }"></div>
  or
  <div v-ui-context="{ tag: 'example-tag', description: 'An example context with text content' }">Text Content Value</div>
```

The context object must have a 'tag' and either a 'path' or a 'value'.
If both 'path' and 'value' are undefined, the element's textContent will be used as the value.

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/config/store.js
+++ b/shell/config/store.js
@@ -40,6 +40,7 @@ let store = {};
   resolveStoreModules(require('../store/cru-resource.ts'), 'cru-resource.ts');
   resolveStoreModules(require('../store/notifications.ts'), 'notifications.ts');
   resolveStoreModules(require('../store/cookies.ts'), 'cookies.ts');
+  resolveStoreModules(require('../store/ui-context.ts'), 'ui-context.ts');
 
   // If the environment supports hot reloading...
 
@@ -71,6 +72,7 @@ let store = {};
       '../store/cru-resource.ts',
       '../store/notifications.ts',
       '../store/cookies.ts',
+      '../store/ui-context.ts',
     ], () => {
       // Update `root.modules` with the latest definitions.
       updateModules();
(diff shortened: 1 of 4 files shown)


codyrancher (commented):
I don't have a problem with the changes but I'm not sure I'm understanding the purpose or when I'd want to use this over say just a store or provide/inject.

Could you maybe elaborate some more on the situations we'd want to use this?

torchiaf:
@codyrancher the purpose is to make available simple key: value objects for extension, for instance the AI assistant.
I don't expect we'll use it extensively - most of the info like current cluster or current namespaces are already available in the store - but consider that we have a page where we want to make accessible specific info, like labels or links, the simplest way is to use this `v-ui-context` (at least this is the way without having to parse the html from the extension).

codyrancher:
> @codyrancher the purpose is to make available simple key: value objects for extension, for instance the AI assistant. I don't expect we'll use it extensively - most of the info like current cluster or current namespaces are already available in the store - but consider that we have a page where we want to make accessible specific info, like labels or links, the simplest way is to use this `v-ui-context` (at least this is the way without having to parse the html from the extension).

Based on that info I would think provide/inject would be sufficient https://vuejs.org/guide/components/provide-inject.

I'm okay with it going in though.

PR #15535, opened 2025-09-29
 Fixed an issue where the current cluster is used when navigating to the auth configuration portion of the app 

### LINKED ISSUE #10709: Users & Authentication keeps the last visited cluster in the URL and causes errors
**Setup**
- Rancher version: v2.8-4861032fa9ab90503d95fd563ef5a7baa331365e-head
- Browser type & version: Google Chrome v122.0.6261.94 (Official Build) (arm64)

**Describe the bug**
When accessing `Users & Authentication`, the URL includes the last visited cluster.

- Accessing from the Local cluster:
image

- Accessing from our Longhorn cluster:
image

This could be causing errors like the following:

**To Reproduce**
Add a user to a Project, e.g. "Default" on the Local cluster:
image

Go to any other cluster and, from there, click on `Users & Authentication`. Look for the user, go to their Details page and open the Project Roles tab:
image

The Project column shows it in a `cluster:project-id` format, in this case `local:p-p5224`. 

Click on the Project link.

It will open the Project page but somehow within the context of the last visited cluster (as you see at the top, we're not on the Local cluster):
image

Use the Back button in your browser. The Project column now shows the Project name:
image

If we click on the Project name, we'll get an error:
image

The links in the Project column are different in the two scenarios.
The first time, when it was `local:p-p5224`, the link was:
`https://r.en3sis.com/dashboard/c/c-m-q6577r5l/explorer/management.cattle.io.project/local%2Fp-p5224`
The second time, coming back from the Project page, the link was:
`https://r.en3sis.com/dashboard/c/c-m-q6577r5l/auth/management.cattle.io.project/c-m-q6577r5l/p-p5224`

There are two changes: 
1. `/explorer/` for `/auth/`
2. `/local%2Fp-p5224` for `/c-m-q6577r5l/p-p5224`

**Expected Result**
- The link should always lead to the Project page within the correct cluster.
- Back and forth navigation should not affect the link.

codyrancher:
### Summary
Fixes #10709

### Technical notes summary
Added a route definition which includes the `_` blank cluster as a param.

Navigating to the auth configuration from both local and downstream clusters and verifying we always see the local cluster in the url.

Navigating to the auth configuration from both local and downstream clusters and verifying we always see the `_` blank cluster in the url.

### Screenshot/Video

https://github.com/user-attachments/assets/8abdb9b9-d551-4e0c-8993-4ec045fa6b87

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/form/ResourceTabs/composable.ts
+++ b/shell/components/form/ResourceTabs/composable.ts
@@ -39,7 +39,7 @@ export const useTabCountWatcher = () => {
 
 export const useTabCountUpdater = () => {
   const tabKey = randomStr();
-  const updateCount = inject<UpdateCountFn>(UPDATE_COUNT_PROVIDER_KEY);
+  const updateCount = inject<UpdateCountFn | undefined>(UPDATE_COUNT_PROVIDER_KEY, undefined);
 
   const updateTabCount = (count: number | undefined) => {
     updateCount?.(tabKey, count);
--- a/shell/config/product/auth.js
+++ b/shell/config/product/auth.js
@@ -6,6 +6,7 @@ import {
   RBAC_BUILTIN, RBAC_DEFAULT, STATE, NAME as HEADER_NAME, AGE, SIMPLE_NAME
 } from '@shell/config/table-headers';
 import { MULTI_CLUSTER } from '@shell/store/features';
+import { BLANK_CLUSTER } from '@shell/store/store-types';
 
 export const NAME = 'auth';
 
@@ -37,7 +38,7 @@ export function init(store) {
     to:                  {
       name:   'c-cluster-product-resource',
       params: {
-        product: 'auth', resource: 'management.cattle.io.user', cluster: 'local'
+        product: NAME, resource: MANAGEMENT.USER, cluster: BLANK_CLUSTER
       }
     }
   });
@@ -62,7 +63,7 @@ export function init(store) {
     route:      {
       name:   'c-cluster-product-resource',
       params: {
-        cluster:  'local',
+        cluster:  BLANK_CLUSTER,
         product:  NAME,
         resource: MANAGEMENT.USER,
       }
--- a/shell/config/product/auth.js
+++ b/shell/config/product/auth.js
@@ -34,6 +34,12 @@ export function init(store) {
     removable:           false,
     showClusterSwitcher: false,
     category:            'configuration',
+    to:                  {
+      name:   'c-cluster-product-resource',
+      params: {
+        product: 'auth', resource: 'management.cattle.io.user', cluster: 'local'
+      }
+    }
   });
 
   virtualType({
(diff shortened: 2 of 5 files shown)


shell/components/form/ResourceTabs/composable.ts
```diff
@@ -39,7 +39,7 @@ export const useTabCountWatcher = () => {
 
 export const useTabCountUpdater = () => {
   const tabKey = randomStr();
-  const updateCount = inject<UpdateCountFn>(UPDATE_COUNT_PROVIDER_KEY);
+  const updateCount = inject<UpdateCountFn | undefined>(UPDATE_COUNT_PROVIDER_KEY, undefined);
```
codyrancher, line 42: This doesn't fix the noted issue but it does resolve a warning I was seeing in the console when a table was rendered outside of a tab.

image

shell/config/product/auth.js
```diff
@@ -34,6 +34,12 @@ export function init(store) {
     removable:           false,
     showClusterSwitcher: false,
     category:            'configuration',
+    to:                  {
+      name:   'c-cluster-product-resource',
+      params: {
+        product: 'auth', resource: 'management.cattle.io.user', cluster: 'local'
```
richard-cox: this doesn't feel like the right fix, and will probably error if the user doesn't have access to the local cluster.

will add more detail tomorrow, but the general approach is to keep the current `cluster` store pointing at the last cluster visited to avoid resetting on navigation when it isn't needed.

in this instance we should use the blank cluster `_` and / or fix the place where it's looking at the `cluster` store

codyrancher: Wont the user need access to the local store to be able to touch anything in auth?

codyrancher: It did appear to be consistent with how we link to the users page in the same file https://github.com/rancher/dashboard/blob/master/shell/config/product/auth.js#L56-L63

But I'll wait for the details tomorrow.

richard-cox: Auth resources are provided by the management store, which does access the local cluster but not via the steve proxy directly to it (ish). That mechanism is how users can create clusters, projects, bindings, etc in the local cluster without being a member of the local cluster. So users could be granted access to see users & auth but not see general local cluster resources.

https://github.com/rancher/dashboard/blob/master/shell/config/product/auth.js#L56-L63 looks like a bug. Think i put that in a looong time ago.

Think the fix here is to update shell/models/management.cattle.io.projectroletemplatebinding.js `projectDetailLocation` and possibly `clusterDetailLocation` to include the correct cluster in `params`

codyrancher: @richard-cox you're right about using the blank cluster; I've gone ahead and switched to blank cluster. However, based on where you think this should be fixed I think you're assuming this pr is trying to fix more than I intended. 

There's no longer a problem with how projects are displayed in the Project Roles tab. I thought it would be nice to make our management pages more consistent instead of just closing the issue as already fixed.

https://github.com/user-attachments/assets/dc00df00-eec0-419a-ac7e-e81ca86e8e56

PR #15481, opened 2025-09-18
Re-work APIs to work properly via options and composition API and be easier to document

nwmac:
### Summary
Fixes #

### Occurred changes and/or fixed issues

WIP

This PR re-works the APIs to a pattern that better supports Vue options and composition API.

### Checklist
- [x] The PR is linked to an issue and the linked issue has a Milestone, or no issue is needed
- [x] The PR has a Milestone 
- [x] The PR template has been filled out
- [x] The PR has been self reviewed 
- [x] The PR has a reviewer assigned
- [x] The PR has automated tests or clear instructions for manual tests and the linked issue has appropriate QA labels, or tests are not needed
- [x] The PR has reviewed with UX and tested in light and dark mode, or there are no UX changes
- [x] The PR has been reviewed in terms of Accessibility

PR #15541, opened 2025-09-30
Implement the kubectl-explain to be consistent with the new drawer design

codyrancher:
### Summary
Partial fix for #14904

Just making our slideouts/drawers more consistent.

### Technical notes summary
I'm considering adding a drawer/open drawer/close commit to the slidein store which has some of the defaults already in place.

Anywhere an explain drawer is present.

The explain drawer.

### Screenshot/Video
Old:

https://github.com/user-attachments/assets/0db43938-77ae-43c9-bb96-d531ea8ae4f2

New:

https://github.com/user-attachments/assets/0075e7c8-6654-4026-81e7-b131c7d46afe

### Checklist
- [ ] The PR has a Milestone 

### DIFF
--- a/pkg/kubectl-explain/components/SlideInPanel.vue
+++ b/pkg/kubectl-explain/components/SlideInPanel.vue
@@ -1,54 +1,41 @@
 <script>
 import ExplainPanel from './ExplainPanel';
-import { KEY } from '@shell/utils/platform';
 import { expandOpenAPIDefinition, getOpenAPISchemaName, makeOpenAPIBreadcrumb } from '../open-api-utils.ts';
-import { useWatcherBasedSetupFocusTrapWithDestroyIncluded } from '@shell/composables/focusTrap';
+import Drawer from '@shell/components/Drawer/Chrome.vue';
+import { OpenAPI } from '../open-api';
 
 const HEADER_HEIGHT = 55;
 
+// Cache of the Open API Data for a cluster
+const openAPI = new OpenAPI();
+
 export default {
-  components: { ExplainPanel },
+  components: { ExplainPanel, Drawer },
+
+  emits: ['close'],
 
   props: {
     schema: {
       type:    Object,
       default: () => {}
+    },
+    cluster: {
+      type:    Object,
+      default: () => undefined
     }
   },
 
   data() {
     return {
-      isOpen:         false,
-      definition:     undefined,
-      busy:           true,
-      error:          false,
-      expandAll:      false,
-      isResizing:     false,
-      resizeLeft:     '',
-      resizePosition: 'absolute',
-      width:          '33%',
-      right:          '-33%',
-      breadcrumbs:    undefined,
-      definitions:    {},
-      noResource:     false,
-      notFound:       false,
+      definition:  undefined,
+      busy:        true,
+      error:       false,
+      expandAll:   false,
+      breadcrumbs: undefined,
+      definitions: {},
+      noResource:  false,
+      notFound:    false,
     };
-  },
-
-  watch: {
-    // better to trigger focus-trap based on "isOpen" rather than using the "created" hook
-    // because re-opening an already loaded resource data would not run the "created" hook
-    // and therefore not triggering the focus trap
-    isOpen(neu, old) {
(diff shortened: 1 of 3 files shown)


oboc-sts:
Oh, this looks (and behaves) so much better. Loving this. A few thoughts from my side:

- The drawers come in 3 sizes (large = 85%, medium = 50%, small = 35% ... of the viewport width)
- The drawers can be chained (linked) in a horizontal navigation:
    - this means you can navigate back and forth between different instances of the same drawer (pages) on the same level, and have the option to go all the way back to the original instance of the drawer
    - because of the way we've designed the drawers, this navigation can have an unlimited number of "breadcrumbs"
    - however, in order for us to reach this kind of scalability, you can only go back 1 breadcrumb at a time, using the "Go Back" button displayed on the RHS of the drawer header
        - the "Go Back" button only appears if you navigate away from the original instance of the drawer to another one
        - the "Go Back" button will have as a tooltip the title of the instance of the drawer you're going back to.
- The drawers can also be chained (linked) in a vertical navigation:
    - this means you can open a new drawer (B) from an existing drawer (A), on the next level. If A is the original drawer, it will open on the default level = 1 (0 = the level of the current page), which means drawer B will open one level higher, on level 2. We can extend this with drawer C on level 3 and so on... 
    - on each level, you can still have a horizontal navigation for the drawer on that level. E.g. you can have B1 > B2 > B3 > ... as chained instances on level 2
    - there are however a few constraints:
        - the drawers you open on the upper levels will inherit the size of the original (you can't set a different size on them)
            - when you open a drawer on a new upper level, all the drawers on the lower levels will be pushed slightly to the left (% or a fixed gap remains TBD)
        - whilst in theory this vertical navigation can spread across unlimited levels, from a UX point of view, this is not an ideal scenario. Which means we should limit this type of navigation to a maximum of 2 or 3 levels.

**The current drawer component in our DS** has already been designed based on these ^ specs and can be found [here](https://www.figma.com/design/K3e7DHHM5TsS0vQELeoIFZ/Rancher-DS?node-id=1106-1945&m=dev&t=GFmvprvUqE13ex6e-1). It looks like this:

image


**The header of the drawer is a DS sub-component** in the same [Figma file](https://www.figma.com/design/K3e7DHHM5TsS0vQELeoIFZ/Rancher-DS?node-id=1094-1269&m=dev&t=GFmvprvUqE13ex6e-1) that looks like this:

image

**And everything can be seen in action** on the SUSE Observability platform - the screencast below showcases all the specs above and their use cases. On SUSE Observability I've decided to limit the levels of vertical navigation to only 2.

https://github.com/user-attachments/assets/bbae6f21-eb80-43b0-8f9f-3b5eea1d34c5

You can play with some of these yourself on [play.stackstate.com](https://play.stackstate.com/#/welcome)

Because of the specs above, I'm afraid having a manual resizing of the drawers would complicate things more than needed (it conflicts with a few of the rules above). There is however a solution: we can add a control (button) in the RHS of the drawer header to switch between the 3 sizes we have for drawers with the following specs to consider:
- this would be visible only for drawers on level 1
- we, as a system, control this and, by default, the control will not be displayed so we would have to "opt-in" to display it on a particular drawer because we believe it could help the user
- will have to take into account the current size of the drawer and offer the user the chance to go wider or narrower, depending on the choices that make sense for the content of the drawer

Let's talk about it...

aalves08:
FYI @codyrancher we are not far from shifting to a new structure of the shell/extensions API based on the work Neil has done in https://github.com/rancher/dashboard/pull/15481. Probably has low impact on your work, but we might need to update the method's signature and comments on our side based on you work.

codyrancher:
> FYI @codyrancher we are not far from shifting to a new structure of the shell/extensions API based on the work Neil has done in #15481. Probably has low impact on your work, but we might need to update the method's signature and comments on our side based on you work.

@aalves08  Thanks for the heads up. I'll wait until the changes get merged before moving forward.

gaktive:
Removing the milestone from this PR. @codyrancher once this is picked up again, either by you or someone else, make sure the PR milestone lines up with the issue.

PR #15560, opened 2025-10-01
Chart Card Updates

### LINKED ISSUE #15559: [Apps & Marketplace] - Card updates
Let’s make sure we apply the following rules on all the cards within the charts catalogue, cluster tools and extensions:

- display the latest available version in the subtitle of the card
- display the timestamp in the subtitle, this will be the “last updated” timestamp
- If the card has the installed flag visible, display the following tooltip on hover: “Installed  ([version number])”

momesgin:
### Summary
Fixes #15559 

Made the cards look more consistent on Charts/Tools/Extensions pages.

### Technical notes summary

- Added the missing last updated date to the cards on the tools page
- Updated the installed tooltip to include the version number: “Installed ([version number])”
- Added flex-wrap to wrap the sub header items
- Updated tests

Check the cards on Charts/Tools/Extensions pages and make sure they are consistent in displaying the version, date and installed tooltip.

Hopefully none.

### Screenshot/Video

Charts page:
image
Extensions page:
image
Tools page:
image

Wrapping sub header items(e.g. version and date):
image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/models/chart.js
+++ b/shell/models/chart.js
@@ -195,7 +195,7 @@ export default class Chart extends SteveModel {
       }
 
       if (this.isInstalled) {
-        const installedVersion = this.matchingInstalledApps[0].spec.chart.metadata.version;
+        const installedVersion = this.matchingInstalledApps[0]?.spec?.chart?.metadata?.version;
 
         statuses.push({
           icon: 'icon-confirmation-alt', color: 'success', tooltip: { text: `${ this.t('generic.installed') } (${ installedVersion })` }


shell/models/__tests__/chart.test.ts
```diff
@@ -270,9 +271,16 @@ describe('class Chart', () => {
 
       const result = chart.cardContent as CardContent;
 
-      const keys = result.statuses.map((s) => s.tooltip.key);
+      const statuses = result.statuses.map((s) => {
+        if (s.tooltip.key) {
```
eva-vashkevich: s.tooltip?.key would be nice?

shell/models/chart.js
```diff
@@ -195,8 +195,10 @@ export default class Chart extends SteveModel {
       }
 
       if (this.isInstalled) {
+        const installedVersion = this.matchingInstalledApps[0].spec.chart.metadata.version;
```
eva-vashkevich: I'd add optional chaining here as well just in case

shell/pages/c/_cluster/apps/charts/chart.vue
```diff
@@ -259,11 +259,11 @@ export default {
               class="status"
             >
               <i
-                v-clean-tooltip="t(status.tooltip.key)"
+                v-clean-tooltip="status.tooltip.key ? t(status.tooltip.key) : status.tooltip.text"
```
eva-vashkevich, line 262: optional chaining?

shell/pages/c/_cluster/explorer/tools/index.vue
```diff
@@ -96,7 +96,7 @@ export default {
             title:    { text: chart.chartNameDisplay },
             statuses: chart.cardContent.statuses
           },
-          subHeaderItems: chart.cardContent.subHeaderItems.slice(0, 1),
+          subHeaderItems: chart.cardContent.subHeaderItems,
```
eva-vashkevich, line 99: optional chaining?

shell/models/__tests__/chart.test.ts
momesgin: thanks for catching this, I'll update the type and add the optional chaining

shell/models/chart.js
momesgin: I think this is fine without it, the "if" block(`if (this.isInstalled) {...)` will make sure you always have a chart and the rest of the chain should be there.

shell/pages/c/_cluster/apps/charts/chart.vue
momesgin: `tooltip` is required, but `key` and `text` are optional, so this should be fine too.

shell/pages/c/_cluster/explorer/tools/index.vue
momesgin: in the model where you get the card content(`get cardContent() {...`) we always return `subHeaderItems` in the result, so we should be fine here as well.

shell/models/__tests__/chart.test.ts
momesgin: done

shell/models/chart.js
eva-vashkevich: should or will tho?

momesgin: In the chart mixin `shell/mixins/chart.js` we have this code:
```
    currentVersion() {
      return this.existing?.spec.chart.metadata.version;
    },

```
and it's a 4 years old code, I think we would have some reports if there was an issue with it

eva-vashkevich: this is how bugs are made

momesgin: I had an assumtion that having too many optional chaining would have a noticible impact on the processing load when dealing with large data(in our case apps/charts page can be quite large), but I did some research and ran some tests myself that showed a negligible impact, so I updated the code to have more optional chaining in place. 

Thanks for your review

PR #15664, opened 2025-10-19
Add CRD support to kubectl-explain drawer

### LINKED ISSUE #15249: Explain drawer does not show crd spec description
**Setup**
- Rancher version: v2.12.0

**Describe the bug**

Not sure if it's a bug or a feature request but when opening the explain drawer for a custom resource (certmanager Certificate for example), I cannot get the description of fields nested in the spec

Image

**To Reproduce**

apply a CRD and open the explain drawer

**Result**

**Expected Result**

every field from the CRD is present in the drawer, similar to built-in resources like services

Image

or kubectl explain

```yaml
kubectl explain certificates.spec.additionalOutputFormats
GROUP:      cert-manager.io
KIND:       Certificate
VERSION:    v1

FIELD: additionalOutputFormats <[]Object>


DESCRIPTION:
    Defines extra output formats of the private key and signed certificate chain
    to be written to this Certificate's target Secret.

    This is a Beta Feature enabled by default. It can be disabled with the
    `--feature-gates=AdditionalCertificateOutputFormats=false` option set on
    both
    the controller and webhook components.
    CertificateAdditionalOutputFormat defines an additional output format of a
    Certificate resource. These contain supplementary data formats of the signed
    certificate chain and paired private key.

FIELDS:
  type  <string> -required-
  enum: DER, CombinedPEM
    Type is the name of the format type that should be written to the
    Certificate's target Secret.

```

**Screenshots**

**Additional context**

enneitex:
### Summary
Fixes [#15249](https://github.com/rancher/dashboard/issues/15249)

this change adds support for displaying CRDs fields definition similar to what is already done for native resources.  
It also fixes the kind displayed for fields like `pod.spec.containers.args` which are `[]string` but the UI previously showed `[]Object`

I tested manually with a few CRDs (see video below)

### Screenshot/Video

before:

https://github.com/user-attachments/assets/f5d5b5aa-d766-4506-9314-999cc60e188c

after:

https://github.com/user-attachments/assets/25fa94d4-e5b8-4a78-99a8-81e58fe6e56e

### Checklist
- [ ] The PR has a Milestone 

### DIFF
--- a/pkg/kubectl-explain/components/SlideInPanel.vue
+++ b/pkg/kubectl-explain/components/SlideInPanel.vue
@@ -84,7 +84,9 @@ export default {
     },
 
     scrollTop() {
-      this.$refs.main.$el.scrollTop = 0;
+      if (this.$refs.main?.$el) {
+        this.$refs.main.$el.scrollTop = 0;
+      }
     },
 
     addCloseKeyHandler() {
--- a/pkg/kubectl-explain/components/ExplainPanel.vue
+++ b/pkg/kubectl-explain/components/ExplainPanel.vue
@@ -161,7 +161,7 @@ export default {
             v-else
             class="field-type"
           >
-            {{ t('kubectl-explain.object') }}
+            {{ field.type === 'array' && field.items?.type ? field.items.type : t('kubectl-explain.object') }}
           </div>
         </div>
       </div>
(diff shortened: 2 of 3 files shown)


pkg/kubectl-explain/components/SlideInPanel.vue
```diff
@@ -84,7 +84,9 @@ export default {
     },
 
     scrollTop() {
-      this.$refs.main.$el.scrollTop = 0;
+      if (this.$refs.main?.$el) {
```
enneitex, line 87: I'm not sure about this, I hit `Uncaught TypeError: Cannot read properties of null (reading '$el')` once

PR #15683, opened 2025-10-21
Replace a modal resize hack with usage of the action-menu store.

### LINKED ISSUE #10559: RollbackWorkloadDialog: Direct Manipulation of CSS in `sizeDialog()` Function
The `sizeDialog()` function in the close workload dialog directly modifies CSS properties of elements owned and maintained by `vue-js-modal`. This approach raises concerns about maintainability for two reasons:

1. `vue-js-modal` or our modal implementation could change - leading to unintended regressions in the behavior of this dialog
2.  Directly manipulating CSS may have unintended side-effects, where these styles may be applied dynamically or overridden by other CSS rules

## Proposed Change

Instead of directly manipulating CSS, consider emitting an event from the component to signal the need for dialog sizing changes.

## Referenced Code 

https://github.com/rancher/dashboard/blob/ddbb569fe5d5423063f0f5451490862b4741dbc1/shell/dialog/RollbackWorkloadDialog.vue#L165-L172

codyrancher:
### Summary
**Note: this is for 2.14.**

Fixes #10559

### Technical notes summary
We were hackily modifying the width of a modal dialog. This uses our existing action-menu store responsible for managing the state of the modals to update the width of the modal which is already reactive.

WorkloadRollbackDialog when switching toggling between show/hide diff.

WorkloadRollbackDialog when switching toggling between show/hide diff.

### Screenshot/Video

https://github.com/user-attachments/assets/2bdd783b-990d-45e4-b555-107d9e0b2ef4

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/store/action-menu.js
+++ b/shell/store/action-menu.js
@@ -111,6 +111,14 @@ export const mutations = {
     state.modalData = data;
   },
 
+  updateModalData(state, data) {
+    state.modalData = state.modalData || {};
+
+    data.forEach(({ key, value }) => {
+      state.modalData[key] = value;
+    });
+  },
+
   clearCallbackData(state) {
     state.performCallbackData = undefined;
   },
--- a/shell/dialog/RollbackWorkloadDialog.vue
+++ b/shell/dialog/RollbackWorkloadDialog.vue
@@ -165,12 +165,9 @@ export default {
       return option.label;
     },
     sizeDialog() {
-      const dialogs = document.getElementsByClassName('modal-container');
-      const width = this.showDiff ? '85%' : '600px';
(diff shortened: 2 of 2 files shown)


shell/store/action-menu.js
```diff
@@ -111,6 +111,14 @@ export const mutations = {
     state.modalData = data;
   },
 
+  updateModalData(state, data) {
+    state.modalData = state.modalData || {};
+
+    data.forEach(({ key, value }) => {
+      state.modalData[key] = value;
```
codyrancher, line 118: We want to modify properties instead of assign a full new object to modalData because it can trigger a full re-render of a modal which can reset component state unintentionally.

PR #15743, opened 2025-10-24
Support displaying multiple ips

### LINKED ISSUE #15588: Nodes Page should show all external and internal IPs on the node
Setup

- Rancher version: v2.13-ec1bc5d858a158c372b7c90ff677cf56c43edddc-head
- Dashboard version: master d17284f
- Rancher UI Extensions: n/a
- Browser type & version: Chrome, any version

In a dual-stack DigitalOcean node-driver cluster, each node has two public IP addresses (one IPv4 and one IPv6) and two private IP addresses (one IPv4 and one IPv6).
 
Currently, the UI displays only the IPv6 address as both the external and internal IPs (see screenshot below). This appears to happen because either the IPv6 address is listed last among the addresses of each type, or ignores that there may be multiple entries of that type with different address families (see the snippet below).

Image

```
status:
  addresses:
    - address: 64.23.177.130
      type: InternalIP
    - address: 2604:a880:4:1d0:0:1:108e:c000
      type: InternalIP
    - address: 64.23.177.130
      type: ExternalIP
    - address: 2604:a880:4:1d0:0:1:108e:c000
      type: ExternalIP
    - address: do-dual-jiaqi-do-d-4j9dd-lp4lf
      type: Hostname
```

UI should display multiple values for external and internal IPs, while still showing “Same as External” for simplicity when applicable.

N/A 

**Additional context**

N/A 

UI Requirement:
- Display a pill with the text: "+{n} more" only when there are more than one extenral ips or more than one internal ips
- On hover the pill should display "Click to show {n} more  ip(s)"
- By clicking on the pill a popover should be displayed listing only the remaining ips(i.e. not including what already being displayed in the table column)
- Like kubectl we should list the first item of the arrays for internal and external ips

momesgin:
### Summary
Fixes #15588 

Please check the UI requirements added to the issue.
### Technical notes summary

- Update the corresponding formatter for displaying the external and internal ips on the table:`InternalExternalIP.vue`
- Updated `node.js` and `cluster.x-k8s.io.machine` models, they now return array of ips to the formatter and the formatter takes care of displaying multiple ips. Also for the single IPs, the first item in the array will be returned like `kubectl`rather than the last one.
- Unit tests

Go to Explore > Cluster > Nodes:
 - make sure we display the same external/internal ips that kubectl would return
 - when multiple external/internal ips are available we display the "+{n} more" pill and it works as expected(like how it's been mentioned in the ticket under UI requirements)

Check the IP columns in the tables on these pages for both single and multiple IPs:
- Explorer => Cluster => Nodes
- Cluster Management => Cluster details

Check the Target column on(finding a case with an external IP would be great):
- Explorer => Services

### Screenshot/Video

https://github.com/user-attachments/assets/344e46d6-48ac-49df-8ee0-c4d77b681ac3

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/formatter/InternalExternalIP.vue
+++ b/shell/components/formatter/InternalExternalIP.vue
@@ -21,7 +21,7 @@ export default {
       return this.row.internalIps?.filter((ip) => this.isIp(ip)) || [];
     },
     internalSameAsExternal() {
-      return this.filteredInternalIps[0] === this.filteredExternalIps[0];
+      return this.externalIp && this.internalIp && this.externalIp === this.internalIp;
     },
     showPopover() {
       return this.filteredExternalIps.length > 1 || this.filteredInternalIps.length > 1;


shell/components/formatter/InternalExternalIP.vue
```diff
@@ -2,19 +2,60 @@
...
+    ...mapGetters({ t: 'i18n/t' }),
+    filteredExternalIps() {
+      return this.row.externalIps?.filter((ip) => this.isIp(ip)) || [];
+    },
+    filteredInternalIps() {
+      return this.row.internalIps?.filter((ip) => this.isIp(ip)) || [];
+    },
     internalSameAsExternal() {
-      return this.row.internalIp === this.row.externalIp;
+      return this.filteredInternalIps[0] === this.filteredExternalIps[0];
```
mantis-toboggan-md: I think this computed property needs to be tweaked a bit. We're now showing 	
`- / Same as External` when neither category of IP is defined; I think we still want to show `- / -`
before
Screenshot 2025-10-24 at 4 52 14 PM

after
Screenshot 2025-10-24 at 4 52 21 PM

mantis-toboggan-md (commented):
The InternalExternalP formatter is also used in the cluster detail page; this PR has broken it such that it always shows `- / Same as External`

I think what needs to happen is for the updated getters to be added to `cluster.x-k8s.io.machine` model to support rke2/k3s clusters detail page. At a glace it looks like they'd be veeery similar to what you have in the node model already. I think the management.cattle.io.node model is only used for rke1 detail page so we should be ok to leave it alone, but worth double-checking that 

Screenshot 2025-10-24 at 4 52 21 PM

momesgin:
> The InternalExternalP formatter is also used in the cluster detail page; this PR has broken it such that it always shows `- / Same as External`
> 
> I think what needs to happen is for the updated getters to be added to `cluster.x-k8s.io.machine` model to support rke2/k3s clusters detail page. At a glace it looks like they'd be veeery similar to what you have in the node model already. I think the management.cattle.io.node model is only used for rke1 detail page so we should be ok to leave it alone, but worth double-checking that
> 
> Screenshot 2025-10-24 at 4 52 21 PM

thanks for catching this, since the formatter was updated to receive an array I've updated `cluster.x-k8s.io.machine` model to return an array as well.

shell/models/cluster.x-k8s.io.machine.js
```diff
@@ -244,29 +244,19 @@ export default class CapiMachine extends SteveModel {
...
+  }
 
-    if (external) {
-      return external;
-    }
+  get internalIp() {
+    return this.internalIps[0];
+  }
 
-    return this.t('generic.none');
```
momesgin, line 270: I intentionally didn't add back returning the translation string for consistency and also the risk of checking for `!!externalIp`, the formatter will take care of absent ips by returning "-".

shell/components/formatter/InternalExternalIP.vue
```diff
@@ -25,40 +66,170 @@ export default {
...
+        />
+      </span>
     </template>
     <template v-else>
       -
     </template>
-    /
-    <template v-if="internalSameAsExternal && isIp(row.internalIp)">
+    <span class="separator">/</span>
+    <template v-if="internalSameAsExternal">
```
mantis-toboggan-md, line 86: This formatter will render `- / Same as External` when neither internal nor external IPs are defined. Though technically they are the same when they're both null, I think we still want the table formatter to render `- / -` in that case. I think one way of doing that would be changing this line to something like

`<template v-if="internalSameAsExternal && internalIp">`

shell/models/cluster.x-k8s.io.machine.js
mantis-toboggan-md: That makes sense to me. I checked usage of `internalIp` and `externalIp` in the dashboard repo looking for possible regressions and I think this change is safe.

shell/components/formatter/InternalExternalIP.vue
momesgin: makes sense, I updated the logic inside the computed property `internalSameAsExternal`

shell/models/cluster.x-k8s.io.machine.js
momesgin: thanks for checking 🙏

PR #15750, opened 2025-10-27
Automation [2.12.4]: skip repository test and fix about page prime test

yonasberhe23:
### Summary
Fixes #

Resolves two issues blocking CI:
1. fixes About page prime test https://github.com/rancher/qa-tasks/issues/1953
2. skips Repositories test

The `Repo Refresh results in correct api requests` test is failing because repository refresh doesn't trigger GET requests on the first click (works on second click).
- Skip the failing test with reference to GitHub issue #15749
- Test will be re-enabled once the underlying issue is fixed

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/cypress/e2e/tests/pages/explorer/apps/repositories.spec.ts
+++ b/cypress/e2e/tests/pages/explorer/apps/repositories.spec.ts
@@ -1,7 +1,7 @@
 import ReposListPagePo from '@/cypress/e2e/po/pages/chart-repositories.po';
 import AppClusterRepoEditPo from '@/cypress/e2e/po/edit/catalog.cattle.io.clusterrepo.po';
-import { ChartPage } from '@/cypress/e2e/po/pages/explorer/charts/chart.po';
-import { ChartsPage } from '@/cypress/e2e/po/pages/explorer/charts/charts.po';
+// import { ChartPage } from '@/cypress/e2e/po/pages/explorer/charts/chart.po';
+// import { ChartsPage } from '@/cypress/e2e/po/pages/explorer/charts/charts.po';
 import { CLUSTER_REPOS_BASE_URL } from '@/cypress/support/utils/api-endpoints';
 
 describe('Apps', () => {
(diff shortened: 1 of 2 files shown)


cypress/e2e/tests/pages/explorer/apps/repositories.spec.ts
```diff
@@ -154,64 +154,67 @@ describe('Apps', () => {
...
-        cy.wait('@rancherCharts3').its('request.url').should('include', '?link=index');
-
-        // Nav to the summary page for a specific chart
-        ChartPage.navTo(clusterId, 'Rancher Backups');
-        chartPage.waitForPage('repo-type=cluster&repo=rancher-charts&chart=rancher-backup');
-        // The specific version of the chart should be fetched (as the cache was cleared)
-        cy.wait('@rancherCharts3').its('request.url').should('include', 'version=');
-      });
-    });
+    // Skipping this test due to https://github.com/rancher/dashboard/issues/15749
```
izaac, line 157: I think making this skipped with skip would be preferable.

yonasberhe23: @izaac do you mean with `.skip`? e2e best practices doc states it's better to just comment out the test instead of using .skip.

https://extensions.rancher.io/internal/testing/e2e-test#best-practices

izaac: Got it. I've read that document, and I can see what they're getting at. It looks like they're suggesting we comment out tests as a specific workaround, probably to avoid some 'fluff' in the test reports from grep tags.

While it's not a standard industry practice (mainly because it can make tests easy to forget and might skew our reports) I'm totally fine with following this note for now to keep things moving.

Let's just make sure we flag this as something to revisit. That way, we can address the underlying issue later and get back to using the standard .skip() method, which is safer for long-term maintenance.

izaac (approved):
Approved with a note.

PR #15691, opened 2025-10-22
Ipv6 support 

mantis-toboggan-md:
### Summary
Fixes #15161
Fixes #15279 
Fixes #15261 
Fixes #15599 

### Occurred changes and/or fixed issues
This PR adds ipv6-related functionality to 3 areas

#### ec2 Machine Configs
A new checkbox 'Enable IPv6' has been added to the ec2 machine pool UI. When the checkbox is unchecked, only ipv4 vpcs and subnets will be shown in the networking dropdown.  When the checkbox is checked, only ipv6-enabled vpcs and subnets will be shown in the networking dropdown, and two additional inputs shown: ipv6 address count and primary ipv6.

VPCs can be either ipv4-only or dual stack (ipv4 and ipv6). Subnets can be ipv4-only, dual-stack, or ipv6 only. If the selected network option is dual-stack, users are shown another checkbox, 'IPv6 Only.' If the selected network is an ipv6-only subnet, the IPv6 Only checkbox is not shown and the property it controls, `<ec2 machine config>.ipv6AddressOnly` is set to `true`

If one pool is using ipv6 all pools must use ipv6. If users select an ipv4 subnet in one pool then select an ipv6 subnet in another pool, the networking sections will show an error banner telling users that they need to make all pools' networking ipv4-only or ipv6/dual-stack.

#### v2 Cluster Provisioning Networking tab
A new dropdown has been added to the networking tab in cluster configuration called "Stack Preference." This dropdown will default to "IPv4" unless the UI detects that node pools have been configured to use IPv6*. If using ipv6 the dropdown will be updated to "dual stack" and validated against selecting 'IPv4' - that is, either IPv6 or Dual-Stack are valid configurations. 

* we can only tell with ec2 and digital ocean nodes right now

#### Cluster Registration Tab
The cluster registration tab has a new warning shown when the UI detects that ipv6 is in use. At this stage nodes haven't been registered so the UI looks to the cluster's stackPreference field to determine if there might be ipv6 nodes. If stackPreference is set to dual or ipv6, the UI will open the 'advanced' section by default and render a warning banner. The warning banner looks off in dark mode; this is a known issue being tracked elsewhere https://github.com/rancher/ux/issues/315#issuecomment-3421078148

### Technical notes summary
[...]

PR #15772, opened 2025-10-28
Automation: fix cluster provisioning tests

yonasberhe23:
### Summary
Fixes #

Fixes two issue:
1. Amazon EC2 cluster provisioning tests were failing after the IPv6 support feature was merged #15691. The test was attempting to select a network option labeled 'maxdualstack-vpc' which is not available when IPv4 configuration is selected (IPv4 is the default Stack Preference). The IPv6 support feature introduced logic that filters available network options based on IPv6 configuration - dual-stack networks like 'maxdualstack-vpc' only appear when IPv6 is enabled. This fix updates the test to use the 'default' network option instead, ensuring the test passes without requiring additional UI interactions to enable IPv6 mode.
2. Additionally, this PR fixes flakiness with cluster state assertions by updating the test to accept both 'Reconciling' and 'Updating' states instead of only checking for 'Reconciling'. This addresses timing issues where the cluster may transition through different states during the provisioning process, ensuring the test reliably passes regardless of which state is observed immediately after cluster creation.

### Checklist
- [x] The PR has a Milestone 


yonasberhe23:
fixed in https://github.com/rancher/dashboard/pull/15773

PR #15791, opened 2025-10-29
Fix missing store on ui-context directive unmount

torchiaf:
### Summary
Fixes #

### Technical notes summary

We noticed that in certain conditions, It can happen that the store variable from the parent where the directive is mounted is undefined.
This happens because the `beforeUnmount` in the directive is called after the element is already unmounted.

We are now saving the remove API in the element inself so that we can eliminate the store dependency on unmount.

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/package.json
+++ b/shell/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@rancher/shell",
-  "version": "3.0.8-rc.5",
+  "version": "3.0.8-rc.6",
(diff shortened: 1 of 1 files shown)


codyrancher (dismissed):

torchiaf:
@codyrancher I realized that the solution was much simpler, I pushed a new commit.

PR #15818, opened 2025-10-30
Render the brand image when the route is not cluster

### LINKED ISSUE #15673: Rancher logo "flashes" when navigating betwen pages
**Setup**
- Rancher version: v2.13-head
- Rancher UI Extensions: NA
- Browser type & version: Firefox & Chrome

**Describe the bug**

The Rancher logo will "flash" when navigating between pages.

**To Reproduce**

1. Provision a downstream cluster
2. Navigate between local cluster and newly provisioned downstream cluster

**Result**

Rancher logo will "flash"

**Expected Result**

The logos will transition smoothly.

**Screenshots**

https://github.com/user-attachments/assets/d1ecf35a-8d54-4cf8-b4e3-ddf83974828f

**Additional context**

rak-phillip:
### Summary

This updates when the brand image is shown so that it will render only when not on a cluster route.

Fixes #15673 

- Render the brand image when the route is not cluster

### Technical notes summary

`currentCluster` is set to `undefined` before receiving a new value when navigating between different clusters, this is the root cause of the "flashing" behavior. Updating the conditional rendering for the brand image so that it's based off of the route helps to remediate this issue.

Navigation throughout dashboard. The correct header image should be shown.

There might be some scenarios where the brand image was shown for a cluster route.

### Screenshot/Video

https://github.com/user-attachments/assets/a203426d-4b4c-4b7c-b936-cca69c3ef398

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/nav/Header.vue
+++ b/shell/components/nav/Header.vue
@@ -495,7 +495,7 @@ export default {
             :alt="t('branding.logos.label')"
           />
           <div
-            v-if="!currentCluster"
+            v-if="!$route.path.startsWith('/c/')"
             class="simple-title"
           >
             <BrandImage


rak-phillip:
> I had a question and a note here.
> 
>     * In what scenario do we actually want to render the image on line 501? While clicking through it seems we never actually render it. It makes me wonder if we could delete this all together. The brand image on line 542 appears to be the one we actually render everywhere.
> 

I think it's just the home layout - we originally specified `simple` via layouts:

https://github.com/rancher/dashboard/pull/2730/files#diff-e7ced6fc7ab620b59c02c41687323bfaaf4269f2c8c50abc1b38e37a48470c7a

And now in templates (plain.vue & `fail-whale.vue` as well):

https://github.com/rancher/dashboard/blob/master/shell/components/templates/home.vue#L76

It could be a vestige.. the history isn't super clear. 

>     * There isn't a scenario that I found where this didn't work, however, it relies on convention  and there's nothing stopping us or extension developers from completely changing the route naming conventions.

I understand the risk of making this change. Maybe we can try to assess other approaches to better guard against. The existing approach of relying on the current cluster isn't working, so my thought process was that we could use the cluster routes as a stand-in.

codyrancher:
> I understand the risk of making this change. Maybe we can try to assess other approaches to better guard against. The existing approach of relying on the current cluster isn't working, so my thought process was that we could use the cluster routes as a stand-in.

I say go ahead and merge

a110605:
@rak-phillip , @codyrancher , this PR changes causes rancher logo to display over the imported harvester logo. 

Screenshot 2025-11-03 at 2 59 14 PM

PR #15886, opened 2025-11-06
Implements RcIcon from the design system

codyrancher:
### Summary
Implements the RcIcon component of our Design System.

### Technical notes summary
I did some refactoring so that our status color/design could more easily be reused.

The storybook for RcIcon, RcStatusBadge and RcStatusIndicator.

See above

### Screenshot/Video
image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/pkg/rancher-components/src/components/RcIcon/RcIcon.test.ts
+++ b/pkg/rancher-components/src/components/RcIcon/RcIcon.test.ts
@@ -13,4 +13,39 @@ describe('rcIcon.vue', () => {
     expect(wrapper.classes()).toContain('icon-search');
     expect(wrapper.classes()).toContain('medium');
   });
+
+  it('sets aria-hidden to true by default', () => {
+    const wrapper = shallowMount(RcIcon, {
+      props: {
+        size: 'medium',
+        type: 'search',
+      },
+    });
+
+    expect(wrapper.attributes('aria-hidden')).toBe('true');
+  });
+
+  it('sets aria-hidden to false when explicitly provided', () => {
+    const wrapper = shallowMount(RcIcon, {
+      props: {
+        size:       'medium',
+        type:       'search',
+        ariaHidden: false,
+      },
+    });
+
+    expect(wrapper.attributes('aria-hidden')).toBe('false');
+  });
+
+  it('sets aria-hidden to true when explicitly provided', () => {
+    const wrapper = shallowMount(RcIcon, {
+      props: {
+        size:       'medium',
+        type:       'search',
+        ariaHidden: true,
+      },
+    });
+
+    expect(wrapper.attributes('aria-hidden')).toBe('true');
+  });
 });


yarn.lock
codyrancher: I'm thinking about removing icons from the base and shell dependencies

shell/pages/c/_cluster/manager/hostedprovider/index.vue
```diff
@@ -196,7 +196,8 @@ export default {
             <span class="mr-10">{{ row.name }}</span>
             <RcStatusBadge
               v-if="row.prime"
-              status="prime"
+              class="prime-badge"
+              status="success"
```
codyrancher, line 200: Prime isn't a standard status so I removed it and added a class so we can override the values locally.

pkg/rancher-components/src/components/utils/status.ts
codyrancher, line 1: I placed utils inside the components library due to extensive issues with referencing imports outside of this directory via aliases or relative paths.
Depending on what you choose I ran into issues with:
- Building extensions
- Building storybook
- Running storybook
- Importing into shell

pkg/rancher-components/src/components/RcIcon/RcIcon.vue
```diff
@@ -0,0 +1,36 @@
...
+  return RcIconType[props.type];
+});
+
+const color = computed(() => {
+  if (props.status && props.status === 'inherit') {
+    return 'inherit';
+  }
+
+  return useStatusColors(props as StatusObject, 'outlined').textColor.value;
+});
```
rak-phillip, line 30: I don't think that type casting is the best approach, I find that this can typically pave over type assertions that would otherwise fail. We might be able to ensure that we invoke `useStatusColors()` with a valid status via something like:

```suggestion
const color = computed(() => {
  if (props.status === 'inherit') {
    return 'inherit';
  }

  const status = props.status || 'none';

  return useStatusColors({ status }, 'outlined').textColor.value;
});
```

pkg/rancher-components/src/components/RcIcon/RcIcon.vue
```diff
@@ -0,0 +1,36 @@
...
+</script>
```
rak-phillip, line 31: nit: After performing a more in-depth review of the `useStatusColors()` composable, I don't think that we need to be invoking `useStatusColors()` every time the color computed is re-evaluated. Since `useStatusColors()` is returning reactive state in the form of a computed, I think that a more idiomatic approach would look something like this:
  

```suggestion
<script setup lang="ts">
import { RcIconProps, RcIconType, RcIconSize } from '@components/RcIcon/types';
import { computed } from 'vue';
import { useStatusColors } from '@components/utils/status';
const props = withDefaults(defineProps<RcIconProps>(), { size: 'small' });

const fontSize = computed(() => {
  return RcIconSize[props.size];
});

const iconClass = computed(() => {
  return RcIconType[props.type];
});

const status = computed(() => {
  if (props.status && props.status !== 'inherit') {
    return props.status;
  }

  return 'none';
});

const { textColor } = useStatusColors({ status: status.value }, 'outlined');

const color = computed(() => {
  if (props.status === 'inherit') {
    return 'inherit';
  }

  return textColor.value;
});
</script>
```

The tradeoff is that we exchanged the nesting with an additional computed.

pkg/rancher-components/src/components/utils/status.ts
rak-phillip: I think that this is fine as long we intend to keep these utils internal to Rancher Components.

shell/pages/c/_cluster/manager/hostedprovider/index.vue
rak-phillip: I would still like to understand why we need to be overriding styles like this. IMO this is a sign there's a gap between the design system, feature designs, and the component library implementation.

pkg/rancher-components/src/components/RcIcon/RcIcon.vue
codyrancher: Yeah, this makes sense. Switching to this.

shell/pages/c/_cluster/manager/hostedprovider/index.vue
codyrancher: I agree, we already had a line in the components doc we're maintaining but I've also notated this comment.

PR #15929, opened 2025-11-12
Fix Tooltips Performance Issue

### LINKED ISSUE #15781: Charts index page sometimes loads very slowly in firefox
**Setup**
- Rancher version: v2.13-b42202a64d6edf623acee4901a5334fcdacb53b3-head
- Rancher UI Extensions: none
- Browser type & version: firefox 144.0

**Describe the bug**
Sometimes the /apps/charts page loads quite slow, around 20 seconds. I don't see slow network requests. Sometimes even after the cards have loaded, the UI freezes and I must close the tab and open a new one. This happens with the local cluster and with downstream clusters. 

**To Reproduce**
1. navigate to apps->charts

**Result**
Page takes ages to load; sometimes the ui locks up when it does load

**Expected Result**
it should not take 20 seconds to load the page

momesgin:
### Summary
Fixes #15781 

This PR addresses a major performance issue with tooltips that was especially noticeable on pages with many elements, like the Apps/Charts page, and was almost causing Firefox to freeze when the developer tools open since it does much more proccess around Vue components. The root cause was that the previous directive implementation instantiated a full Vue component for every single tooltip upon page load. This created hundreds of component instances memory, each with its own lifecycle and watchers, leading to a massive upfront processing cost that could freeze the browser.

The new implementation avoids this by using a lightweight, event-driven approach. It no longer creates Vue components for each tooltip. Instead it creates and destroys a single, simple tooltip instance on-demand like on mouse hover, which reduces memory usage and fixes the slowdown.

### Technical notes summary

I've **tried** to cover all the current functinalities we have for the tooltips(e.g. modifiers, delay, triggers, etc.), also:

* Converted`shell/directives/clean-tooltip.js` to typescript
* Added unit tests and updated e2e tests

- Verify that tooltips appear correctly on hover/focus, display the content, and disappear on mouse leave/blur across different UI elements on different pages.
- Navigate to Explorer -> Apps/Charts in Firefox, open developer tools and refresh the page. Monitor performance during page load, scrolling, and tooltip interactions.
- Test tooltips with different placement modifiers (e.g., `v-clean-tooltip.bottom`) to ensure they render in the correct positions.
- Verify that tooltips respect the delays.
- Test the tooltips that their content changes dynamically, I had to mock this situation by adding a `setTimeout` function to manually update the content, but we probably have some examples in the app.
- Check keyboard navigation (focus/blur) triggers tooltips as expected.
- Check the side menu for the tooltips on the icons, it should behave the same as before, e.g. hovering on the icons should only trigger the tooltips if the side menu is collapsed

Testing the above items should cover most of the possible regressions too. Also make sure the styling is identical or super close.

### Screenshot/Video

https://github.com/user-attachments/assets/27044981-3c5b-44de-930f-e7618543415f

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/pages/c/_cluster/explorer/__tests__/index.test.ts
+++ b/shell/pages/c/_cluster/explorer/__tests__/index.test.ts
@@ -211,7 +211,7 @@ describe('page: cluster dashboard', () => {
 
       expect(box.element).toBeDefined();
       expect(box.element.classList).toContain(status);
-      expect(!!(box.element as any).__tooltipValue__).toBe(clickable);
+      expect(!!(box.element as any).__tooltipOptions__?.content).toBe(clickable);
       expect(icon.element.classList).toContain(iconClass);
 
       await box.trigger('click');


cypress/e2e/tests/pages/fleet/gitrepo.spec.ts
```diff
@@ -4,7 +4,6 @@ import { generateFakeClusterDataAndIntercepts } from '@/cypress/e2e/blueprints/n
 import PreferencesPagePo from '@/cypress/e2e/po/pages/preferences.po';
 import { EXTRA_LONG_TIMEOUT_OPT } from '@/cypress/support/utils/timeouts';
 import { HeaderPo } from '@/cypress/e2e/po/components/header.po';
-import 'cypress-real-events/support';
```
momesgin, line 7: Moved to `cypress/support/e2e.ts` so it becomes available for all the e2e tests

shell/directives/clean-tooltip.ts
```diff
@@ -0,0 +1,271 @@
...
+
+  // Don't show the tooltip if the content is empty.
+  const finalContent =
+    typeof content === 'object' && content.content ? content.content.trim() : typeof content === 'string' ? content.trim() : '';
+
+  if (!finalContent) {
+    return;
+  }
+  // Create a new tooltip instance.
+  singleton = createTooltip(target, {
```
rak-phillip: It looks like `creatTooltip` expects 3 arguments:

```
Expected 3 arguments, but got 2.ts(2554)
v-tooltip.d.ts(7, 60): An argument for 'modifiers' was not provided.
```

shell/directives/clean-tooltip.ts
```diff
@@ -0,0 +1,271 @@
...
+ * @param {string|object} value The content to be purified.
+ * @returns {string|object} The purified content.
+ */
+function purifyContent(value: string | { content: string }): string | { content: string } {
+  if (typeof value === 'string') {
+    return purifyHTML(value);
+  } else if (
+    value &&
+    typeof value === 'object' &&
+    typeof value.content === 'string'
```
rak-phillip: Will this always evaluate to `true`? It looks like `content` is always a type of string.

shell/directives/clean-tooltip.ts
```diff
@@ -0,0 +1,271 @@
+import { DirectiveBinding, Directive } from 'vue';
+import { destroyTooltip, createTooltip } from 'floating-vue';
+import { purifyHTML } from '@shell/plugins/clean-html';
+
+// This is a singleton tooltip implementation that improves performance on pages with many tooltips.
+// Instead of instantiating a Vue component for every tooltip on the page, this directive attaches lightweight event listeners.
+// It then imperatively creates and destroys a single tooltip instance as needed, avoiding the high upfront memory and processing cost of many Vue components.
+let singleton: any = null;
```
rak-phillip: Instead of `any` we should be able to specify a type for `singleton` based off the return type of `createTooltip()`


```suggestion
let singleton: ReturnType<typeof createTooltip> | null = null;
```

shell/directives/clean-tooltip.ts
momesgin: thanks, I missed that in my env.

shell/directives/clean-tooltip.ts
momesgin: great! thanks for the code suggestion

shell/directives/clean-tooltip.ts
momesgin: I basically got this condition from the previous directive implementation and didn't really question it:

```
function purifyContent(value) {
  const type = typeof value;

  if (type === 'string') {
    return purifyHTML(value);
  } else if (value && type === 'object' && typeof value.content === 'string') {
    return { ...value, content: purifyHTML(value.content) };
  } else {
    return value;
  }
}
```

but I think you're right, and I believe the fact that content is always string can simplify the code

pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.vue
```diff
@@ -355,7 +355,7 @@ export default defineComponent({
       disabled: isDisabled,
       [status]: status,
       suffix: hasSuffix,
-      'v-popper--has-tooltip': hasTooltip,
+      'has-clean-tooltip': hasTooltip,
```
richard-cox, line 358: there;s a couple of references to v-popper--has-tooltip still in cypress/e2e/tests/pages/explorer2/namespace-picker.spec.ts, do those need updating?

(starting review, got pulled off to something else, didn't want to lose this comment)

shell/directives/clean-tooltip.ts
```diff
@@ -0,0 +1,260 @@
+import { DirectiveBinding, Directive } from 'vue';
+import { destroyTooltip, createTooltip } from 'floating-vue';
+import { purifyHTML } from '@shell/plugins/clean-html';
+
+// This is a singleton tooltip implementation that improves performance on pages with many tooltips.
```
richard-cox, line 5: It feels like there's a lot of work / lines of code, doesn't floating-vue's popper already have something specifically for this?

shell/directives/clean-tooltip.ts
```diff
@@ -0,0 +1,260 @@
...
+  let placement = 'top';
+  let popperClass = '';
+  let delay: TooltipDelay = { show: 250, hide: 100 };
+  let triggers: string[];
+
+  if (typeof value === 'string') {
+    content = value;
+    triggers = ['hover'];
+  } else if (value && typeof value === 'object') {
+    content = value.content || '';
```
richard-cox: all of these `|| ..` duplicate the initialisation above.

```
content = value.content || content
```

the above pattern should work for all of these

shell/directives/clean-tooltip.ts
```diff
@@ -0,0 +1,260 @@
...
+ * @param {string|object} value The value of the directive.
+ * @param {object} modifiers The modifiers of the directive.
+ * @returns {object} The parsed tooltip options.
+ */
+function getTooltipOptions(value: string | { content?: string, placement?: string, popperClass?: string, delay?: TooltipDelay, triggers?: string[] }, modifiers: Partial<Record<string, boolean>>): TooltipOptions {
+  let content = '';
+  let placement = 'top';
+  let popperClass = '';
+  let delay: TooltipDelay = { show: 250, hide: 100 };
+  let triggers: string[];
```
richard-cox:
```suggestion
  let triggers: string[] = ['hover'];
```
then all `triggers = ['hover'];` from below can be removed

shell/directives/clean-tooltip.ts
```diff
@@ -0,0 +1,260 @@
...
+  }
+}
+
+/**
+ * Parses the tooltip options from the directive's value and modifiers.
+ * @param {string|object} value The value of the directive.
+ * @param {object} modifiers The modifiers of the directive.
+ * @returns {object} The parsed tooltip options.
+ */
+function getTooltipOptions(value: string | { content?: string, placement?: string, popperClass?: string, delay?: TooltipDelay, triggers?: string[] }, modifiers: Partial<Record<string, boolean>>): TooltipOptions {
```
richard-cox: apart from the string / object and content conversion trying to work out why all this is done?

looks like we're just strictly defining/typing the config and providing new defaults.... instead of just using the underlying components built-in defaults

most of config is optional, can that be reflected in the typing and we drop any defaults from here?

cypress/e2e/po/side-bars/burger-side-menu.po.ts
```diff
@@ -87,12 +87,12 @@ export default class BurgerMenuPo extends ComponentPo {
     this.sideMenu().should('have.class', 'menu-close');
   }
 
-  static checkIconTooltipOn(): Cypress.Chainable {
-    return cy.get('.option').get('.cluster-icon-menu').first().should('have.class', 'v-popper--has-tooltip');
+  static checkIconTooltipOn(content: string): Cypress.Chainable {
+    return cy.get('.v-popper__popper .v-popper__inner').should('be.visible').and('contain.text', content);
```
richard-cox, line 91: comment only.

ironically `v-popper--has-tooltip` previously was on the element even if the tooltip was not visible. now though it's only there if the tooltip IS visbile

momesgin:
> Didn't the old method not attach a component but a directive, and the fix here is to swap out that memory intensive directive for a lightweight wrapper directive instead?
> 
> I've

seems you didn't finish your comment

shell/directives/clean-tooltip.ts
momesgin: No, floating-vue does not have a built-in option for this

shell/directives/clean-tooltip.ts
momesgin: that's great feedback 🙏

pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.vue
momesgin: you're right, we should replace them as well

PR #15998, opened 2025-11-21
Automation: Add fleet helmops tests

yonasberhe23:
### Summary
Fixes #https://github.com/rancher/qa-tasks/issues/2089

Adds E2E tests for creating and editing Fleet HelmOps with Secrets and ConfigMaps. 
- Verifies that downstreamResources are correctly placed in spec (not spec.helm). 
- Verifies changes persist after removing existing resources and adding new ones, with proper handling of 409 conflict retries.

### Screenshot/Video
image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/cypress/e2e/po/components/labeled-select.po.ts
+++ b/cypress/e2e/po/components/labeled-select.po.ts
@@ -77,6 +77,22 @@ export default class LabeledSelectPo extends ComponentPo {
     return this.self().type(name);
   }
 
+  /**
+   * Click the deselect X button for a selected item
+   * @param label The label of the selected item to deselect
+   * @returns
+   */
+  clickDeselectButton(label: string): Cypress.Chainable {
+    return this.self()
+      .contains(label)
+      .then(($selectedItem) => {
+        // Use within() to scope the search to the selected item and click the button within that scope
+        return cy.wrap($selectedItem).within(() => {
+          cy.get('button.vs__deselect').click();
+        });
+      });
+  }
+
   static byLabel(self: CypressChainable, label: string): LabeledSelectPo {
     return new LabeledSelectPo(
       self
--- a/cypress/e2e/tests/pages/fleet/helmop.spec.ts
+++ b/cypress/e2e/tests/pages/fleet/helmop.spec.ts
@@ -1,0 +1,345 @@
+import { FleetApplicationCreatePo, FleetApplicationListPagePo } from '~/cypress/e2e/po/pages/fleet/fleet.cattle.io.application.po';
+import { FleetHelmOpCreateEditPo } from '@/cypress/e2e/po/pages/fleet/fleet.cattle.io.helmop.po';
+import { generateFakeClusterDataAndIntercepts } from '@/cypress/e2e/blueprints/nav/fake-cluster';
+import { HeaderPo } from '@/cypress/e2e/po/components/header.po';
+import { EXTRA_LONG_TIMEOUT_OPT, MEDIUM_TIMEOUT_OPT } from '@/cypress/support/utils/timeouts';
+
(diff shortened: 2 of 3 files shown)


cypress/e2e/tests/pages/fleet/helmop.spec.ts
```diff
@@ -0,0 +1,340 @@
...
+      // Metadata step
+      helmOpCreatePage.resourceDetail().createEditView().nameNsDescription()
+        .name()
+        .set(helmOpName);
+      helmOpCreatePage.resourceDetail().createEditView().nextPage();
+
+      // Chart step
+      helmOpCreatePage.setChart('jenkins');
+      helmOpCreatePage.setRepository('https://charts.jenkins.io');
+      helmOpCreatePage.setVersion('5.8.61');
```
izaac: I noticed we are using an external chart (Jenkins) here. This introduces a dependency on `charts.jenkins.io` being reachable and version `5.8.61` existing.

To avoid network flakes or upstream changes breaking our pipeline, could we either:

Use a standard 'test' chart that exists within our own repo/setup?
Or, if we must use an external one, use a generic one that we know is extremely stable? i.e. from `https://artifacthub.io`, a lot of them hosted in Quay which is very stable and redundant.

cypress/e2e/tests/pages/fleet/helmop.spec.ts
```diff
@@ -0,0 +1,340 @@
...
+      helmOpEditPage.configMapsSelector().clickLabel(configMapName);
+
+      // Save the HelmOp
+      helmOpEditPage.resourceDetail().createEditView().save();
+
+      // Wait for update request (handles 409 conflicts with automatic retries)
+      cy.wait('@updateHelmOp').then((interception) => {
+        // If 409 conflict, wait for retry
+        if (interception.response?.statusCode === 409) {
+          cy.log('HelmOp update failed with 409 error, waiting for retry...');
```
izaac: If the application handles the retry logic automatically, can we simply wait for the final success state? If the application doesn't retry automatically, masking the 409 might be hiding a race condition issue in the Dashboard.

cypress/e2e/po/components/labeled-select.po.ts
```diff
@@ -77,6 +77,20 @@ export default class LabeledSelectPo extends ComponentPo {
...
+   * Click the deselect X button for a selected item
+   * @param label The label of the selected item to deselect
+   * @returns
+   */
+  clickDeselectButton(label: string): Cypress.Chainable {
+    return this.self()
+      .find('.vs__selected-options')
+      .contains('span.vs__selected', label)
+      .closest('.vs__selected')
+      .find('.vs__deselect')
```
izaac: Traversing up and down the DOM (closest -> find) is brittle, could we use a testid ?

cypress/e2e/tests/pages/fleet/helmop.spec.ts
yonasberhe23: I've updated the test to wait for the first request, and if it's not 200, wait for the next request (which should be Dashboard's automatic retry). We don't explicitly check for 409, we just wait for a 200 response. If Dashboard doesn't automatically retry, the second cy.wait will timeout and the test will fail, exposing the bug.

This is what it looks like:
Image

cypress/e2e/po/components/labeled-select.po.ts
yonasberhe23: Removed the closest() + find() traversal. The deselect button is rendered by vue-select (third-party), so we can't add testids. Updated to use within() to scope the search, avoiding upward DOM traversal

PR #16023, opened 2025-11-24
Implement the RcButton to better reflect the design system

codyrancher:
### Summary
Implements the RcButton component of our Design System.

### Technical notes summary
I did backslide on one of the RcIcon changes we made in that PR to make the status/color reactive.

The storybook for RcIcon and RcButton.

See above

### Screenshot/Video
image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/pkg/rancher-components/src/components/RcButton/RcButton.vue
+++ b/pkg/rancher-components/src/components/RcButton/RcButton.vue
@@ -36,6 +36,10 @@ const buttonSizesNew: { size: ButtonSize, className: string }[] = [
 const props = withDefaults(defineProps<ButtonVariantProps & ButtonSizeProps & ButtonVariantNewProps & ButtonSizeNewProps & IconProps>(), { size: 'medium' });
 
 const activeVariantClassName = computed(() => {
+  if (props.variant === 'multiAction' || props.multiAction) {
+    console.warn('[RcButton] The "multiAction" variant is deprecated and will be removed in a future version.'); // eslint-disable-line no-console
+  }
+
   const activeVariant = buttonVariants.find(({ variant }) => props[variant]);
 
   if (activeVariant) {
--- a/shell/components/Drawer/Chrome.vue
+++ b/shell/components/Drawer/Chrome.vue
@@ -39,7 +39,7 @@ const ariaLabel = computed(() => i18n.t('component.drawer.chrome.ariaLabel.close
       <slot name="footer">
         <div class="actions">
           <RcButton
-            role="secondary"
+            variant="secondary"
             size="large"
             :aria-label="ariaLabel"
             @click="emit('close')"
--- a/shell/components/Drawer/ResourceDetailDrawer/index.vue
+++ b/shell/components/Drawer/ResourceDetailDrawer/index.vue
@@ -92,7 +92,7 @@ useResourceDetailDrawerProvider();
     <template #additional-actions>
       <RcButton
         v-if="canEdit"
-        role="primary"
+        variant="primary"
         size="large"
         :aria-label="action.ariaLabel"
         :data-testid="editBttnDataTestId"


pkg/rancher-components/src/components/RcIcon/types.ts
```diff
@@ -145,16 +145,20 @@ export const RcIconType = {
...
-export const RcIconSize = {
-  large:  '25px',
-  medium: '18px',
-  small:  '14px',
-  none:   undefined
+export const RcIconSizeToCSS = {
+  large:   '25px',
+  medium:  '18px',
+  small:   '14px',
+  inherit: 'inherit'
```
codyrancher, line 152: Switched from none to inherit since I think it describes and behaves better.

pkg/rancher-components/src/components/RcIcon/types.ts
```diff
@@ -145,16 +145,20 @@ export const RcIconType = {
...
+export type RcIconSize = keyof typeof RcIconSizeToCSS;
+
+export type RcIconType = keyof typeof RcIconTypeToClass;
+
 export interface RcIconProps {
-  size: keyof typeof RcIconSize;
-  type: keyof typeof RcIconType;
+  type: RcIconType;
+  size?: RcIconSize;
   ariaHidden?: boolean;
```
codyrancher, line 162: I just wanted to clean up the sub types so they can be reused.

pkg/rancher-components/src/components/RcIcon/RcIcon.vue
```diff
@@ -1,14 +1,17 @@
 <script setup lang="ts">
-import { RcIconProps, RcIconType, RcIconSize } from '@components/RcIcon/types';
+import { RcIconSizeToCSS, RcIconTypeToClass, RcIconProps } from './types';
```
codyrancher, line 2: Because we have problems with referencing using the alias.

I want to address this separately either as separating rancher components fully or moving them into shell.

pkg/rancher-components/src/components/RcIcon/RcIcon.vue
```diff
@@ -19,13 +22,13 @@ const status = computed(() => {
...
 
-const { textColor } = useStatusColors({ status: status.value }, 'outlined');
-
 const color = computed(() => {
   if (props.status === 'inherit') {
     return 'inherit';
   }
 
+  const { textColor } = useStatusColors({ status: status.value }, 'outlined');
+
```
codyrancher: We moved this in the RcIcon PR but I brought it back because while working on this I realized the reason I moved this in was to make the color reactive to the status prop changing.

There are some other ways we could make this reactive so maybe there's another route you'd be happy with.

rak-phillip: We can update the type of `StatusObject` to `export type StatusObject = { status: Ref<Status> };` and pass pass a reactive ref instead. This approach allows the internal computed properties of `useStatusColors()` to track changes reactively and avoids recreating computed properties inside `useStatusColors()` on each eval.

pkg/rancher-components/src/components/RcButton/RcButton.vue
```diff
@@ -53,19 +96,45 @@ defineExpose({ focus });
...
 <style lang="scss" scoped>
+@mixin iconMargins($size) {
+  .left-icon {
+    margin-right: $size;
+  }
+
+  .right-icon {
+    margin-left: $size;
+  }
+}
```
rak-phillip: Relying on margins attached to the icon component forces developers to be aware of the size as well as the specifics of spacing when slots are used. Can we achieve a similar result utilizing flex gap instead?

pkg/rancher-components/src/components/RcIcon/RcIcon.vue
codyrancher: Good idea, I switched to this and it looks like we're good.

pkg/rancher-components/src/components/RcButton/RcButton.vue
codyrancher: Fair, switched to using gap.

pkg/rancher-components/src/components/RcIcon/RcIcon.vue
codyrancher: This ended up causing the e2e tests failures I was seeing. I wanted to avoid forcing consumers to make status reactive but because this has caused a couple bugs to crop up I've decided to simplify and just accept a Ref<Status>.

pkg/rancher-components/src/components/Pill/RcStatusBadge/RcStatusBadge.vue
```diff
@@ -1,10 +1,12 @@
 <script setup lang="ts">
+import { computed } from 'vue';
 import { RcStatusBadgeProps } from './types';
 import { useStatusColors } from '@components/utils/status';
 
 const props = defineProps<RcStatusBadgeProps>();
 
-const { backgroundColor, borderColor, textColor } = useStatusColors(props, 'outlined');
+const status = computed(() => props.status);
+const { backgroundColor, borderColor, textColor } = useStatusColors(status, 'outlined');
```
rak-phillip: nit: Would `toRef()` be preferable to `computed()` in this case[^1]?

```suggestion
const { backgroundColor, borderColor, textColor } = useStatusColors(toRef(() => props.status), 'outlined');'outlined');
```

Passing a ref of a prop to a composable function is explicitly stated as one of the purposes of `toRef()`. That is to say, I don't think there's an issue with the `computed()` usage in this instance.

[^1]: https://vuejs.org/api/reactivity-utilities#toref

pkg/rancher-components/src/components/Pill/RcStatusIndicator/RcStatusIndicator.vue
```diff
@@ -1,10 +1,12 @@
 <script setup lang="ts">
+import { computed } from 'vue';
 import { RcStatusIndicatorProps } from './types';
 import { useStatusColors } from '@components/utils/status';
 
 const props = defineProps<RcStatusIndicatorProps>();
 
-const { backgroundColor, borderColor } = useStatusColors(props, 'solid');
+const status = computed(() => props.status);
+const { backgroundColor, borderColor } = useStatusColors(status, 'solid');
```
rak-phillip, line 9: nit: same comment as https://github.com/rancher/dashboard/pull/16023/changes#r2683660841

rak-phillip (commented):
I only have a minor nitpick at this point.

pkg/rancher-components/src/components/Pill/RcStatusIndicator/RcStatusIndicator.vue
codyrancher: Fixed both

pkg/rancher-components/src/components/RcButton/RcButton.vue
```diff
@@ -5,36 +5,78 @@
...
+  IconProps
+} from './types';
+import RcIcon from '@components/RcIcon/RcIcon.vue';
+
+const buttonVariants: { variant: keyof ButtonVariantProps, className: string }[] = [
+  { variant: 'primary', className: 'variant-primary' },
+  { variant: 'secondary', className: 'variant-secondary' },
+  { variant: 'tertiary', className: 'variant-tertiary' },
+  { variant: 'link', className: 'variant-link' },
+  { variant: 'multiAction', className: 'variant-multi-action' },
```
rak-phillip, line 22: NOTE: I think that we need to track this multiAction variant and plan for its removal. If I recall, this is used in a select few places and I don't think this is identified in our design system.

shell/components/Drawer/ResourceDetailDrawer/index.vue
```diff
@@ -92,7 +92,8 @@ useResourceDetailDrawerProvider();
     <template #additional-actions>
       <RcButton
         v-if="canEdit"
-        :primary="true"
+        role="primary"
```
rak-phillip: Should this be `variant` now?

```suggestion
        variant="primary"
```

shell/components/Drawer/Chrome.vue
```diff
@@ -37,13 +38,14 @@ const ariaLabel = computed(() => i18n.t('component.drawer.chrome.ariaLabel.close
     <div class="footer pp-4">
       <slot name="footer">
         <div class="actions">
-          <button
-            class="btn role-secondary"
+          <RcButton
+            role="secondary"
```
rak-phillip: Should this be `variant`?

```suggestion
            variant="secondary"
```

codyrancher: Must've missed, fixed in my update.

shell/components/Drawer/ResourceDetailDrawer/index.vue
codyrancher: Must've missed, fixed in my update.

pkg/rancher-components/src/components/RcButton/RcButton.vue
codyrancher: Added a deprecation message and created an issue.

https://github.com/rancher/dashboard/issues/16450

PR #16120, opened 2025-12-08
Map the csp, federal, and rgs themes to suse

### LINKED ISSUE #15794: Alias additional themes to 'suse'
Additional themes like `csp`, `rgs`, & `federal` can be aliased so that they just use the `suse` them when active.

rak-phillip:
### Summary
This maps the csp, federal, and rgs themes to suse by introducing a new `brand` getter. 

Fixes #15794

- Map csp and federal brands to suse
- Preserve brand logos 
- Add consts for brands

### Technical notes summary

The mapping of the brand is performed through a new `brand` getter that will translate "csp", "federal", and "rgs" brands to "suse". 

This also aims to preserve the included logos for the affected themes, if one exists, by adding a special case to fetch the original logo for `rancher-logo.svg` in `shell/components/BrandImage.vue`

Updating the theme and assert the following produce the expected result

- Banners
- Logos
- Color scheme
- Login image
- Favicon

Unknown edge cases might be used throughout Dashboard where custom behavior or styles are applied based on the brand. 

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/plugins/dashboard-store/getters.js
+++ b/shell/plugins/dashboard-store/getters.js
@@ -210,11 +210,15 @@ export default {
   brand: (state, getters) => {
     const brand = getters['byId'](MANAGEMENT.SETTING, SETTING.BRAND);
 
-    if ([BRAND.CSP, BRAND.FEDERAL, BRAND.RGS].includes(brand?.value)) {
+    if (!brand?.value) {
+      return undefined;
+    }
+
+    if ([BRAND.CSP, BRAND.FEDERAL, BRAND.RGS].includes(brand.value)) {
       return BRAND.SUSE;
     }
 
-    return brand?.value;
+    return brand.value;
   },
 
   /**


shell/plugins/dashboard-store/getters.js
```diff
@@ -204,6 +207,17 @@ export default {
     }
   },
 
+  brand: (_state, getters) => {
+    const brand = getters['all'](MANAGEMENT.SETTING)
```
richard-cox:
```suggestion
    const brand = getters['byId'](MANAGEMENT.SETTING, SETTING.BRAND)
```

shell/plugins/dashboard-store/getters.js
```diff
@@ -204,6 +207,17 @@ export default {
     }
   },
 
+  brand: (_state, getters) => {
```
richard-cox:
```suggestion
  brand: (state, getters) => {
```

shell/plugins/dashboard-store/getters.js
```diff
@@ -204,6 +207,17 @@ export default {
...
+    const brand = getters['all'](MANAGEMENT.SETTING)
+      ?.find((setting) => setting.id === SETTING.BRAND);
+
+    if ([BRAND.CSP, BRAND.FEDERAL, BRAND.RGS].includes(brand?.value)) {
```
richard-cox: Comment only.

I don't think this is called often (so clarity / readability wins here) but if it were would suggest two changes

1. Move the array into a const somewhere (avoids recreating the array every time)
2. exit early on !brand.value (it won't ever exist in the array, no need to search)

shell/utils/favicon.js
```diff
@@ -9,17 +9,15 @@ export function haveSetFavIcon() {
...
   const link = findIconLink(document.head.getElementsByTagName('link'));
 
   if (link) {
     let brandImage;
 
-    if (brandSetting?.value === 'suse') {
+    if (brandSetting === 'suse') {
       brandImage = require('~shell/assets/brand/suse/favicon.png');
-    } else if (brandSetting?.value === 'csp') {
-      brandImage = require('~shell/assets/brand/csp/favicon.png');
```
richard-cox, line 21: shall we remove this now, or create an issue to remove unused brand assets later?

shell/components/BrandImage.vue
```diff
@@ -78,19 +78,30 @@ export default {
...
+        if (this.isDark && this.uiLogoDark) {
           return this.uiLogoDark;
         }
 
         if (this.uiLogoLight) {
           return this.uiLogoLight;
         }
+
+        // csp, rgs, and federal map to SUSE, but have their own custom logos
+        if (this.brandBase !== this.brand) {
```
richard-cox, line 96: this could be folded into line 123

if (!this.brand || this.brandBase !== this.brand) {

rak-phillip: hmm.. moving this into line 123 removes the edge case that we are trying to handle for the rancher logo and will apply to both banner and login landscape as well.

shell/plugins/dashboard-store/getters.js
rak-phillip: Do we just simply not prefer the underscore prefix? I find it useful to communicate when a given parameter is intentionally left unused. 

Either way, I've accepted the change to move this along.

shell/utils/favicon.js
rak-phillip: Let's defer to later. I have an in progress issue created under #16143

shell/plugins/dashboard-store/getters.js
rak-phillip: With the first point, I would prefer to favor clarity/readability. We're only dealing with 3 members in the array and any improvements that we can make to impact time or space complexity will be unrealized unless this list becomes arbitrarily large; I don't anticipate that this array will ever exceed these 3 members.

I have fixed point 2.

rak-phillip:
> With this approach the favcicon and hasStylesheet parts are lost for csp. I think favicon is fine (they're both very similar) but not sure on hasStylesheet. looking at what it does... it just adds the brand name csp as a style to the body. we could risk it and see what breaks, will leave that one with you.

I think that this is fine. The csp stylesheet is just overriding some color variables and supplying a few of its own, which we don't want if it is to match the suse theme. FYI @nwmac to confirm.

shell/plugins/dashboard-store/getters.js
richard-cox: Just a pattern that we haven't started with or used in other places.

shell/components/BrandImage.vue
richard-cox: ahh sorry was rushing, didn't spot the three if's for three images.. and the fall back is line 123.

and we're not bothering with banner.svg and login-landscape.svg given none of rsg, federal or csp brands contain them.

PR #16139, opened 2025-12-10
Improve type generation of the i18n plugin

rak-phillip:
### Summary
This updates the Rancher Shell type generation to include the i18n plugin.

- Add the i18n plugin to the typegen script
- Annotate `stringFor()` with a dostring to include types
- Fix typescript warnings in the `t()` function of `useI18n`

### Technical notes summary

Kubewarden UI will fail with the following error when updating to the latest versions of `@rancher/shell`:

```
  ERROR in node_modules/@rancher/shell/composables/useI18n.ts:3:27
  TS7016: Could not find a declaration file for module '@shell/plugins/i18n'.
  '/home/phillip/Development/kubewarden-ui/node_modules/@rancher/shell/plugins/i18n.j
  s' implicitly has an 'any' type.
    If the '@rancher/shell' package actually exposes this module, try adding a new
  declaration (.d.ts) file containing `declare module '@shell/plugins/i18n';`
      1 | import { Store } from 'vuex';
      2 |
    > 3 | import { stringFor } from '@shell/plugins/i18n';
        |                           ^^^^^^^^^^^^^^^^^^^^^
      4 |
      5 | let store: Store<any> | null = null;
      6 | /**
```

This issue appears to be isolated to Kubewarden, but the root cause can be linked to kubewarden depending on later versions of typescript than shell and other extensions, so this failure could be linked to better type checking overall. 

There's also a risk that we will encounter these issues for other extensions in the future as we update dependencies in shell.

After updating the types for the i18n plugin, we received a new typescript warning:

```
ERROR in node_modules/@rancher/shell/composables/useI18n.ts:14:20
  TS2345: Argument of type 'Store<any> | null' is not assignable to parameter of type
  'Store<any>'.
    Type 'null' is not assignable to type 'Store<any>'.
      12 |  */
      13 | const t = (key: string, args?: unknown, raw?: boolean): string => {
    > 14 |   return stringFor(store, key, args, raw);
         |                    ^^^^^
      15 | };
      16 |
      17 | export type I18n = { t: typeof t };
```

I added some guards to ensure that the store is available before invoking `stringFor()`.

Publish an updated versions of shell to a local registry or link the package. Kubewarden should no longer show type errors. 

We will want to test with a few other extensions, but that should be covered by CI.

This adds better type support for the i18n plugin. Other extensions might rely on ill defined types and will need to updated accordingly to support the change. 

### Screenshot/Video

NA

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/composables/useI18n.ts
+++ b/shell/composables/useI18n.ts
@@ -11,7 +11,16 @@ let store: Store<any> | null = null;
  * @returns A translated string or the raw value if the raw parameter is set to true.
  */
 const t = (key: string, args?: unknown, raw?: boolean): string => {
-  return stringFor(store, key, args, raw);
+  if (!store) {
+    if (!!process.env.dev) {
+      // eslint-disable-next-line no-console
+      console.warn('useI18n: store not available');
+    }
+
+    return key;
+  }
+
+  return stringFor(store, key, args as any, raw);
 };
 
 export type I18n = { t: typeof t };
--- a/shell/package.json
+++ b/shell/package.json
@@ -6,6 +6,7 @@
   "license": "Apache-2.0",
   "author": "SUSE",
   "private": false,
+  "types": "types/shell/index.d.ts",
   "engines": {
     "node": ">=20.0.0"
   },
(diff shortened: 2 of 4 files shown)


cnotv:
We need to bind the use of dependencies better here.
I predicted this issue to happen back in May, looks like it took longer than expected to happen 😄 
https://github.com/rancher/kubewarden-ui/issues/1186

cnotv:
Can confirm issue solved, thanks 🥂

PR #16268, opened 2025-12-23
Remove `bg-` prefix from the state background

### LINKED ISSUE #16267: Console Error: TypeError: can't access property "primary", statusColors.value is undefined
**Setup**
- Rancher version: v2.14-head
- Rancher UI Extensions: NA
- Browser type & version: Firefox & Chrome

**Describe the bug**

And error is printed to the console when navigating the the deployments page:

```
TypeError: can't access property "primary", statusColors.value is undefined
    backgroundColor status.ts:61
    refreshComputed reactivity.esm-bundler.js:391
    get value reactivity.esm-bundler.js:1648
    unref reactivity.esm-bundler.js:1493
    setup RcStatusIndicator.vue:16
```

**To Reproduce**

1. Create a deployment and wait for it to enter an active state
2. Navigate to the deployment details page

**Result**

An error is printed to the console. The status indicator is hollow:

Image

**Expected Result**

No error printed to the console. The status indicator is success:

Image

**Additional context**

It appears that the `stateBackground` is prefixed with a `bg-`

https://github.com/rancher/dashboard/blob/bac4bc16aa0222115660a0b0d9b90cbb6712766b/shell/plugins/dashboard-store/resource-class.js#L758-L760

While the status indicator expects no prefix:

https://github.com/rancher/dashboard/blob/bac4bc16aa0222115660a0b0d9b90cbb6712766b/pkg/rancher-components/src/components/utils/status.ts#L3-L28

rak-phillip:
### Summary
This resolves a console error by removing the `bg-` prefix from the state background.

Fixes #16267 

Remove `bg-` prefix from the state background

### Technical notes summary

It appears that the `stateBackground` is prefixed with a `bg-`

https://github.com/rancher/dashboard/blob/bac4bc16aa0222115660a0b0d9b90cbb6712766b/shell/plugins/dashboard-store/resource-class.js#L758-L760

While the status indicator expects no prefix:

https://github.com/rancher/dashboard/blob/bac4bc16aa0222115660a0b0d9b90cbb6712766b/pkg/rancher-components/src/components/utils/status.ts#L3-L28

See issue - there should be no error printed to the browser console when navigating to a deployment detail page. The status indicator should be correct.

Any valid status that contains `bg-` will have the matching string stripped.

### Screenshot/Video

Image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/Resource/Detail/ResourcePopover/index.vue
+++ b/shell/components/Resource/Detail/ResourcePopover/index.vue
@@ -32,7 +32,7 @@ const fetch = useFetch(async() => {
 });
 
 const stateBackground = computed(() => {
-  return fetch.value.data?.stateBackground?.replace('bg-', '') || 'unknown';
+  return fetch.value.data?.stateSimpleColor || 'unknown';
 });
 
 const resourceTypeLabel = computed(() => {


rak-phillip:
> I've noticed we use bg- in quite a few places in the code. Would it make more sense to rename statusDefinitions instead to avoid this happening in the future as we migrate more components?

That's a good point. I'll defer to @codyrancher to see if he has any feedback on if we want to modify `statusDefinitions`, as he was the original author.

codyrancher:
I'm not a big fan of adding the 'bg' prefix to the StatusDefinitions. 

I see how it would make migrating easier but I'd prefer to make the new components as free from our legacy cruft as we can. I don't think it makes sense for a status definition to have any mention of `background`. 

One alternative is to use the [stateSimpleColor](https://github.com/rancher/dashboard/blob/master/shell/plugins/dashboard-store/resource-class.js#L754-L756) which should help bridge the gap from the old to the new.

rak-phillip:
@eva-vashkevich @codyrancher this has been updated to utilize `stateSimpleColor` and is ready for another review.

PR #16294, opened 2026-01-02
Improved warning for imported clusters with externally upgraded version

### LINKED ISSUE #14401: Improve Warning Message on Edit Page for Externally Upgraded Imported RKE2/K3s Clusters

This issue is for requesting an enhancement of an existing warning message 

If an imported RKE2/k3s cluster is upgraded externally, the following warning message shows up in the cluster editing page: "The selected version is lower than the actual version of this cluster".  See screenshot below.

Image

To reduce confusion and improve clarity, it's recommended to enhance the warning message by including the cluster's actual version. For example:

```
The selected version below is lower than the actual version (<value from status.version.gitVersion>) of this cluster. 
Please consider disabling the version management feature if the cluster has been upgraded externally. 
```

Recommendations for a clearer message are welcome. 

SURE-7219

eva-vashkevich:
### Summary
Fixes #14401

Our k8s version dropdown was showing a version from config.kubernetesVersion, which was incorrect for externally upgraded clusters. Because of the external upgrade, our config version and the real version (value.version.gitVersion) were different. This lead to user confusion. Even though this is not a supported use case, we should be showing correct version of the cluster, not the version Rancher thinks it should be.

Updated the "Kubernetes version" dropdown to include the version from value.version.gitVersion if it doesn't match config.kubernetesVersion
Updated banner message

- Import a cluster (making 2 versions lower than latest)
- Upgrade it via Rancher, make sure no warning is shown and the versions are shown correctly.
- Upgrade it manually. Make sure new warning is shown. Make sure that dropdown indicates that this version was added manually. Make sure that the manual version doesn't disappear from the dropdown when version is changed (make sure user can undo the changes)

Shouldn't be any

### Screenshot/Video

Screenshot 2026-01-06 at 3 27 59 PM

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/pkg/imported/l10n/en-us.yaml
+++ b/pkg/imported/l10n/en-us.yaml
@@ -5,7 +5,7 @@ imported:
   agentEnv:
     header: Agent Environment Variables
   basics:
-    versionMismatch: "You have upgraded the cluster manually. Managing the cluster version via Rancher and external means at the same time is not supported and may cause unexpected behavior and failures. <br>Please consider disabling the version management feature if the cluster has been upgraded externally."
+    versionMismatch: "There is a mismatch between the expected and the current versions of the cluster. This usually happens if the imported cluster was upgraded outside of this Rancher instance. Managing the cluster version via Rancher and external means at the same time is not supported and may cause unexpected behavior and failures. <br>Please consider disabling the version management feature if the cluster has been upgraded externally."
     versionManagement:
       title: K8s version management
       globalEnabled: Global default (enabled)


pkg/imported/l10n/en-us.yaml
```diff
@@ -5,7 +5,7 @@ imported:
   agentEnv:
     header: Agent Environment Variables
   basics:
-    versionMismatch: The selected version is lower than the actual version of this cluster.
+    versionMismatch: "You have upgraded the cluster manually. Managing the cluster version via Rancher and external means at the same time is not supported and may cause unexpected behavior and failures. <br>Please consider disabling the version management feature if the cluster has been upgraded externally."
```
rak-phillip: nit: The message begins with an assertion that the user upgraded the cluster manually and it ends with less certainty about what has happened ("if the cluster has been upgraded externally"). In my opinion, this change in tone is quite jarring. 

nit: We lose some context about why the warning is displayed with the update message: the actual version doesn't match the expected version. Are there scenarios where this is warning can be raised without external manipulation of versions?

Finally, this part of the message doesn't read clearly. I think that we need to give this another pass.

> Managing the cluster version via Rancher and external means at the same time is not supported

pkg/imported/components/Basics.vue
```diff
@@ -127,9 +133,6 @@ export default defineComponent({
     },
     versionInformationDisabled() {
       return this.versionManagement === VERSION_MANAGEMENT_DEFAULT && !this.versionManagementGlobalSetting;
-    },
-    versionMismatch() {
-      return compare(this.config.kubernetesVersion, this.value.version.gitVersion) < 0;
```
rak-phillip, line 132: What is the motivation for moving `versionMismatch` out of the computed properties and into the data prop, was there a change in behavior with the added conditional that required moving it to the `created()` hook?

rak-phillip (commented):
I think that we need to improve the warning message before merging.

pkg/imported/components/Basics.vue
eva-vashkevich: before the banner wasn't persisted when the selection changed, this ensures the original discrepancy is persisted

PR #16306, opened 2026-01-05
Display warnings for Azure pools that use unmanaged disks

### LINKED ISSUE #16288: Azure Provisioning Form: default "Use managed disks" based on existing node pools to avoid availability set mismatch
When editing an existing Azure cluster that was created with unmanaged disks, the node pool form in the UI still defaults "Use managed disks" to `true` when adding a new node pool.

If the cluster uses an availability set with unmanaged disks, Azure does not allow mixing managed and unmanaged disks in the same availability set, and node creation can fail with an error like:

`error creating machine: error in driver during machine creation: cannot convert non-managed availability set jferraz-test-as to managed availability set`

When updating or editing an existing Azure cluster, the default value of "Use managed disks" in the node pool form should match the current cluster / node pool configuration.

For example:

* If the existing node pool(s) are using unmanaged disks, then when the user clicks "Add node pool", the "Use managed disks" option should default to `false`.
* If the existing node pool(s) are using managed disks, then the option can stay `true` by default.

This will help avoid mixing disk types inside the same availability set and prevent node creation errors.

**Additional context**
This is a follow up to [https://github.com/rancher/dashboard/issues/16184](https://github.com/rancher/dashboard/issues/16184), which changed the default to use managed disks.

The issue appears when:

1. A cluster is created on Azure using an availability set with unmanaged disks.
2. Later, in the UI, the user clicks "Add node pool" for this cluster.
3. The "Use managed disks" option is `true` by default and the user keeps this value.
4. Node creation fails with the error about converting a non-managed availability set to a managed availability set.

rak-phillip:
### Summary
This adds deprecation warnings to the RKE2 Azure form when the "Unmanaged" disk option is used.

Fixes #16288

- Remove `unmanaged` from the availability set label
- Add an Unmanaged Disks deprecation banner
- Add a prominent warning when the mode is edit

### Technical notes summary

There has been a lot of discussion on how to best handle this deprecation. We have opted to perform the following:

- Keep the "Use Managed Disks" checkbox defaulted to true
- Display a deprecation warning when this checkbox is false

See https://github.com/rancher/dashboard/issues/16288#issuecomment-3747017131

See the issue for more details.

1. Creating a new Azure rke2 cluster
   2. Change the default value for managed disks
   3. Assert that the warning banner is rendered

Regression risk is very low. This includes banners to provide more context about a change that is coming in Azure.

### Screenshot/Video

image

image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/machine-config/azure.vue
+++ b/shell/machine-config/azure.vue
@@ -209,6 +209,10 @@ export default {
     }
   },
 
+  setup() {
+    return { _EDIT };
+  },
+
   data() {
     return {
       azureEnvironments,
@@ -518,6 +522,11 @@ export default {
     </div>
   </div>
   <div v-else>
+    <Banner
+      v-if="mode === _EDIT && !value.managedDisks"
+      color="warning"
+      :label="t('cluster.machineConfig.azure.managedDisks.deprecationWarning', {}, true)"
+    />
     <div class="row mt-20">
       <div class="col span-6">
         <LabeledSelect


rak-phillip:
I had an oob discussion with @oboc-sts to review the validation implementation. We arrived at the conclusion that the preferred approach is to implement a pattern that will keep the pools in sync for the user. We will accomplish this by:

- adding an informational tooltip that explains how this values needs to be the same for all pools to the managed disks checkbox
- adding a modal for the user to confirm taking action on updating all of the pools if managed disks changes

rak-phillip:
Converting to draft as we gather more details on the implementation: https://github.com/rancher/dashboard/issues/16288#issuecomment-3730577920

shell/machine-config/azure.vue
```diff
@@ -518,6 +518,11 @@ export default {
     </div>
   </div>
   <div v-else>
+    <Banner
+      v-if="mode === _EDIT && !value.managedDisks"
```
mantis-toboggan-md, line 526: This banner never shows up for me. It looks like `_EDIT` is not defined in the template context; it works when I add a computed property

`_EDIT: () => _EDIT,`

mantis-toboggan-md (changes requested):
There is a minor issue with the edit banner

shell/machine-config/azure.vue
rak-phillip: Indeed, I'm much too comfortable with the affordances that come with the composition api. Rather than expose the `_EDIT` constant via a computed prop or data, I opted to expose it via the setup hook; this will ensure that it remains a constant, without any reactivity attached.

mantis-toboggan-md (approved):

PR #16318, opened 2026-01-06
normalize whitespace-only CodeMirror input

### LINKED ISSUE #16317: When creating a configmap in GUI a space is added to a empty value
**Setup**
- Rancher version: 2.13.1, master
- Rancher UI Extensions:  NA
- Browser type & version: Any

**Describe the bug**

When creating a config map if you have entered anything in the GUI's "value" field and then deleted it the subsequent yaml will contain a space.

**To Reproduce**

- Create a new configmap in any namespace.
- While creating add a key "test" and add something to the value field and then delete it, this results in a dot character appearing. 
- Delete again to remove the dot
- Save.
- Edit the yaml to see a space has been added between the quotes ' ' 

**Result**
Space added to the value

**Expected Result**

no space added to the value

**Screenshots**

**Additional context**

SURE-11126

skanakal:
### Summary
Fixes https://github.com/rancher/dashboard/issues/16317

validate the input at the CodeMirror component

### Technical notes summary

The CodeMirror.vue component’s onInput() handler forwarded editor output directly..
Whitespace-only values were not detected.

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/CodeMirror.vue
+++ b/shell/components/CodeMirror.vue
@@ -283,7 +283,7 @@ export default {
         :disabled="isDisabled"
         :original-style="true"
         @ready="onReady"
-        @input="onInput"
+        @update:value="onInput"
         @changes="onChanges"
         @focus="onFocus"
         @blur="onBlur"


shell/components/CodeMirror.vue
```diff
@@ -219,6 +219,9 @@ export default {
     },
 
     onInput(newCode) {
+      if (typeof newCode === 'string' && newCode.trim() === '') {
```
richard-cox: how do we know that users will never need a whitespace value?

richard-cox (changes requested):
what do the codemirror docs say, anything in their repo issues about this? if a bug is it in their implementation or our integration?

shell/components/CodeMirror.vue
skanakal: Initially, I assumed Kubernetes treated whitespaces values as empty and added trimming logic based on that assumption but that wasn't for configmap... 
It doesn't look like a bug to me, deleting the last character often leaves a "ghost space" in the editor buffer...our CSS renders a "dot" for every space, that ghost space creates a visual bug...
- doc.getValue() to get the content and doc.setValue(content: string) to set it to empty which strips the .cm-markdown-single-trailing-space class from the DOM, which causes the "dot" to disappear...

richard-cox: The question revolved around this being a generic component to display text where a blank space might be expected or required. So i think we can move away from kube specific requirements.

Something is adding that white space when it shouldn't do, feels like the solution should be to address that part rather than working around it.

skanakal: @richard-cox I instrumented the CodeMirror instance to log the raw document value and character codes directly from cm.getValue().

After typing text(In the value Tab) and deleting all characters, the editor does not return an empty string (""). Instead, it consistently returns:

```
CM INTERNAL VALUE: " " [32]
RAW FROM CODEMIRROR: { stringified: '" "', length: 1, charCodes: [32] }
```

What I have verified so far, 

- The value was logged directly from CodeMirror (not from Vue state).
- The space appears immediately after deleting the final character.
- This behavior was reproduced across multiple test cases(like changing the mode, wrapperline,etc...)
- No additional transformation logic is introducing the space on our side...

What my hypothesis since start but couldn't confirm;

It appears that CM 5 may be normalizing an empty document to a single space character, possibly due to its internal line model, which maintains at least one line for cursor stability...

As a result, The editor never reaches a truly empty document state....

Could someone please investigate further to confirm on 

Whether this is expected CM5 behavior?

richard-cox: The goal would be to stop code mirror adding the space to the value we receive. Hopefully via config. If it's a bug their side then a fix their side. Failing that we'd need an alternative solution like a very targeted version of this PR

votdev: The bug is caused by https://github.com/rennzhang/codemirror-editor-vue3/blob/ce80d1a618e41bbfaabb7e9d2a68fefeed5c5c18/packages/src/hooks/useEvents.ts#L60

a110605: If the bug is caused by upstream [codemirror-editor](https://github.com/rennzhang/codemirror-editor-vue3), we should report to package author and fix there.

skanakal: @richard-cox @a110605 

The codemirror-editor-vue3 wrapper's internal change listener ([useEvents.ts](https://github.com/rennzhang/codemirror-editor-vue3/blob/ce80d1a618e41bbfaabb7e9d2a68fefeed5c5c18/packages/src/hooks/useEvents.ts#L60)) fires and emits three separate Vue events from the same code path, all derived from that same "":

```
emit("update:value", content.value || "")   // "" || "" → ""
emit("input",        content.value || " ")  // "" || " " → " "   ← the bug
emit("change",       content.value, cm)      // "" (no fallback)
```

Currently, CodeMirror.vue listened on @input, so onInput received " " and re-emitted 'onInput' with " " up to the ConfigMap form. That space got persisted, producing value: ' ' in the saved YAML instead of value: ''.

With the [fix](https://github.com/rancher/dashboard/pull/16318/changes#diff-056160ac018ea8f019405781fd9f0c4e7e73c7007fcbeb3c41ea87286e35e03fR286) , CodeMirror.vue listens on @update:value, which fires from the identical guarded block, at the identical moment, with the identical payload shape ((value: string)), just without the || " " fallback bug. So onInput now receives "", and the correct empty value flows all the way up.  Henec use the `update:value`, not input: codemirror-editor-vue3's input event turns an empty doc into a space and update doesn't...

@votdev Thank you for the pointers 


https://github.com/user-attachments/assets/84d49101-7f1a-4eaa-beb9-a57ce026fc13

richard-cox:
@rak-phillip / @marcelofukumoto what are the 2.16 plans for codemirror? are we bumping the version, replacing or neithe (yet)?

rak-phillip:
@richard-cox I understand that the current plan is to resume work on #17201 and move forward with bumping codemirror to the latest version. The codemirror component that we use will be integrated into Rancher Components.

richard-cox (approved):

PR #16334, opened 2026-01-07
Added the ability to remove namespaces from projects entirely, not just move them between projects.

### LINKED ISSUE #11128: Move out Namespaces from Projects

It should be possible to move out a namespace from the related project.

Currently, it's only possible to move a namespace from a project to another one.

image

A namespace can be removed from a project by deleting the fields below:

- label `field.cattle.io/projectId`
- annotation `field.cattle.io/projectId`

**Additional context**

codyrancher:
### Summary
Fixes #11128

Added the ability to remove namespaces from projects entirely, not just move them between projects. Previously, the "Move" action on namespaces only allowed moving to another project. Now users can select "None (remove from project)" to disassociate a namespace from any project.

- Added "None (remove from project)" option to the target project dropdown in the Move Namespace dialog
- When "None" is selected, the `field.cattle.io/projectId` label and annotation are cleared from the namespace
- Added unit tests for the new functionality

1. Select a namespace that is currently in a project, click "Move", verify "None (remove from project)" appears as the first option and is selected by default
2. Click "Move" with "None" selected - verify namespace is removed from the project (no longer appears under that project in the UI)
3. Select a different project from the dropdown and click "Move" - verify namespace moves to the new project (existing functionality still works)
4. Test bulk selection: select multiple namespaces from the same project, click "Move", select "None", verify all are removed from the project

- Moving namespaces between projects (existing functionality)
- Bulk move operations for namespaces
- Namespace list filtering by project (namespaces removed from projects should appear in "Not in a Project" group)
- Project resource counts (should decrease when namespace is removed)

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -4620,7 +4620,7 @@ moveModal:
   description: 'You are moving the following namespaces:'
   moveButtonLabel: Move
   targetProject: Target Project
-  noProject: None (remove from project)
+  noProject: None (remove from any project)
 
 nameNsDescription:
   name:
--- a/shell/dialog/MoveNamespaceDialog.vue
+++ b/shell/dialog/MoveNamespaceDialog.vue
@@ -6,6 +6,8 @@ import LabeledSelect from '@shell/components/form/LabeledSelect';
 import { MANAGEMENT } from '@shell/config/types';
 import { PROJECT } from '@shell/config/labels-annotations';
 
+const NONE_VALUE = ' ';
+
 export default {
   emits: ['close'],
 
@@ -37,7 +39,7 @@ export default {
     return {
       projects:      [],
       toMove:        [],
-      targetProject: ''
+      targetProject: null
     };
   },
 
@@ -46,6 +48,10 @@ export default {
 
     excludedProjects() {
       return this.toMove.filter((namespace) => !!namespace.project).map((namespace) => namespace.project.shortId);
+    },
+
+    isAllInProject() {
+      return this.toMove.every((namespace) => !!namespace.project);
     },
 
     projectOptions() {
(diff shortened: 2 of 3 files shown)


shell/assets/translations/en-us.yaml
```diff
@@ -4620,6 +4620,7 @@ moveModal:
   description: 'You are moving the following namespaces:'
   moveButtonLabel: Move
   targetProject: Target Project
+  noProject: None (remove from project)
```
eva-vashkevich: I think 'None (remove from any project)' would be better? 
If my ns is already not associated with any project, this is still the first dropdown choice, and it feels misleading

eva-vashkevich (commented):
I think bulk move is broken. It seems to only apply to the first ns being moved.
ie if I have one ns without a project and one in a project, bulk moving them to none doesn't work. I don't think it was caused by this change, but would be good to fix it. Or we can open a new issue.

https://github.com/user-attachments/assets/c7994ed8-d1b9-4631-be08-fabe7ae5dfd8


https://github.com/user-attachments/assets/e9e29b0e-d880-4031-83b9-81105f865728

shell/dialog/MoveNamespaceDialog.vue
```diff
@@ -59,6 +59,14 @@ export default {
 
         return inCluster;
       }, []);
+
+      // Add "None" option to allow removing namespace from project
+      options.unshift({
```
eva-vashkevich: if currently we do not display current project when moving, should we hide it in the case of moving namespace without a project?

shell/assets/translations/en-us.yaml
codyrancher: Updated the text and moved the default value back to null to force the user to choose just like in the `master` version.

shell/dialog/MoveNamespaceDialog.vue
codyrancher: Updated to be consistent with the projects where we only provide the option if all the of the namespaces have a project.

eva-vashkevich (approved):
lgtm

PR #16341, opened 2026-01-08
Refactor ActionMenuShell to emit single action-invoked event

### LINKED ISSUE #13553: Vue warnings in extensions action menu

codyrancher:
## Summary
We wanted to refactor ActionMenuShell emits so we can avoid warnings caused by emitting new messages that weren't in the compile time template defintion of defineEmits().

Fixes #13553

## Occurred Changes/Fixed Issues
- Refactored `ActionMenuShell.vue` to emit a single `action-invoked` event with an `ActionMenuSelection` payload instead of dynamic events
- Updated the consumers of this interface.

## Technical Notes
- `ActionMenuSelection` interface provides: `action` (string name), `actionData` (full action object), `event` (MouseEvent), `route`
- The `action-invoked` event now fires for all action types (Vuex, custom, extension invoke)
- Existing consumers using `@action-invoked` without payload (e.g., `close`) continue to work since payload is optional

## Areas for Testing
- Extensions page: Test menu actions (Manage Repos, Add Rancher Repos, Manage Extension View, Developer Load)
- OIDC Client detail page: Test regenerate/remove secret actions on client secret cards
- Any component using RcItemCard with actions prop

## Regression Risk Areas
- ActionMenu behavior in SortableTable rows
- ResourcePopover action menu
- Any extension using ActionMenuShell with customActions

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/pkg/rancher-components/src/components/RcItemCard/RcItemCard.vue
+++ b/pkg/rancher-components/src/components/RcItemCard/RcItemCard.vue
@@ -82,8 +82,8 @@ interface RcItemCardProps {
 
   /** Optional actions that will be displayed inside an action-menu
    *
-   * Each action should include an `action` name, which is emitted as a custom event when selected.
-   * To respond to the event, you must also register a matching event listener using the `@` syntax.
+   * Each action should include an `action` name, which is included in the payload
+   * of the 'action-invoked' event when the action is selected.
    *
    * Example:
    * <rc-item-card
@@ -94,8 +94,10 @@ interface RcItemCardProps {
    *       enabled: true
    *     }
    *   ]"
-   *   @focusSearch="focusSearch"
+   *   @action-invoked="handleActionInvoked"
    * />
+   *
+   * The handler receives a payload with { action: string, actionData: object, event: MouseEvent, ... }
    */
   actions?: DropdownOption[];
 
@@ -117,23 +119,9 @@ const props = defineProps<RcItemCardProps>();
 /**
  * Emits:
  * - 'card-click' when card is clicked or activated via keyboard.
- * - custom events defined in the `actions` prop, but only if the corresponding event listener is explicitly declared on the component.
+ * - 'action-invoked' when an action is selected from the action menu (only when `actions` prop is used).
  */
-const emit = defineEmits<{(e: 'card-click', value: ItemValue): void; (e: string, payload: unknown): void;}>();
-
-const actionListeners = computed(() => {
-  if (!props.actions) return {};
-
-  const listeners: Record<string, (payload: unknown) => void> = {};
-
-  for (const a of props.actions) {
-    if (a.action) {
-      listeners[a.action] = (payload: unknown) => emit(a.action, payload);
-    }
-  }
-
-  return listeners;
-});
+const emit = defineEmits<{(e: 'card-click', value?: ItemValue): void; (e: 'action-invoked', payload: unknown): void;}>();
 
 /**
  * Handles the card click while avoiding nested interactive elements
@@ -282,7 +270,7 @@ const cardMeta = computed(() => ({
                 <ActionMenu
                   data-testid="item-card-header-action-menu"
                   :custom-actions="actions"
-                  v-on="actionListeners"
+                  @action-invoked="(payload) => emit('action-invoked', payload)"
                 />
               </rc-item-card-action>
             </template>
--- a/shell/components/ActionMenu.vue
+++ b/shell/components/ActionMenu.vue
@@ -11,7 +11,7 @@ const SHOW = 'show';
 export default {
   name: 'ActionMenu',
 
-  emits: ['close'],
+  emits: ['close', 'action-invoked'],
 
   components: { IconOrSvg },
   props:      {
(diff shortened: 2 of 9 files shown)


pkg/rancher-components/src/components/RcItemCard/RcItemCard.vue
```diff
@@ -117,23 +119,9 @@ const props = defineProps<RcItemCardProps>();
...
-  const listeners: Record<string, (payload: unknown) => void> = {};
-
-  for (const a of props.actions) {
-    if (a.action) {
-      listeners[a.action] = (payload: unknown) => emit(a.action, payload);
-    }
-  }
-
-  return listeners;
-});
```
codyrancher, line 136: I debated whether or not we want to keep this and deprecate the interface slowly because of extensions.

I went and looked at harvester, kubewarden and elemental extensions and I didn't find any uses of `:actions` or `:custom-actions` so I decided it was safe to make the change instead of deprecating.

shell/components/ActionMenuShell.vue
```diff
@@ -56,24 +72,7 @@ const execute = (action: any, event: MouseEvent, args?: any) => {
...
-      action.action,
-      {
-        action,
-        event,
-        ...args,
-        route,
-      }
-    );
-  } else {
+  } else if (!props.customActions) {
```
codyrancher, line 75: We no longer use the dynamic emit but we had a default `else` before that should only execute if there aren't any custom logic to maintain the original code paths.

shell/components/ActionMenuShell.vue
rak-phillip, line 1: We might also be able to drop the `inheritAttrs: false` that was recently introduced to suppress the warnings:

https://github.com/rancher/dashboard/pull/15526/changes#diff-ec172234db7799c8195e978c6e4f8fac288412099983afaedb91caab0d9e7246

shell/components/ActionMenuShell.vue
rak-phillip: If I recall correctly, this exists only to support an edge case from the original ActionMenu. This shouldn't really be an execution path with this replacement.

shell/components/ActionMenu.vue
rak-phillip, line 1: We should be deleting this version of the ActionMenu at some point. The intent for keeping this around was to maintain backwards compatibility with extensions in 2.10 environments.

aalves08:
> LGTM - I think this looks great and this change doesn't look as disruptive as I initially thought. It appears that most components in Dashboard generate actions via the `resources` prop.
> 
> @aalves08 this is technically a breaking change for any extensions that might be consuming the ActionMenuShell component. Do we have official channels for communicating this when we release a new version of shell?

@rak-phillip @codyrancher I can't say 100% that there isn't a single extension using this, but if I had to bet, nobody is using this. It's not documented anywhere... @mantis-toboggan-md @eva-vashkevich @torchiaf are you using this? I believe that the risk is very low, imho.

> We should be deleting this version of the ActionMenu at some point. The intent for keeping this around was to maintain backwards compatibility with extensions in 2.10 environments.

In regards to this, `2.10` is EOM. If I had to decide, I would remove it in the next opportunity or create an issue for it. Legacy stuff that only creates entropy. 

But I'll defer that decision to @nwmac @gaktive

mantis-toboggan-md:
> I can't say 100% that there isn't a single extension using this, but if I had to bet, nobody is using this. It's not documented anywhere... @mantis-toboggan-md @eva-vashkevich @torchiaf are you using this? 

No I'm not. ActionMenu/ActionMenuShell are not currently being used by virtual-clusters-ui, capi-ui-extension, nor any of the built-in provisioning extensions

aalves08:
Given the reply from @mantis-toboggan-md , I think this can go in @codyrancher @rak-phillip . Even if @torchiaf is using it, it's still not GA, so it can accommodate any change thrown at it.

PR #16354, opened 2026-01-10
Changed mechanism to see if hosted provider extension is prime only

### LINKED ISSUE #15972: Update Hosted providers page to check if extension is prime using a new mechanism
As a follow-up to https://github.com/rancher/dashboard/issues/15843, update https://github.com/eva-vashkevich/dashboard/blob/master/shell/pages/c/_cluster/manager/hostedprovider/index.vue#L112 to utilize the new prime detection mechanism

Edit: After some discussion, we decided to just remove any checks for prime from Hosted providers page.
Added manual test label because we do not have an easy way to test with prime extensions

eva-vashkevich:
### Summary
Fixes #15972 

Removed checks for "prime"ness from Hosted Providers page.

After adding gates on extension installation, checking for prime on Hosted providers page became redundant. To simplify code management, we decided to remove the check alltogether.

Removed all mentions of prime.

Hosted providers page
Install Alibaba extension. 
Check that 'prime only' label is no longer shown. 
Note: there is a bug in ali-ui that makes extension show up on cluster creation even if it was disabled here. That is unrelated to this issue

None.

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -2726,7 +2726,6 @@ cruResource:
 providers:
   hosted: 
     title: Hosted Providers
-    prime: Prime only
     warning: Disabling providers will make any clusters that use them uneditable
 drivers:
   kontainer:
--- a/shell/core/extension-manager-impl.js
+++ b/shell/core/extension-manager-impl.js
@@ -306,7 +306,7 @@ export const createExtensionManager = (context) => {
 
     // Apply the plugin based on its metadata
     applyPlugin(plugin) {
-      // Types
+    // Types
       Object.keys(plugin.types).forEach((typ) => {
         Object.keys(plugin.types[typ]).forEach((name) => {
           this.register(typ, name, plugin.types[typ][name]);
@@ -459,9 +459,11 @@ export const createExtensionManager = (context) => {
     getProviders(context) {
       // Custom Providers from extensions - initialize each with the store and the i18n service
       // Wrap in try ... catch, to prevent errors in an extension breaking the page
+
       const extensions = context.$extension.listDynamic('provisioner').map((name) => {
         try {
           const provisioner = context.$extension.getDynamic('provisioner', name);
+
           const defaults = {
             isCreate: false,
             isEdit:   false,


shell/pages/c/_cluster/manager/hostedprovider/index.vue
```diff
@@ -107,7 +107,7 @@ export default {
     async generateRows() {
       this.rows = this.allProviders.map((p) => {
         const active = p.id in this.settings ? this.settings[p.id] : true;
-        const canNotPrime = p.prime && !this.prime;
+        const canNotPrime = p.context.isPrime && !this.prime;
```
richard-cox: We might need to get together as a team and discuss the load but silently fail approach.

Shouldn't marking a whole extension as prime restrict all of it's functionality, which means ALL extension points (API level, including dynamic components like models, edits, lists, etc)? If so i think we should be doing the restriction wholesale in the extension mechanism.

eva-vashkevich: Hi @richard-cox 
I don't think it is silently failing in this case because we do show to the user that provider is prime-only.
Currently, we only enforce it within the extensions themself, so I do agree that centralizing it inside Rancher will make it easier to maintain.

richard-cox: Ok, went back to the original issue. 

https://github.com/rancher/dashboard/issues/15843 / https://github.com/rancher/dashboard/pull/15934 means users should not be able to install extensions that have the prime only annotation (there's a conversation regarding it there as well).

Is this mechanism a second line of preventing community users from using prime only features? Have we hit a use case for it yet?

eva-vashkevich: @richard-cox  the original implementation was needed because we didn't have prime gates yet, so you bring up the good point that we can probably just remove it now from Hosted providers page.
But, having it here also adds a bit of a reminder of Prime value

richard-cox: Gotcha.

Alex also came up with an interesting point, that this would be needed to support extensions that haven't yet been updated to a version that has the annotation. 

Given this is a bit edge case will leave the decision with you. 
- Goes in - covers edge case, adds more code to maintain
- Doesn't go in - edge case is exposed, less code to maintain

eva-vashkevich: @richard-cox I don't think this is true. This change makes the annotation the only source of truth. So if extension doesn't have the annotation, it will bypass this.
Additionally:
- the only place this code affects right now is Hosted providers page. We can reuse it for other areas but currently have no expressed need for that
- the only prime extension that is displayed on Hosted providers page is Alibaba and it had the annotation from the first release.

PR #16381, opened 2026-01-15
Creating unit tests for javascript utils to reduce risk when converting to typescript

### LINKED ISSUE #9859: Convert utils to Typescript
# Description

Utils are the most straightforward parts of the code to convert to TS and we should tackle them down.
This is an attempt to have more efforts in this direction, to have less bug prone code.

https://github.com/rancher/dashboard/tree/master/shell/utils
[So far we have 45 files](https://github.com/search?q=repo%3Arancher%2Fdashboard+path%3Autils%2F*.js&type=code).

codyrancher:
### Summary
One of a handful of prs to add unit tests for javascript utils to reduce risk when converting to typescript.

Targets but doesn't fix #9859

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/utils/__tests__/queue.test.ts
+++ b/shell/utils/__tests__/queue.test.ts
@@ -1,0 +1,208 @@
+import Queue from '@shell/utils/queue';
+
+describe('class: Queue', () => {
+  let queue: any;
+
+  beforeEach(() => {
+    queue = new Queue();
+  });
+
+  describe('constructor', () => {
+    it('should create empty queue', () => {
+      expect(queue.getLength()).toBe(0);
+      expect(queue.isEmpty()).toBe(true);
+    });
+
+    it('should initialize offset to 0', () => {
+      expect(queue.offset).toBe(0);
+    });
+  });
+
+  describe('getLength', () => {
+    it('should return 0 for empty queue', () => {
+      expect(queue.getLength()).toBe(0);
+    });
+
+    it('should return correct length after enqueue', () => {
+      queue.enqueue('item1');
+      queue.enqueue('item2');
+
+      expect(queue.getLength()).toBe(2);
+    });
+
+    it('should return correct length after dequeue', () => {
+      queue.enqueue('item1');
+      queue.enqueue('item2');
+      queue.dequeue();
+
+      expect(queue.getLength()).toBe(1);
+    });
+  });
+
+  describe('isEmpty', () => {
+    it('should return true for empty queue', () => {
+      expect(queue.isEmpty()).toBe(true);
+    });
+
+    it('should return false after enqueue', () => {
+      queue.enqueue('item');
+
+      expect(queue.isEmpty()).toBe(false);
+    });
+
+    it('should return true after dequeueing all items', () => {
+      queue.enqueue('item');
+      queue.dequeue();
+
+      expect(queue.isEmpty()).toBe(true);
+    });
+  });
+
+  describe('enqueue', () => {
+    it('should add item to queue', () => {
+      queue.enqueue('item');
+
+      expect(queue.getLength()).toBe(1);
+    });
+
+    it('should add multiple items in order', () => {
+      queue.enqueue('first');
+      queue.enqueue('second');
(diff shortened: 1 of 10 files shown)


shell/utils/__tests__/queue.test.ts
```diff
@@ -0,0 +1,208 @@
...
+
+describe('class: Queue', () => {
+  let queue: any;
+
+  beforeEach(() => {
+    queue = new Queue();
+  });
+
+  describe('constructor', () => {
+    it('should create empty queue', () => {
```
cnotv, line 11: Please use explicit initial state assertion, so `this.queue` will always return a list and initialize as empty.

shell/utils/__tests__/queue.test.ts
```diff
@@ -0,0 +1,208 @@
...
+      expect(queue.offset).toBe(0);
+    });
+  });
+
+  describe('getLength', () => {
+    it('should return 0 for empty queue', () => {
+      expect(queue.getLength()).toBe(0);
+    });
+
+    it('should return correct length after enqueue', () => {
```
cnotv, line 26: Please ensure that the same tests for `enqueue()` are used also for `dequeue()`

shell/utils/__tests__/units.test.ts
```diff
@@ -0,0 +1,279 @@
...
+    expect(UNITS).toStrictEqual(['', 'K', 'M', 'G', 'T', 'P', 'E', 'Z', 'Y']);
+  });
+
+  it('should export FRACTIONAL array with correct values', () => {
+    expect(FRACTIONAL).toStrictEqual(['', 'm', 'u', 'n', 'p', 'f', 'a', 'z', 'y']);
+  });
+});
+
+describe('fx: formatSi', () => {
+  it('should format small values without suffix when addSuffix is false', () => {
```
cnotv, line 23: Default formatting is missing.

shell/utils/__tests__/units.test.ts
```diff
@@ -0,0 +1,279 @@
...
+  it('should respect maxExponent option', () => {
+    const result = formatSi(1500000000, { maxExponent: 1 });
+
+    expect(result).toBe('1500000 K');
+  });
+
+  it('should format zero correctly', () => {
+    const result = formatSi(0);
+
+    expect(result).toBe('0 ');
```
cnotv, line 86: This does not look correct.

codyrancher:
I'm having some difficulty understanding what you're trying to say here so if you'll bare with me I'd just like to confirm some of the points.

> I am personally contrary on adding TS later, especially for lengthy 2k LoT code change, as this is the highest level of human prone errors VS generating code with AI.

Are you saying you'd like to skip converting to typescript here? You're mostly wary because of the size of the change involved?

> It is also required to be sure what the functions are returning, in order to write the tests, so not sure how you could skip such a necessary step.

My goal was to enforce invariants so that when these files were changed to typescript they would stay logically the same bugs and all. That doesn't require or involved verifying the correctness of ALL outputs.

cnotv:
> I'm having some difficulty understanding what you're trying to say here so if you'll bare with me I'd just like to confirm some of the points.
> 
> > I am personally contrary on adding TS later, especially for lengthy 2k LoT code change, as this is the highest level of human prone errors VS generating code with AI.
> 
> Are you saying you'd like to skip converting to typescript here? You're mostly wary because of the size of the change involved?

No, I said that it would make more sense in this sequence:
- Convert to TS with any or temporary types, to define what the utility does
- Add the tests with some cases you defined in the TS
- Correct the TS with all the code smell crossed

> > It is also required to be sure what the functions are returning, in order to write the tests, so not sure how you could skip such a necessary step.
> 
> My goal was to enforce invariants so that when these files were changed to typescript they would stay logically the same bugs and all. That doesn't require or involved verifying the correctness of ALL outputs.

I find it very difficult to think about adding TS properly without fixing, because that's the actual purpose of TS.

cnotv:
In any case, that was my opinion, this whole conversation is not a blocker at all.

PR #16403, opened 2026-01-20
Automation: slack notification fix

yonasberhe23:
### Summary
Fixes #https://github.com/rancher/qa-tasks/issues/2148

Prevents Slack notifications when tests didn't run or when only post-processing steps fail. The pipeline now tracks test execution results separately from overall build status and only sends notifications when tests actually ran and failed. 

- Added testExecutionResult variable to track test execution status separately from build status
- Only process results.xml and send notifications when tests actually ran (verified by <testcase> elements)
- Made HTML report operations non-fatal to prevent false build failures

### Checklist
- [x] The PR has a Milestone 


cypress/jenkins/Jenkinsfile
```diff
@@ -71,11 +73,13 @@ node {
             try {
               stage('Grab Results') {
                   sh "cypress/jenkins/scpget.sh dashboard/results.xml"
-                  sh "cypress/jenkins/scpget.sh dashboard/cypress/reports/html/* html"
+                  sh "cypress/jenkins/scpget.sh dashboard/cypress/reports/html/* html || true"
```
izaac: We should fail the build if an HTML reporting error occurs. It is better to surface and fix the root cause rather than allowing it to fail silently. cc @ShylajaDevadiga

yonasberhe23: Good call, will remove that.

cypress/jenkins/Jenkinsfile
```diff
@@ -71,11 +73,13 @@ node {
...
               }
             } catch(err) {
                 echo "Error: " + err
-                currentBuild.result = 'FAILURE'
+                if (!fileExists('results.xml')) {
+                    currentBuild.result = 'FAILURE'
+                }
```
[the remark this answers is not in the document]
yonasberhe23: Stale results concern is not applicable. Each run starts with a clean workspace (deleteDir()), so results.xml won't exist from a previous run. However i agree with the point that there's inconsistent behavior; we should always fail the build if the "Grab Results" stage fails.

cypress/jenkins/Jenkinsfile
```diff
@@ -90,8 +94,18 @@ node {
...
-                  sh "pwd && ls -al && ls -al cypress"
+                  def hasTestResults = false
+                  if (fileExists('results.xml')) {
+                      def resultsContent = readFile('results.xml').trim()
+                      hasTestResults = resultsContent.contains('<testcase')
+                  }
+                  
+                  if (hasTestResults) {
+                      step([$class: 'JUnitResultArchiver', testResults: '**/results.xml'])
+                      testExecutionResult = currentBuild.result
```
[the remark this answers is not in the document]
yonasberhe23: This is a good point about currentBuild.result potentially being affected by previous stages is valid. However, comparing build results before/after JUnitResultArchiver can misclassify when the result was already UNSTABLE from a previous stage and tests also fail—the result doesn't change, so it would incorrectly assume tests passed. Instead, we can check the XML directly for <failure elements, which reflects the actual test outcome regardless of prior build status. This is more reliable and explicit about what we're checking.

cypress/jenkins/Jenkinsfile
```diff
@@ -103,7 +117,13 @@ node {
...
             }
             catch(err) {
-              echo "Error: " + err
+              echo "Error in Test Report stage: " + err
+              if (testExecutionResult == null && fileExists('results.xml')) {
+                  def resultsContent = readFile('results.xml').trim()
+                  if (resultsContent.contains('<testcase')) {
+                      testExecutionResult = 'FAILURE'
+                  }
+              }
```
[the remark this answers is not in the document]
yonasberhe23: Agreed. The catch block incorrectly assumed FAILURE when testExecutionResult was null and tests ran, even if tests passed. Updated the catch block to check the XML for <failure elements to determine the actual test outcome.

cypress/jenkins/slack-notification.sh
```diff
@@ -14,7 +14,7 @@ send_slack_notification() {
     
     if [ -z "$bot_token" ]; then
         echo "Warning: UI_SLACK_BOT_TOKEN not set, skipping Slack notification"
-        return 0
+        return 1
```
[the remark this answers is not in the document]
yonasberhe23: Not making changes here. 

Returning 1 when the token is missing is correct: the operation didn't succeed.
The function name is fine: it attempts to send; if it can't, it fails.
The Jenkinsfile doesn't check the exit code, so this won't break the build.

cypress/jenkins/Jenkinsfile
izaac: I've been looking closer at this `error()` call. Since it terminates the pipeline immediately, the downstream stages (`Test Report`, `Notifications`) will never execute. We should remove it or wrap the reporting logic in a finally block to ensure we still get alerted on failure.

izaac: To achieve your goal of "Fail Fast but Report," you must move the Reporting stages into a finally block. This guarantees they run even if Run Tests throws an `error()`.

izaac:
```groovy
node {
    // ...
    try {
        // If this fails, we want to stop immediately.
        stage('Checkout') { ... }
        
        stage('Run Tests') {
            try {
               sh 'cypress/jenkins/init.sh'
            } catch (err) {
               // This captures the script failure
               currentBuild.result = 'FAILURE'
               throw err // Throw not error()
            }
        }
        
        // This stage will not run if 'Run Tests' fails (Saving resources)
        stage('Next Step') { ... }

    } catch (err) {
        echo "Pipeline aborted early: ${err}"
        currentBuild.result = 'FAILURE' 
    } finally {
        // Reporting Block
        // This always runs, even if we threw an error above.
        
        stage('Grab Results') {
             // Safe to run even if init.sh failed (scp fails gracefully)
             sh "cypress/jenkins/scpget.sh dashboard/results.xml || true" 
        }

        stage('Test Report') {
             // The XML parsing logic here
        }

        stage('Notifications') {
             // The Slack script call
             sh "cypress/jenkins/slack-notification.sh ${currentBuild.result}"
        }
    }
}
```

cypress/jenkins/Jenkinsfile
```diff
@@ -90,8 +92,25 @@ node {
             }
             try {
               stage('Test Report') {
-                  step([$class: 'JUnitResultArchiver', testResults: '**/results.xml'])
-                  sh "pwd && ls -al && ls -al cypress"
+                  def hasTestResults = false
```
izaac: I think the logic here can be simplified to read results.xml once.

```groovy
stage('Test Report') {
    try {
        if (fileExists('results.xml')) {
            // archive immediately
            junit testResults: '**/results.xml', allowEmptyResults: true

            // read once
            def content = readFile('results.xml').trim()
            
            if (content.contains('<failure')) {
                testExecutionResult = 'UNSTABLE'
            } else if (content.contains('<testcase')) {
                testExecutionResult = 'SUCCESS'
            } else {
                testExecutionResult = null // file exists but is empty/invalid
            }
        } else {
            echo "No results.xml found."
            testExecutionResult = null
        }

        // publish if this fails, we already have testExecutionResult
        publishHTML(target : [
            allowMissing: false,
            alwaysLinkToLastBuild: true,
            keepAll: true,
            reportDir: 'html',
            reportFiles: 'index.html',
            reportName: 'html-report',
            reportTitles: 'Report'
        ])

    } catch(err) {
        echo "Error in Test Report stage: " + err
        currentBuild.result = 'FAILURE'
    }
}
```

cypress/jenkins/Jenkinsfile
yonasberhe23: Thanks for this feedback. Changes made and tested. ready for review

cypress/jenkins/init.sh
```diff
@@ -356,7 +356,7 @@ corral config vars set cypress_tags "${CYPRESS_TAGS}"
 # Save all values to a file for the Slack notification script
 cat > "${WORKSPACE}/notification_values.txt" << EOF
 RANCHER_VERSION=${RANCHER_VERSION}
-RANCHER_IMAGE_TAG=${RANCHER_IMAGE_TAG}
+RANCHER_IMAGE_TAG=${RANCHER_IMAGE_TAG_FOR_CORRAL:-${RANCHER_IMAGE_TAG}}
```
yonasberhe23, line 359: snuck this in to account for Prime builds

PR #16433, opened 2026-01-22
Fix long secret values causing page horizontal scroll

### LINKED ISSUE #16408: When displaying opaque secrets, we should manage the length of the secret to avoid horizontal scroll on the page

When displaying Opaque secrets, we replace characters with dots, but keep the length unchanged. This leads to us adding a horizontal scroll to the pages that show the secret. This scroll doesn't provide any value, since users cannot see the content and make it look awkward.

Image

We can either truncate the secret we show to a standardized length, which will also obscure the length of the secret, or add some wrapping to prevent scrolling.

**Additional context**
A sample secret that can be used to test:

```apiVersion: v1
kind: Secret
metadata:
  name: test-long-secret
type: Opaque
data:
  short-key: c2hvcnQgdmFsdWU=
  long-key: YWJjZGVmZ2hpamtsbW5vcHFyc3R1dnd4eXowMTIzNDU2Nzg5QUJDREVGR0hJSktMTU5PUFFSU1RVVldYWVphYmNkZWZnaGlqa2xtbm9wcXJzdHV2d3h5ejAxMjM0NTY3ODlBQkNERUZHSElKS0xNTk9QUVJTVFVWV1hZWmFiY2RlZmdoaWprbG1ub3BxcnN0dXZ3eHl6MDEyMzQ1Njc4OUFCQ0RFRkdISUpLTE1OT1BRUlNUVVZXWFlaYWJjZGVmZ2hpamtsbW5vcHFyc3R1dnd4eXowMTIzNDU2Nzg5QUJDREVGR0hJSktMTU5PUFFSU1RVVldYWVphYmNkZWZnaGlqa2xtbm9wcXJzdHV2d3h5ejAxMjM0NTY3ODlBQkNERUZHSElKS0xNTk9QUVJTVFVWV1hZWg==

```

codyrancher:
## Summary
- Long opaque secrets now display with a horizontal scrollbar within their container instead of causing page-level horizontal scroll
- Uses `contain: inline-size` to constrain the container width and `overflow-x: auto` on the concealed content

Fixes #16408

- Handled overflow better
- Added some spacing between the action header and the buttons so they buttons don't overlap the text,

- Create an opaque secret with a very long value (300+ characters, no spaces)
- View the secret detail page
- Verify the secret container stays within the page bounds
- Verify a horizontal scrollbar appears on the concealed content row
- Verify short secrets display normally without a scrollbar

A secret you can use
```apiVersion: v1
kind: Secret
metadata:
  name: test-long-secret
type: Opaque
data:
  short-key: c2hvcnQgdmFsdWU=
  long-key: YWJjZGVmZ2hpamtsbW5vcHFyc3R1dnd4eXowMTIzNDU2Nzg5QUJDREVGR0hJSktMTU5PUFFSU1RVVldYWVphYmNkZWZnaGlqa2xtbm9wcXJzdHV2d3h5ejAxMjM0NTY3ODlBQkNERUZHSElKS0xNTk9QUVJTVFVWV1hZWmFiY2RlZmdoaWprbG1ub3BxcnN0dXZ3eHl6MDEyMzQ1Njc4OUFCQ0RFRkdISUpLTE1OT1BRUlNUVVZXWFlaYWJjZGVmZ2hpamtsbW5vcHFyc3R1dnd4eXowMTIzNDU2Nzg5QUJDREVGR0hJSktMTU5PUFFSU1RVVldYWVphYmNkZWZnaGlqa2xtbm9wcXJzdHV2d3h5ejAxMjM0NTY3ODlBQkNERUZHSElKS0xNTk9QUVJTVFVWV1hZWg==
```

Displaying or editing various secret types.

### Screenshot/Video

https://github.com/user-attachments/assets/6f5d5047-c57f-4eb0-be66-e6b4f0c8df2c

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/DetailText.vue
+++ b/shell/components/DetailText.vue
@@ -167,11 +167,11 @@ export default {
     <h5
       v-if="labelKey"
       v-t="labelKey"
-      v-clean-tooltip="itemLabel"
+      v-clean-tooltip="{content: itemLabel, popperClass: 'detail-text-tooltip'}"
     />
     <h5
       v-else-if="label"
-      v-clean-tooltip="label"
+      v-clean-tooltip="{content: label, popperClass: 'detail-text-tooltip'}"
     >
       {{ label }}
     </h5>
@@ -209,6 +209,7 @@ export default {
     <template v-if="!isBinary && !jsonStr && isLong && !expanded">
       <a
         href="#"
+        class="more-characters"
         @click.prevent="expand"
       >{{ plusMore }}</a>
     </template>
@@ -298,5 +299,19 @@ export default {
 .monospace {
   white-space: pre-wrap;
   word-wrap: break-all
+}
+
+.more-characters {
+  margin-top: 8px;
+  display: inline-block;
+}
+</style>
+
+<style lang="scss">
+// The global styles for tooltips are in dashboard/shell/assets/styles/global/_tooltip.scss.
+// I don't want to make this change for all tooltips since there's 149 instances as of writing this
+// so I'm adding a global style here that's scoped to the class we're adding to the tooltips we have in this component.
+.detail-text-tooltip.v-popper__popper.v-popper--theme-tooltip {
+  overflow-wrap: anywhere;
 }
 </style>


shell/components/DetailText.vue
```diff
@@ -238,6 +238,18 @@ export default {
   background-color: var(--input-bg);
   border-radius: var(--border-radius);
   border: solid var(--border-width) var(--input-border);
+  contain: inline-size;
+  width: 100%;
+
+  h5 {
+    margin-bottom: 15px;
```
codyrancher, line 253: The top right buttons would overlap the text below when mouseover styling was applied.

codyrancher:
> I think there is some room for improvement
> Screen.Recording.2026-01-26.at.2.41.36.PM.mov

Fixed, accounting for label now as well.

https://github.com/user-attachments/assets/721562e7-53be-4e47-bd06-58ea144c86ca

eva-vashkevich:
@codyrancher don't hate me, but should we add a tooltip or some other way to see the full key?

codyrancher:
> @codyrancher don't hate me, but should we add a tooltip or some other way to see the full key?

No problem. I probably should've done it anyway. 

While switching to the tooltip I noticed two other issues that I also addressed.
1. It would center the tooltip over the bar even if the label was small which looked odd.
2. The scrollbar would overlap the secret text


https://github.com/user-attachments/assets/7853effb-8320-40f7-ba85-5196af0aa224

eva-vashkevich:
@codyrancher this still seems odd. I don't think you can scroll the tooltip, and we don't show the key fully in config either.

https://github.com/user-attachments/assets/d1a0eee1-cec6-4fbb-9c5e-da7c85d3f8b2

codyrancher:
> @codyrancher this still seems odd. I don't think you can scroll the tooltip, and we don't show the key fully in config either.
> Screen.Recording.2026-01-28.at.3.47.13.PM.mov

Interesting, so it seems the difference between your example and mine is the hyphenation. When it's hyphenated we will word wrap so the tooltip shows multiple lines which shows the entire key without scrolling.

codyrancher:
Now word-wrap even if the keys don't have any natural breaks.
Also re-positioned the more characters link to not interfere with the scrollbar.

https://github.com/user-attachments/assets/0bfd9381-ddb9-4083-8076-3d2ffe197108

eva-vashkevich (approved):
lgtm

PR #16443, opened 2026-01-22
Fix project scoped secrets validation

momesgin:
### Summary
Fixes rancher/rancher#53744 

- While I was unable to reproduce the issue locally, analysis of the provided screen recording on the JIRA ticket and the codebase revealed a race condition where the `backingNamespace` could be missing or invalid when the form validation runs.
- Implemented logic to intelligently select the first project with a valid `backingNamespace` as the default during secret creation.
- Updated validation rules to dynamically adjust based on the secret scope, preventing premature validation errors.
- Added a manual safety check to ensure a namespace is present before saving a project-scoped secret.

### Technical notes summary
- In `shell/edit/secret/index.vue`:
    - Modified `fetch()` to find the first project with a `backingNamespace` instead of blindly picking the first project.
    - Added a watcher for `isProjectScoped` to remove the 'required' validation rule for `metadata.namespace` when in project-scoped mode. This allows the UI to remain valid while the user selects a project.
    - Updated the `projectName` watcher to set `metadata.namespace` to `null` if the selected project lacks a `backingNamespace` (e.g., if it's still initializing), preventing the use of stale or invalid namespaces.
    - Added a manual check in `saveSecret()` to block saving if `metadata.namespace` is missing when `isProjectScoped` is true.
- Unit tests.

- Navigate to the Project Scoped Secret Creation: Explorer > Storage > Project Secrets
- Since the issue is intermittent, repeated testing is valuable. Maybe loading the page directly via URL would help? vs navigating from the UI.
- Verify that the "Create" button is enabled once a project is selected (and other required fields are filled).
- Switch between different projects and ensure the secret is created in the correct backing namespace.

- Ensure that the validation for standard namespaced secrets isn't broken.
- For Project Scoped Secrets, verify that switching projects quickly doesn't leave the form in an inconsistent state.

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/CodeMirror.vue
+++ b/shell/components/CodeMirror.vue
@@ -330,9 +330,6 @@ export default {
         border-color: var(--primary-border);
       }
 
-      .CodeMirror-wrap pre {
-        word-break: break-word;
-      }
       .CodeMirror-code {
         .CodeMirror-line {
           &:not(:last-child)>span:after,
@@ -425,6 +422,10 @@ export default {
 
       .CodeMirror-gutters {
         background: inherit;
+      }
+
+      .CodeMirror-wrap pre {
+        word-break: break-word;
       }
     }
 
--- a/shell/components/ResourceTable.vue
+++ b/shell/components/ResourceTable.vue
@@ -427,7 +427,7 @@ export default {
     },
 
     _applicableExtensionTableHooks() {
-      if (this.$store.$plugin?.getUIConfig) {
+      if (this.$store.$extension?.getUIConfig) {
         const extensionTableHooks = getApplicableExtensionEnhancements(this, ExtensionPoint.TABLE, TableLocation.RESOURCE, this.$route);
 
         return extensionTableHooks;
--- a/shell/package.json
+++ b/shell/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@rancher/shell",
-  "version": "3.0.9-rc.1",
+  "version": "3.0.9-rc.2",
   "description": "Rancher Dashboard Shell",
   "repository": "https://github.com/rancher/dashboard",
   "license": "Apache-2.0",


nwmac:
@momesgin Rather than put a manual-test label on the PR, can you update the linked issue and change from dev-automation to manual - the issues get picked up by QA, not the PRs. Generally we don't put labels on PRs, other than those created by bots to indicate such.

shell/edit/secret/index.vue
```diff
@@ -60,19 +60,24 @@ export default {
...
 
     if (isProjectScoped) {
-      // If ssp is enabled the store not have all projects. ensure we have them all
+      // If SSP is enabled the store may not have all projects. Ensure we have them all
       await this.$store.dispatch('management/findAll', { type: MANAGEMENT.PROJECT });
       if (this.isCreate) {
-        // Pick first project as default
-        this.projectName = this.filteredProjects[0].metadata.name;
+        // Pick first project with a backing namespace as default.
+        // Without this, we might pick a project that is still initializing, causing validation errors.
```
richard-cox: Something isn't right here.

All projects should have a backingNamepace otherwise they wouldn't exist (a project is just a proxy onto an upstream namespace).

I think the problem does revolve around a missing namespace and initialisation, but it shouldn't be this

shell/edit/secret/index.vue
```diff
@@ -378,14 +392,39 @@ export default {
...
+      }
+    },
+
     projectName(neu) {
       if (this.isCreate && neu) {
         this.value.metadata.labels = this.value.metadata.labels || {};
         this.value.metadata.labels[UI_PROJECT_SECRET] = neu;
 
-        const projectScopedNamespace = this.filteredProjects.find((p) => p.metadata.name === neu).status.backingNamespace;
+        const project = this.filteredProjects.find((p) => p.metadata.name === neu);
```
richard-cox: Changes to this function aren't needed, unless for our purposes we need null instead of undefined, or it's failing to find a project and throwing an error (not seen in jira video)

if the latter it could just be `const projectScopedNamespace = this.filteredProjects.find((p) => p.metadata.name === neu)?.status.backingNamespace;`

momesgin: done

shell/edit/secret/index.vue
momesgin: @richard-cox Thanks for this info. Since I can't reproduce the issue it's hard to pinpoint the problem.

Should we ask and wait for reliable reproduction steps?

momesgin: we're making other changes to the PSS feature they might actually fix this issue too 🤷‍♂️

PR #16444, opened 2026-01-22
Enable editing of custom claims for keycloak provider

### LINKED ISSUE #16241: Keycloak custom mapping
backend rancher/rancher 52405

rak-phillip:
### Summary
This enables the editing of custom claims for keycloak provider.

Fixes #16241

- Enable editing of custom claims for keycloak provider
- Add unit tests to assert new behavior

### Technical notes summary

rancher/rancher#53465 contains fixes for custom userName claims in keycloak. This change requires that custom claims be enabled for keycloak as well as generic oidc.

Keycloak OIDC should render and save custom claims

This change should be additive to the keycloak provider config form.

### Screenshot/Video

image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/edit/auth/__tests__/oidc.test.ts
+++ b/shell/edit/auth/__tests__/oidc.test.ts
@@ -247,5 +247,55 @@ describe('oidc.vue', () => {
       expect(groupsClaim.exists()).toBe(true);
       expect(emailClaim.exists()).toBe(true);
     });
+
+    it('should render addCustomClaims and supportsGroupSearch  checkbox when provider is keycloak', async() => {
+      wrapper.vm.model.id = 'keycloakoidc';
+      await nextTick();
+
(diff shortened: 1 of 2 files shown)


rak-phillip:
> should we surface the claims info on the non-edit screen?

I don't think we should change this. The scope of the change is to surface existing behavior for keycloak.

> should we validate the email field (we probably have it somewhere in our code...)?

If the ask is to treat this field as an email address, no. The custom claim would be something like `user_email` instead of `email`.

> if we "add custom claims" (check the checkbox) are any of the fields required? right now the form reacts as none being required

I don't believe so. Custom claims are for manually mapping when a provider doesn't use standard claims in a token. There's no guarantee that we will need to map all three currently represented.

aalves08 (approved):
Given the replies, 👍

PR #16517, opened 2026-01-30
Improve agents unit test info

richard-cox:
### Summary
- Improve agents unit test
- Based on some trial runs of creating unit tests
- We should continue to improve and expand these

Confluence page has also been updated

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -84,29 +84,60 @@ To get started, follow the `Getting Started` section.
   - `storybook/`: Component documentation source
 
 
+## Unit Tests
+
+Please see [Rancher UI Internal Documentation - Testing - Unit Tests](https://extensions.rancher.io/internal/testing/unit-test) for more information
+
+### Writing Unit Tests
+
+- Tests use the Jest framework (Chai based assertions in a Mocha like structure)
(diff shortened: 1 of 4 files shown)


AGENTS.md
codyrancher, line 1: All of this global information may start polluting the global context even though it's not relevant for all tasks.

I'm fine to accept so we can experiment but it's starting to look like we should parcel out to writing [skills](https://github.blog/changelog/2025-12-18-github-copilot-now-supports-agent-skills/). I'm not seeing an open standard for it though https://github.com/agentsmd/agents.md

richard-cox: That's the conundrum, making all of this tool agnostic. There's a note in the agemts.md dir readme about supporting multiple agents. The only idea i've got is breaking the mono agent.md into multiple root agents files and more tool specific symlinks...at the cost of not supporting tools that we don't have the symlinks for.

codyrancher: Yeah it is. The only idea I have is to add a 'build' script for each tool that assembles the md files into the expected format but it seems overkill for now.

PR #16525, opened 2026-01-30
Remove a deprecated warning we were getting on ResourceTable pages.

codyrancher:
### Summary
Just swap from $plugin to $extension in order to get rid of this warning.

image

Question, do you think we can remove [this ](https://github.com/rancher/dashboard/blob/master/shell/store/index.js#L776)reference now too? 

### Technical notes summary
The objects are identical and have the same methods so this is fairly low risk.

List pages can be used to verify this change.

See above

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/ResourceTable.vue
+++ b/shell/components/ResourceTable.vue
@@ -427,7 +427,7 @@ export default {
     },
 
     _applicableExtensionTableHooks() {
-      if (this.$store.$plugin?.getUIConfig) {
+      if (this.$store.$extension?.getUIConfig) {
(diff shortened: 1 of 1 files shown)


codyrancher:
> I think we are ok to do this @codyrancher . I agree with you. I searched for the code and this was the last lingering hit if I am not mistaken. Let's get rid of $plugin in the place you mentioned in your question.
> 
> We still have a month to adjust for `2.14` IF we hit any issue.

Sounds good, I think we just need to verify that removing $plugin won't affect extensions that weren't built with this change I'm about to merge.

PR #16568, opened 2026-02-04
Bypass checking home URL to match charts when repo name exists

### LINKED ISSUE #16567: Installed Apps, Edit/Upgrade failed to show the installed Helm Chart version
SURE-11006

**Setup**
- Rancher version: 2.12.2 and later
- Rancher UI Extensions:
- Browser type & version:

**Describe the bug**

In Explorer > Apps > Installed Apps, selecting Edit/Upgrade(Change Version) from the menu fails to pre-select the currently installed version on the next page.

After some investigation, we found the issue to be the home URL check inside `matchingCharts` in `shell/models/catalog.cattle.io.app.js`. This check was originally added to prevent false upgrade suggestions for cert-manager([the issue](https://github.com/rancher/dashboard/issues/11465)).

The check should be removed while ensuring we do not introduce regressions.
 
**To Reproduce**

**Result**

**Expected Result**

**Screenshots**

**Additional context**

NOTE: Testing this is a bit tricky, if you need help ping @momesgin

momesgin:
### Summary
Fixes #16567

Skipped `home` URL validation when a chart has a repository and exactly one match. This allows upgrades to proceed when the upstream chart's `home` URL changes in a newer version.

### Technical notes summary
I've wrapped the existing `home` URL filtering logic inside an `if (!repoName || matchingCharts.length > 1)` condition. This ensures we still correctly prevent suggesting updates from a UI-added repo for charts installed via CLI. At the same time, it avoids unnecessary home URL checks when only a single match is found among the existing repositories. Also added a unit test.

Testing the real scenario explained in SURE-11006 is tricky. If you can't test it by yourself you might need to do this: 

1. Apps > Repositories > Create a repo using my testing helm repo URL `https://momesgin.github.io/mo-charts/`
2. Install version **0.2.0**
3. Make sure on **Installed Apps** page the **Upgradable** column displays **0.3.0** in front of the chart
4. Ping me to update the repo by replacing the old home URL
5. Refresh the repo and check the Installed Apps page again
6. Without my fix the **Upgradable** column should NOT display **0.3.0** in front of the chart, but with my fix it **should**

- Make sure `cert-manager` installed via CLI doesn't get an upgrade suggestion from a repository added to the UI that also has `cert-manager` chart(e.g. AppCo). For more info you can check [the original issue](https://github.com/rancher/dashboard/issues/11465).

### Checklist
- [x] The PR has a Milestone 


shell/models/__tests__/catalog.cattle.io.app.test.ts
```diff
@@ -120,7 +120,7 @@ describe('class CatalogApp', () => {
       [installedCertManagerOfficialFromCli, [], APP_UPGRADE_STATUS.NO_UPGRADE],
       [installedCertManagerOfficialFromCli, [appCoMatchingChart1], APP_UPGRADE_STATUS.NO_UPGRADE],
       [installedCertManagerOfficialFromCli, [appCoMatchingChart1, appCoMatchingChart2], APP_UPGRADE_STATUS.NO_UPGRADE],
-      [installedCertManagerOfficialFromCli, [appCoMatchingChart1, appCoMatchingChart2, certManagerOfficialMatchingChart1], APP_UPGRADE_STATUS.SINGLE_UPGRADE],
+      [installedCertManagerOfficialFromCli, [appCoMatchingChart1, appCoMatchingChart2, certManagerOfficialMatchingChart1], APP_UPGRADE_STATUS.NO_UPGRADE],
```
momesgin: @richard-cox This is a side-effect of this PR which I think it's fine, but need to know your opinion.

To explain this side-effect:

1. Install `cert-manager` through CLI from `https://charts.jetstack.io/`:
```
helm repo add jetstack https://charts.jetstack.io/
helm repo update

helm install cert-manager jetstack/cert-manager --namespace cert-manager --version v1.7.1
```
2. Go to Explorer > Apps > Repositories and add a new repo using the same repo URL(`https://charts.jetstack.io/`)

We used to find a match and display an upgradeable version(e.g. 1.20.0-alpha.1) for these scenarios, but after this change we won't.

momesgin: So far this is the only side-effect I've found, still looking.

shell/models/catalog.cattle.io.app.js
```diff
@@ -67,39 +67,18 @@ export default class CatalogApp extends SteveModel {
     const chartName = chart.metadata?.name;
     const repoName = chart.metadata?.annotations?.[CATALOG_ANNOTATIONS.SOURCE_REPO_NAME] || this.metadata?.labels?.[CATALOG_ANNOTATIONS.CLUSTER_REPO_NAME];
 
+    if (!repoName) {
```
richard-cox: I thought the plan was 

1. if the app references a specific repo and there's a match we do not do anything with home url
    - This removes the scenario where in the ui someone installed a chart ... but the chart / repo home urls don't align
2. if the app does not reference a specific repo then we continue as before

momesgin: @richard-cox I’ve updated the code to check the home URL only when no repository is found (which ensures we don’t suggest incorrect updates for `cert-manager`) or when multiple repositories are detected.

richard-cox (approved):
Think we should also get QA to validate both angles mentioned in the description
- SURE-11006 - this one needs a `mo-charts-2` repo (can be a fork of mo-charts just containing a version bump and the specific change form the JIRA that caused the issue)
- https://github.com/rancher/dashboard/issues/11465 for regressions.

I think we could probably risk / process the backports too, will create them. Those will also need manual QA

PR #16587, opened 2026-02-05
Implements RcAccordion to reflect the styles in the design system.

### LINKED ISSUE #16119: Add a new accordion component following the new design

Add a new accordion component
Rewrite generic cluster import to use new accordion component

**Additional context**

codyrancher:
### Summary
Fixes #16119

Ask for a link to the design system if you'd like it. I think we can either implement the `shell/components/Accordion.vue` using this component or we could replace it.

We will still need some feedback from UX since there wasn't a dark theme implementation.

The RcAccordion and storybook entries

RcButton ghost variant focus styling.

### Screenshot/Video

https://github.com/user-attachments/assets/7a71ecf0-a622-4839-8d0a-113fc93db134

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/pkg/rancher-components/src/components/RcButton/RcButton.vue
+++ b/pkg/rancher-components/src/components/RcButton/RcButton.vue
@@ -238,12 +238,7 @@ button {
     padding: 0;
     background-color: transparent;
 
-    &:focus, &.focused {
-      @include focus-outline;
-      outline-offset: 0;
-    }
-
-    &:focus-visible {
+    &:focus-visible, &.focused {
       @include focus-outline;
       outline-offset: 0;
     }
--- a/pkg/rancher-components/src/components/RcAccordion/RcAccordion.test.ts
+++ b/pkg/rancher-components/src/components/RcAccordion/RcAccordion.test.ts
@@ -52,13 +52,11 @@ describe('rcAccordion.vue', () => {
     await nextTick();
     expect(isBodyVisible(wrapper)).toBe(false);
 
-    // Click to expand
     await wrapper.find('.rc-accordion-header').trigger('click');
     await nextTick();
     await flushPromises();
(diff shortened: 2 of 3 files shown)


pkg/rancher-components/src/components/RcButton/RcButton.vue
```diff
@@ -238,12 +238,7 @@ button {
     padding: 0;
     background-color: transparent;
 
-    &:focus, &.focused {
-      @include focus-outline;
-      outline-offset: 0;
-    }
-
-    &:focus-visible {
```
codyrancher, line 246: We can move this change specifically into RcAccordion but I think it may be more broadly applicable.

When clicking on a ghost button with a mouse we would end up with the keyboard focus styling which appears out of place.

pkg/rancher-components/src/components/RcAccordion/RcAccordion.vue
```diff
@@ -0,0 +1,159 @@
...
+    font-size: 18px;
+    line-height: 24px;
+    color: var(--body-text);
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+  }
+
+  &-body {
+    padding: var(--gap) calc(var(--gap-md) + var(--gap-lg)) var(--gap-md) calc(var(--gap-md) + var(--gap-lg));
```
codyrancher, line 156: I admit this look a little odd with all of the calc() invocations. If we'd like to change this to just straight values we can.

The reason I left this in is because I deviated from how the gaps and padding was done in the DS. 

The DS used multiple layers of gaps and padding in the open state vs just padding in the closed state. I wanted to replicate the same spacing without doing the layering. It allowed us to style the header and body separately which code wise felt cleaner to me. But I still wanted to be able reference the initial layered values.

PR #16596, opened 2026-02-06
Improve screen reader compatibility of card component

### LINKED ISSUE #16581: [a11y] Charts: Filters in cards are not announced by screen readers
**Setup**
- Rancher version: v2.14-head
- Rancher UI Extensions: NA
- Browser type & version: Firefox/Chrome

**Describe the bug**

Each card is accessible by keyboard; the card contents are also accessible by keyboard. When a tag like "Partners" or "Linux Only" receives focus, the screen reader announces the parent card action.

**To Reproduce**

1. Navigate to Local Cluster => Apps => Charts
2. Navigate to a tag within a card, like "Linux Only", using the keyboard while screen reader is active

**Result**

The screen reader will announce the contents of the parent card.

**Expected Result**

The screen reader should announce the contents of the tag that has focus, for example: "Linux Only"

**Screenshots**

NA

**Additional context**

A workaround exists: Filters can be applied and announced via the sidebar to the right.

rak-phillip:
### Summary
This improves screen reader compatibility of the card component by flattening the tab order of interactive elements in the dom.

Fixes #16581 

### Technical notes summary

The previous implementation nested interactive elements, making screen reader behavior unreliable depending on which product a customer might choose. The card container was the primary interaction target (`role="button"`) and could nest several interactive elements within, like tags to apply filters (`role="link"`) or dropdown menus to apply actions. While nesting interactive elements is a technically valid HTML5 implementation, it is strongly discouraged when designing with accessibility in mind.

This change also aims to improve roles of interactive elements in the cards to better express their intent.

Cards should support the same keyboard/mouse interactions.

Screen readers should accurately announce all interactive card elements.

Areas to test include:

- Cluster Tools
- App Charts
- Extensions

Card interactions can be affected in the following ways:

- Mouse interactions for cards and card actions
- Keyboard interaction for cards and card actions
- Card interactions when the optional title is omitted
- Visual elements when hovering/activating a card should remain consistent

### Screenshot/Video

image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/pkg/rancher-components/src/components/RcItemCard/RcItemCard.vue
+++ b/pkg/rancher-components/src/components/RcItemCard/RcItemCard.vue
@@ -167,6 +167,7 @@ const cardMeta = computed(() => ({
   actionMenuLabel: props.actions && t('itemCard.actionMenu.label', { cardTitle: labelText(props.header.title) }),
 }));
 
+const cursorValue = computed(() => props.clickable ? 'pointer' : 'auto');
 </script>
 
 <template>
@@ -321,6 +322,7 @@ $image-medium-box-width: 48px;
   border-radius: var(--border-radius-md);
   border: 1px solid var(--border);
   background: var(--body-bg);
+  cursor: v-bind(cursorValue);
 
   &.clickable:hover {
     border-color: var(--primary);


shell/pages/c/_cluster/apps/charts/index.vue
```diff
@@ -696,6 +696,7 @@ export default {
             :content="card.content"
             :value="card.rawChart"
             variant="medium"
+            role="link"
```
rak-phillip, line 699: This card previously received a role of "button", where a role of "link" better describes its behavior - it routes to a new page instead of performing an action on the given page.

shell/pages/c/_cluster/apps/charts/AppChartCardFooter.vue
```diff
@@ -30,26 +31,34 @@ function onClickItem(type: string, label: string) {
...
-          {{ label }}
-          <span v-if="footerItem.labels.length > 1 && j !== footerItem.labels.length - 1">, </span>
+          <rc-button
+            v-clean-tooltip="footerItem.labelTooltip"
+            variant="ghost"
+            class="app-chart-card-footer-button secondary-text-link"
+            data-testid="app-chart-card-footer-item-text"
+            :aria-label="t('catalog.charts.appChartCard.footerItem.ariaLabel', { filter: label })"
+            @click="onClickItem(footerItem.type, label)"
+          >
```
rak-phillip, line 49: The previous implementation had `rc-item-card-action` with a tabindex of 0 and no role. These filters perform an action on the page, so a button best fits their behavior. I opted for the ghost variant so that we can control the non-standard styling of the filter buttons.

shell/assets/translations/en-us.yaml
```diff
@@ -1183,7 +1183,7 @@ catalog:
       subHeaderItem:
         missingVersionDate: Last updated date is not available for this chart
       footerItem:
-        ariaLabel: Apply filter
+        ariaLabel: 'Apply {filter} filter'
```
rak-phillip, line 1186: This has been updated so that screen readers can better announce which filter will be applied.

pkg/rancher-components/src/components/RcItemCard/RcItemCard.vue
```diff
@@ -214,7 +210,16 @@ const cardMeta = computed(() => ({
...
+          <div
+            class="item-card-header-left"
+            :role="cardMeta.role"
+            :tabindex="cardMeta.tabIndex"
+            :aria-label="cardMeta.ariaLabel"
+            :data-testid="`item-card-${id}`"
+            @click.self="_handleCardClick"
+            @keydown.enter="_handleCardClick"
+            @keydown.space.prevent="_handleCardClick"
+          >
```
rak-phillip, line 224: This is the most impactful change. We move the tabindex, role, and primary click behavior to the `item-card-header-left`. The primary goal of performing this change is to flatten the tab order in the dom so that screen readers can more reliably announce all content in the card - certain screen readers can have issues reliably announcing nested interactive content, so this change helps improve reliability across all products while retaining existing keyboard/mouse interactions.

rak-phillip:
@momesgin you're the original author of the cards, so I want to run this by you. My concern is that there's an overlooked edge case that can cause the card behavior to regress. Let me know if you have any questions about the approach.

momesgin:
@rak-phillip seems we don't have the mouse pointer style on hovering on the cards anymore. The whole card is clickable on both Charts and Extensions pages.

rak-phillip:
@momesgin good catch! This has been fixed in 5fbe97bc3daaf6ed4ef4f3b8d71a945b468e60db

momesgin (dismissed):
LGTM

PR #16614, opened 2026-02-10
[16625][e2e] Deployment fix, add structuredClone to all tests

### LINKED ISSUE #16625: [e2e] Fix the deployments.spec.ts having a constant fail
The deployments.spec.ts is having a flaky test that is causing it to almost always fail after the addition of the cards.

The problem happens because the RUNNING is on the screen so the condition to have the RUNNING it is not enough. Than clicking on the DECREASE causes a conflict and it doesnt work.

We need to wait for the correct UI to be updated with the 2 PODs as RUNNING instead.

marcelofukumoto:
### Summary
Fixes #16625

- The flaky is caused by the UI changed
- To fix it was necessary to update the logic to follow the new UI

- Cypress E2E test for deployments.spec.ts

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/cypress/e2e/po/pages/explorer/workloads/workloads.po.ts
+++ b/cypress/e2e/po/pages/explorer/workloads/workloads.po.ts
@@ -90,6 +90,13 @@ export class WorkloadDetailsPageBasePo extends BaseDetailPagePo {
   }
 
   /**
+   * Wait until there's exactly one pods status count element
+   */
+  waitForSinglePodsStatusCount(): Cypress.Chainable {
+    return cy.get('[data-testid="rc-counter-badge"]').should('have.length', 1);
+  }
+
+  /**
    * Wait for the workload details page to be fully loaded
    * This ensures the masthead title is present and the scaling controls are ready
    */
--- a/cypress/e2e/tests/pages/explorer2/workloads/deployments.spec.ts
+++ b/cypress/e2e/tests/pages/explorer2/workloads/deployments.spec.ts
@@ -125,7 +125,7 @@ describe('Deployments', { testIsolation: 'off', tags: ['@explorer2', '@adminUser
         .should('contain.text', 'Running');
 
       // Wait until there's only one pods status count element
-      cy.get('[data-testid="rc-counter-badge"]').should('have.length', 1);
+      workloadDetailsPage.waitForSinglePodsStatusCount();
 
       workloadDetailsPage.podsStatusCount().should('be.visible', MEDIUM_TIMEOUT_OPT)
         .should('contain', '2');


cypress/e2e/tests/pages/explorer2/workloads/deployments.spec.ts
```diff
@@ -118,13 +118,18 @@ describe('Deployments', { testIsolation: 'off', tags: ['@explorer2', '@adminUser
...
-
       workloadDetailsPage.replicaCount().should('contain', '2', MEDIUM_TIMEOUT_OPT);
 
       // Verify pod status shows healthy scaling state
       workloadDetailsPage.podsStatus().should('be.visible', MEDIUM_TIMEOUT_OPT)
         .should('contain.text', 'Running');
-      workloadDetailsPage.podScaleDown().should('be.enabled').click();
+
+      // Wait until there's only one pods status count element
+      cy.get('[data-testid="rc-counter-badge"]').should('have.length', 1);
```
richard-cox: this is one of the places where the `cy.get` should be in a PO somewhere (which can then be used by other tests). 

given that we really need this merged though lets leave as is

marcelofukumoto: Updated. I will run the tests and see if all good.

cypress/e2e/po/pages/explorer/workloads/workloads.po.ts
```diff
@@ -85,6 +85,17 @@ export class WorkloadDetailsPageBasePo extends BaseDetailPagePo {
...
 
+  podsStatusCount(): Cypress.Chainable {
+    return this.card().podsStatusCount();
+  }
+
+  /**
+   * Wait until there's exactly one pods status count element
+   */
+  waitForSinglePodsStatusCount(): Cypress.Chainable {
+    return cy.get('[data-testid="rc-counter-badge"]').should('have.length', 1);
```
richard-cox, line 96: Sorry, misleading advice. There would be a new PO which represented RCCounterBadges in the target area and then this function would count them.

This really needs to go in though, so lets address at some point in the future

marcelofukumoto: A yes, not sure why I added to this one and not to the CardPO too. And for that one as well should already have. I just realized I created the same function just to get the same badge as well.
I will create another issue to improve it.

marcelofukumoto: Created https://github.com/rancher/dashboard/issues/16631

PR #16644, opened 2026-02-12
Fix incorrect parent nullification in diff() when nested children exist

### LINKED ISSUE #16414: diff() nulls parent objects when only nested keys exist
**Setup**
- Rancher version: 2.13.1, master
- Rancher UI Extensions: NA
- Browser type & version: Any

**Describe the bug**

When modifying nested values in rancher-logging chart, specifically `hostTailer.workloadMetaOverrides` the UI correctly generates the new data but then immediately overwrites the entire object with `null` during the object diffing process. This results  (labels/annotations) failin to save.

**To Reproduce**

1. Install rancher-logging (rancher-logging 4.10.0-rancher.19)
2. edit helm values and add 
~~~
logging:
  enabled: true
  fluentbitDisabled: true
  hostTailer:
    enabled: true
    workloadMetaOverrides:
      annotations:
        hello: world
      labels:
        key: hellowrold
~~~
3. Save

**Result**

~~~
root@master-0:~# kubectl get hosttailer -A -o yaml
apiVersion: v1
items:
- apiVersion: logging-extensions.banzaicloud.io/v1alpha1
  kind: HostTailer
  metadata:
    annotations:
      meta.helm.sh/release-name: rancher-logging
      meta.helm.sh/release-namespace: cattle-logging-system
    creationTimestamp: "2026-01-21T06:49:53Z"
    generation: 1
    labels:
      app.kubernetes.io/managed-by: Helm
    name: hosttailer
    namespace: cattle-logging-system
    resourceVersion: "532722"
    uid: fa01ff80-d3b4-4e63-8547-21916d896fad
  spec:
    enableRecreateWorkloadOnImmutableFieldChange: false
    image:
      imagePullSecrets: []
    workloadMetaOverrides: {}
    workloadOverrides: {}
~~~

**Expected Result**

```
root@master-0:~# kubectl get hosttailer -A -o yaml
apiVersion: v1
items:
- apiVersion: logging-extensions.banzaicloud.io/v1alpha1
  kind: HostTailer
  metadata:
    annotations:
      meta.helm.sh/release-name: rancher-logging
      meta.helm.sh/release-namespace: cattle-logging-system
    creationTimestamp: "2026-01-21T06:22:54Z"
    generation: 4
    labels:
      app.kubernetes.io/managed-by: Helm
    name: hosttailer
    namespace: cattle-logging-system
    resourceVersion: "528992"
    uid: 7a0d06d4-f1c9-4e41-bf03-718dc7c13592
  spec:
    enableRecreateWorkloadOnImmutableFieldChange: false
    image:
      imagePullSecrets: []
    workloadMetaOverrides:
      annotations:
         hello: world
      labels:
         key: hellowrold
    workloadOverrides: {}
kind: List

```

**Screenshots**

On the rancher UI, edit charts -> rancher-logging -> edit values.yaml
Image

**Additional context**

[SURE-10926]

skanakal:
### Summary
Fixes  https://github.com/rancher/dashboard/issues/16414

### Technical notes summary

Currently the diff logic flattened nested keys and compared them as strings... When only nested children existed in the updated object, the parent key was incorrectly treated as missing and nullified... This resulted in valid nested updates being overwritten with null.(screenshot attached on the issue)
- The fix addresses the, parent keys are not treated as missing when nested descendants exist.
- preventNull behavior remain unchanged.

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/utils/__tests__/object.test.ts
+++ b/shell/utils/__tests__/object.test.ts
@@ -210,6 +210,83 @@ describe('fx: diff', () => {
 
     expect(result).toStrictEqual(expected);
   });
+  it('should preserve nested object values when transitioning from empty object to populated object', () => {
+    const from = { parent: { child: { config: {} } } };
+    const to = {
+      parent: {
+        child: {
+          config: {
+            annotations: { hello: 'world' },
+            labels:      { key: 'value' }
+          }
+        }
+      }
+    };
+
+    const result = diff(from, to);
+    const expected = {
+      parent: {
+        child: {
+          config: {
+            annotations: { hello: 'world' },
+            labels:      { key: 'value' }
+          }
+        }
+      }
+    };
+
+    expect(result).toStrictEqual(expected);
+  });
+
+  it('should preserve explicit empty object when clearing nested object values', () => {
+    const from = { parent: { child: { config: { annotations: { hello: 'world' } } } } };
+    const to = { parent: { child: { config: {} } } };
+
+    const result = diff(from, to);
+    const expected = { parent: { child: { config: {} } } };
+
+    expect(result).toStrictEqual(expected);
+  });
+
+  it('should not emit nested empty objects when there are no differences', () => {
+    const from = { parent: { child: { config: { annotations: { hello: 'world' } } } } };
+    const to = { parent: { child: { config: { annotations: { hello: 'world' } } } } };
+
+    const result = diff(from, to);
+    const expected = {};
+
+    expect(result).toStrictEqual(expected);
+  });
+
+  it('should correctly nullify nested keys when removed', () => {
+    const from = {
+      parent: {
+        child: {
+          config: {
+            annotations: { hello: 'world' },
+            labels:      { key: 'value' }
+          }
+        }
+      }
+    };
+    const to = { parent: { child: { config: { annotations: { hello: 'world' } } } } };
+
+    const result = diff(from, to);
+    const expected = { parent: { child: { config: { labels: { key: null } } } } };
+
+    expect(result).toStrictEqual(expected);
+  });
+
+  it('should not nullify child keys when parent object is updated', () => {
+    const from = { parent: { child: { config: {} } } };
+    const to = { parent: { child: { config: { annotations: { hello: 'world' } } } } };
+
+    const result = diff(from, to);
+    const expected = { parent: { child: { config: { annotations: { hello: 'world' } } } } };
+
+    expect(result).toStrictEqual(expected);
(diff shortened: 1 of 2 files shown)


shell/utils/object.js
```diff
@@ -220,8 +220,13 @@ export function diff(from, to, preventNull = false) {
     if ( Array.isArray(toVal) || Array.isArray(fromVal) ) {
       // Don't diff arrays, just use the whole value
       res[k] = toVal;
-    } else if ( isObject(toVal) && isObject(from[k]) ) {
-      res[k] = diff(fromVal, toVal);
+    } else if ( isObject(toVal) && isObject(fromVal) ) {
+      const nested = diff(fromVal, toVal, preventNull);
+
+      // Only assign if there are actual differences
+      if (!isEmpty(nested)) {
```
aalves08: I think we'll hit problems if we want to set a nested object to an **empty** object, like:

```
from = { config: { setting: 'value' } }
to = { config: {} }  // User explicitly wants to clear config

// Expected: { config: {} }
// With fix: {} (nothing returned because isEmpty({}) is true)
```

Please add some unit tests to validate not only your usecase, but also cover the "angle" described above 🙏

aalves08 (changes requested):
left a comment

shell/utils/object.js
skanakal: @aalves08 Thank you for review. I have added a check to force {} when a populated object is cleared (isEmpty(toVal) && !isEmpty(fromVal)), this will ensures the diff returns { config: {} } instead of an empty result... and also have added some unti tests that covers the scenario... Thank you!

shell/utils/__tests__/object.test.ts
```diff
@@ -212,6 +212,56 @@ describe('fx: diff', () => {
   });
 });
 
+describe('nested object behavior', () => {
```
aalves08: This is an improvement, but it's outside of the `describe('fx: diff'` block, where it should belong.

Also, this a pure function unit test and not a test covering the scenario described in https://github.com/rancher/dashboard/issues/16414.

We should cover the scenario in https://github.com/rancher/dashboard/issues/16414, which is what QA is going to be looking at.

aalves08 (changes requested):
reviewed

shell/utils/__tests__/object.test.ts
skanakal: @aalves08 thank you for the review.
I've moved the tests under `describe('fx: diff'` block and updated it. 
PTAL.

aalves08 (approved):
LGTM

PR #16686, opened 2026-02-18
Improve custom resource quota validation

### LINKED ISSUE #16608: Project Custom Resource Quotas being created without identifiers
**Setup**
- Rancher version: 

[Rancher](https://github.com/rancher/rancher)	v2.14-f52b06b8ebfd9a2fde9daf54612d74569e8a7bc4-head
[Dashboard](https://github.com/rancher/dashboard)	master bee169a
[Rancher CLI](https://github.com/rancher/cli)	v2.12.0-rc.1

**Describe the bug**
Found while testing https://github.com/rancher/dashboard/issues/14797

**To Reproduce**
1- Select a cluster
2- Go to Cluster > Projects/Namespaces
3- Create a Project
4- Under Resource Quotas, create a custom resource with only the mandatory field 
5- Observe that the object is posted to the backend, but the UI doesn't exhibit the resource 
6- Try to create a custom resource with all fields empty

**Result**

https://github.com/user-attachments/assets/dac5a3c6-e5b2-467a-85ca-642f95084e2d

Image

https://github.com/user-attachments/assets/b9ef485b-34a7-4e4d-9620-2ea1359a7c9d

Image

**Expected Result**
UI must disable the Create button if the mandatory fields are not set

rak-phillip:
### Summary
This adds more robust validation to the custom resource quotas so that the submit button will be disabled when a resource identifier is not provided by the user.

Fixes #16608 

- Add validation for custom resource identifiers
- Fix issue with removing custom quotas
- Prevent deletion of values for duplicate keys
- Write unit tests for the change Resource Quota behavior

### Technical notes summary

I encountered some edge cases that have been improved upon. I left comments throughout to explain the changes. 

See issue for more details:

- Creating custom quotas should require a Resource Identifier for all line items
- Creating predefined quotas should behave as the form did before

Custom quota creation could still have scenarios where empty resource identifiers are allowed. 

### Screenshot/Video

image

image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/form/ResourceQuota/ProjectRow.vue
+++ b/shell/components/form/ResourceQuota/ProjectRow.vue
@@ -47,8 +47,8 @@ export default {
 
   data() {
     return {
-      customType:      '',
-      localTypeValues: [],
+      customType:         '',
+      previousTypeValues: [],
     };
   },
 
@@ -58,7 +58,7 @@ export default {
     } else {
       this.customType = this.type;
     }
-    this.localTypeValues = [...this.typeValues];
+    this.previousTypeValues = [...this.typeValues];
   },
 
   computed: {
@@ -112,8 +112,15 @@ export default {
 
   watch: {
     typeValues: {
+      /**
+       * Intentionally uses `oldValues` (not `_newValues`) so that
+       * `previousTypeValues` always reflects the committed state before the
+       * current change. `deleteResourceLimits` relies on this snapshot to
+       * detect duplicate keys: if a key still appears twice in the old list
+       * we know another row owns it and must not delete the shared limit.
+       */
       handler(_newValues, oldValues) {
-        this.localTypeValues = [...oldValues];
+        this.previousTypeValues = [...oldValues];
       },
       deep: true
     }
@@ -173,7 +180,10 @@ export default {
       const usedLimit = this.value?.spec.namespaceDefaultResourceQuota?.limit;
 
       if (isExtended) {
-        const matchesForKey = this.localTypeValues.filter((typeValue) => {
+        // `previousTypeValues` holds the previous snapshot of typeValues.
+        // Counting matches against the old list lets us detect whether another
+        // row already owns this key before we delete the shared limit entry.
+        const matchesForKey = this.previousTypeValues.filter((typeValue) => {
           const [, typeKey] = typeValue.split('.');
 
           return resourceKey === typeKey;


shell/edit/management.cattle.io.project.vue
```diff
@@ -158,7 +162,7 @@ export default {
     },
 
     removeQuota(key) {
-      const isExtended = key.startsWith(`${ TYPES.EXTENDED }.`);
+      const isExtended = key.startsWith(`${ TYPES.EXTENDED }`);
```
rak-phillip, line 165: The key can sometimes be "extended", without any resource identifier. We want to ensure that keys are not erroneously removed in this event.

shell/edit/management.cattle.io.project.vue
```diff
@@ -110,6 +111,9 @@ export default {
     }
   },
   methods: {
+    validateQuotas(isValid) {
+      this.isQuotasValid = isValid;
+    },
```
rak-phillip, line 116: The quotas validation logic is primarily in `shell/components/form/ResourceQuota/Project.vue`. Having a simple boolean here allows for easily toggling the submit button.

shell/components/form/ResourceQuota/ProjectRow.vue
```diff
@@ -140,7 +168,35 @@ export default {
...
+
+        if (limit?.extended && typeof this.value.spec.resourceQuota?.limit?.extended[resourceKey] !== 'undefined') {
+          delete this.value.spec.resourceQuota.limit.extended[resourceKey];
+        }
+        if (usedLimit?.extended && typeof this.value.spec.namespaceDefaultResourceQuota?.limit?.extended[resourceKey] !== 'undefined') {
+          delete this.value.spec.namespaceDefaultResourceQuota.limit.extended[resourceKey];
+        }
+
+        return;
+      }
```
rak-phillip, line 208: I encountered some edge cases that made editing custom quotas quite frustrating:

1. Custom quotas would not be deleted on edit. For example, I entered a Resource Identifier of test4, along with limits. I then decide that I want this to be test5 and I specify the limits again. Upon saving and editing, both test4 and test5 would exist with quotas defined
2. Duplicate custom quota keys could clear values for a previously specified quota. For example, create a resource identifier of test1 and specify limits; save the form. Edit the form and add a new resource identifier of test1234 - the original "test1" should retain its limits.

shell/components/form/ResourceQuota/ProjectRow.vue
```diff
@@ -99,6 +110,15 @@ export default {
     }
   },
 
+  watch: {
+    typeValues: {
+      handler(_newValues, oldValues) {
+        this.localTypeValues = [...oldValues];
+      },
```
rak-phillip, line 124: This is used in `deleteResourceLimits()` for prevention of key deletion while typing.

shell/components/form/ResourceQuota/ProjectRow.vue
codyrancher: I found it a little difficult to understand this and part of the confusion came from not realizing that localTypeValues was using the `oldValues` instead of the new values.

I'm wondering if we'd like to put comments in the code in both places to at least highlight the need to use more scrutiny when reading in the future.

shell/components/form/ResourceQuota/ProjectRow.vue
```diff
@@ -140,7 +168,35 @@ export default {
...
+         */
+        if (matchesForKey.length >= 2) {
+          return;
+        }
+
+        if (limit?.extended && typeof this.value.spec.resourceQuota?.limit?.extended[resourceKey] !== 'undefined') {
+          delete this.value.spec.resourceQuota.limit.extended[resourceKey];
+        }
+        if (usedLimit?.extended && typeof this.value.spec.namespaceDefaultResourceQuota?.limit?.extended[resourceKey] !== 'undefined') {
+          delete this.value.spec.namespaceDefaultResourceQuota.limit.extended[resourceKey];
```
codyrancher, line 204: I know it's already elsewhere in the code but should we use Vue.delete for reactivity?

shell/components/form/ResourceQuota/Project.vue
```diff
@@ -122,16 +154,19 @@ export default {
       :add-allowed="remainingTypes().length > 0"
       :mode="mode"
       @remove="emitRemove"
```
codyrancher, line 162: Would we need to emit the validation-changed event when removing too? If something was invalid but we removed it I think we need to trigger it.

https://github.com/user-attachments/assets/716753ee-e5e9-4e41-9855-33c2d8b8f82f

shell/components/form/ResourceQuota/ProjectRow.vue
rak-phillip: No, the `Vue.delete` method has been removed since the Vue3 migration[^1]. It shouldn't be required anymore with the proxy-based change detection. 

With that said, we really shouldn't be mutating the value prop like we do here, but I've opted to keep it intact with instead of performing a larger refactor of the component interactions.

[^1]: https://v3-migration.vuejs.org/breaking-changes/#removed-apis

shell/components/form/ResourceQuota/ProjectRow.vue
rak-phillip: This is a good point. I've done the following to help clarify the intent of this code:

1. Rename `localTypeValues` to `previousTypeValues`
2. Add comments to this function and the associated watcher to clarify what we intend to do in order to ensure that this works

shell/components/form/ResourceQuota/Project.vue
rak-phillip: Good catch. This should be fixed now via 338af72b6221ed9f797016e69f1b35db78bc1b61

PR #16690, opened 2026-02-18
[16676] appco improve the flow to create a repository and install an application from suse app collection

### LINKED ISSUE #16676: [AppCo] Improve the flow to create a Repository and install an application from SUSE App Collection

Implement the App Collection based on the requirements provided by Neil.
The problem is that the flow to add App Collection charts is not user friendly, so we want to improve the UX.

- We want to be able to remove the step of the ImagePullRequest
- Facilitate the creation of the OCI Repository, adding a hardcoded option and specific configuration for SUSE App Collection
- On the chart list, add a Banner to facilitate the creation of the OCI Repository configuration if doesn't exist
- When installing, give options for the user to select the previous created authentication so it can automatically create an Image Pull Request. Or select a previously created one.

- There is a lot of improvements on this, so for now it is to just improve that simple flow, moving to Fleet later on.

**Additional context**

- That is part of a bigger plan for App Collection https://github.com/rancher/dashboard/issues/14627

marcelofukumoto:
### Summary
Fixes #16676

- Changes to the List of Repositories to add a new column
- Changes to the List of Installed Apps to add an Icon on the chart to represent SUSE AppCo
- Changes to Create/Edit of Repositories to add SUSE App Co option
- Changes to List of Charts to add the icon to represent SUSE AppCo
- Changes to Install Chart, added banner for SUSE AppCo
- Added functional flow for it as well to crate an ImagePullSecret

### Technical notes summary

- The ImagePullSecret is used using the Secret from the Repository
- You can't select a new ImagePullSecret you automatically will get one created first time using installation of that, automatically that one will be used for all the other cases. It is one per ImagePullSecret

- Create new repositories for Suse App Co
  - suse-application-collection as default name with default description
  - It should be maintained only when Suse App Co
  - SHOW CONFIGURATION should have the CARDS disabled
- List of Repositories should have the new TYPE SUSE App Co
- List of charts should have a tooltip on the Chart List when it is a SUSE App Co repository
- install charts Suse App Co
  - First: If there is a IMAGE PULL SECRET already created, it will default to that one (remembering that this is specific to NAMESPACE)
  - Second: If you have access to the DECODED DATA of the REPO SECRET, it will say it will generate one using it.
  - Third: If you dont have access and there is no IMAGE PULL SECRET, you will be locked to MANUAL creation which will create a new IMAGE PULL SECRET.
  - Important that changing NAMESPACE changes the conditions as well because the FIRST consider the IMAGE PULL SECRET of the selected namespace.
- Upgrade charts Suse App Co
  - Edit mode should be default to MANUAL selection
  - Same previous rules apply but you cant change namespace

- Install and upgrade of all types of charts
- Create and edit repositories, changed the way that selects the types

### Checklist
- [x] The PR has a Milestone 


cypress/e2e/po/edit/catalog.cattle.io.clusterrepo.po.ts
```diff
@@ -1,13 +1,13 @@
...
 import NameNsDescription from '@/cypress/e2e/po/components/name-ns-description.po';
 import ResourceDetailPo from '@/cypress/e2e/po/edit/resource-detail.po';
-import RadioGroupInputPo from '@/cypress/e2e/po/components/radio-group-input.po';
 import LabeledInputPo from '@/cypress/e2e/po/components/labeled-input.po';
 import UnitInputPo from '@/cypress/e2e/po/components/unit-input.po';
 import AsyncButtonPo from '@/cypress/e2e/po/components/async-button.po';
 import CheckboxInputPo from '@/cypress/e2e/po/components/checkbox-input.po';
 import SelectOrCreateAuthPo from '@/cypress/e2e/po/components/select-or-create-auth.po';
 import FooterPo from '@/cypress/e2e/po/components/footer.po';
+import RcItemCardPo from '~/cypress/e2e/po/components/rc-item-card.po';
```
richard-cox:
```suggestion
import RcItemCardPo from '@/cypress/e2e/po/components/rc-item-card.po';
```

cypress/e2e/po/pages/extensions.po.ts
```diff
@@ -83,7 +83,7 @@ export default class ExtensionsPagePo extends PagePo {
     appRepoCreate.waitForPage();
 
     // fill the form
-    appRepoCreate.repoRadioBtn().set(1);
+    appRepoCreate.repoRcItemCard('git-repo').click();
```
richard-cox: nit

I like this change because it explictely ties the option with the value we want. However we make these calls a lot with just a text string, to avoid issues with that in the future we could make that the strings into an enum or have hardcoded helpers. For example

appRepoCreate.gitRepoCard().click()

appRepoCreate.selectGitRepoCard()
appRepoCreate.selectHelmUrlCard()

shell/pages/c/_cluster/apps/charts/index.vue
```diff
@@ -570,6 +608,29 @@ export default {
       color="error"
       :label="err"
     />
+    <Banner
+      v-if="canCreateRepos && showAppCollectionBanner && !hideBannerPref"
```
richard-cox: appCo is a prime only thing, so i think we should only show if `const isPrime = getVersionData().RancherPrime === 'true';`

Might need double checking though

shell/pages/c/_cluster/apps/charts/index.vue
```diff
@@ -570,6 +608,29 @@ export default {
       color="error"
       :label="err"
     />
+    <Banner
```
richard-cox, line 626: this introduces a lot of white space. the gap between banner and search should be the same as banner and dark repo/category/status card

Image

vs

Image

shell/pages/c/_cluster/apps/charts/index.vue
```diff
@@ -570,6 +608,29 @@ export default {
...
+      @close="closeSuseAppCollectionBanner"
+    >
+      <RichTranslation
+        k="catalog.charts.appCollectionRepoMissing"
+        tag="div"
+        :raw="true"
+      >
+        <template #repoCreate="{ content }">
+          <router-link
+            :to="{ name: 'c-cluster-product-resource-create', params: { resource: 'catalog.cattle.io.clusterrepo', cluster: $route.params.cluster, product: $store.getters['productId'] }, query: { target: CLUSTER_REPO_TYPES.SUSE_APP_COLLECTION } }"
```
richard-cox: 'catalog.cattle.io.clusterrepo' --> CATALOG.CLUSTER_REPO

pkg/rancher-components/src/components/RcItemCard/RcItemCard.vue
```diff
@@ -25,9 +27,11 @@ type Label = {
 
 /**
  * Represents an image used in the card.
+ * Can be either a traditional image (src) or an icon (icon).
  */
 type Image = {
```
richard-cox: the src or icon stipulation can be achieved with

```
type Image =
 | { src?: never, icon: RcIconType; alt: Label; }
 | { src: string; icon?: never, alt: Label;  };
```

shell/assets/translations/en-us.yaml
```diff
@@ -1386,9 +1394,18 @@ catalog:
...
+        description: Git repository containing Helm charts or cluster templates
+      http: 
+        title: Helm Repository
+        description: HTTP(S) URL pointing to a Helm chart repository index
+      oci: 
+        title: OCI Repository
+        description: OCI-compliant registry hosting Helm charts as artifacts
+      suseAppCollection: 
+        title: SUSE App Collection
+        description: Curated, enterprise-ready applications from SUSE® Application Collection
```
```suggestion
        description: Curated, enterprise-ready applications from SUSE Application Collection
```

i think, as we don't `R` any other reference

shell/models/catalog.cattle.io.clusterrepo.js
```diff
@@ -143,14 +143,22 @@ export default class ClusterRepo extends SteveModel {
     return this.metadata?.state?.name === 'active';
   }
 
+  get isSuseAppCollection() {
+    return this.metadata?.annotations?.[CATALOG.SUSE_APP_COLLECTION];
```
richard-cox, line 147: This is a little brittle, as it only covers repos added by the UI and not via terraform, api or another gitops like way.

Can we primarily look at the annotation, and if not present also check the value.spec.url starts with the appco url (we could hardcode the url for the moment, possibly tweak in future)

shell/models/catalog.cattle.io.app.js
```diff
@@ -106,6 +106,14 @@ export default class CatalogApp extends SteveModel {
     return this.spec?.chart?.metadata?.version;
   }
 
+  /**
+   * Get if it is an installation from SUSE App Collection
+   * Specific installation done on the UI
+   **/
+  get isSuseAppCollection() {
```
richard-cox: This is another tricky one.

If someone installs the chart outside of the UI it won't have this annotation?

To cover that case can we fall back on if the repo has the app co url?

shell/edit/catalog.cattle.io.clusterrepo.vue
```diff
@@ -1,15 +1,19 @@
-<script>
+<script lang="ts">
```
richard-cox, line 1: comment only.

+1 to this sort of thing

shell/edit/catalog.cattle.io.clusterrepo.vue
```diff
@@ -18,30 +22,77 @@ export default {
...
+    UnitInput,
+    RcItemCard,
   },
 
   mixins: [CreateEditView],
 
   data() {
-    const clusterRepoType = !!this.value.spec.gitRepo ? CLUSTER_REPO_TYPES.GIT_REPO : this.value.isOciType ? CLUSTER_REPO_TYPES.OCI_URL : CLUSTER_REPO_TYPES.HELM_URL;
+    // Determine the cluster repo type based on existing values (for edit mode)
+    const clusterRepoType = !!this.value.spec.gitRepo ? CLUSTER_REPO_TYPES.GIT_REPO : this.value.isOciType ? this.value.metadata.annotations[CATALOG.SUSE_APP_COLLECTION] ? CLUSTER_REPO_TYPES.SUSE_APP_COLLECTION : CLUSTER_REPO_TYPES.OCI_URL : CLUSTER_REPO_TYPES.HELM_URL;
```
richard-cox: This is getting a bit verbose, can it be split into if/else?

shell/edit/catalog.cattle.io.clusterrepo.vue
```diff
@@ -18,30 +22,77 @@ export default {
...
+      {
+        id:      CLUSTER_REPO_TYPES.OCI_URL,
+        header:  { title: { key: 'catalog.repo.target.oci.title' } },
+        image:   { src: require('@shell/assets/images/providers/oci-open-containers.svg'), alt: { key: 'catalog.repo.target.oci.title' } },
+        content: { key: 'catalog.repo.target.oci.description' },
+      },
+    ];
+
+    // Only show SUSE App Collection option for RancherPrime
+    if (getVersionData()?.RancherPrime) {
```
richard-cox: strickly speaking this should be `getVersionData().RancherPrime === 'true'` (RancherPrime could be `"false"`)

shell/edit/catalog.cattle.io.clusterrepo.vue
```diff
@@ -18,30 +22,77 @@ export default {
...
   data() {
-    const clusterRepoType = !!this.value.spec.gitRepo ? CLUSTER_REPO_TYPES.GIT_REPO : this.value.isOciType ? CLUSTER_REPO_TYPES.OCI_URL : CLUSTER_REPO_TYPES.HELM_URL;
+    // Determine the cluster repo type based on existing values (for edit mode)
+    const clusterRepoType = !!this.value.spec.gitRepo ? CLUSTER_REPO_TYPES.GIT_REPO : this.value.isOciType ? this.value.metadata.annotations[CATALOG.SUSE_APP_COLLECTION] ? CLUSTER_REPO_TYPES.SUSE_APP_COLLECTION : CLUSTER_REPO_TYPES.OCI_URL : CLUSTER_REPO_TYPES.HELM_URL;
+
+    const clusterRepoTargets = [
+      {
+        id:      CLUSTER_REPO_TYPES.HELM_URL,
+        header:  { title: { key: 'catalog.repo.target.http.title' } },
+        image:   { icon: 'helm' as RcIconType, alt: { key: 'catalog.repo.target.http.title' } },
```
richard-cox, line 55: for me helm + git don't show an icon/image

Image

shell/edit/catalog.cattle.io.clusterrepo.vue
richard-cox: which is the case on community docker builds (the is card shown)

```
console.warn(`getVersionData()?.RancherPrime: "${ getVersionData()?.RancherPrime }"` );
...
catalog.cattle.io.clusterrepo.vue:64 getVersionData()?.RancherPrime: "false"
```

shell/edit/catalog.cattle.io.clusterrepo.vue
```diff
@@ -18,30 +22,77 @@ export default {
...
+      getVersionData,
+      clusterRepoTargets,
+      previousName:        '',
+      previousDescription: '',
     };
   },
 
+  created() {
+    // When creating a new repo with a target query parameter, initialize the form properly
+    const targetFromQuery = this.mode === _CREATE ? this.$route.query.target : null;
```
richard-cox: It's a bit odd but we track query params over in shell/config/query-params.js. so `target` needs adding there in a new `//` section

shell/edit/catalog.cattle.io.clusterrepo.vue
```diff
@@ -201,6 +288,20 @@ export default {
...
+      <template v-else-if="clusterRepoType === CLUSTER_REPO_TYPES.SUSE_APP_COLLECTION">
+        <div class="col span-6">
+          <LabeledInput
+            v-model:value.trim="value.spec.url"
+            :required="true"
+            :label="t('catalog.repo.oci.urlLabel')"
+            :placeholder="t('catalog.repo.oci.placeholder', null, true)"
+            :mode="mode"
+            data-testid="clusterrepo-oci-url-input"
+            :disabled="true"
```
richard-cox, line 314: We need to open this one up to more.

OCI repos can either be all (root url) or some (root url + qualifier) charts in the repo. This helps if app co grows very large and users want to avoid fetching them all. There's a bit of info in the rancher docs - https://ranchermanager.docs.rancher.com/how-to-guides/new-user-guides/helm-charts-in-rancher/oci-repositories#add-an-oci-based-helm-chart-repository

So we could visual split the url into two

domain - `oci://dp.apps.rancher.io`, not editable
path - `/charts`, editable

shell/edit/catalog.cattle.io.clusterrepo.vue
```diff
@@ -240,8 +341,11 @@ export default {
       :limit-to-namespace="false"
       :in-store="inStore"
       :allow-ssh="clusterRepoType !== CLUSTER_REPO_TYPES.OCI_URL"
-      generate-name="clusterrepo-auth-"
+      :pre-select="clusterRepoType === CLUSTER_REPO_TYPES.SUSE_APP_COLLECTION ? { selected: AUTH_TYPE._BASIC } : null"
+      :allow-none="clusterRepoType !== CLUSTER_REPO_TYPES.SUSE_APP_COLLECTION"
+      :generate-name="clusterRepoType === CLUSTER_REPO_TYPES.SUSE_APP_COLLECTION ? 'clusterrepo-appco-auth-' : 'clusterrepo-auth-'"
```
richard-cox: `clusterrepo-appco-auth-` is used in about three places in three different files. can it be a const somewhere?

shell/assets/images/providers/oci-open-containers.svg
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
```
richard-cox, line 1: did we get this one from UX?

shell/edit/catalog.cattle.io.clusterrepo.vue
```diff
@@ -124,35 +188,58 @@ export default {
...
   },
 };
 </script>
 
 <template>
   <form>
+    <h2>{{ t('catalog.repo.target.label') }}</h2>
+    <div class="row mb-10">
+      <div class="col span-12 target-groups">
+        <RcItemCard
```
richard-cox, line 233: in view mode (repo --> show config) all cards are clickable.

in that world the user should not be able to change from their selected type

shell/components/form/SelectOrCreateAuthSecret.vue
```diff
@@ -366,6 +390,18 @@ export default {
...
+      if ( this.fixedImagePullSecret ) {
+        out.unshift({
+          label: this.t('selectOrCreateAuthSecret.createImagePullSecret'),
+          value: AUTH_TYPE._IMAGE_PULL_SECRET,
+          kind:  'highlighted'
+        });
+      }
+
+      if (this.fixedHttpBasicAuth) {
+        out = out.filter((o) => o.label.search('clusterrepo-appco-auth-') === 0 || ['title', 'divider'].includes(o.kind) || o.value === AUTH_TYPE._BASIC);
```
richard-cox: does `clusterrepo-appco-auth-` need to passed in? this is a generic component and we should try to keep implemention specific stuff at the level above

shell/components/form/SelectOrCreateAuthSecret.vue
```diff
@@ -150,14 +155,24 @@ export default {
       default: false,
     },
 
+    fixedHttpBasicAuth: {
```
richard-cox, line 162: we don't do a good job of this everywhere, but these two should have jsdoc (it's not immediately clear what they do)

shell/components/form/SelectOrCreateAuthSecret.vue
```diff
@@ -581,6 +628,20 @@ export default {
           if ((this.selected === AUTH_TYPE._SSH) && this.showSshKnownHosts) {
             secret.data.known_hosts = base64Encode(this.sshKnownHosts || '');
           }
+
+          if (this.selected === AUTH_TYPE._IMAGE_PULL_SECRET) {
+            const config = {
+              auths: {
+                'dp.apps.rancher.io': {
```
richard-cox: this is another where where we have specific features inside here.

i think if we go down this path we should make it clear, so something like AUTH_TYPE._IMAGE_PULL_SECRET --> AUTH_TYPE._APP_CO_IMAGE_PULL_SECRET

richard-cox: That also applies to other references like selectOrCreateAuthSecret.imagePullSecret

shell/pages/c/_cluster/apps/charts/install.vue
```diff
@@ -370,49 +376,78 @@ export default {
 
     /* Look for annotation to say this app is a legacy migrated app (we look in either place for now) */
     this.migratedApp = (this.existing?.spec?.chart?.metadata?.annotations?.[CATALOG_ANNOTATIONS.MIGRATED] === 'true');
+
+    if (this.repo.isSuseAppCollection) {
+      let defaultSelectedSecret = await this.$store.getters['cluster/byId'](SECRET, `cattle-system/${ this.repo.spec.clientSecret.name }`);
```
richard-cox, line 383: only a few (admin) users have access to the cattle-system namespace.

so this line will return undefined. which is ok.. except throws an error on `this.defaultGeneratedNameForImagePullSecret = `${ this.selectedSecret.id.split('/')[1] }-image-pull-secret`;` below

shell/pages/c/_cluster/apps/charts/install.vue
```diff
@@ -370,49 +376,78 @@ export default {
...
+
+      if (!defaultSelectedSecret) {
+        try {
+          defaultSelectedSecret = (await this.$store.dispatch('cluster/find', { type: SECRET, id: `cattle-system/${ this.repo.spec.clientSecret.name }` }));
+        } catch (e) {
+          // If the secret doesn't exist, technically should be an impossible case
```
richard-cox: needs updating, given that this will be the case for most users.

if this fails, should we then attempt to get a secret from the default namespace (whatever's selected in the ns filter?)

shell/pages/c/_cluster/apps/charts/install.vue
```diff
@@ -370,49 +376,78 @@ export default {
...
+        }
+      }
+
+      this.selectedSecret = defaultSelectedSecret;
+      this.defaultGeneratedNameForImagePullSecret = `${ this.selectedSecret.id.split('/')[1] }-image-pull-secret`;
```
richard-cox: need to handle the case where defaultGeneratedNameForImagePullSecret and generatedNameForImagePullSecret could be undefined (standard user installing their first app co chart)

shell/assets/translations/en-us.yaml
```diff
@@ -1335,6 +1337,12 @@ catalog:
           }. Start by setting some basic information used by {vendor} to manage the App.
         nsCreationDescription: "To install the app into a new namespace enter it's name in the Namespace field and select it."
         createNamespace: "Namespace <code>{namespace}</code> will be created."
+        requiresImagePullSecret: "This Chart requires an Image Pull Secret in order for images to be succesfully pulled."
+        dontUseDefaultImagePullSecret: "Manually select an Image Pull Secret or create a new one with Username/Password"
+        selectedRepoAuth: "The authentication that will be used is based on: <b>{ repoAuthenticationName }</b>.<br>"
```
richard-cox: i think we can drop this line

shell/assets/translations/en-us.yaml
```diff
@@ -1335,6 +1337,12 @@ catalog:
...
+        generatedImagePullSecretBanner: "A new Image Pull Secret will be generated with the name: <b>{ imagePullSecretName }</b>."
```
richard-cox: then beef up this one

```suggestion
        generatedImagePullSecretBanner: "A new Image Pull Secret <b>{ imagePullSecretName }</b> will be generated from the Repository secret <b>{ repoAuthenticationName }</b>."
```

cypress/e2e/po/edit/catalog.cattle.io.clusterrepo.po.ts
marcelofukumoto: I need to pay more attention to those. Thanks!

shell/pages/c/_cluster/apps/charts/index.vue
marcelofukumoto: There is an error banner as well on this case.
image

Currently is:
SEARCH
24px
15px 
ERROR BANNER
15px
15px
SUSE BANNER
15px
MAINPAGE

I will change to be
SEARCH
24px
ERROR BANNER
15px
SUSE BANNER
24px
MAINPAGE

And 
SEARCH
24px
SUSE BANNER
24px
MAIN PAGE

shell/models/catalog.cattle.io.clusterrepo.js
marcelofukumoto: My way of thought is that for now that would be a special case for the UI for now.
But I will check to add this additional condition. I wanted to separated the OCI SUSE and the new SUSE option to avoid regression. But let me check on this.

shell/edit/catalog.cattle.io.clusterrepo.vue
marcelofukumoto: Oh thank you! I prefer to change, let me change it.

shell/edit/catalog.cattle.io.clusterrepo.vue
marcelofukumoto: It works for me... hmmm I wonder why

shell/edit/catalog.cattle.io.clusterrepo.vue
marcelofukumoto: Changed to === 'true'

shell/edit/catalog.cattle.io.clusterrepo.vue
marcelofukumoto: Done!

shell/assets/translations/en-us.yaml
marcelofukumoto: Changed

shell/pages/c/_cluster/apps/charts/index.vue
marcelofukumoto: Changed!

shell/pages/c/_cluster/apps/charts/index.vue
marcelofukumoto: Fixed using some 15px + 9px space when there are banners

shell/pages/c/_cluster/apps/charts/index.vue
marcelofukumoto: Missed this one. I removed to test and forgot to put back.
Thanks! FIxed

cypress/e2e/po/pages/extensions.po.ts
marcelofukumoto: Done!

shell/edit/catalog.cattle.io.clusterrepo.vue
marcelofukumoto: Changed!

shell/assets/images/providers/oci-open-containers.svg
marcelofukumoto: Yes, from @kwwii .

shell/edit/catalog.cattle.io.clusterrepo.vue
marcelofukumoto: That is true! That is a common mistake we have on other screens for not default fields. Let me fix this.

shell/components/form/SelectOrCreateAuthSecret.vue
marcelofukumoto: I will move as a parameter to be used together with the fixedHttpBasicAuth to fix the labels that will show as well.

marcelofukumoto: Changed

shell/components/form/SelectOrCreateAuthSecret.vue
marcelofukumoto: Added!

shell/components/form/SelectOrCreateAuthSecret.vue
marcelofukumoto: Ok! Done!

shell/edit/catalog.cattle.io.clusterrepo.vue
marcelofukumoto: Turned into it in the TYPES file, didnt file a better place. I hope it is ok.

pkg/rancher-components/src/components/RcItemCard/RcItemCard.vue
marcelofukumoto: That will make it a lot clearer, thanks! Changed

shell/assets/translations/en-us.yaml
marcelofukumoto: We can't drop, because there are 3 possibilities on this one:

The selectedRepoAuth + usePreviouslyGeneratedImagePullSecretBanner and the selectedRepoAuth + generatedImagePullSecretBanner.

Some cases a new one will be generated, some it is going to use the last one already created.

But I updated the text a little to be one line instead.
So instead of having selectedRepoAuth, I just made 2 different texts.

shell/assets/translations/en-us.yaml
marcelofukumoto: Used this as a bse and created the other 3 possibilities

shell/pages/c/_cluster/apps/charts/index.vue
marcelofukumoto: Oh, made me realize I removed the logic "if exists suseAppCollection"

shell/pages/c/_cluster/apps/charts/install.vue
marcelofukumoto: That is interesting. I tried with the standard users, so that is how I concluded that most people should have access to that one.
But I will add a case for users that doesn't have the access.
image

marcelofukumoto: But that was because as standard user I couldnt see clusters created by the ADMIN, so there is some permissions that could cause the case where standard user cannot see the secret.

shell/pages/c/_cluster/apps/charts/install.vue
marcelofukumoto: The secret here is the secret used to create the REPOSITORY. So technically there is no alternative.
The name is there on the repo: this.repo.spec.clientSecret.name
So technically I could just use that name instead, even not finding the selectedSecret.

marcelofukumoto: If the name is there but the DECODED DATA is not, it will ignore the option to CREATE from the AUTH.
If the name is there it will try to fetch the IMAGE PULL SECRET with the same name, which is ideal, and use it. If not it will ask to create using the same name. That works as well.

marcelofukumoto: So tested only using the name, that is the only data required so seems to work fine.
If there is no Decoded Data it will activate the mode where the user can only create one manually. Or use one based on the name if can retrieve that one.

shell/pages/c/_cluster/apps/charts/install.vue
marcelofukumoto: So fixed only using the name. seems to work fine and go to the no Decoded Data options.

shell/pages/c/_cluster/apps/charts/install.vue
marcelofukumoto: So fixed only using the name. seems to work fine and go to the no Decoded Data options.

shell/edit/catalog.cattle.io.clusterrepo.vue
marcelofukumoto: Funny enough, I actually tested them to be clickable haha so I was totally off on this one.
I added the disabled (it didnt exist on RcItemCard) I used the overall colors so it should be ok in terms of UX) 
I asked Ken @kwwii on this one.
image

marcelofukumoto:
> * For the app co helm repo clicking 'refresh' puts the repo into error `GET "https://dp.apps.rancher.io/v2/charts/tags/list": basic credential not found` , and then there's a full page dev warning `Failed waiting for: state=active`

That is an old bug, I faced it before.
image

I think it is when it takes long? But I have faced it with OCI previously as well, I guess it is related to how long it takes.

> * We should have a nice name for app co in the charts repo list
>   
>   * SUSE App Collection
>   * Image

Ok! Let me add this one!

> * when editing a chart that was installed via appco the page warns that a new secret will be created. if we already have a secret in the target ns I don't think we should create a new one each time. can the steps to test in pr description be updated with the edit app case?

Yes, that is correct. If we already have a secret it will use the previously created one, you still have an option to create one manually with a different secret, but the one created automatically is only one. Let me update on the description.

> * i didn't see the 'from suse app co' tooltip next to an app that was installed via app co (think this was dropped for spec?)

A yes, installed APP we removed it. It dropped.

> 
> I didn't quite get through all of shell/pages/c/_cluster/apps/charts/install.vue, will tackle that in next round

shell/edit/catalog.cattle.io.clusterrepo.vue
marcelofukumoto: For this one we discussed on this before with me, @nwmac  and @kwwii. 
I shared this concern since you shared with me before.
The conclusion was that we should keep this one on the SUSE APP CO, at least for now.
If user wants to make it more dynamic the user can always register OCI manually.

I hope that is ok. But I believe that is for now, as you said it would be better if the user can filter better, but @nwmac specifically shared than it would need to know the names and exactly what wants to filter. So maybe if we have the proper "urls" so the user doesn't need to type, it would be a better solution.

tldr.:
- For now we can keep as it is and opinionated the SUSE App Co usage.
- Users can choose other options if they want using normal OCI.
- If we want to have paths editable, maybe the best option is to give them options. I think in the end what we can do is create a main repository that can fetch all data, and sub repository that can fetch detailed one with specific options.

shell/models/catalog.cattle.io.clusterrepo.js
marcelofukumoto: I checked on this. Some concerns is that the SUSE App Collection is a separated option from the OCI Registry only separated by this annotation. I tried to change and worked at some extend but my main concerns:

- Before I can isolated in a really shallowed manner only not allowing options being visible on the UI for NOT PRIME. So the separation was: PRIME can see SUSE App Collection, and COMMUNITY cannot see it. With the URL OCI can be SUSE App Collection, I would need to separate: An OCI Registry with the SUSE Url on Community would be OCI Registry, and on PRIME would be SUSE App Co. It is technically an impossible case, but ideally would need a logic in the code to separate it.

- Second concern is that I am hardcoding some AUTH names, so I would need to revisit the LOGIC. From what I tested it seems ok since the INSTALL logic uses the name as default so it doesn't really affect. 

- Now main concern is: If we allow TERRAFORM or API or any other way to be treated as SUSE App Collection, we need to consider the case where the user in the UI can create a OCI Registry with the SUSE App Collection URL and that one will turn into a SUSE App Collection. I would say ideally would be possible, but that would require even some changes on the UX for UI cases to allow that possibility as well. If the USER types the SUSE URL it would trigger for it to change the type or something like that.

It would be better to not add logic that interacts with the other types too much to reduce regression.

shell/models/catalog.cattle.io.app.js
marcelofukumoto: Same as the other comment.

shell/edit/catalog.cattle.io.clusterrepo.vue
marcelofukumoto: Oh, Dark Mode doesnt swap the colors.
But I guess it makes sense to always be white bg and black color.
Fixed! Thank you for finding it! I need to do this swap light/dark at the end of the issue! My apologies!

shell/models/catalog.cattle.io.clusterrepo.js
richard-cox: Think my main concern is that we flag certain charts from appco which will be missed, and more importantly won't support the image pull secrets.

This would block users from using app co in the gitops world (repo added via terraform / api) and custom world (repo added as oci registry to reduce size impact). This also supports the multiple repo world as well (repo added as more than one small oci registry)

Can we split the functionality and https://github.com/rancher/dashboard/pull/16690#discussion_r2841087611 (not sure both locations makes sense now)
- isSuseAppCollection
  - is annotation OR url
  - does not show the appco specific stuff in CRUD registry world
  - does show appco specific stuff in chart CRUD world
- isSuseAppCollectionFromUI
  - is annotation
  - does show the appco specific stuff in CRUD registry world
  - does show appco specific stuff in chart CRUD world

shell/edit/catalog.cattle.io.clusterrepo.vue
richard-cox: No problem!

If it helps you can quickly toggle between modes (with shift + t) if you enable a preference (user avatar --> preferences --> `Enable Dark/Light Theme keyboard shortcut toggle (shift+T)`

shell/pages/c/_cluster/apps/charts/install.vue
richard-cox: I tried this again, and as a standard user who is a member of a downstream cluster with the appco repo added
- without access to namespaces/projects - /k8s/clusters/c-x6x5z/v1/secrets 404s
- with access to ns/projects - /k8s/clusters/c-x6x5z/v1/secrets returns an empty list

For the local / upstream cluster i also see the same.

I tried again ... with a user who was cluster owner and they CAN see the secrets in cattle-system. So it's the cluster membership level that's the key.

However as a cluster owner AND rancher admin the screen prompts for user and password but states a new one will be created

image

Has the checkbox and 'a new image pull secret' banner become disconnected?

richard-cox: i tried switching the ns and then i think get the correct flow

image

marcelofukumoto: Oh, I had this bug before for different reasons when checking the data. But I fixed for some cases, let me check again.

marcelofukumoto: Ok... I fixed the case for the opposite. When you dont have a decodedData.
So let me check what is wrong and add the conditions.

marcelofukumoto: Ok... it works the second time it opens the screen. But not the first one. Some missing data.

marcelofukumoto: Sometimes the trigger for the namespace triggers before the fetch, causing the problem. Let me fix that

marcelofukumoto: Fixed, the problem happened when the NAMESPACE change flow happens before the data being fetched.
And it is a racing problem. So making sure that the namespace process is only valid when the data is previously fetched fixed the problem.

shell/models/catalog.cattle.io.app.js
marcelofukumoto: Not used anymore !
It was used on the Installed Apps page, so I will remove it.

shell/models/catalog.cattle.io.clusterrepo.js
marcelofukumoto: Ok, let me try.
So it will only change the CHART page. Let me check.

marcelofukumoto: I tried this one.
Seems to be working fine!
image
It is an OCI repo only here and on this world.

Charts level it is considered the same as the other one. So it triggers the flow with the Image Pull Secret.
image

So I remove the entire ANNOTATION logic right now since we are using the the URL as a base. That would make it possible for all cases.

The only difference is when you are on RANCHER PRIME, the repository will be shown as SUSE App Co and all the proper features. If not RANCHER PRIME it will remove all the features and will show as OCI. Not a real case that we were worried but now works.

With that if you create anywhere with the correct URL (starting with) It should turn into SUSE App Co repository.
The only concern now is the flow when you create as an OCI Registry, normal one, and suddenly when you go see it turns into a SUSE App Collection, but I guess it is ok.

If we dont want that, we can revert my latest change.

shell/config/types.js
```diff
@@ -391,21 +391,27 @@ export const ADDRESSES = {
...
-  HELM_URL: 'helm-url',
-  GIT_REPO: 'git-repo',
-  OCI_URL:  'oci-url'
+  HELM_URL:            'helm-url',
+  GIT_REPO:            'git-repo',
+  OCI_URL:             'oci-url',
+  SUSE_APP_COLLECTION: 'suse-application-collection'
 };
 
+export const CLUSTER_REPO_APPCO_AUTH_GENERATE_NAME = 'clusterrepo-appco-auth-';
```
richard-cox, line 415: Could we add some jsdocs to these

for each prop something covering...
- used to name catalog cluster repo and app image pull secrets
- used to name catalog app image pull secrets
- used to manage catalog cluster repos for AppCo

shell/components/form/SelectOrCreateAuthSecret.vue
richard-cox: ah, now that we pass in the app co specific stuff this is just AUTH_TYPE._IMAGE_PULL_SECRET again (and selectOrCreateAuthSecret.imagePullSecret)

shell/pages/c/_cluster/apps/charts/install.vue
```diff
@@ -370,49 +378,84 @@ export default {
...
+      let defaultSelectedSecret = await this.$store.getters['cluster/byId'](SECRET, `cattle-system/${ this.repo.spec.clientSecret.name }`);
+
+      if (!defaultSelectedSecret) {
+        try {
+          defaultSelectedSecret = (await this.$store.dispatch('cluster/find', { type: SECRET, id: `cattle-system/${ this.repo.spec.clientSecret.name }` }));
+        } catch (e) {
+          // If cannot get the secret for any reason, permission or doesn't exist
+          // We can fallback to use the name only and with that name move forward.
+          // On only other required data is the DecodedData but not having it will only trigger a different flow.
+          defaultSelectedSecret = { name: this.repo.spec.clientSecret.name };
```
richard-cox: Little bit uncomfortable with this. 

It's also brittle, if the secret it'll just fail to install.

So we have to choose between an improved UX or a slightly strange one (users without access to the secret will use the secret)

If we do go down this path it's worth warning the admin that the secret will be used by rancher as the default image pull secret when installing charts. would be a banner somewhere in the create appco repo page.

shell/pages/c/_cluster/apps/charts/install.vue
```diff
@@ -370,49 +378,84 @@ export default {
...
+        }
+      }
+
+      this.selectedSecret = defaultSelectedSecret;
+      this.defaultGeneratedNameForImagePullSecret = `${ this.selectedSecret.name }-image-pull-secret`;
+      this.generatedNameForImagePullSecret = `${ this.selectedSecret.name }-image-pull-secret-${ generateRandomAlphaString(5) }`;
+      this.appCoDataFetched = true;
```
richard-cox, line 399: as this is all in the fetch you can just use `!$fetchState.pending` instead

shell/pages/c/_cluster/apps/charts/install.vue
richard-cox: Little bit uncomfortable with this. 

It's also brittle, if the secret is deleted it'll just fail to install.

So we have to choose between an improved UX (all users don't need to enter image pull secrets) or a slightly strange one (users without access to the secret will use the secret)

If we do go down this path it's worth warning the admin that the secret will be used by rancher as the default image pull secret when installing charts. would be a banner somewhere in the create appco repo page.

richard-cox: This also didn't work for me. As a standard-user who is a member of a cluster with appco setup i can only create a new secret

Image

shell/models/catalog.cattle.io.clusterrepo.js
richard-cox: this doesn't quite work, given we've lost the ability to edit the oci url and other settings
- create oci repo with url oci://dp.apps.rancher.io/charts
- labelling, creating charts with secrets, etc all works great
- edit oci repo - classed as app co so get limited form

i think we can still have the best of all worlds

keep the annotation when creating the app co specific repo. if the annotation is there it's edit experience is app co

this is the split was trying to get at in the previous comment

shell/edit/catalog.cattle.io.clusterrepo.vue
richard-cox: Given we can support app co with custom urls now via oci (and covering in that other comment thread https://github.com/rancher/dashboard/pull/16690#discussion_r2841053733) we can close this one

shell/models/catalog.cattle.io.clusterrepo.js
richard-cox: if we also use the label first, then check for url we cover other scenarios like users who mirror appco and have a different url

marcelofukumoto:
https://github.com/rancher/dashboard/pull/16690#discussion_r2852678109
Ok. For this one I will revert my changes back to the previous logic which was the first one you shared before I tried to change.

shell/pages/c/_cluster/apps/charts/install.vue
marcelofukumoto: Fixed, it works well.

shell/config/types.js
marcelofukumoto: Done

marcelofukumoto:
https://github.com/rancher/dashboard/pull/16690#discussion_r2852369378

Fixed this one and changed as well for the URL to be dynamic

marcelofukumoto:
https://github.com/rancher/dashboard/pull/16690#discussion_r2852493362

I tried and with the fetchState pending it doesnt work. With the one I added it works. Maybe you tested with that? I think the timing that set those are different.

shell/pages/c/_cluster/apps/charts/install.vue
richard-cox: ok, think i'm aligned now.

we won't submit the secret name, we're just using it for the default name of the secret that the user will create. 

however still can't create an app as a cluster member. right afterwards i started getting the too many request errors again. could be related. could you give a try?

richard-cox: i was using habor, could be a problem with that specific chart?

richard-cox: Validated 

- downstream cluster
- admin setups up standard appco repo + supplies registry secret
- standard-user added as member to downstream cluster
- standard-user creates a project and namespace
- standard-user installs alertmanager (supplies their own image pull secret)
- alertmanager successfully deploys

shell/models/catalog.cattle.io.clusterrepo.js
richard-cox: Validated

- downstream cluster
- admin setups up standard OCI repo + supplies registry secret
- charts are identified as appco
- admin installs alertmanager (uses registry secret as image pull secret)
- alertmanager successfully deploys

richard-cox (approved):
Ship it!

richard-cox:
@marcelofukumoto some follow on

- The points from https://github.com/rancher/dashboard/pull/16690#pullrequestreview-3853439302 should be captured in new gh issues
- Can we add something for QA to take a look at? I'm concerned that the image pull secret we create is not working with non-rancher charts. This could expose a fault in the way we're creating the image pull secret and assigning it, a fault in how Rancher BE is processing it or the chart itself. Would be good to validate at least one works. So...
  - as an admin create the AppCo repository in a downstream cluster
  - as an admin find a chart that is not from Rancher / SUSE and install it
  - the app should successfully install

Issue #16668, opened 2026-02-13
AWS VPCs missing a Name tag are not shown in ec2 provisioning, and labeled "undefined (vpc id)" in EKS provisioning


mantis-toboggan-md:
**Describe the bug**
The dashboard expects aws VPCs to have a Tags property that includes an object with a Key property set to "name." Apparently some VPCs do not have these tags. In EKS provisioning they're labeled undefined and they are excluded as options from ec2 provisioning. We should fall back to showing the vpc id when we don't have a name tag.

**Screenshots**

<img width="619" height="455" alt="Image" src="https://github.com/user-attachments/assets/5e666e57-5a14-4971-a16d-2e4eb0147212" />

yonasberhe23:
Tested on Rancher v2.14-67d0e7cbff8374d216a1bb0c44d59688ce1020f0-head / Dashboard master `e601906`

Reopening: Partially fixed; VPCs that do not have "Name" tag are no longer excluded from ec2 provisioning however, in EKS provisioning, VPCs that do not have the tag are still labeled `undefined`

---

# EKS provisioning

VPCs that do not have the tag are labeled `undefined` ❌

<img width="770" height="331" alt="Image" src="https://github.com/user-attachments/assets/9e102f30-fc3c-42c3-8170-04fa5c13046c" />

---

# ec2 provisioning

VPCs that do not have "Name" tag are no longer excluded from ec2 provisioning. ✅

### VPC with a "Name" tag
<img width="564" height="255" alt="Image" src="https://github.com/user-attachments/assets/a47c25cd-2771-4ab0-88fb-4f2f635d8a10" />

### VPC without a "Name" tag

<img width="639" height="260" alt="Image" src="https://github.com/user-attachments/assets/0ce4ad4b-6f24-4350-8e90-f359ad7b7e7a" />

yonasberhe23:
Tested on Rancher v2.14-16b1922c97a15d33e8c7af6ae0a38976abdc53af-head / Dashboard master `4f31d4b`

In EC2 provisioning, VPCs that do not have "Name" tag are no longer excluded. ✅
in EKS provisioning, VPCs that do not have the tag are no longer  labeled undefined; We fall back to showing the VPC id. ✅


### EKS provisioning

VPC with a "Name" tag
<img width="690" height="336" alt="Image" src="https://github.com/user-attachments/assets/7e69f188-4579-463d-b233-29e9e05636a4" />

VPC without a "Name" tag
[...]

Issue #15788, opened 2025-10-29
When IPv6-only cluster is specified, the stack preference automatically is set to `Dual-Stack` instead of `IPv6`


markusewalker:
**Setup**
- Rancher version: `head` https://github.com/rancher/rancher/commit/b42202a64d6edf623acee4901a5334fcdacb53b3

**Describe the bug**

In testing https://github.com/rancher/dashboard/issues/15161, there appears to be a functional issue when you select IPv6. See below:

<img width="833" height="150" alt="Image" src="https://github.com/user-attachments/assets/258039df-a1a2-45b8-acd1-d8f4f766328c" />

Once you click that and you navigate to the stack preference, you see the following:

<img width="1332" height="291" alt="Image" src="https://github.com/user-attachments/assets/d8e24dbb-a08b-44cf-94c6-f27a21016765" />

I would expect `IPv6` to be chosen instead of `Dual-Stack`, especially as I noted that I want to use an IPv6-only cluster.

**To Reproduce**

1. Setup an IPv6 Rancher environment.
2. Create an AWS node driver cluster.
     - Specify `Use IPv6 Only` and navigate to the networking tab and note the stack preference is `Dual-Stack` and not `IPv6`.

**Result**
The stack preference is automatically set to `Dual-Stack`.

**Expected Result**

The stack preference should be automatically set to `IPv6`.

mantis-toboggan-md:
Does the cluster fail to provision or otherwise provision incorrectly if the stack preference is set to `dual`? 

@jiaqiluo maybe you could chime in on this one: if a user has selected "ipv6 only" for every machine pool, is there some difference in behavior between ipv6 and dual stack preference settings?

We've made the ui select `dual` not `ipv6` because we think that if the UI is going to change the value of off-screen inputs we should do so in a manner least-likely to cause issues if our ui logic has misinterpreted user intent. As far as what people might expect - the description under the checkbox in the ec2 config page informs users that the stack preference will be set to `dual`.

jiaqiluo:
Hi @mantis-toboggan-md 

>  if a user has selected "ipv6 only" for every machine pool, is there some difference in behavior between ipv6 and dual stack preference settings?

The stack preference must match the cluster’s networking configuration:

- Set to `ipv4` for IPv4-only clusters
- Set to `ipv6` for IPv6-only clusters
- Set to `dual` for dual-stack clusters



[...]

Issue #15787, opened 2025-10-29
Improve language for IPv6-related options in EC2 Machine Pools


jiaqiluo:
While testing the IPv6-related options in EC2 Machine Pools in the current UI (as shown in the screenshot), I noticed a few sentences that could be improved for clarity.

1/ 
Change from "Use IPv6 for instance metadata service" to "**Enable** IPv6 **endpoint** for instance metadata service"

2/ 
Change from "This allows IPv6 communication for components and workloads"  to "This designates the first assigned IPv6 address as the primary address, ensuring a consistent, non-changing IPv6 address for the instance. This option does not control whether IPv6 addresses are assigned."
 

<img width="1552" height="714" alt="Image" src="https://github.com/user-attachments/assets/14eab2fb-9c3c-40ef-a92d-a302d4d2c61b" />

markusewalker:
Validated that this has been addressed on `v2.14-head` https://github.com/rancher/rancher/commit/abe4709419d926e765cabc404abe82d28a6d7e56. Please see testing considered below:

|#|Scenario|Result|
|---|---|---|
|1|Validate prior `Enable IPv6` checkbox now is set to `Enable Dual-Stack`|:white_check_mark:|
|2|Validating clicking `Enable Dual-Stack` reveals the `Enable IPv6` checkbox`|:white_check_mark:|
|3|Validate prior `Use IPv6 for instance metadata service` checkbox now is set to `Enable IPv6 endpoint for instance metadata service`|:white_check_mark:|

Issue #15786, opened 2025-10-29
Improve language for Stack Preference in the cluster networking section


markusewalker:
**Is your feature request related to a problem? Please describe.**

In testing https://github.com/rancher/dashboard/issues/15279, here is the current language for `Stack Preference`:

<img width="1338" height="165" alt="Image" src="https://github.com/user-attachments/assets/faf8c2c3-b9e0-4d9a-8f42-c4b360ce7c64" />

The last line in particular could be cleaned up a little bit.

**Describe the solution you'd like**

It would be nice to see the following for the last sentence:

`Depending on the stack preference, the cluster is updated accordingly. IPv4 uses 127.0.0.1, Dual-Stack uses localhost and IPv6 uses [::1].`

Also, unsure if there is going to be an external Rancher documentation that discusses IPv6, stack preference in the UI in particular. But if so, it would also be nice to take the majority of the description and just link it to that documentation. That way, customers can read more about it in detail, while not over-bloating the UI.

So specifically, it would look like this altogether:

`Stack preference determines the networking stack used by the cluster. For more information, please read more here (here would be a hyperlink to the external Rancher documentation).`

`Depending on the stack preference, the cluster is updated accordingly. IPv4 uses 127.0.0.1, Dual-Stack uses localhost and IPv6 uses [::1].`

jiaqiluo:
Rancher documentation will contain the explanation of the Stack Preference field on the [k3s-cluster-configuration](https://ranchermanager.docs.rancher.com/reference-guides/cluster-configuration/rancher-server-configuration/k3s-cluster-configuration) and [rke2-cluster-configuration](https://ranchermanager.docs.rancher.com/reference-guides/cluster-configuration/rancher-server-configuration/rke2-cluster-configuration) pages. 

Here is a preview:
 

> ##### Stack Preference
> 
> Choose the networking stack for the cluster. This option affects:
> 
> - The address used for health and readiness probes of components such as Calico, etcd, kube-apiserver, kube-scheduler, kube-controller-manager, and kubelet
> - The server URL in the `authentication-token-webhook-config-file` for the Authorized Cluster Endpoint
> - The `advertise-client-urls` setting for etcd during snapshot restoration
> 
> Options are `dual`, `ipv4`, `ipv6`:
> 
> - When set to `dual`, the cluster uses `localhost`
> - When set to `ipv6`, the cluster uses `[::1]`
> - When set to `ipv4`, the cluster uses `127.0.0.1`

[...]

PR #16694, opened 2026-02-18
ec2 ipv6 updates

### LINKED ISSUE #16391: Warn users creating ipv6 rke2/k3s clusters if flannel masq or stack preference aren't configured as recommended
Requires https://github.com/rancher/dashboard/issues/15734

There are multiple fields which must be configured to enable an rke2 or k3s cluster to use only ipv6 networking. Setting these fields automatically presents UX issues because they are contained in an area of the provisioning form that is not visible while configuring machine pool networking. 

Warn users before saving if they are using ipv6 and the stack preference or flannel masq inputs are not configured as recommended. Show a confirmation modal when the user clicks save, and update the modal shown when the user clicks 'edit as yaml' to include the same warning.

The following messaging should be updated to reflect that both flannel masq and stack pref should be configured:

Image

Automatically set flannel masq and stack preference inputs depending on networking configuration

**Additional context**
issue to add flannel masq input: https://github.com/rancher/dashboard/issues/15734
issue to add stack pref input: https://github.com/rancher/dashboard/issues/15279

Modal currently show when users click 'edit as yaml'

Image

mantis-toboggan-md:
### Summary
Fixes #16391 #15788 #15787 #15786 #15785 #16668 https://github.com/rancher/dashboard/issues/15734#issuecomment-3862249061

A confirmation modal should appear when creating ec2 clusters if any of the following are true:
- some/all pools have 'enable dual-stack' checked and cluster or service CIDR ranges don't include an ipv6 cidr
- some/all pools have 'enable dual-stack' checked, not 'enable ipv6' checked and stack preference is not 'dual'
- all pools have 'enable ipv6' checked and stack preference is not 'ipv6'
- a k3s version is selected, all pools have 'enable ipv6' checked, and 'enable flannel masq' is not checked

The pre-existing 'enable ipv6' checkbox adds ipv6 networking capability to droplets; they also have ipv4 addresses https://docs.digitalocean.com/products/networking/ipv6/. So this is treated like checking 'enable dual stack' in ec2. A confirmation modal should appear if:
- some/all pools have 'enable ipv6' checked and cluster or service CIDR ranges don't include an ipv6 cidr
~- some/all pools have 'enable ipv6'  checked and stack preference is not 'dual'~
mistake in the PR description, that requirement should read:
- some/all pools have 'enable ipv6'  checked and stack preference is not 'dual'/'ipv6'

https://github.com/user-attachments/assets/0d90e650-a464-45b0-aa0c-c920245f5135

Screenshot 2026-02-19 at 7 58 37 AM

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/edit/provisioning.cattle.io.cluster/rke2.vue
+++ b/shell/edit/provisioning.cattle.io.cluster/rke2.vue
@@ -1528,7 +1528,7 @@ export default {
     },
 
     async showIpv6Warning(hookContext) {
-      if (this.mode !== _CREATE) {
+      if (this.mode !== _CREATE || !this.machinePools?.length) {
         return;
       }
       const stackPreference = this.value.spec.rkeConfig.networking.stackPreference;
@@ -1541,7 +1541,7 @@ export default {
       const isIpv6 = this.hasOnlyIpv6Pools;
 
       const flannelMasqInvalid = isIpv6 && isK3s && !flannelMasqEnabled;
-      const stackPrefInvalid = (isIpv6 && stackPreference !== STACK_PREFS.IPV6) || (isDualStack && stackPreference !== STACK_PREFS.DUAL);
+      const stackPrefInvalid = (isIpv6 && stackPreference !== STACK_PREFS.IPV6) || (isDualStack && ![STACK_PREFS.IPV6, STACK_PREFS.DUAL].includes(stackPreference));
 
       const clusterCIDRInvalid = (isIpv6 || isDualStack) && !clusterCIDR.includes(':');
       const serviceCIDRInvalid = (isIpv6 || isDualStack) && !serviceCIDR.includes(':');
--- a/cypress/e2e/tests/pages/manager/cluster-provisioning-amazon-ec2-rke2.spec.ts
+++ b/cypress/e2e/tests/pages/manager/cluster-provisioning-amazon-ec2-rke2.spec.ts
@@ -26,7 +26,7 @@ describe('Deploy RKE2 cluster using node driver on Amazon EC2', { tags: ['@manag
   before(() => {
     cy.login();
 
-    // // clean up amazon cloud credentials
+    // clean up amazon cloud credentials
     cy.getRancherResource('v3', 'cloudcredentials', null, null).then((resp: Cypress.Response<any>) => {
       const body = resp.body;
 
@@ -154,11 +154,14 @@ describe('Deploy RKE2 cluster using node driver on Amazon EC2', { tags: ['@manag
     });
 
     // check provider
+
     clusterList.list().provider(this.rke2Ec2ClusterName).should('contain.text', 'Amazon EC2');
     clusterList.list().providerSubType(this.rke2Ec2ClusterName).should('contain.text', 'RKE2');
 
     // check machines
     clusterList.list().machines(this.rke2Ec2ClusterName).should('contain.text', '1');
+    // check that the machine progress bar is 1 color
+    clusterList.list().machines(this.rke2Ec2ClusterName).find('.piece').should('have.length', 1);
   });
 
   it('cluster details page', function() {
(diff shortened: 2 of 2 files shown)


oboc-sts (approved):
LGTM! Super nicely done...

shell/edit/provisioning.cattle.io.cluster/rke2.vue
```diff
@@ -849,8 +850,12 @@ export default {
       }
     },
 
-    hasSomeIpv6Pools() {
-      return !!(this.machinePools || []).find((p) => p.hasIpv6);
+    hasOnlyIpv6Pools() {
+      return !(this.machinePools || []).find((p) => !p.isIpv6 || p.isDualStack)?.length;
```
codyrancher: I don't think this is working as intended. 

1. `machinePools` is a list of objects which means when applicable...
2. `(this.machinePools || []).find((p) => !p.isIpv6 || p.isDualStack)` will return a `machinePool` object ...
3. `machinePool` doesn't have a length property which in turn makes this return true.

cypress/e2e/po/edit/provisioning.cattle.io.cluster/create/cluster-create-rke2-amazon.po.ts
```diff
@@ -24,11 +26,27 @@ export default class ClusterManagerCreateRke2AmazonPagePo extends ClusterManager
...
+
+  networkTab(): NetworkRke2 {
+    return new NetworkRke2();
+  }
+
+  ipv6ConfirmationDialog() {
+    return cy.get('body').find('[data-testid="ipv6-dialog"]');
+  }
+
+  ipv6Recommentations() {
```
codyrancher: Minor typo

shell/edit/provisioning.cattle.io.cluster/__tests__/rke2.test.ts
```diff
@@ -188,7 +188,7 @@ describe('component: rke2', () => {
     wrapper.vm.machinePools.forEach((p: any) => expect(p.drainBeforeDelete).toBe(true));
   });
 
-  it('should set distro root directory from k8sDistro on a Harvester cluster creation on save override (_doSaveOverride)', async() => {
+  it.skip('should set distro root directory from k8sDistro on a Harvester cluster creation on save override (_doSaveOverride)', async() => {
```
codyrancher: Do we want to keep this skipped? If we don't intend to fix it we can remove otherwise maybe a TODO with an issue linked.

shell/assets/translations/en-us.yaml
```diff
@@ -2577,12 +2587,11 @@ cluster:
...
+      description: Stack preference determines the networking stack used by the cluster. IPv4 uses <code>127.0.0.1</code>, Dual-Stack uses <code>localhost</code> and IPv6 uses <code>[::1]</code>. Read more about <a href="https://ranchermanager.docs.rancher.com/reference-guides/cluster-configuration/rancher-server-configuration/rke2-cluster-configuration#stack-preference"  "target="_blank" rel="noopener noreferrer nofollow" >RKE2</a> and <a href="https://ranchermanager.docs.rancher.com/reference-guides/cluster-configuration/rancher-server-configuration/k3s-cluster-configuration#stack-preference"  "target="_blank" rel="noopener noreferrer nofollow" >K3s</a> networking in the Rancher documentation.
```
codyrancher: Two issues I believe:

1. Both times we use `target` we're prefixing the `target` key with a quote `"target` instead of just the value.
    - It doesn't break anything though
2. Do we need to use the DOCS_BASE when linking to our docs or are we okay with not referencing the current version? I see other instances in the translations that just hardcodes it.
    - example: https://github.com/rancher/dashboard/blob/master/shell/config/home-links.js#L30-L34

mantis-toboggan-md: Good catch, fixed the target attribute. I think you're right that using DOCS_BASE is the proper way to link out to rancher documentation, though we have a mix of hardcoded urls already. I updated what I'd added in this PR

shell/edit/provisioning.cattle.io.cluster/__tests__/rke2.test.ts
mantis-toboggan-md: I've fixed the underlying issue and un-skipped this

shell/edit/provisioning.cattle.io.cluster/rke2.vue
mantis-toboggan-md: You're right, this is totally broken and my new e2e test missed it, I think a combo of this bug and an issue identifying dual stack vs ipv6-only machines. I've fixed this and I expanded test coverage so the bug would actually cause test failures

PR #16707, opened 2026-02-19
Fix findLabelSelector and matchingLabelSelector

### LINKED ISSUE #15529: UI shows incorrect cached elements for Pods list in Service details view
SURE-10760

**Setup**
- Rancher version: `2.12.1`
- Rancher UI Extensions: n/a
- Browser type & version: any

**Describe the bug**

With `ui-sql-cache` **off**, the list of pods is cached (at least when viewing a `Service` details view) incorrectly. Check repro steps for more information

**To Reproduce**

- Provision a Rancher v2.12.1 cluster.
- Disable the `ui-sql-cache` Feature Flag.
- Create a `Service` without any matching selector like the following (import YAML):
 
```
apiVersion: v1
kind: Service
metadata:
  name: emptyservice
spec:
  selector:
    bla: irrelevant
  type: ClusterIP
  ports:
  - name: http
    port: 80
    protocol: TCP
    targetPort: http
```

- Click on the `emptyservice` (details view for Service). The list under the pods tab should be empty.
- Click on `Workloads` -> `Pods`:
- Click again on `Service Discovery` -> `Services` -> `emptyservice`  (details view for Service).
- See the list of all pods (which shouldn't be related to this Service)

**Result**
`emptyservice` shows incorrectly the list of all Pods

**Expected Result**

`emptyservice` should not show any `Pod` associated

**Screenshots**

https://github.com/user-attachments/assets/95f78d78-cec9-4041-897f-6860e0ce7b38

**Additional context**

- With `ui-sql-cache` **on** the Service details view Pod's list briefly shows the incorrectly the list of all Pods, but right after updates to the correct value (this was validated in the scenario where we've disabled `ui-sql-cache` first, then re-enabled it)

- Pertinent question is if `findLabelSelector` action and `matchingLabelSelector` getter are ready to work with `ui-sql-cache` **off**

### LINKED ISSUE #16732: Deployment Detail Page: Services list is always empty
**Setup**
- Rancher version: v2.14.0 (dev of)

**Describe the bug**
- Deployment services list is never popukate

**To Reproduce**
- Install Longhorn
-  cluster explorer --> Deployment --> `longhorn-ui` Detail page --> services list

**Result**

Image

**Expected Result**

Image

**Additional context**
https://github.com/rancher/dashboard/pull/16364/changes#diff-8feb714b5665e764f50a90fd673235de56b8be1d1956d023d8226094ef36df91L60-L61

richard-cox:
### Summary
Fixes #15529
Fixes #16732

- findLabelSelector
  - Ensure we correctly set the record of what's in the cache... with what's actually in the cache
  - This became out of sync
    - we were clearing the cache and setting the collection to one relevant to a labelSelector
    - we were then adding labelSelector to a record saying we had this
    - the record was ADDED to after every call instead of being reset
  - fixes vai off world
    - we fetch pods for service 1 --> change to service 2 --> we fetch pods for service 2 (cache clears pods for service 1 BUT not the record of them) --> change to service 1 --> cache has record of service 1 pods so returns all in cache, however pods are for service 2
    - we now ensure the cache list and record match matching pods
    - this doesn't affect via on world because we handle labelSelector filters like normal filters in a findPage request
- matchingLabelSelector
  - this getter is only used in services and workload views to get pods associated with them
  - we look at the cache and get applicable resources given the labelSelector
  - the cache can have just the results of the labelSelector, or more and we should then apply the labelSelector to it
  - the fix here is to ensure that we're applying the label selector to the findPage scenario
  - fixes vai on world
    - we fetch pods for service 1 --> change to service 2 --> show cache containing pods from service 1 --> fetch pods for service 2 --> cache updates and we show the correct ones
    - to avoid this ensure we always run the selector on the pods in the cache
    - this does not affect vai off world because the cache only contains the results of the labelSelector
- empty services list in deployment detail page
  - https://github.com/rancher/dashboard/issues/16732, https://github.com/rancher/dashboard/pull/16364/changes#diff-8feb714b5665e764f50a90fd673235de56b8be1d1956d023d8226094ef36df91L60-L61
  - I think the lines to fetch the services was mistakenly removed when refactoring the selector to a model relatedServices getter
  - i fixed this here to unblock one of the regression tests

### Technical notes summary
Also fixed an issue where the selector was ignored when fetching `matching` pods

With vai off and on
- Core bug
  - Find two services with different amounts of associated Pods
  - Flick multiple times between the detail page of each service and ensure the Pods list contains the correct values
  - Also refresh on a service detail page and ensure Pods list contains correct values
 

With vai off and on

- matchingLabelSelector
  - Find two deployments with different amount of pods
  - Flick multiple times between the detail page of each deployment and ensure the Pods list contains the correct values
  - Also refresh on a detail page and ensure Pods list contains correct values
- Longhorn - (after install)
  -  --> on cluster explorer Longhorn `Overview` page should show card with link to longhorn service --> click --> should load long horn ui (page just needs to load, it can show errors)
  -  cluster explorer --> Deployment --> `longhorn-ui` Detail page --> services list should be populated
- Compliance (replacement app for CIS Benchmark)
  - cluster tools --> install compliance app
  - load https://raw.githubusercontent.com/rancher/compliance-operator/refs/heads/main/tests/k3s-bench-test.yaml, copy url, apply yaml in target cluster
  - wait for scan to complete --> click on a scan --> Shows entries
  - Unrelated issue - https://github.com/rancher/dashboard/issues/16735

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/plugins/dashboard-store/mutations.js
+++ b/shell/plugins/dashboard-store/mutations.js
@@ -440,7 +440,10 @@ export default {
       cache.map.set(proxies[i][keyField], proxies[i]);
     }
 
+    cache.havePage = undefined;
+    cache.haveNamespace = undefined;
     cache.haveSelector = { [selector]: true };
+    cache.haveAll = undefined;
     cache.revision = revision || 0;
   },
 
--- a/shell/plugins/dashboard-store/__tests__/getters.test.ts
+++ b/shell/plugins/dashboard-store/__tests__/getters.test.ts
@@ -47,6 +47,7 @@ describe('dashboard-store: getters', () => {
       expect(urlForGetter('typeFoo', 'idBar')).toBe('protocol/urlFoo/idBar');
     });
   });
+
   describe('dashboard-store > getters > urlOptions', () => {
     // we're not testing function output based off of state or getter inputs here since they are dependencies
     const state = { config: { baseUrl: 'protocol' } };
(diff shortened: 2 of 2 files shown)


shell/plugins/dashboard-store/mutations.js
```diff
@@ -440,7 +440,7 @@ export default {
       cache.map.set(proxies[i][keyField], proxies[i]);
     }
 
-    cache.haveSelector[selector] = true;
+    cache.haveSelector = { [selector]: true };
```
richard-cox, line 445: at some point we should remove the map part and just use `cache.haveSelector=selector` i chickened out though given this is bug complete week.

i can create a tech debt issue if requested

shell/plugins/dashboard-store/getters.js
```diff
@@ -177,7 +179,7 @@ export default {
 
     // Filter first by namespace if one is provided, since this is efficient
     if (namespace && typeof namespace === 'string') {
-      matching = type === POD ? getters['podsByNamespace'](namespace) : matching.filter((obj) => obj.namespace === namespace);
+      matching = type === POD && !selector ? getters['podsByNamespace'](namespace) : matching.filter((obj) => obj.namespace === namespace);
```
richard-cox, line 182: this is a secondary fix. matchingLabelSelector calls matching. if there's a selector we shouldn't return all pods in a namespace but go on to apply it

shell/detail/workload/index.vue
```diff
@@ -43,6 +43,10 @@ export default {
       hash.pods = this.value.fetchPods();
     }
 
+    if (this.serviceSchema) {
```
richard-cox, line 46: this was the fix for https://github.com/rancher/dashboard/issues/16732

PR #16723, opened 2026-02-20
[16688] improve notification on generate name resource creation

### LINKED ISSUE #16688: Improve notification on generate-name resource creation
- we now show a growl (and notification) when we create a resource with generate name
- this lets the user know the actual name of the resource created
- however we only show the name and not the resource type, so this can be confusing in scenarios where the user hasn't explicitly created a resource of that time. For example when provisioning a cluster the ui creates a few secondary resources the user might not be aware of... and sees notifications for them

- we include the resource type in the growl + notification

**Additional context**
- screenshot shown after creating an ec2 cluster

Image

marcelofukumoto:
### Summary
Fixes #16688

- Changed the text used by the growl when a resource is auto-generated
- It uses now the type-map/label for for it
- Still fallbacks to the previously used translation OR the .kind  

- Tested on KubeConfig in the Cluster Edit screen
- Tested on previous case on Audit Log Policy
- It triggers for every resource created that has autogenerated name  

### Screenshot/Video

image
image
image

### Checklist
- [x] The PR has a Milestone 


shell/plugins/steve/steve-class.js
```diff
@@ -64,19 +64,37 @@ export default class SteveModel extends HybridModel {
     return this.$getters['paginationEnabled'](this.type);
   }
 
+  /**
+   * Get the label for the resource type, with fallbacks to typeLabel translation and kind
+   * @returns {string}
+   *
+   */
+  get resourceTypeLabel() {
+    return this.$rootGetters['type-map/labelFor'](this.schema) || this.t(`typeLabel."${ this.type }"`) || this.kind;
```
richard-cox: in theory `labelFor` should be enough (it looks at typeLabel and then falls back on schema)

i just looked at resource-class though to see if that used ~typeLabel~ labelFor and there's a nice `typeDisplay` that does. 

so we can skip resourceTypeLabel and directly use typeDisplay

marcelofukumoto: Oh, that makes things easier, let me try.

marcelofukumoto: Worked well after the changes
image
image

marcelofukumoto:
Updated the ISSUE and PR.

PR #16776, opened 2026-02-26
fix issue where we couldnt filter by projects with the same name when on groupBy projects in list view

### LINKED ISSUE #15336: Filter in Project/Namespace is not working correctly if there are projects with the same name
**Setup**
- Rancher version: 2.12.0
- Rancher UI Extensions: NA
- Browser type & version: Firefox/Chrome

**Describe the bug**

 Filter field in Rancher 2.12 doesn't work as expected with projects of the same name.

**To Reproduce**

- Go to Projects/Namespaces
- Create 3 projects with the same name e.g. "samename"
- Create one namespace in the first 2 projects, e.g. test1 in the first and test2 in the second project
- Leave the 3rd without namespace

1. Now select Group by Project and filter for "samename"
   Only one project will be shown. The two others are invisible.

2. Now select Flat List and filter for "samename"
   Only the projects with a namespace in it are shown. The one without a namespace is invisible.

**Result**

Filtering in Rancher 2.12 doesn't work as expected

**Expected Result**

Filtering in Rancher 2.12 works as expected

**Screenshots**

**Additional context**

SURE-10659

See notes in associated Jira ticket.

The results are **technically** correct based on the data available. We might need to adjust how we display projects and namespaces to ensure that there is more certainty surrounding the search results. I think this issue warrants some more in-depth review before we set out to fix.

aalves08:
### Summary
Fixes #15336

- fix issue where we couldnt filter by projects with the same name when on groupBy projects in list view
- adds e2e tests

Check issue for test case

### Screenshot/Video

Screenshot 2026-02-26 at 15 13 32

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/config/table-headers.js
+++ b/shell/config/table-headers.js
@@ -56,14 +56,8 @@ export const NAME = {
 };
 
 export const PROJECT_NAMESPACES_NAME = {
-  name:          'name',
-  labelKey:      'tableHeaders.name',
-  value:         'nameDisplay',
-  getValue:      (row) => row.nameDisplay,
-  search:        ['nameDisplay', 'projectNameDisplay'],
-  sort:          ['nameSort'],
-  formatter:     'LinkDetail',
-  canBeVariable: true,
+  ...NAME,
+  search: ['nameDisplay', 'projectNameDisplay'],
 };
 
 export const LOGGING_OUTPUT_PROVIDERS = {


aalves08:
> I tested the fix we still only get 2 results in "Flat List" (test1, test2), but we get all 3 results in "Group by Project". It's been a while since I triaged this issue, but I think that we expect to see the inconsistent results between the different views, is that correct?
> 

The flat list is just a list of `namespaces` with the col for `project`. If there's an empty project (no namespaces) with most definitely won't appear on that flat list, so imho, there's nothing to fix there. You could already search fine by `namespaces`

So, in my understanding, the only thing to fix was the search on the `groupBy` scenario where one can definitely "see" those empty projects and where the search/filtering was definitely wrong for the groups of projects WITH namespaces (we were employing a gimmick for the empty groups to have the same name as the col header search field 😛 ).

This list is a weird one because it aggregates two very different resources/concepts.

See if you agree 🙏

shell/config/table-headers.js
```diff
@@ -55,6 +55,17 @@ export const NAME = {
   canBeVariable: true,
 };
 
+export const PROJECT_NAMESPACES_NAME = {
```
richard-cox, line 58: To avoid scenarions where we tweak the name column and have to update here as well it's worth updating this to

```
export const PROJECT_NAMESPACES_NAME = {
  ...NAME,
  search: ['nameDisplay', 'projectNameDisplay'],
};
```

richard-cox:
I'd agree, it's a weird experience and we've hit issues before (users cannot create namespaces in empty projects when the list isn't grouped by projects). This PR hasn't broken or fixed it though, so ok on that front

PR #16780, opened 2026-02-26
Improve data binding patterns and logic for Resource Quotas

### LINKED ISSUE #16769: Input values of Project Custom Resource Quotas are being deleted as identifier is added
**Setup**
- Rancher version: [Rancher](https://github.com/rancher/rancher)	v2.14.0-alpha6
[Dashboard](https://github.com/rancher/dashboard)	v2.14.0-alpha4
[Rancher Prime CLI](https://github.com/rancher/cli)	v2.12.0-rc.1
- Rancher UI Extensions: none
- Browser type & version: Version 145.0.7632.109 (Official Build) (arm64)

**Describe the bug**
The input values of Project Limit and Namespace Default Limit are being deleted as the user adds the resource identifier of project custom resource quotas, and are being passed to the next row that is added.

**To Reproduce**
1- Select a cluster
2- Go to Cluster > Projects/Namespaces
3- Create a Project
4- Under Resource Quotas, create a custom resource and input valid values for Project Limit and Namespace Default Limit 
5- Then, type a value for the resource identifier
6- Observe how the values for the Project Limit and Namespace Default Limit are deleted
7- Add a new resource 
8- Observe that the values of Project Limit and Namespace Default Limit from the previous row are added to the new row

**Result**
Observe how the values for the Project Limit and Namespace Default Limit are deleted and are added to the new row

**Expected Result**
All input values must be displayed and preserved in their respective original fields, regardless of the order they're filled and the rows added.

**Screenshots**
https://github.com/user-attachments/assets/ef31db6d-b5f8-471e-8534-a2f0479ac725
https://github.com/user-attachments/assets/45a58fac-0bb5-496c-88e4-33ee23b22815
Image

### LINKED ISSUE #16770: Deleting the last row of Project Custom Resource Quotas is also deleting the first row
**Setup**
- Rancher version: [Rancher](https://github.com/rancher/rancher) v2.14.0-alpha6
[Dashboard](https://github.com/rancher/dashboard) v2.14.0-alpha4
[Rancher Prime CLI](https://github.com/rancher/cli) v2.12.0-rc.1
- Rancher UI Extensions: none
- Browser type & version: Version 145.0.7632.109 (Official Build) (arm64)

**Describe the bug**

Removing the last row of Project Resource Quotas also removes the first row.

**To Reproduce**
1- Select a cluster
2- Go to Cluster > Projects/Namespaces
3- Create a Project
4- Under Resource Quotas, create a custom resource 
5- Add more rows of custom resources
6- Remove the last row
7- Observe that the first row is also deleted

**Result**
Removing the last row of Project Resource Quotas also removes the first row.

**Expected Result**
Removed row should not delete or modify the other rows

**Screenshots**

https://github.com/user-attachments/assets/2c6e2833-3cc0-4d6e-8629-acd10ceadd0d

rak-phillip:
### Summary
This improves data binding patterns and fixes logic with managing quotas by promoting a uni-directional data flow.

Fixes #16769
Fixes #16770

- Add replacement component for Project Row
- Replace ProjectRow.vue with ResourceQuotaEntry.vue
- Add tests for ResourceQuotaEntry.vue
- Update tests for Project.vue and management.cattle.io.project.vue

### Technical notes summary

The aim of this PR is to eliminate edge cases and increase certainty by solving data-flow issues with the existing ProjectRow.vue:

1. ProjectRow.vue makes use of ArrayList.vue, which tracks keys by index; this makes it difficult to reliably mutate collections
2. ProjectRow.vue mutates the value prop, making it difficult to track and assert where changes need to be made
3. ProjectRow.vue contains a lot of edge cases that are well outside the realm of rendering component data and managing its own state
 
This change solves these problems by adhering to a props-down, events up pattern for data binding; offsetting the resource quota data management to the parent component. 

This also replaces ArrayList.vue so that unique ids can properly be assigned for internal change tracking.

With this change in place:

- ResourceQuotaEntry.vue no longer mutates the value prop to manage quotas
- Project.vue no longer mutates the value prop in any way, instead it parses spec to produce quotas and vice-versa
- Project.vue now emits generated spec so that management.cattle.io.project.vue can perform updates
- management.cattle.io.project.vue performs straightforward updates with the generated spec provided to it

Creating, editing, and deleting resource quotas. 

The design of ProjectRow.vue was quite complex, there could be missed edge cases that this component covered. e2e tests should help to assert that these cases have not regressed as a result of this change.

#### Before

image

#### After

image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/form/ResourceQuota/Project.vue
+++ b/shell/components/form/ResourceQuota/Project.vue
@@ -151,14 +151,23 @@ export default {
 };
 </script>
 <template>
-  <div>
+  <div
+    role="grid"
+    :aria-label="t('resourceQuota.ariaLabel.grid')"
+  >
     <Banner
       color="info"
       label-key="resourceQuota.banner"
       class="mb-20"
     />
-    <div class="headers mb-10">
-      <div class="mr-10">
+    <div
+      role="row"
+      class="headers mb-10"
+    >
+      <div
+        role="columnheader"
+        class="mr-10"
+      >
         <label>
           {{ t('resourceQuota.headers.resourceType') }}
           <span
@@ -167,7 +176,10 @@ export default {
           >*</span>
         </label>
       </div>
-      <div class="mr-20">
+      <div
+        role="columnheader"
+        class="mr-20"
+      >
         <label>
           {{ t('resourceQuota.headers.resourceIdentifier') }}
           <span
@@ -180,15 +192,21 @@ export default {
           />
         </label>
       </div>
-      <div class="mr-20">
+      <div
+        role="columnheader"
+        class="mr-20"
+      >
         <label>{{ t('resourceQuota.headers.projectLimit') }}</label>
       </div>
-      <div class="mr-10">
+      <div
+        role="columnheader"
+        class="mr-10"
+      >
         <label>{{ t('resourceQuota.headers.namespaceDefaultLimit') }}</label>
       </div>
     </div>
     <template
-      v-for="resourceQuota in resourceQuotas"
+      v-for="(resourceQuota, resourceQuotaIndex) in resourceQuotas"
       :key="resourceQuota.id"
     >
       <ResourceQuota
@@ -197,6 +215,7 @@ export default {
         v-model:resource-identifier="resourceQuota.resourceIdentifier"
         v-model:project-limit="resourceQuota.projectLimit"
         v-model:namespace-default-limit="resourceQuota.namespaceDefaultLimit"
+        :index="resourceQuotaIndex + 1"
         :types="remainingTypes(resourceQuota.resourceType)"
         :mode="mode"
         @remove="removeResource"
@@ -208,7 +227,7 @@ export default {
         data-testid="btn-add-resource"
         @click="addResource"
       >
-        Add Resource
+        {{ t('resourceQuota.add.label') }}
       </rc-button>
     </div>
   </div>
--- a/shell/components/form/ResourceQuota/ResourceQuotaEntry.vue
+++ b/shell/components/form/ResourceQuota/ResourceQuotaEntry.vue
@@ -12,11 +12,15 @@ import { useI18n } from '@shell/composables/useI18n';
 
 const props = defineProps<{
   id: string,
+  index: number,
   mode: string,
   types: any[],
 }>();
 
 const emit = defineEmits(['remove']);
+
+const store = useStore();
+const { t } = useI18n(store);
 
 const resourceType = defineModel<string>('resourceType');
 const resourceIdentifier = defineModel<string>('resourceIdentifier');
@@ -31,8 +35,17 @@ const isCustom = computed(() => {
   return resourceType.value === TYPES.EXTENDED;
 });
 
-const store = useStore();
-const { t } = useI18n(store);
+const removeAriaLabel = computed(() => {
+  let identifier;
+
+  if (isCustom.value) {
+    identifier = resourceIdentifier.value || t('generic.ariaLabel.genericRow', { index: String(props.index) });
+  } else {
+    identifier = typeOption.value?.label || t('generic.ariaLabel.genericRow', { index: String(props.index) });
+  }
+
+  return t('resourceQuota.ariaLabel.remove', { identifier });
+});
 
 const customTypeRules = computed(() => {
   // Return a validation rule that makes the field required when isCustom is true
@@ -66,6 +79,7 @@ const updateResourceIdentifier = (resourceType: string) => {
 
 <template>
   <div
+    role="row"
     class="row mb-10"
   >
     <Select
@@ -73,6 +87,7 @@ const updateResourceIdentifier = (resourceType: string) => {
       class="mr-10"
       :mode="mode"
       :options="types"
+      :aria-label="t('resourceQuota.ariaLabel.resourceType', { row: index })"
       data-testid="projectrow-type-input"
       @update:value="updateResourceIdentifier"
     />
@@ -84,6 +99,7 @@ const updateResourceIdentifier = (resourceType: string) => {
       :placeholder="t('resourceQuota.resourceIdentifier.placeholder')"
       :rules="customTypeRules"
       :require-dirty="false"
+      :aria-label="t('resourceQuota.ariaLabel.resourceIdentifier', { row: index })"
       class="mr-10"
       data-testid="projectrow-custom-type-input"
     />
@@ -96,6 +112,7 @@ const updateResourceIdentifier = (resourceType: string) => {
       :input-exponent="typeOption.inputExponent"
       :base-unit="typeOption.baseUnit"
       :output-modifier="true"
+      :aria-label="t('resourceQuota.ariaLabel.projectLimit', { row: index })"
       data-testid="projectrow-project-quota-input"
     />
     <UnitInput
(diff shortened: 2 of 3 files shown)


shell/components/form/ResourceQuota/__tests__/ResourceQuotaEntry.test.ts
```diff
@@ -0,0 +1,221 @@
...
+      const wrapper = createWrapper({ resourceType: TYPES.EXTENDED });
+
+      expect((wrapper.vm as any).typeOption).toStrictEqual(extendedType);
+    });
+
+    it('should match on the resourceType', () => {
+      const wrapper = createWrapper({ resourceType: 'configMaps' });
+
+      expect((wrapper.vm as any).typeOption).toStrictEqual(mockType);
+    });
```
rak-phillip: I think this test is redundant after I fixed this to remove some outdated logic.

shell/components/form/ResourceQuota/ResourceQuotaEntry.vue
```diff
@@ -0,0 +1,126 @@
...
+const emit = defineEmits(['remove']);
+
+const resourceType = defineModel<string>('resourceType');
+const resourceIdentifier = defineModel<string>('resourceIdentifier');
+const projectLimit = defineModel<string>('projectLimit');
+const namespaceDefaultLimit = defineModel<string>('namespaceDefaultLimit');
+
+const typeOption = computed(() => {
+  return props.types.find((type) => type.value === resourceType.value);
+});
```
codyrancher, line 32: typeOption can become undefined/null and I don't see any checks for it before using it.

shell/components/form/ResourceQuota/ResourceQuotaEntry.vue
```diff
@@ -0,0 +1,126 @@
...
+      :input-exponent="typeOption.inputExponent"
+      :base-unit="typeOption.baseUnit"
+      :output-modifier="true"
+      data-testid="projectrow-namespace-quota-input"
+    />
+    <RcButton
+      variant="tertiary"
+      @click="remove(id)"
+    >
+      Remove
```
codyrancher: Translation

shell/components/form/ResourceQuota/Project.vue
```diff
@@ -37,80 +37,112 @@ export default {
...
+      const nsLimit = this.value.spec.namespaceDefaultResourceQuota.limit;
+
+      Object.keys(projectLimit).forEach((key) => {
+        if (key !== TYPES.EXTENDED) {
+          this.resourceQuotas.push({
+            id:                    uniqueId(),
+            resourceType:          key,
+            resourceIdentifier:    key,
+            projectLimit:          projectLimit[key],
+            namespaceDefaultLimit: nsLimit[key] || '',
```
codyrancher, line 135: Might need a conditional access here. It looks like nsLimit could be undefined but I'm not positive.

shell/components/form/ResourceQuota/Project.vue
```diff
@@ -37,80 +37,112 @@ export default {
...
+    this.quotasFromSpec();
   },
 
   computed: { ...QUOTA_COMPUTED },
 
-  methods: {
-    updateType(event) {
-      const { index, type } = event;
+  watch: {
+    resourceQuotas: {
```
codyrancher: Do you care that this watcher will trigger and emit input/validateChanged when quotasFromSpec() is invoked in created()? Maybe use `immediate: false`?

shell/components/form/ResourceQuota/Project.vue
```diff
@@ -152,30 +184,30 @@ export default {
...
+        v-model:resource-type="resourceQuota.resourceType"
+        v-model:resource-identifier="resourceQuota.resourceIdentifier"
+        v-model:project-limit="resourceQuota.projectLimit"
+        v-model:namespace-default-limit="resourceQuota.namespaceDefaultLimit"
+        :types="remainingTypes(resourceQuota.resourceType)"
+        :mode="mode"
+        @remove="removeResource"
+      />
+    </template>
+    <div class="mmt-6">
```
codyrancher: I added mmt and I think that was a mistake. I wanted to migrate everything to the new size variables but based on how much that has changed I think I should just remove mmt all together so it doesn't get adopted wider.

rak-phillip: I lifted `mmt` from ArrayList to keep consistency with the component that we replaced. It appears that we also have an `mt-6` defined, so I'll use that for now.

shell/components/form/ResourceQuota/Project.vue
rak-phillip: Yes, that's a good catch. This should be fixed.

shell/components/form/ResourceQuota/Project.vue
rak-phillip: hmmm... I don't think that `immediate: false` would have an effect on the outcome here, that should be the default state when registering a watcher. It is a bit of unnecessary churn when mounting the component, and we can probably solve this using `$watch()`[^1] instead.

[^1]: https://vuejs.org/api/component-instance#watch

shell/components/form/ResourceQuota/ResourceQuotaEntry.vue
rak-phillip: Yes, that's a good observation. I have added a default fallback to guard against this.

shell/components/form/ResourceQuota/ResourceQuotaEntry.vue
rak-phillip: Fixed

shell/components/form/ResourceQuota/Project.vue
rak-phillip: Scratch that, it looks like other `mt-6` usages could be done in error. I'll just specify a class instead.

shell/components/form/ResourceQuota/ResourceQuotaEntry.vue
```diff
@@ -0,0 +1,145 @@
...
+    identifier = resourceIdentifier.value || t('generic.ariaLabel.genericRow', { index: String(props.index) });
+  } else {
+    identifier = typeOption.value?.label || t('generic.ariaLabel.genericRow', { index: String(props.index) });
+  }
+
+  return t('resourceQuota.ariaLabel.remove', { identifier });
+});
+
+const store = useStore();
+const { t } = useI18n(store);
```
codyrancher: I know we have hoisting and lazy evaluation but we are binding these variables after they're being used above. I'm not sure if you want to move it or not.

shell/components/form/ResourceQuota/ResourceQuotaEntry.vue
```diff
@@ -0,0 +1,145 @@
...
+      :aria-label="t('resourceQuota.ariaLabel.projectLimit', { row: index })"
+      data-testid="projectrow-project-quota-input"
+    />
+    <UnitInput
+      v-model:value="namespaceDefaultLimit"
+      :mode="mode"
+      :placeholder="typeOption.placeholder"
+      :increment="typeOption.increment"
+      :input-exponent="typeOption.inputExponent"
+      :base-unit="typeOption.baseUnit"
```
codyrancher, line 124: Do we care that these could get `undefined` when typeoption is `{}`?

shell/components/form/ResourceQuota/ResourceQuotaEntry.vue
rak-phillip: Yeah, let's move them.

shell/components/form/ResourceQuota/ResourceQuotaEntry.vue
rak-phillip: We *should* always have at least one type option - extended. This maps closely to how the form previously behaved.. but I agree, if we're encountering some undefined behavior then we need to assure that the form will function in a predicable way. I think we can attempt the following:

1. Clearly define some error state/behavior so that we can relay to the user that something has failed
2. Default to extended instead of {}. Doing this, we might be able to continue operating, but I'm unsure what else might be wrong if we enter this state..

I'm leaning toward option 2, a sane default. What do you think?

codyrancher (approved):

PR #16789, opened 2026-02-26
Fixing several issues that stem from a race condition access issue around namespaces

### LINKED ISSUE #16442: Namespace label is not displayed correctly for users that can't list namespaces
**Setup**
- Rancher version: 2.14-head
- Rancher UI Extensions: N/A
- Browser type & version: N/A

**Describe the bug**

**To Reproduce**

1. Create a standard user
2. Log in as that standard users and create  (ie an rke2) cluster
3. Go into that clusuter's details

**Result**
"Namespace: false..." is shown. Hovering over it just adds a blank white rectangle

**Expected Result**

We should show an appropriate message and should not add a tooltip if it's blank

**Screenshots**

Image

**Additional context**

codyrancher:
### Summary
Fixes #16442

The namespace `ResourcePopover` could crash when the user lacked permission to list namespaces or when `currentProduct`/`currentCluster` were unresolved due to a race condition.

### Technical notes summary

- `useNamespace` now evaluates `currentStore`/`canList` inside the `computed` so they remain reactive and aren't accessed before the store is ready. The popover is gated behind `canList`, falling back to plain text.
- `ResourcePopover` renders a plain text when its fetch fails
- Conditional check in `Namespace.hideDetailLocation` and `Resource._glance` for `currentProduct`/`currentCluster`

The namespace popover on cluster pages immediately after the cluster was created.

- Reproducing was pretty inconsistent, you have to immediately visit the cluster detail page after creating it to repro.

- **Resource detail masthead** — namespace/project/workspace identifying fields
- **`ResourcePopover`** — the `v-if="!fetch.error"` guard affects all consumers, not just namespace
- **Glance panel** — `_glance` the conditional changes could affect other resources

### Screenshot/Video
If the request fails we default to showing the id of the resource we're attempting to show without a popover.

image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/Resource/Detail/ResourcePopover/__tests__/index.test.ts
+++ b/shell/components/Resource/Detail/ResourcePopover/__tests__/index.test.ts
@@ -79,6 +79,18 @@ describe('component: ResourcePopover/index.vue', () => {
       expect(wrapper.find('.display').exists()).toBe(false);
     });
 
+    it('should show plain text and no PopoverCard when fetch fails', async() => {
+      mockClusterFind.mockRejectedValue(new Error('Not found'));
+      const wrapper = createWrapper(undefined, undefined, PopoverCardStub);
+
+      await wrapper.vm.$nextTick();
+      await wrapper.vm.$nextTick();
+      await wrapper.vm.$nextTick();
+
+      expect(wrapper.findComponent(PopoverCard).exists()).toBe(false);
+      expect(wrapper.text()).toBe('test-ns/test-pod');
+    });
+
     it('should fetch data using the default store', async() => {
       const wrapper = createWrapper();
 
--- a/shell/models/namespace.js
+++ b/shell/models/namespace.js
@@ -267,7 +267,9 @@ export default class Namespace extends SteveModel {
   }
 
   get hideDetailLocation() {
-    return !!this.$rootGetters['currentProduct'].hideNamespaceLocation;
+    const currentProduct = this.$rootGetters['currentProduct'];
+
+    return currentProduct ? !!currentProduct.hideNamespaceLocation : true;
   }
 
   get glance() {
--- a/shell/plugins/dashboard-store/resource-class.js
+++ b/shell/plugins/dashboard-store/resource-class.js
@@ -1956,12 +1956,12 @@ export default class Resource {
       {
         name:          'namespace',
         label:         this.t('component.resource.detail.glance.namespace'),
-        formatter:     'Link',
+        formatter:     this.$rootGetters['currentProduct']?.id && this.$rootGetters['currentCluster']?.id ? 'Link' : undefined,
         formatterOpts: {
           to: {
             name:     `c-cluster-product-resource-id`,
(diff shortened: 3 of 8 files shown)


shell/plugins/dashboard-store/resource-class.js
```diff
@@ -1956,12 +1956,12 @@ export default class Resource {
...
         label:         this.t('component.resource.detail.glance.namespace'),
-        formatter:     'Link',
+        formatter:     this.$rootGetters['currentProduct']?.id && this.$rootGetters['currentCluster']?.id ? 'Link' : undefined,
         formatterOpts: {
           to: {
             name:     `c-cluster-product-resource-id`,
-            product:  this.$rootGetters['currentProduct'].id,
-            cluster:  this.$rootGetters['currentCluster'].id,
+            product:  this.$rootGetters['currentProduct']?.id,
+            cluster:  this.$rootGetters['currentCluster']?.id,
```
codyrancher, line 1964: Ran into an issue where currentProduct and currentCluster were both undefined.

We now just default to text if the location options aren't available.

shell/models/namespace.js
```diff
@@ -267,7 +267,9 @@ export default class Namespace extends SteveModel {
   }
 
   get hideDetailLocation() {
-    return !!this.$rootGetters['currentProduct'].hideNamespaceLocation;
+    const currentProduct = this.$rootGetters['currentProduct'];
+
+    return currentProduct ? !!currentProduct.hideNamespaceLocation : true;
```
codyrancher, line 272: ran into an issue where currentProduct wasn't defined. If that occurs I just default to hiding the location since it's safer and we won't have the data to link to it anyway.

shell/components/Resource/Detail/ResourcePopover/__tests__/index.test.ts
```diff
@@ -79,6 +79,18 @@ describe('component: ResourcePopover/index.vue', () => {
       expect(wrapper.find('.display').exists()).toBe(false);
     });
 
+    it('should show plain text and no PopoverCard when fetch fails', async() => {
+      mockClusterFind.mockRejectedValue(new Error('Not found'));
+      const wrapper = createWrapper(undefined, undefined, PopoverCardStub);
+
+      await wrapper.vm.$nextTick();
```
eva-vashkevich, line 86: why does it need 3 waits?

codyrancher: It's odd but it's actually needed if we keep this structure. The ticks are for:

1. The mockRejectedValue promise settles
2. The useFetch catch/finally block runs, setting error and clearing loading
3. Re-renders  based on the updated reactive state (v-if="!fetch.error" switches to the v-else span)

codyrancher: It's a little pervasive in this file and it's mostly because of the popover/fetch.

PR #16829, opened 2026-03-04
Implementing RcSection as specified by the Design System

### LINKED ISSUE #16119: Add a new accordion component following the new design

codyrancher:
### Summary
Fixes #16119

- New `RcSection` component with header (title, badges, actions, counter, errors slots), expand/collapse, and primary/secondary type and background variants.
- New `RcSectionBadges` helper — renders up to 3 `RcStatusBadge` warns and truncates if exceeded.
- New `RcSectionActions` helper — shows first 2 actions as buttons, overflows the rest into a dropdown.
- *Side fixes*: `RcIcon` - sizing, `RcCounterBadge`/`RcStatusBadge` sizing, fixed some of the badge colors.

### Technical notes summary
- Created the helper components instead of adding everything as props to `RcSection` to maintain both flexibility and ease of use.

- Dark theme: verify section background colors and toggle icon visibility.
- Badge truncation: only 3 render with console warning.
- Actions overflow: only 2 render then we add the secondary actions.
- Verify `RcCounterBadge`/`RcStatusBadge` sizing changes 
- Verify swapped colors

- **RcStatusBadge / RcCounterBadge** 
- **RcIcon sizing** 

### Screenshot/Video

https://github.com/user-attachments/assets/63a13230-b105-4512-a9c3-4c0a7cfe7e5c

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/pkg/rancher-components/src/components/RcSection/RcSection.vue
+++ b/pkg/rancher-components/src/components/RcSection/RcSection.vue
@@ -239,6 +239,7 @@ button.btn-medium.toggle-button {
   flex-direction: column;
   gap: 24px;
   padding: 0 0 16px;
+  color: var(--body-text);
 
   &.expandable-content {
     padding: 0 0 16px 24px;
--- a/storybook/src/stories/Components/RcSection.stories.ts
+++ b/storybook/src/stories/Components/RcSection.stories.ts
@@ -193,10 +193,10 @@ export const SecondaryFixed: Story = {
     template: `
       <div style="background: #EFEFEF; padding: 24px;">
         ${ withSlotContent(
-    secondaryFixedTemplate,
-    `${ contentGroup('Content Group 1', 'First group content goes here.', true) }
+      secondaryFixedTemplate,
+      `${ contentGroup('Content Group 1', 'First group content goes here.', true) }
            ${ contentGroup('Content Group 2', 'Second group content goes here.') }`,
-  ) }
+    ) }
       </div>
     `,
   }),
@@ -396,6 +396,80 @@ export const FullHeader: Story = {
     docs:     {
       canvas: { sourceState: 'shown' },
       source: { code: fullHeaderTemplate },
+    },
+  },
+};
+
+/**
+ * **Automatic alternating backgrounds**
+ *
+ * When the `background` prop is omitted, nested sections automatically
+ * alternate between `primary` and `secondary` backgrounds. The outermost
+ * section defaults to `primary` (no parent), its child resolves to
+ * `secondary`, the grandchild back to `primary`, and so on. This creates
+ * visual depth without any manual configuration.
+ */
+const alternatingBgTemplate = `<!-- No background prop — backgrounds alternate automatically -->
+<RcSection title="Level 1" type="primary" mode="with-header" :expandable="false">
+  <p>No parent — defaults to primary</p>
+  <RcSection title="Level 2" type="secondary" mode="with-header" :expandable="false">
+    <p>Parent is primary — auto-resolves to secondary</p>
+    <RcSection title="Level 3" type="secondary" mode="with-header" :expandable="false">
+      <p>Parent is secondary — auto-resolves back to primary</p>
+    </RcSection>
+  </RcSection>
+</RcSection>`;
+
+export const AlternatingBackgrounds: Story = {
+  render: (args: any) => ({
+    components: { RcSection },
+    setup() {
+      return { args };
+    },
+    template: alternatingBgTemplate,
+  }),
+  parameters: {
+    controls: { disabled: true },
+    docs:     {
+      canvas: { sourceState: 'shown' },
+      source: { code: alternatingBgTemplate },
+    },
+  },
+};
+
+/**
+ * **Overriding automatic backgrounds**
+ *
+ * You can override the automatic alternation by explicitly setting the
+ * `background` prop. Here the outermost section is `primary`, and we force
+ * both the child and grandchild to `primary` as well — breaking the
+ * automatic alternation. This is useful when you need consecutive sections
+ * to share the same visual weight regardless of nesting depth.
+ */
+const overrideBgTemplate = `<!-- Explicit background prop overrides the automatic alternation -->
+<RcSection title="Level 1" type="secondary" mode="with-header" :expandable="false" background="secondary">
+  <p>Explicit background="secondary"</p>
+  <RcSection title="Level 2" type="secondary" mode="with-header" :expandable="false" background="secondary">
+    <p>Would auto-resolve to primary, but forced to secondary</p>
+    <RcSection title="Level 3" type="secondary" mode="with-header" :expandable="false" background="secondary">
+      <p>Also forced to secondary — all three levels share the same background</p>
+    </RcSection>
+  </RcSection>
+</RcSection>`;
+
+export const BackgroundOverride: Story = {
+  render: (args: any) => ({
+    components: { RcSection },
+    setup() {
+      return { args };
+    },
+    template: overrideBgTemplate,
+  }),
+  parameters: {
+    controls: { disabled: true },
+    docs:     {
+      canvas: { sourceState: 'shown' },
+      source: { code: overrideBgTemplate },
     },
   },
 };


codyrancher:
## Open Questions - @oboc-sts 
- What region do we want to be clickable for expanding the section?
- What colors for dark mode?
- What should the styling for labeled actions be? (We wanted buttons and link buttons don't quite look like the design system)

pkg/rancher-components/src/components/Pill/RcCounterBadge/RcCounterBadge.vue
codyrancher, line 1: I noticed several size/color discrepancies in the previous DS components we worked on. I resolved all the issues I found here and the other components.

pkg/rancher-components/src/components/RcSection/RcSection.vue
```diff
@@ -0,0 +1,219 @@
...
+import { computed } from 'vue';
+import RcButton from '@components/RcButton/RcButton.vue';
+import RcIcon from '@components/RcIcon/RcIcon.vue';
+
+const props = withDefaults(defineProps<{
+  type: 'primary' | 'secondary';
+  mode: 'with-header' | 'no-header';
+  expandable: boolean;
+  expanded?: boolean;
+  background: 'primary' | 'secondary';
```
codyrancher: We have the same values for type. I think we may want to change one of the sets but I hadn't thought of a better convention.

pkg/rancher-components/src/components/RcSection/RcSection.vue
```diff
@@ -0,0 +1,219 @@
...
+
+.title {
+  display: inline-flex;
+  gap: 12px;
+  font-size: 18px;
+  line-height: 1.2;
+  color: var(--body-text, inherit);
+}
+
+button.btn-medium.toggle-button {
```
codyrancher, line 206: I know that this restyles the button component. We're being very specific about the sizes and spacing of the components so everything aligns in vertical columns when the sections are nested.

pkg/rancher-components/src/components/RcSection/RcSectionActions.vue
```diff
@@ -0,0 +1,93 @@
...
+
+const overflowActions = computed(() => (props.actions.length < 3 ? [] : props.actions.slice(2)));
+</script>
+
+<template>
+  <RcButton
+    v-for="(action, index) in primaryActions"
+    :key="index"
+    :class="{ 'icon-action': !action.label }"
+    variant="tertiary"
```
codyrancher: This was discussed with Ovi, we may end up changing the look and feel in the future but we wanted to have some interaction visibility so we went with something other than ghost.

shell/assets/styles/themes/_light.scss
codyrancher, line 1: Some of the colors were wrong/inverted in the status badges so I went through and corrected them. I think it only actually matters in modern.scss.

pkg/rancher-components/src/components/RcSection/RcSectionActions.vue
codyrancher, line 1: I was undecided as to whether we should do stories for these helper components are not.

pkg/rancher-components/src/components/RcSection/RcSection.vue
```diff
@@ -0,0 +1,219 @@
...
+ */
+import { computed } from 'vue';
+import RcButton from '@components/RcButton/RcButton.vue';
+import RcIcon from '@components/RcIcon/RcIcon.vue';
+
+const props = withDefaults(defineProps<{
+  type: 'primary' | 'secondary';
+  mode: 'with-header' | 'no-header';
+  expandable: boolean;
+  expanded?: boolean;
```
rak-phillip: It appears that we require the parent to keep track of the expanded state. I'm wondering if we can utilize `defineModel()`[^1] to allow either the parent to determine the expanded state or to track the state internally by default. 

[^1]: https://vuejs.org/api/sfc-script-setup.html#definemodel

pkg/rancher-components/src/components/RcSection/RcSection.vue
rak-phillip: We can consider alternatives to restyling the button itself, but I think that this is the intent of the ghost button overall - to provide custom button styles that our base buttons do not cover.

We also may want to consider if there's a better way to expose the ghost button for styling so that it doesn't require us to restyle the button component. Perhaps there's too many button details still associate with this variant.

pkg/rancher-components/src/components/RcSection/RcSection.vue
rak-phillip: I think that this is fine. My only other idea is to have this be a boolean that acts like a toggle, where "secondary" would be the "on" value.

rak-phillip (commented):
I haven't gotten far in my review, but I was considering the usage of the component. We intend for this component to be deeply nested and for the background color to alternate depending on its position in the hierarchy. It looks to me that the current implementation requires developers to track where a section is in relation to other sections and supply a background color as a slot. 

I think that this could be prone to mistakes, especially in scenarios where sections are deeply nested or a form might require repositioning/refactoring of sections in the future. My current proposal: What if we had some "root" or "base" section that defined the initial background color? We can then utilize provide/inject for children of the root section to automatically toggle the background color based on their position in the hierarchy. We can allow developers to override the background color via props.

This might increase complexity a great deal or be outside the scope for this component.. my current goal is to enhance the QOL of working with this component. What do you think?

codyrancher:
> I haven't gotten far in my review, but I was considering the usage of the component. We intend for this component to be deeply nested and for the background color to alternate depending on its position in the hierarchy. It looks to me that the current implementation requires developers to track where a section is in relation to other sections and supply a background color as a slot.
> 
> I think that this could be prone to mistakes, especially in scenarios where sections are deeply nested or a form might require repositioning/refactoring of sections in the future. My current proposal: What if we had some "root" or "base" section that defined the initial background color? We can then utilize provide/inject for children of the root section to automatically toggle the background color based on their position in the hierarchy. We can allow developers to override the background color via props.
> 
> This might increase complexity a great deal or be outside the scope for this component.. my current goal is to enhance the QOL of working with this component. What do you think?

My initial reaction is that I'd rather not add the complexity to the interface but given the first use case I think you may have a point. Let me mull it over some.

Edit: We now auto switch the background anytime it's not specified.

There is one odd condition that I don't think we can cover but the workaround is simple as a consumer:
```
<RcSection>
    <OtherComponent>
        <RcSection />
    <OtherComponent>
</RcSection>
```
The second RcSection will still alternate the background color even though it's not directly nested in the parent. The nested one can always just treat itself like a root and specify the background.

pkg/rancher-components/src/components/RcSection/RcSection.vue
codyrancher: Yeah, I originally did that but I was having issues with weback complaining about it being a macro inside storybook. That seems to no longer be an issue so... done.

pkg/rancher-components/src/components/RcSection/RcSection.vue
rak-phillip, line 1: The toggle for the background color using provide/inject works great! It does looks like text color for the section content is white by default in my tests, are you seeing the same?

Image

pkg/rancher-components/src/components/RcSection/RcSection.vue
```diff
@@ -0,0 +1,233 @@
+<script setup lang="ts">
+/**
+ * A section container used for grouping and organizing content, with an
+ * optional header that supports title, badges, actions, and expandability.
+ *
+ * Example:
+ *
+ * <RcSection title="Details" type="primary">
+ *   <p>Section content here</p>
+ * </RcSection>
```
rak-phillip: nit: this example is incomplete. Based on the `RcSectionProps` type, the `type`, `mode`, and `expandable` props are required.

pkg/rancher-components/src/components/RcSection/RcSection.vue
```diff
@@ -0,0 +1,233 @@
...
+  if (props.background) {
+    return props.background;
+  }
+
+  const parent = parentBackground?.value ?? null;
+
+  return parent === 'primary' ? 'secondary' : 'primary';
+});
+
+provide(RC_SECTION_BG_KEY, resolvedBackground);
```
rak-phillip, line 57: I really like the way this works. When I was working through the problem in my head, I didn't consider that we could nest `provide` in such a way and thought that we'd require a root component with a single `provide`.

pkg/rancher-components/src/components/RcSection/RcSectionActions.vue
rak-phillip: In thinking about this a little more, I think it would help to improve discoverability by having complete usage example to demonstrate the helper components in the RcSection story.

storybook/src/stories/Components/RcSection.stories.ts
```diff
@@ -0,0 +1,421 @@
...
+      description: 'Whether the section can be expanded/collapsed via the header.',
+    },
+    expanded: {
+      control:     { type: 'boolean' },
+      description: 'Whether the section is currently expanded. Supports v-model:expanded.',
+    },
+    background: {
+      options:     ['primary', 'secondary'] as SectionBackground[],
+      control:     { type: 'select' },
+      description: 'Background color of the section. Primary - a color that blends in with the background. Secondary - a color that contrasts the background. When omitted, nested sections automatically alternate between primary and secondary.',
```
rak-phillip, line 57: I think that we should have explicit examples to demonstrate this to really drive the usage home:

1. Automatic alternating backgrounds 
2. Overriding backgrounds that would otherwise alternate automatically 

Documentation/communication will be key for this behavior

pkg/rancher-components/src/components/RcSection/RcSection.vue
codyrancher: Fixed this example and provided a more exhaustive one as well.

storybook/src/stories/Components/RcSection.stories.ts
codyrancher: Added both 
image

codyrancher: Also addressing the following since I can't replay where the comment was made:
> In thinking about this a little more, I think it would help to improve discoverability by having complete usage example to demonstrate the helper components in the RcSection story.

I think this is addressed by resolving the show code on the page.
image

pkg/rancher-components/src/components/RcSection/RcSection.vue
codyrancher: Ended up just setting it to body-text. I have a feeling it'll mostly be dictated by the content.

PR #14187, opened 2025-04-24
Refactor AssignTo, MoveModal and feature flag modal as prompt modals

aalves08:
### Summary
Fixes #14221

### Occurred changes and/or fixed issues

- Refactor `AssignTo`, `MoveModal` and `feature flag` modal as prompt modals
- add unit tests to assert modals display
- remove store mechanisms that were dedicated to these modals
- create store mechanism to handle callbacks from `promptModal` on modal close

### Areas or cases that should be tested

- `AssignTo` modal - Fleet clusters list view table action (assign cluster to another workspace - `Change Workspace`)
- `MoveModal` modal - Move namespace to another project
- `feature flag` modal - Feature flags in settings. Table/menu action. Check `legacy` and `ui-sql-cache` for different scenarios

### Checklist
- [x] The PR is linked to an issue and the linked issue has a Milestone, or no issue is needed
- [x] The PR has a Milestone 
- [x] The PR template has been filled out
- [x] The PR has been self reviewed 
- [x] The PR has a reviewer assigned
- [x] The PR has automated tests or clear instructions for manual tests and the linked issue has appropriate QA labels, or tests are not needed
- [x] The PR has reviewed with UX and tested in light and dark mode, or there are no UX changes

PR #16833, opened 2026-03-04
Vite fixes we could make prior to the migration

codyrancher:
### Summary
Fixes we can make prior to the Vite migration. These were surfaced when working on https://github.com/rancher/dashboard/pull/16710.
- Removed unnecessary parentheses wrapping component definitions.
- Fixed `<stye>` tag typo to `<style>`
- Added `@shell/` prefix to import paths.
- Fixed existing bug in `FeatureFlagListDialog.vue` 
- Wrapped SCSS division in `calc()` 

### Technical notes summary
- See inline comments

- The `FeatureFlagListDialog.vue` change is the only one that could cause a regression.

- See above

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/SortableTable/index.vue
+++ b/shell/components/SortableTable/index.vue
@@ -2099,13 +2099,13 @@ export default {
 
   $header-padding: 20px;
   .sub-header-row {
-    padding: 0 0 $header-padding / 2 0;
+    padding: 0 0 calc($header-padding / 2) 0;
   }
 
   .fixed-header-actions {
     padding: 0 0 $header-padding 0;
     &.with-sub-header {
-      padding: 0 0 $header-padding / 4 0;
+      padding: 0 0 calc($header-padding / 4) 0;
     }
 
     width: 100%;


shell/components/form/BannerSettings.vue
```diff
@@ -9,7 +9,7 @@ import { ToggleSwitch } from '@components/Form/ToggleSwitch';
 import { TextAreaAutoGrow } from '@components/Form/TextArea';
 import { Banner } from '@components/Banner';
 
-export default ({
+export default {
```
codyrancher, line 12: Though this is valid javascript it was causing problems with static analysis inside `@vitejs/plugin-vue` (may have also been `vue-compiler-sfc` as well)

shell/detail/fleet.cattle.io.cluster.vue
```diff
@@ -6,7 +6,7 @@ import ResourceTabs from '@shell/components/form/ResourceTabs';
 import Tab from '@shell/components/Tabbed/Tab';
 import { MANAGEMENT, FLEET, SCHEMA } from '@shell/config/types';
 import { FLEET as FLEET_LABELS } from '@shell/config/labels-annotations';
-import { allHash } from 'utils/promise';
+import { allHash } from '@shell/utils/promise';
```
codyrancher, line 9: Without adding an additional alias we couldn't resolve the utils path. We could've made it relative or we could just use an existing `@shell` alias.

shell/components/SortableTable/index.vue
```diff
@@ -2099,13 +2099,13 @@ export default {
 
   $header-padding: 20px;
   .sub-header-row {
-    padding: 0 0 $header-padding / 2 0;
+    padding: 0 0 calc($header-padding / 2) 0;
```
codyrancher, line 2102: Division without calc is deprecated

https://sass-lang.com/documentation/breaking-changes/slash-div/#:~:text=Automatic%20Migration,for%20each%20stylesheet%20that's%20compiled.

shell/dialog/FeatureFlagListDialog.vue
```diff
@@ -134,7 +134,7 @@ export default {
             this.waiting = false;
             this.close();
 
-            await this.$store.dispatch('management/findAll', { type: this.resource, opt: { force: true } });
+            await this.$store.dispatch('management/findAll', { type: this.update.type, opt: { force: true } });
```
codyrancher, line 137: So this was a latent bug and we may want to resolve this another way.  `this.resource` is never defined in this component so this dispatch always fails and is swallowed by the empty `catch()`.

You can reproduce this by toggling a feature flag that requires a restart. It's easier if you add a console.log inside the catch that's swallowing the errors.

I debated deleting this all together but I decided to attempt to restore the original behavior.

This was surfaced when trying to fix e2e changes and may not strictly be necessary for the vite migration.

rak-phillip: I think this makes sense. It looks like the offending change was introduced via modal refactoring #14187; `shell/list/management.cattle.io.feature.vue` originally had a `resource` prop that was dropped in the transfer.

rak-phillip (approved):
LGTM

PR #16588, opened 2026-02-05
Add a label to RcItemCard actions

rak-phillip:
### Summary
This improves the accessibility of RcItemCard by adding an aria label for the action dropdown menu. 

Fixes #16579

### Occurred changes and/or fixed issues

- Add a label to RcItemCard actions

### Technical notes summary

The previous implementation would be announced as "Button Menu" by screen readers. This update makes it more clear which button is associated with any given card. For example, screen readers will now announce "Alerting Drivers actions", "Elemental actions", "Logging actions", etc... 

### Areas or cases that should be tested

With a screen reader active:

1. Navigate to Local Cluster Explorer => Cluster => Tools
2. Use the keyboard to navigate through each action associated with the cards rendered on page

### Areas which could experience regressions

This should be providing more context to the buttons. If a card title is long or confusing, this approach could be unintuitive. 

### Screenshot/Video

NA

### Checklist
- [x] The PR is linked to an issue and the linked issue has a Milestone, or no issue is needed
- [x] The PR has a Milestone 
- [x] The PR template has been filled out
- [x] The PR has been self reviewed 
- [x] The PR has a reviewer assigned
- [x] The PR has automated tests or clear instructions for manual tests and the linked issue has appropriate QA labels, or tests are not needed
- [x] The PR has reviewed with UX and tested in light and dark mode, or there are no UX changes
- [x] The PR has been reviewed in terms of Accessibility
- [x] The PR has considered, and if applicable tested with, the three Global Roles `Admin`, `Standard User` and `User Base`

PR #16871, opened 2026-03-06
Repositories dynamic tooltip

### LINKED ISSUE #15661: Helper to Detect Text Ellipsis

Sometimes we want to display a different UI when a text is truncated, for example, on Apps/Charts page, on the chart cards we display the chart's repository name as a tag which if the name is too long it gets truncated. When the text is not truncated we want to display a tooltip with the message "Find similar repositories...", but if truncated we want to display the whole text for that repository instead on the tooltip. Since this behaviour will be needed in more cases we need to create a helper to be re-used throughout the app.

This can be either a helper function or a directive.

**Additional context**

https://github.com/rancher/dashboard/issues/14798

momesgin:
### Summary
Fixes #15661 

Added dynamic tooltips to `AppChartCardFooter` labels. When a label text is truncated, hovering over it will display a tooltip containing the full label text prepended to the base tooltip.

### Technical notes summary
- Created an `isTruncated` utility function in `shell/utils/style.ts` to detect if an HTML element's text overflows its container.
- Added `mouseenter` event listeners and element refs in `AppChartCardFooter.vue` to dynamically update the tooltip content based on the truncation state of the label elements.

- Verify that hovering over truncated labels in the App Chart Card Footer displays the expanded text in the tooltip.
- Verify that non-truncated labels only show the default tooltip (if any) and do not show their own text in the tooltip.
- Tested on Chrome. Reviewers can test on Firefox or Safari.

- Tooltip behavior on App Chart Cards.

### Screenshot/Video

image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/pages/c/_cluster/apps/charts/AppChartCardFooter.vue
+++ b/shell/pages/c/_cluster/apps/charts/AppChartCardFooter.vue
@@ -3,10 +3,11 @@ import { reactive } from 'vue';
 import { RcItemCardAction } from '@components/RcItemCard';
 import { RcButton } from '@components/RcButton';
 import { isTruncated } from '@shell/utils/style';
-import { RcIcon } from '@components/RcIcon';
+import RcIcon from '@components/RcIcon/RcIcon.vue';
+import type { RcIconType } from '@components/RcIcon/types';
 
 interface FooterItem {
-  icon?: string;
+  icon?: RcIconType;
   iconTooltip?: { key?: string; text?: string };
   labels: string[];
   labelTooltip?: string;
--- a/shell/utils/style.ts
+++ b/shell/utils/style.ts
@@ -52,5 +52,8 @@ export function isTruncated(element: HTMLElement | null): boolean {
     return false;
   }
 
-  return element.scrollWidth > element.clientWidth || (element.scrollHeight - element.clientHeight > 1);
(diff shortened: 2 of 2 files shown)


momesgin:
> After fixing the context for the radio buttons in #16588, I think it's worth exploring making the tooltips consistent regardless of if they are truncated. For example:
> 
> * "Find charts based on the RKE2 repository" instead of "this repository"
> * "Find charts based on the Linux Only tag" instead of "this tag"
> 
> This would also work with long repo names and would reduce on the amount of code needed to toggle the correct tooltip.

I think that's a good idea! If @oboc-sts approves I can work on it

oboc-sts:
> After fixing the context for the radio buttons in #16588, I think it's worth exploring making the tooltips consistent regardless of if they are truncated. For example:
> 
> * "Find charts based on the RKE2 repository" instead of "this repository"
> * "Find charts based on the Linux Only tag" instead of "this tag"
> 
> This would also work with long repo names and would reduce on the amount of code needed to toggle the correct tooltip.

This ^ was the original idea. But we had to move away from this specific approach to a more generic one. I'm failing to  remember the reasons right now, but there was a reason 😬 . However, I'm pretty sure keeping consistency across multiple types of links had something to do with it.

Let's not change it please.

PR #16881, opened 2026-03-06
[16817] Fixed condition for appCo to allow namespace changes correctly

### LINKED ISSUE #16817: AppCo: Changing namespace does not update image pull secret UI for SUSE AppCo chart install
**Setup**
- Rancher version: v2.14.0-alpha7 (prime build)
- Rancher UI Extensions: Dashboard v2.14.0-alpha5
- Browser type & version:

**Describe the bug**
When installing a SUSE AppCo chart, changing the Namespace in the form does not update the image pull secret section.
Example: Switching from a namespace that has an existing image pull secret to a namespace that does not still shows the previous namespace’s messaging (e.g., “generate from repo secret”) instead of updating to the correct behavior for the new namespace (e.g., locked to manual creation).

**To Reproduce**

1. Create a Rancher Prime instance
2. Go to Cluster > Apps > Repositories > Create the SUSE AppCo repository
3. During creation of the repo, create Basic Auth secret (this should be created in the `cattle-system`namespace automatically)
4. Go to Apps & Marketplace → Charts and start installing a SUSE App Co chart (e.g. `alertmanager`).
5. Select namespace with the secret (`cattle-system`). Confirm the image pull secret section shows the expected message (e.g., "A new Image Pull Secret <image-pull-secret-name> will be generated from the Repository secret <basic-auth-secret-name>").
6. Change the namespace to one that has no image pull secret and no basic auth secret (no decoded repo secret)
7. Observe the image pull secret section.

**Result**
The image pull secret messaging stays as it was for the previously selected namespace. The UI does not refresh when the namespace is changed.

**Expected Result**
After changing the namespace, the UI should re-evaluate and show the correct messaging for the newly selected namespace:

- Namespace with existing image pull secret → default to that secret (Case 1)
- Namespace with no secret but decoded repo secret → offer to generate from repo secret (Case 2)
- Namespace with neither → locked to manual creation (Case 3)

**Screenshots**

https://github.com/user-attachments/assets/c0b6be6a-4ac3-419b-8923-469738a2ea25

**Additional context**
Discovered while testing https://github.com/rancher/dashboard/issues/16676

marcelofukumoto:
### Summary
Fixes #16817

- The condition for the fetched data is not correct, it should fix an edge case but now it is blocking normal namespace changes

### Technical notes summary

- I had reverted from trying to use fetchState pending but didn't revert the logic

- AppCo installation

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/pages/c/_cluster/apps/charts/install.vue
+++ b/shell/pages/c/_cluster/apps/charts/install.vue
@@ -938,7 +938,7 @@ export default {
 
     async initializeDataForNamespaceChanges() {
       // Skip the flow if the data still not fetched, it will trigger after fetching manually
-      if (!this.appCoDataFetched) {
+      if (this.appCoDataFetched) {
         try {
           this.defaultImagePullSecret = await this.$store.dispatch('cluster/find', { type: SECRET, id: `${ this.targetNamespace }/${ this.repo.spec.clientSecret.name }-image-pull-secret` });
         } catch (e) {


momesgin:
@marcelofukumoto I can reproduce the issue while I'm getting a 404 error:

https://github.com/user-attachments/assets/60c0f09c-9f3b-48c0-a828-94af24b8d62d

marcelofukumoto:
That seems to be correct. It is trying to fetch the imagePullSecret, returning 404.
Since it is not created it is going to create a new one.
When you change namespace it fetches again for the next one (which shows that the bug has been fixed)

The bug is that when you change namespace it was not fetching the data again.

momesgin:
> That seems to be correct. It is trying to fetch the imagePullSecret, returning 404. Since it is not created it is going to create a new one. When you change namespace it fetches again for the next one (which shows that the bug has been fixed)
> 
> The bug is that when you change namespace it was not fetching the data again.

I shared the error just as an extra info, maybe I'm not understanding the issue's description:

"_When installing a SUSE AppCo chart, changing the Namespace in the form does not update the image pull secret section.
Example: Switching from a namespace that has an existing image pull secret to a namespace that does not still shows the previous namespace’s messaging (e.g., “generate from repo secret”) instead of updating to the correct behavior for the new namespace (e.g., locked to manual creation)._"

I have created my secret in two namespaces(e.g. "cattle-system" during the repo creation and "default" manually). The message doesn't change no matter what namespace I choose from the dropdown, is this expected?

marcelofukumoto:
Yes! It is expected since the namespaces could have the same status.

You have a secret used in the REPO.
When you install the first APP it will create an image pull secret in the selected namespace.
When you install the second APP on the same namespace, it will use the previous created one.
If you create in one namespace it will be created only on that namespace. So the message will be different based on the namespace.

The bug was that changing namespace was not fetching the data from the newly selected namespace.

---
To make a clear example for a simple case for the bug fixed:
- install one app on namespace1
- namespace1 now has the image pull secret
- go to install again in a different app
- select namespace1
  - message should show: image pull secret will **not be created**
- select namespace2
  - message should show: image pull secret will **be created**  

The bug was the message was not refreshing, so changing namespace was not working.

momesgin:
> Yes! It is expected since the namespaces could have the same status.
> 
> You have a secret used in the REPO. When you install the first APP it will create an image pull secret in the selected namespace. When you install the second APP on the same namespace, it will use the previous created one. If you create in one namespace it will be created only on that namespace. So the message will be different based on the namespace.
> 
> The bug was that changing namespace was not fetching the data from the newly selected namespace.
> 
> To make a clear example for a simple case for the bug fixed:
> 
> * install one app on namespace1
> * namespace1 now has the image pull secret
> * go to install again in a different app
> * select namespace1
>   
>   * message should show: image pull secret will **not be created**
> * select namespace2
>   
>   * message should show: image pull secret will **be created**
> 
> The bug was the message was not refreshing, so changing namespace was not working.

I get it now, thanks for the clarification 🙏

PR #15605, opened 2025-10-09
Update Shell API's method signatures and enable typedoc integration

aalves08:
### Summary
Fixes #15597
Fixes #15850

**Important notice:** needs a rebase from https://github.com/rancher/dashboard/pull/15532 once that's merged

### Occurred changes and/or fixed issues

-  Update Shell API's methods usage signature to a pattern more "up-to-date" and easy to be consumed
- Integrate the `Notification Center` with the Shell API
- Add information for when the Shell API is available to be consumed
- Update typedoc integration to auto-generate new documentation (remove commented code so it's enabled by default)
- Remove `growls` from Shell API (deprecated)
- Add new `System` API to Shell API
- Fixes minor issues with focus trap on Shell API methods
- add unit tests for Shell API implementation

### Areas or cases that should be tested

- Check if documentation looks and builds fine
- Check if all Shell API methods are working fine and display intellisense 

### Checklist
- [x] The PR is linked to an issue and the linked issue has a Milestone, or no issue is needed
- [x] The PR has a Milestone 
- [x] The PR template has been filled out
- [x] The PR has been self reviewed 
- [x] The PR has a reviewer assigned
- [x] The PR has automated tests or clear instructions for manual tests and the linked issue has appropriate QA labels, or tests are not needed
- [x] The PR has reviewed with UX and tested in light and dark mode, or there are no UX changes
- [x] The PR has been reviewed in terms of Accessibility

PR #16895, opened 2026-03-10
Create API layer to consume k8s resources - part 1

### LINKED ISSUE #10097: Create Rancher API for extensions to access/manage resources

aalves08:
### Summary
Fixes #10097 

- Creates the foundation for the resource/k8s api abstraction
- Adjusts docusaurus documentation with new entry point for resource/k8s api abstraction
- Covers `find` store action
- Covers `findAll` store action
- Covers `findPage` store action
- Covers `findLabelSelector` store action
- adds unit tests

### Technical notes summary

- decision went for "store" separation between `cluster` and `mgmt` as it's easier to implement and maintain
- since Norman will be deprecated some time in the future, not covering `rancher` store
- implementation pattern follows the same as the [Shell API](https://github.com/rancher/dashboard/pull/15605)

- check if are in agreement with the api contract
- test all methods available in `cluster` and `mgmt`
- check generated documentation

None expected

### Screenshot/Video

Screenshot 2026-03-11 at 16 46 22
Screenshot 2026-03-11 at 16 46 12
Screenshot 2026-03-11 at 16 46 05

### Checklist
- [x] The PR has a Milestone 


shell/apis/resources/mgmt.ts
codyrancher, line 1: cluster.ts and mgmt.ts are basically identical. I think we could do a single implementation and just change the store/namespacing.

shell/apis/resources/mgmt.ts
```diff
@@ -0,0 +1,170 @@
...
+   * const clusters = await resources.mgmt.labelSelector<Cluster>('status=active');
+   * clusters.forEach(c => console.log(c.metadata.name));
+   * ```
+   */
+  async labelSelector<T extends ResourceBase = ResourceBase>(
+    selector: string,
+    options?: any
+  ): Promise<T[]> {
+    try {
+      const resources = await this.store.dispatch('management/findLabelSelector', {
```
codyrancher: We're not passing a `type` here.

shell/apis/resources/mgmt.ts
```diff
@@ -0,0 +1,170 @@
...
+        opt:  options || {}
+      });
+
+      return resources as T[];
+    } catch (e) {
+      // eslint-disable-next-line no-console
+      console.error(`Failed to list resources ${ resourceType }:`, e);
+
+      return [];
+    }
```
codyrancher: We're swallowing all of the errors in this file which prevents the consumers from being able to react to those errors. 

From the perspective of the consumer they'll have to assume that things worked correctly and we're just returning an empty list for no matches.

I'm not sure we want to do that.

shell/apis/resources/mgmt.ts
```diff
@@ -0,0 +1,170 @@
...
+   */
+  async list<T extends ResourceBase = ResourceBase>(
+    resourceType: ResourceType,
+    options?: any
+  ): Promise<T[]> {
+    try {
+      const resources = await this.store.dispatch('management/findPage', {
+        type: resourceType,
+        opt:  options || {}
+      });
```
codyrancher: While in practice I don't think it's going to matter given the timelines, I think we're going to have to gate the pagination API behind the feature flags.

shell/types/resources.ts
codyrancher, line 1: Is this generated or is this something we should maintain? I'm wondering if this should be broken up into separate files or if should just type the models.

codyrancher (commented):
I think purely from an interface level I'm good with the direction here but I have noticed a handful of things in my first pass on this. I still need some time to review further but I wanted to leave some of the current comments I have.

shell/types/resources.ts
richard-cox: Agree on model side. We expose instances of those that inherit all the native kube properties

There's some work on some parts of this via shell/types/rancher/steve.api.ts and shell/types/kube/kube-api.ts, but i think at model and general / common kube properties we should stick

shell/types/store/dashboard-store.types.ts
```diff
@@ -9,6 +9,23 @@ export type ActionCoreFindArgs = {
...
+   * Watch for changes
+   *
+   * false = no, all other values = yes
+   */
+  watch?: boolean,
+  /**
+   * If true don't persist the http response to the store, just pass it back (default: false)
+   */
+  transient?: boolean,
+  //  invalidatePageCache: boolean - whether to invalidate the page cache (default: false) ? is desc and default right ? ok
```
richard-cox: we're in an odd world where these are used internally so should have all the props used... and now also externally

we can either
- add all and mark them as internal
- create external interfaces that `omit` internal props

shell/plugins/dashboard-store/actions.js
```diff
@@ -660,11 +660,16 @@ export default {
   },
 
   // opt:
-  //  filter: Filter by fields, e.g. {field: value, anotherField: anotherValue} (default: none)
-  //  limit: Number of records to return per page (default: 1000)
-  //  sortBy: Sort by field
-  //  sortOrder: asc or desc
-  //  url: Use this specific URL instead of looking up the URL for the type/id.  This should only be used for bootstrapping schemas on startup.
+  //  watch: boolean - whether to start a watch on the selector (default: true) ok
```
richard-cox: we can chop these out and reference `@ActionFindArgs`

shell/apis/intf/resources-api/resource-base.ts
```diff
@@ -0,0 +1,36 @@
...
+ * resources.cluster.list('mycompany.io.customresource');
+ * ```
+ */
+export type ResourceType = K8SResourceType | string;
+
+/**
+ * Base interface for all resource objects returned from the API
+ * @internal
+ */
+export interface ResourceBase {
```
richard-cox: This is a steve resource, given it has id.

It probably needs to be in shell/types/rancher/steve.api.ts, called `RancherResourceBase`, and use metadata: RancherKubeMetadata

shell/apis/intf/resources-api/resource-constants.ts
```diff
@@ -0,0 +1,190 @@
+/**
```
richard-cox, line 1: I don't think we should be hardcoding kube resource types, just `string` should be enough

shell/apis/resources/mgmt.ts
richard-cox: Agree with this.

This would also unlock exposing any dashboard store coming from extensions (see addDashboardStore, we should be able to expose the same functionality for entries there. for example harvester has it's own dashboard based store that could still use the API)

shell/apis/intf/resources-api/cluster-api.ts
```diff
@@ -0,0 +1,127 @@
+import { ResourceType, ResourceBase } from './resource-base';
+import { ActionFindPageArgs, ActionFindLabelSelectorArgs, ActionFindArgs, ActionFindAllArgs } from '@shell/types/store/dashboard-store.types';
+
+/**
+ * Provides access to the Cluster API which can be used for managing cluster resources in Rancher UI
```
richard-cox, line 5:
```suggestion
 * Provides access to the Cluster API which can be used for managing Kubernetes cluster resources
```

shell/apis/intf/resources-api/mgmt-api.ts
```diff
@@ -0,0 +1,123 @@
+import { ResourceType, ResourceBase } from './resource-base';
+import { ActionFindPageArgs, ActionFindLabelSelectorArgs, ActionFindArgs, ActionFindAllArgs } from '@shell/types/store/dashboard-store.types';
+
+/**
+ * Provides access to the Management layer in Rancher UI (users, global settings, etc.)
```
richard-cox, line 5:
```suggestion
 * Provides access to the Management API which can be used for managing Kubernetes resources in Rancher's local cluster. This includes (users, global settings, etc.)
```

shell/apis/resources/cluster.ts
```diff
@@ -0,0 +1,175 @@
...
+   * import type { Pod } from '@shell/types/resources';
+   *
+   * const pod = await resources.cluster.get<Pod>(K8S.POD, 'my-pod-123');
+   * console.log(pod.spec.containers);
+   * ```
+   */
+  async get<T extends ResourceBase = ResourceBase>(
+    resourceType: ResourceType,
+    resourceId: string,
+    options?: any
```
richard-cox: options for all these are typed (see shell/types/store/dashboard-store.types.ts)

shell/apis/resources/mgmt.ts
richard-cox: Agree we need to pass back an error.

We should discuss whether we have an specific $resource Error class which either wraps the underlying error object or just prefixes the message, both give the calling dev more context

shell/apis/resources/cluster.ts
```diff
@@ -0,0 +1,175 @@
...
+   * @example
+   * ```ts
+   * import { K8S } from '@shell/apis';
+   * import type { Pod } from '@shell/types/resources';
+   *
+   * const pod = await resources.cluster.get<Pod>(K8S.POD, 'my-pod-123');
+   * console.log(pod.spec.containers);
+   * ```
+   */
+  async get<T extends ResourceBase = ResourceBase>(
```
richard-cox: we're going to get this muddled with `get` and `find`

Suggestion
- we stick with the find/findX notation
  - this covers both of it's use cases
    - if in cache it returns it
    - if not in cache it populates and returns it
- we leave the option to expose get later 
  - these just return the cache value. nothing fancy on watches, pages, etc

So in theory something like

- fetch single resource --> find
- fetch single or multiple resource via vai based request --> findFiltered
- fetch all of a type - findAll
- fetch all that are applicable to a labelSelector --> findLabelSelector

but they're still not right. a clean look would be something like

- fetch a single resource
- fetch some resources
- fetch all resources

which means something like

- find
- findFiltered (covers both findPage and fineLabelselector)
- findAll

to get findFiltered working in 2.15.0 means some weird plumbing which would go away in  2.16.0 (findPage will fail if vai is off)

richard-cox: `T extends ResourceBase = ResourceBase` i think would force anyone calling this to ensure their type extended our ResourceBase. i think instead we should default to it but not force them. which i think is just `T = ResourceBase`

shell/apis/resources/cluster.ts
```diff
@@ -0,0 +1,175 @@
...
+   * @example
+   * ```ts
+   * import { K8S } from '@shell/apis';
+   * import type { Deployment } from '@shell/types/resources';
+   *
+   * const deployments = await resources.cluster.list<Deployment>(K8S.DEPLOYMENT);
+   * deployments.forEach(d => console.log(d.spec.replicas));
+   * ```
+   */
+  async list<T extends ResourceBase = ResourceBase>(
```
richard-cox: this won't work when vai is disabled (depending on props). in theory this would only need taking into account in 2.15.0

richard-cox (commented):
Done a visual first pass. there's a LOT of really good polish here (thinking docs and inline comments with examples). We might think about doing this in two parts in the future, there might be a lot of churn with those as we tweak the interface in review

shell/apis/resources/cluster.ts
richard-cox: didn't see Cody's comment, closing and taking discussion there

shell/apis/resources/mgmt.ts
richard-cox: Yeah frustrating this one.

In theory 2.15.0 vai will be configurable via feature flag, then 2.16.0 always on.

shell/types/resources.ts
richard-cox: Adding the issue of resource changes as well.

They change between kube versions, mostly not a lot, but sometimes do. So keeping up to date would be a pain. We also then have the issue of downstream clusters being on different versions and thus have/don't have changes

shell/apis/resources/mgmt.ts
aalves08: Implementation now stems from the same class. Will push the code changes. Resolving

shell/apis/resources/mgmt.ts
aalves08: Missed that! Thank you! Will push the code changes. Resolving

shell/apis/resources/mgmt.ts
aalves08: Created a class method that will throw the error and also log it in the console. Will push the code changes. Resolving

shell/apis/resources/mgmt.ts
aalves08: What should be done here? we throw an error if the feature flag is disabled? Is the feature flag available to all user types?

shell/apis/intf/resources-api/cluster-api.ts
aalves08: Updated. Code will be pushed. Resolving

shell/apis/intf/resources-api/mgmt-api.ts
aalves08: Updated. Code will be pushed. Resolving

shell/apis/resources/cluster.ts
aalves08: Done as part of a previous thread. Code will be pushed. Resolving

shell/apis/resources/cluster.ts
aalves08: The implementation of what I've named `get` is the `find` action.

We can rename it if you disagree with the naming here. Any suggestions? 

Otherwise, what's implemented is the `find`, `findAll`, `findPage` and `findLabelSelector`.

These 4 store actions we need to cover on this first pass.

`findFiltered ` needing some "weird plumbing" doesn't work on this scope of changes.

The naming was pretty simple in my head: `get`, `list` and `labelSelector` (this last one is weird, I agree, but I don't have a better name for it). 

When we chatted and I found we had to cover the `findAll` as well, that where the names got a bit more confusing.

To recap, the implementation is:

```
`get` -> `find`
`list` -> `findPage`
`listAll` -> `findAll`
`labelSelector` -> `findLabelSelector`
```

What's the best course of action here?

aalves08: Fixed. Code will be pushed. Resolving

shell/types/store/dashboard-store.types.ts
aalves08: I've surfaced them to be present on the documentation + add all and mark as `internal` the ones we don't want to surface on the docs

shell/plugins/dashboard-store/actions.js
aalves08: Updated. Resolving

shell/apis/intf/resources-api/resource-constants.ts
aalves08: One of @nwmac goals was to have these well defined as a constant for usage with this API abstraction.

They are already defined in `shell/config/types` and in this file we are just surfacing them using those same "constants".

I think the key here is the generated documentation:

Screenshot 2026-03-25 at 13 15 40

Even thought it might look like a bit gibberish, I took the time to add comments to all the values so that devs know exactly what resource they are targeting. 

We also allow `string` as a fallback plan to fetch those CRD's that we don't know about

shell/types/resources.ts
aalves08: Okay, so I've removed all those definitions. I agree that maintaining them is hard and flimsy.

For now the responses of the API are based on `shell/types/rancher/steve.api`.

We have a really old issue https://github.com/rancher/dashboard/issues/5254 to address the conversion of models to TS.

Resolving

shell/apis/intf/resources-api/resource-base.ts
aalves08: Will be using `shell/types/rancher/steve.api.ts`, on which I'll extend `SteveListResponse` `T` as being `SteveGetResponse`. 

Given this, will chuck away `ResourceBase` as it doesn't make sense.

Resolving

shell/types/store/dashboard-store.types.ts
aalves08: @richard-cox I was able to get the deconstruction of the `Omit`to show all the relevant information the docs. There's a manageable middle layer with this, but now intelisense won't show properties that devs shouldn't use. Resolving

shell/apis/resources/cluster.ts
```diff
@@ -0,0 +1,175 @@
...
+   * ```
+   */
+  async labelSelector<T extends ResourceBase = ResourceBase>(
+    selector: string,
+    options?: any
+  ): Promise<T[]> {
+    try {
+      const resources = await this.store.dispatch('cluster/findLabelSelector', {
+        selector,
+        opt: options || {}
```
codyrancher: We're not passing the correct arguments to findLabelSelector, we're missing `type`, we're passing `selector` as a string when it should be a `matching` object.

shell/types/resources.ts
```diff
@@ -0,0 +1,1512 @@
...
+  };
+  status?: {
+    phase?: string;
+    conditions?: Array<any>;
+    podIP?: string;
+    startTime?: string;
+    containerStatuses?: Array<any>;
+    [key: string]: any;
+  };
+}
```
codyrancher: If we do want to keep these types in is there any desire to get more specific:
- conditions: type, status, reason, message, lastTransitionTime. https://pkg.go.dev/k8s.io/api/core/v1#PodCondition
- volumes: name, volumeSource https://pkg.go.dev/k8s.io/api/core/v1#Volume
- containerStatuses: https://pkg.go.dev/k8s.io/api/core/v1#ContainerStatus

It's probably too much effort but I figured I'd bring it up.

shell/apis/resources/mgmt.ts
codyrancher: I'm thinking we put the checks inside the store method(action).

I don't think we want to make all of the consumers check if the feature flag is enabled and check if the resource type is supported before they can use this. It might be better to either throw an exception the user can handle or fallback on fetching everything and filtering locally. 

@richard-cox is probably better to make suggestions here.

shell/apis/resources/cluster.ts
```diff
@@ -0,0 +1,175 @@
...
+   *
+   * // Fetch resources in specific namespaces
+   * const namespacedPods = await resources.cluster.listAll<Pod>(K8S.POD, {
+   *   namespaced: ['default', 'kube-system']
+   * });
+   * ```
+   */
+  async listAll<T extends ResourceBase = ResourceBase>(
+    resourceType: ResourceType,
+    options?: any
```
codyrancher: Minor, butt he interface in cluster-api has the options types for all of the methods. Here for example it's `ActionFindAllArgs`.

codyrancher (commented):
I left a few more comments, I think the only two that need to be fixed for approval for me is the labelSelector arguments and I think we need a solution to gating/handling the pagination apis.

shell/apis/resources/cluster.ts
aalves08: This file doesn't exist anymore... Apparently I asked for a re-review and forgot to push 🤦

shell/types/resources.ts
aalves08: Also removed this file... Sorry @codyrancher 🤦

shell/apis/resources/cluster.ts
aalves08: Updated... Didn't push the code 🤦

docusaurus/docs/extensions/resources-api.md
```diff
@@ -0,0 +1,102 @@
...
+The API is organized into two contexts:
+
+- **Cluster Context** (`cluster`) - For cluster-scoped Kubernetes resources (Pods, Deployments, Services, etc.)
+- **Management Context** (`mgmt`) - For global Rancher resources (Users, Clusters, Settings, etc.)
+
+## How to use the Resources API
+
+### Using Options API in Vue
+
+To use the Resources API in the context of the **Options API** of a Vue component, we globally expose the `$resources` property, which is available under the `this` object:
```
richard-cox, line 24: hate to say it, but we should be using the composition api in examples. in theory all new components should use it

shell/apis/resources/mgmt.ts
richard-cox: the root dashboard store has a convenience getter `paginationEnabled` which wraps checks for both feature flag and specific resource type config.

shell/apis/intf/resources-api/cluster-api.ts
```diff
@@ -0,0 +1,129 @@
...
+ * const deployment = await resources.cluster.get<Deployment>(K8S.DEPLOYMENT, 'my-app');
+ * deployment.spec.replicas; // ✅ typed!
+ * ```
+ */
+export interface ClusterApi {
+  /**
+   * Finds a specific resource by its type and ID.
+   *
+   * @template T - The type of the resource (defaults to SteveGetResponse)
+   * @param resourceType - The type of the resource to find (use **{@link K8S}** contant). See also {@link ResourceType}.
```
richard-cox: `contant` --> `constant` (there's quite a few of these

shell/plugins/dashboard-store/actions.js
```diff
@@ -659,12 +659,7 @@ export default {
     return getters.all(type);
   },
 
-  // opt:
-  //  filter: Filter by fields, e.g. {field: value, anotherField: anotherValue} (default: none)
-  //  limit: Number of records to return per page (default: 1000)
-  //  sortBy: Sort by field
-  //  sortOrder: asc or desc
-  //  url: Use this specific URL instead of looking up the URL for the type/id.  This should only be used for bootstrapping schemas on startup.
+  // opt:ActionFindArgs
```
richard-cox:
```suggestion
  // opt: @ActionFindArgs
```

shell/apis/resources/resources-api-class.ts
```diff
@@ -0,0 +1,177 @@
...
+import { ClusterApi } from '@shell/apis/intf/resources-api/cluster-api';
+import { MgmtApi } from '@shell/apis/intf/resources-api/mgmt-api';
+import { SteveListResponse, SteveGetResponse } from '@shell/types/rancher/steve.api';
+import { GetMethodOptions, ListAllMethodOptions, ListMethodOptions, LabelSelectorMethodOptions } from '@shell/types/store/dashboard-store.types';
+import { Store } from 'vuex';
+
+export class ResourcesApiClassImpl implements ClusterApi, MgmtApi {
+  private store: Store<any>;
+
+  private storeType: 'cluster' | 'management';
```
richard-cox: i think we should open it up to `private storeType: 'cluster' | 'management' | string;` so any store inherited from dashboard-store could have their own strongly typed accessor. then if the harvester team wanted the same API they could just then do `new ResourcesApiClassImpl(store, 'harvester')`

this covers extensions that use addDashboardStore

shell/apis/resources/mgmt.ts
richard-cox: we have interfaces for cluster and management that are functionally identical (given they're implemented by the same class) and only different by the examples in jsdoc. 

with https://github.com/rancher/dashboard/pull/16895/changes#r3064624812 we can open this up to extensions that use addDashboardStore (harvester, epinio). these aren't cluster or management stores though, so being an object that implements cluster/management interfaces would be odd

propose we have 

- ResourcesApiClass implements ResourcesApi
  - new ResourcesApi contains methods from ClusterApi/MgmtApi
  - we can skip docs
- ClusterApi extends ResourcesApi
  - if we need to keep this for store specific examples the `extends ResourceApi` would be the only change
- MgmtApi extends ResourcesApi
  - as per clusterapi

If we change ResourcesApi everything would need updating, but would give clear TS errors

shell/types/store/dashboard-store.types.ts
```diff
@@ -2,21 +2,88 @@ import { KubeLabelSelector } from '@shell/types/kube/kube-api';
...
+  method?: string,
+  /**
+   * Used if we want to specify custom headers for the request
+   */
+  headers?: object,
+  /**
+   * @internal
+   * Array of namespaces to filter by (used in url path, not part of pagination params)
+   */
+  namespaced?: string[],
```
richard-cox: namespaced, sortby and sortorder aren't applicable to find requests

shell/types/store/dashboard-store.types.ts
```diff
@@ -2,21 +2,88 @@ import { KubeLabelSelector } from '@shell/types/kube/kube-api';
...
+   * Use this to override URL instead of looking up the URL for the type/id
+   */
+  url?: string,
+}
+
+/**
+ * @interface
+ * Resources API "get" options
+ */
+export type GetMethodOptions = Omit<ActionFindArgs, 'invalidatePageCache' | 'namespaced' | 'sortBy' | 'sortOrder'>;
```
richard-cox: i think these get/list etc options are good change, but to make it clearer could they be in a file in shell/apis/intf/resources-api?

shell/types/store/dashboard-store.types.ts
```diff
@@ -38,24 +105,38 @@ export interface ActionFindAllArgs extends ActionCoreFindArgs {
...
    * Specifies the name to use if we should save the count returned in the paginated request
    */
   saveCountAs?: string,
 }
 
 /**
+ * @interface
+ * Resources API "listAll" options
+ */
+export type ListAllMethodOptions = Omit<ActionFindAllArgs, 'hasManualRefresh' | 'depaginate' | 'saveCountAs'>;
```
richard-cox: also needs incremental (it'll be removed when incremental loading goes)

shell/apis/intf/resources.ts
```diff
@@ -0,0 +1,47 @@
...
+ * // Cluster-scoped resources (current cluster context)
+ * const pods = await resources.cluster.list<Pod>(K8S.POD);
+ * const deployment = await resources.cluster.get<Deployment>(K8S.DEPLOYMENT, 'my-app');
+ *
+ * // Management/global resources
+ * const users = await resources.mgmt.list(K8S.USER);
+ * const cluster = await resources.mgmt.get(K8S.CLUSTER, 'c-abc123');
+ * ```
+ */
+export interface ResourcesApi {
```
richard-cox: do we need an interface for this?

shell/plugins/dashboard-store/actions.js
richard-cox: this should also have `@returns @ActionFindResponse`

shell/types/store/dashboard-store.types.ts
```diff
@@ -2,21 +2,88 @@ import { KubeLabelSelector } from '@shell/types/kube/kube-api';
...
+  watch?: boolean,
+  /**
+   * @internal
+   * Whether to invalidate the page cache (default: false)
+   */
+  invalidatePageCache?: boolean,
+  /**
+   * XHR request verbs `get`, `post`, `put`, `delete` (default: get)
+   */
+  method?: string,
```
richard-cox, line 38: we don't specify method or headers for any other method. sure they're needed?

shell/apis/resources/cluster.ts
richard-cox: The mapping is clear, however the surface will get messy. So i was suggesting

- find
- findFiltered
  - covers both findPage and findMatching, 
- findAll

for when we implement 'get' we could then have

- cache
- cacheFiltered
- cacheAll

having them grouped by a prefix makes it clear
- find - get from cache, or populate it and return cache
- cache - get from cache

In terms of findPage/findAll vs list/listAll
- findPage
  - requires SSP to exercise the filter/sort features
  - watches resources in a very different way (if ANY change, make a new http request)
- findAll
  - does not support filter/sort features
  - watches resources in a very different way (if one changes, change is received over socket)

so i don't think we should push users to use `list` as findPage if they're not interested in it's advanced sorting/filtering features. which means we flip it from list/listAll to listFilter/listAll

in terms of combining findPage and findMatching, this fits nicely in the use case of 'i'd like to fetch resources filtered by something'. we could split them into something more confusing findFiltered,findFilteredLabelSelector, though we can work together to find a nicer method overload to cover them both

docusaurus/docs/extensions/resources-api.md
```diff
@@ -0,0 +1,102 @@
+---
+id: resources-api
+---
+
+# Resources API
+
+> Available from Rancher `2.15` and onwards
```
richard-cox: can we tag this a experimental as well, just so we can get some feedback as per the product api

shell/apis/intf/resources-api/cluster-api.ts
```diff
@@ -0,0 +1,129 @@
...
+   * @example
+   * ```ts
+   * import { K8S } from '@shell/apis';
+   * import type { Pod } from '@shell/types/resources';
+   *
+   * const pod = await resources.cluster.get<Pod>(K8S.POD, 'my-pod-123');
+   * console.log(pod.spec.containers);
+   * ```
+   */
+  get<T = SteveGetResponse>(
```
richard-cox: these all don't return the SteveXResponse objects, but the resources/s themselves

shell/apis/resources/index.ts
```diff
@@ -0,0 +1,21 @@
...
+import { ClusterApi, MgmtApi, ResourcesApi } from '@shell/apis/intf/resources';
+import { ResourcesApiClassImpl } from './resources-api-class';
+
+export class ResourcesApiImpl implements ResourcesApi {
+  private clusterApi: ClusterApi;
+  private mgmtApi: MgmtApi;
+
+  constructor(store: Store<any>) {
+    this.clusterApi = new ResourcesApiClassImpl(store, 'cluster');
+    this.mgmtApi = new ResourcesApiClassImpl(store, 'management');
```
richard-cox: one for the future

it would be good to run through the stores added via addDashboardStore and expose them here

shell/apis/intf/resources-api/cluster-api.ts
richard-cox: there's one caveat, find and findPage have a 'transient' prop which directly returns the result without touching / storing in the cache. i don't think we need to cover that now though, and if we do it'd be a `T | Response` change which shouldn't break anyone

shell/apis/intf/resources.ts
aalves08: Yep. We use it in `shell/apis/index.ts` and `shell/apis/resources/index.ts`. It's marked as `@internal` so it doesn't appear on the generated docs

shell/apis/resources/index.ts
aalves08: updated

shell/types/store/dashboard-store.types.ts
aalves08: moved the "options" definitions to `shell/apis/intf/resources-api/resource-base`

shell/apis/intf/resources-api/cluster-api.ts
aalves08: fixed

shell/plugins/dashboard-store/actions.js
aalves08: updated

shell/types/store/dashboard-store.types.ts
aalves08: fixed

shell/types/store/dashboard-store.types.ts
aalves08: omitted `incremental` as requested. updated

shell/types/store/dashboard-store.types.ts
aalves08: Idk... The properties are hooked up to the store method though... I went through the code there and picked up what was connected. 

I'll mark them as internal and also omit them from the `GetMethodOptions`. Updated

shell/apis/resources/resources-api-class.ts
aalves08: updated the type to accept `string`.

docusaurus/docs/extensions/resources-api.md
aalves08: Updated

shell/apis/resources/mgmt.ts
aalves08: updated as per your indication. It's cleaner 🙏

docusaurus/docs/extensions/resources-api.md
aalves08: updated, even the shell API docs, to make sure our examples reflect Composition API usage and not Options API.

shell/apis/intf/resources-api/cluster-api.ts
aalves08: SteveGetResponse is just a very sparse definition, which has `id` and a wildcard `[key: string]: any`, which I think fits the bill still...

shell/apis/resources/cluster.ts
aalves08: Instead of:

`get` -> `find`
`list` -> `findPage`
`listAll` -> `findAll`
`labelSelector` -> `findLabelSelector`

We now have:

`find` -> `find`
`findFiltered` -> `findPage` and `findMatching`
`findAll` -> `findAll`

aalves08:
@codyrancher @richard-cox ready for another round of review 🙏

shell/apis/intf/resources-api/resources-api.ts
```diff
@@ -0,0 +1,98 @@
...
+   * const resources = useResources();
+   * const pod = await resources.cluster.find(K8S.POD, 'my-pod-123');
+   * ```
+   */
+  find<T = SteveGetResponse>(
+    resourceType: ResourceType,
+    resourceId: string,
+    options?: FindMethodOptions
+  ): Promise<T | null>;
+
```
codyrancher, line 51: The return type `Promise<T | null>` advertises a nullable result, but no code path in `resources-api-class.ts#find` actually returns `null`

docusaurus/src/css/custom.css
```diff
@@ -189,6 +189,43 @@ h5.anchor {
...
+  margin-top: 0.35rem;
+  margin-bottom: 0.2rem;
+}
+
+/* h4[id^="inherited-from"] + div,
+h5[id^="inherited-from"] + div {
+  background-color: transparent;
+  box-shadow: none;
+  border: 1px solid var(--ifm-color-emphasis-600);
+} */
```
codyrancher: Commented-out CSS block left in

shell/apis/resources/resources-api-class.ts
```diff
@@ -0,0 +1,139 @@
...
+    resourceType: ResourceType,
+    options?: FindFilteredMethodOptions
+  ): Promise<T[]> {
+    try {
+      const { selector, ...rest } = options || {};
+
+      const resources = await this.store.dispatch(`${ this.storeType }/findLabelSelector`, {
+        type: resourceType,
+        selector,
+        opt:  rest || {}
```
codyrancher: `opt: rest || {}` - rest is the result of` const { selector, ...rest } = options || {}`, so it's always an object (never null/undefined). The || {} is dead. opt: rest is sufficient.

shell/apis/resources/resources-api-class.ts
```diff
@@ -0,0 +1,139 @@
...
+import { Store } from 'vuex';
+
+export class ResourcesApiClassImpl implements ResourcesApi {
+  private store: Store<any>;
+
+  private storeType: 'cluster' | 'management' | string;
+
+  private surfaceError(message: string): never {
+    console.error(`Resource API error - ${ this.storeType } - ${ message }`); // eslint-disable-line no-console
+    throw new Error(`Resource API error - ${ this.storeType } - ${ message }`);
```
codyrancher: Maybe for future, I wonder if we'd like to type some of our exceptions to make exception handling more robust.

Image

codyrancher (dismissed):
I only found minor things this time through so feel free to merge unless you'd really like to address them.

docusaurus/src/css/custom.css
aalves08: Removed

shell/apis/intf/resources-api/resources-api.ts
aalves08: Good point. Now a `find` that doesn't yield any results will return `null` as advertised. Helps users deal with a "result not found". Thanks for pointing it out! Resolving

shell/apis/resources/resources-api-class.ts
aalves08: True. Fixed

codyrancher (dismissed):

shell/apis/resources/resources-api-class.ts
```diff
@@ -0,0 +1,139 @@
...
+   * ```
+   */
+  async findFiltered<T = SteveListResponse>(
+    resourceType: ResourceType,
+    options?: FindFilteredMethodOptions
+  ): Promise<T[]> {
+    try {
+      const { selector, ...rest } = options || {};
+
+      const resources = await this.store.dispatch(`${ this.storeType }/findLabelSelector`, {
```
richard-cox: The arguments for findLabelSelector are

```
 {
    type,
    matching: {
      namespace,
      labelSelector
    },
    opt
  }
```

so `selector` arg in FindFilteredMethodOptions isn't right.

ActionFindMatchingArgs (which is FindFilteredMethodOptions) has labelSelector and namespaced which are the mathcing.labelSelector and matching.namespace

shell/apis/resources/cluster.ts
richard-cox: findFiltered at the moment just calls findLabelSelector. it should be doing...

SSP enabled for type | findPage Arg | findLabelSelector Arg
---|---|---
false | returns error | returns findLabelSelector
true | returns findPage | returns findLabelSelector

Given this the FindFilteredMethodOptions should just be PageOptions | MatchingOptions, we then check in findFiltered and if pagination is enabled for the `type` and call the appropriate method

shell/apis/intf/resources-api/resources-api.ts
```diff
@@ -0,0 +1,98 @@
...
+   * // Paginated listing
+   * const deployments = await resources.cluster.findFiltered(K8S.DEPLOYMENT);
+   *
+   * // With label selector
+   * const pods = await resources.cluster.findFiltered(K8S.POD, { selector: 'app=nginx,env=prod' });
+   * ```
+   */
+  findFiltered<T = SteveListResponse>(
+    resourceType: ResourceType,
+    options?: FindFilteredMethodOptions
```
richard-cox: i think there's only one reason to call findFiltered without an arg, which would be with SSP enabled and the developer wanted to fetch all resources but update them via polling instead of individual socket messages. if so that would be part of findAll. i don't think we need to cover that either

so `options` should be mandatory i.e to find find filtered resources they should always provide a filter

shell/apis/resources/resources-api-class.ts
```diff
@@ -0,0 +1,139 @@
...
+import { ResourcesApi } from '@shell/apis/intf/resources-api/resources-api';
+import { SteveListResponse, SteveGetResponse } from '@shell/types/rancher/steve.api';
+import { Store } from 'vuex';
+
+export class ResourcesApiClassImpl implements ResourcesApi {
+  private store: Store<any>;
+
+  private storeType: 'cluster' | 'management' | string;
+
+  private surfaceError(message: string): never {
```
richard-cox: i thought i tested it, but just hit it.

we're swallowing the stack trace. so the error is good but doesn't help developers work out where in code the error is.

Image

think we're better off with 

```
  private surfaceError(message: string, e: any): never {
    console.error(`Resource API error - ${ this.storeType } - ${ message }`); // eslint-disable-line no-console
    throw new Error(`Resource API error - ${ this.storeType } - ${ message }`, { cause: e });
  }
```

Image

shell/apis/intf/resources-api/resources-api.ts
richard-cox: there's lots of `.findFiltered(<type>)` examples / docs that need updating too

richard-cox: i also tried this out 

const pods = await this.$resources.cluster.findFiltered('pods');

and it threw an exception 

Image

note that the all the test in `shell/apis/resources/__tests__/resources-api-class.test.ts` pass because they don't actually call the findLabelSelector action. as a seperate issue we should have unit tests that work through the store level and we mock the store contents (instead of store dispatch fn)

docusaurus/docs/extensions/resources-api.md
```diff
@@ -0,0 +1,103 @@
...
+
+To use the Resources API in the context of the **Options API** of a Vue component, we globally expose the `$resources` property, which is available under the `this` object:
+
+```ts
+import { K8S } from '@shell/apis';
+
+export default {
+  async mounted() {
+    // Cluster context
+    const pods = await this.$resources.cluster.findFiltered(K8S.POD);
```
richard-cox, line 32: when these are updated to work with either the findPage or findLabelSelector (probably best to stick with the former) we should follow the same approach as product api

```
const pageArgs: PageOptions = { ... }
const pods = await this.$resources.cluster.findFiltered(K8S.POD, pageArgs);
```

docusaurus/docs/extensions/resources-api.md
```diff
@@ -0,0 +1,103 @@
...
+id: resources-api
+---
+
+# Resources API (Experimental)
+
+> Note: This API is experimental and may change in future releases. Available from Rancher `2.15` and onwards
+
+## What is the Resources API?
+
+The Resources API is a functional API that helps extension developers to interact with Kubernetes and Rancher resources programmatically. This API provides a type-safe interface for common resource operations such as listing, retrieving, and filtering resources with full TypeScript support and autocomplete.
```
richard-cox:
```suggestion
The Resources API helps extension developers interact with Kubernetes and Rancher resources programmatically. This API provides a type-safe interface for common resource operations such as listing, retrieving, and filtering resources with full TypeScript support and autocomplete.
```

shell/apis/intf/resources-api/resource-constants.ts
```diff
@@ -0,0 +1,188 @@
...
+
+  // Policy
+  /** Kubernetes PodDisruptionBudget policy resource - `policy.poddisruptionbudget` */
+  POD_DISRUPTION_BUDGET,
+
+  // Autoscaling
+  /** Kubernetes HorizontalPodAutoscaler resource - `autoscaling.horizontalpodautoscaler` */
+  HPA,
+
+  // Management/Rancher resources
```
richard-cox: we should only externally expose the ones that are in the public api

https://ranchermanager.docs.rancher.com/api/api-reference

shell/apis/intf/resources-api/resource-constants.ts
```diff
@@ -0,0 +1,188 @@
...
+
+  // Catalog
+  /** Catalog App resource - `catalog.cattle.io.app` */
+  APP:          CATALOG.APP,
+  /** Catalog ClusterRepo resource - `catalog.cattle.io.clusterrepo` */
+  CLUSTER_REPO: CATALOG.CLUSTER_REPO,
+
+  // Provisioning (CAPI)
+  /** Rancher Provisioning Cluster resource - `provisioning.cattle.io.cluster` */
+  RANCHER_CLUSTER:    CAPI.RANCHER_CLUSTER,
```
richard-cox: this one is private, others public

shell/apis/intf/resources-api/resource-constants.ts
richard-cox: same for Fleet and Catalog

docusaurus/docs/extensions/resources-api.md
```diff
@@ -0,0 +1,103 @@
...
+K8S.NODE
+
+// Workload resources
+K8S.DEPLOYMENT
+K8S.STATEFUL_SET
+K8S.DAEMON_SET
+
+// Rancher Management
+K8S.USER
+K8S.CLUSTER
```
richard-cox: this isn't in the public api

shell/apis/intf/resources-api/resources-api.ts
```diff
@@ -0,0 +1,98 @@
...
+ * const deployment = await resources.cluster.find(K8S.DEPLOYMENT, 'my-app');
+ * ```
+ */
+export interface ResourcesApi {
+  /**
+   * Finds a specific resource by its type and ID.
+   *
+   * @template T - The type of the resource (defaults to SteveGetResponse)
+   * @param resourceType - The type of the resource to find (use **{@link K8S}** constant). See also {@link ResourceType}.
+   * @param resourceId - The unique identifier of the resource to find.
```
richard-cox: don't know if we call this out anywhere, but the id is the formate namespace(if resource is namespaced)/ name

shell/apis/intf/resources-api/resources-api.ts
```diff
@@ -0,0 +1,98 @@
...
+   * @param options - Optional find arguments
+   * @returns The found resource item or null if not found.
+   *
+   * @example
+   * ```ts
+   * import { useResources, K8S } from '@shell/apis';
+   *
+   * const resources = useResources();
+   * const pod = await resources.cluster.find(K8S.POD, 'my-pod-123');
```
richard-cox: this example has a namespaced resource without a namespace (see https://github.com/rancher/dashboard/pull/16895/changes#r3161275307)

if users use this interestingly it silently succeeds, as the kube api thinks it's a fetch all so returns an empty list response (i checked if the same happens with $store.dispatch('cluster/find' and it does).

so this should be something like `my-namespace/my-pod-123`.

a nice value add for future would be to check if the resource is namespaced and if so ensure the id contains an `/`

shell/apis/resources/index.ts
```diff
@@ -0,0 +1,21 @@
+import { Store } from 'vuex';
+import { ClusterApi, MgmtApi, ResourcesApiProvider } from '@shell/apis/intf/resources';
+import { ResourcesApiClassImpl } from './resources-api-class';
+
+export class ResourcesApiImpl implements ResourcesApiProvider {
+  private clusterApi: ClusterApi;
+  private mgmtApi: MgmtApi;
+
+  constructor(store: Store<any>) {
+    this.clusterApi = new ResourcesApiClassImpl(store, 'cluster');
```
richard-cox:
```suggestion
    this.clusterApi = new ResourcesApiClassImpl(store, STORE.CLUSTER);
```

shell/apis/resources/index.ts
```diff
@@ -0,0 +1,21 @@
...
+    this.mgmtApi = new ResourcesApiClassImpl(store, 'management');
```
richard-cox:
```suggestion
    this.mgmtApi = new ResourcesApiClassImpl(store, STORE.MANAGEMENT);
```

shell/types/store/dashboard-store.types.ts
```diff
@@ -38,24 +85,32 @@ export interface ActionFindAllArgs extends ActionCoreFindArgs {
...
+   */
   hasManualRefresh?: boolean,
+  /**
+   * Number of records to return per page
+   */
   limit?: number,
   /**
+   * @internal
    * Iterate over all pages and return all resources.
-   *
```
richard-cox, line 45: lost comment

richard-cox (commented):
Will we hit issues when extensions that use a new shell with this in are loaded by an old rancher without the api?

I think there was something like that for the plugin --> extension change?

shell/apis/resources/resources-api-class.ts
aalves08: Fixed and also created https://github.com/rancher/dashboard/issues/17405 to address the same issue on the new prod registration.

shell/apis/intf/resources-api/resources-api.ts
aalves08: That's an important piece of information 😬  thanks for the feedback on that 🙏

docusaurus/docs/extensions/resources-api.md
aalves08: Fixed

shell/apis/intf/resources-api/resource-constants.ts
aalves08: updated

shell/apis/intf/resources-api/resource-constants.ts
aalves08: updated

docusaurus/docs/extensions/resources-api.md
aalves08: updated

shell/apis/intf/resources-api/resources-api.ts
aalves08: updated description/comment so that docs tell the developer how to get the id if it's namespaced

shell/apis/intf/resources-api/resources-api.ts
aalves08: added validation + updated unit tests + comments + examples

shell/apis/resources/index.ts
aalves08: updated

shell/apis/resources/index.ts
aalves08: updated

shell/types/store/dashboard-store.types.ts
aalves08: That's fine. The correct ending bracket is there

shell/apis/resources/resources-api-class.ts
aalves08: There was a lot of weirdness here... args weren't matching, `option` was an optional arg which it shouldn't be, the mashup of the types for `findPage` and `findMatching` was wrong and didn't reproduce the type needed for `findLabelSelector`, etc.

Now `findLabelSelector` uses a signature overload (as agreed) to reproduce both internal scenarios correctly. I've also exposed many more types related to pagination for `findPage` and all.

Also updated examples (mostly removed them because this method is complicated enough and was used in a bunch of places where it just introduced complication...)

Should be much better implemented. Have a look

docusaurus/docs/extensions/resources-api.md
aalves08: Check https://github.com/rancher/dashboard/pull/16895#discussion_r3167695554. Closing this one and keeping the linked comment open just as an umbrella comment since `findFiltered` was wrong in numerous ways

shell/apis/intf/resources-api/resources-api.ts
aalves08: Check https://github.com/rancher/dashboard/pull/16895#discussion_r3167695554.

I think there's value on the current unit tests we have but definitely we could improve them. I've created https://github.com/rancher/dashboard/issues/17321 to create a separate set of tests that go in depth to the store.

shell/apis/resources/cluster.ts
richard-cox: @aalves08 this one is still pending?

the `findFiltered` method itself should...

- if `options` is of type for findPage, call findPage
  - options is a type that extends from ActionFindPageArgs
  - options.pagination will be a truthy
- if `options` is of type for findLabelSelector, call findLabelSelector
  - options is of a type that extends from ActionFindLabelSelectorArgs
  - options.labelSelector will be a truthy

i can understand why there's a bit of a misunderstanding regarding determining types, in `master` these should be

```
export type ActionFindLabelSelectorArgs = ActionFindPageArgs | ActionFindMatchingArgs;

// should be

export type ActionFindLabelSelectorArgs = Omit<ActionFindPageArgs, 'pagination' | 'namespace'> | ActionFindMatchingArgs;
```

if you look at findLabelSelector it does call findPage under the hood but specifically to make the a request to vai with label selectors specific to it (and overrides pagination and namespace)

richard-cox: FindFilteredPageOptions then also should `= ActionFindPageArgs `

shell/apis/intf/resources-api/resources-api.ts
richard-cox: Agreed! keep the current set of unit tests, create some more in a different PR.

Kept this one open given that there's still some examples without the `options` object.

FWIW here's a couple of the cleanest examples for each type

```
    const pods1 = await this.$resources.cluster.findFiltered('pod', {
      pagination: {
        page:                 1,
        projectsOrNamespaces: [],
        filters:              [],
        sort:                 []
      }
    });

    const pods2 = await this.$resources.cluster.findFiltered('pod', { labelSelector: { matchLabels: { type: 'my-type' } } });
```
i think there's a good argument for making some of the properties in PaginationArgs optional (i think all of them, and the UI will provide defaults). that's an issue in `master` though

richard-cox: going on https://github.com/rancher/dashboard/pull/16895#discussion_r3161204265 the examples would be

```
const pods1Args: find page type = {
      pagination: {
        page:                 1,
        projectsOrNamespaces: [],
        filters:              [],
        sort:                 []
      }
    }
const pods1 = await this.$resources.cluster.findFiltered('pod', pods1Args);

const pods2Args: find label selector type = { labelSelector: { matchLabels: { type: 'my-type' } } }
const pods2 = await this.$resources.cluster.findFiltered('pod', pods2Args);

```

shell/apis/intf/resources-api/resources-api.ts
richard-cox: one left

```
resources.cluster.find(K8S.POD, 'my-pod');
```

shell/types/store/dashboard-store.types.ts
richard-cox:
```
This is done via the native kube pagination api, not steve
```
is the lost comment

aalves08: updated

shell/apis/intf/resources-api/resources-api.ts
aalves08: updated

shell/apis/intf/resources-api/resources-api.ts
aalves08: updated

shell/apis/resources/resources-api-class.ts
```diff
@@ -0,0 +1,187 @@
...
+  async findFiltered<T = SteveListResponse>(
+    resourceType: ResourceType,
+    options: FindFilteredPageOptions | FindFilteredLabelSelectorOptions
+  ): Promise<FindFilteredPageResponse<T> | FindFilteredLabelSelectorResponse<T>> {
+    try {
+      if ('pagination' in options) { // pagination mode
+        const canPaginate = this.store.getters[`${ this.storeType }/paginationEnabled`]?.(resourceType);
+
+        if (!canPaginate) {
+          return this.surfaceError('findFiltered requests with FindFilteredPageOptions are only supported when ui-sql-cache is enabled');
```
richard-cox, line 119: by changing these from a thrown error to calling surfaceError we lose the wrapping message in the catch below (which includes `type`).

these also don't need to return, given surfaceError should always throw an error. i'm surprised there's no TS errors for it

richard-cox (approved):
There's one comment that would be good to address in the future, but lets get this one in now

PR #16910, opened 2026-03-11
Split Pod restarts column into two columns

### LINKED ISSUE #16399: Split Pod column `Restarts` into two columns
- Content contains two parts which users should be able to sort / filter on 
  - `2 (28d ago)`
- Column 1
  - `Restart Count`
- Column 2
  - `Last Restart`
- Only applicable to SSP world

Blocked on https://github.com/rancher/rancher/issues/53405

richard-cox:
### Summary
Fixes #16399

- Split Pod restarts column into two columns
  - `1 (2m47s ago)` --> `1` and `2m47s ago`

### Technical notes summary
- Makes use of new features to sort pod on subset of the pod restarts field
  - https://github.com/rancher/rancher/issues/52217
  - https://github.com/rancher/dashboard/issues/16399#issuecomment-3966876820

- View Pods list with containers that restart. Correct restarts and last restart column values
- Sort by restarts + last restart should result in correct order
- Search should not cover new columns, and not break search

- Sort/Search should not break on existing columns

### Screenshot/Video
Before
image

After
image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/config/product/explorer.js
+++ b/shell/config/product/explorer.js
@@ -34,6 +34,7 @@ import { COLUMN_BREAKPOINTS } from '@shell/types/store/type-map';
 import { STEVE_CACHE } from '@shell/store/features';
 import { configureConditionalDepaginate } from '@shell/store/type-map.utils';
 import { CATTLE_PUBLIC_ENDPOINTS, STORAGE } from '@shell/config/labels-annotations';
+import { POD_LAST_RESTART_FIELD as POD_RESTARTS_LAST_FIELD, POD_RESTART_FIELD as POD_RESTARTS_COUNT_FIELD } from '@shell/types/resources/pod';
 
 export const NAME = 'explorer';
 
@@ -419,8 +420,6 @@ export function init(store) {
     [STEVE_STATE_COL, STEVE_NAME_COL, STEVE_NAMESPACE_COL, createSteveWorkloadImageCol(6), STEVE_WORKLOAD_ENDPOINTS, 'Ready', 'Current', 'Desired', STEVE_AGE_COL],
   );
 
-  const POD_RESTART_REG_EX = /^(\d+)\s*\(([^)]+)\)/;
-
   headers(POD,
     [STATE, NAME_COL, NAMESPACE_COL, POD_IMAGES, 'Ready', 'Restarts', 'IP', NODE_COL, AGE],
     [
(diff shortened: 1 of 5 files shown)


shell/config/product/explorer.js
```diff
@@ -419,6 +419,8 @@ export function init(store) {
     [STEVE_STATE_COL, STEVE_NAME_COL, STEVE_NAMESPACE_COL, createSteveWorkloadImageCol(6), STEVE_WORKLOAD_ENDPOINTS, 'Ready', 'Current', 'Desired', STEVE_AGE_COL],
   );
 
+  const POD_RESTART_REG_EX = /^(\d+)\s*\(([^)]+)\)/;
```
codyrancher: Maybe a comment of the format you're trying to parse. I know it's in the pr description but it might help having it in the code.

shell/plugins/steve/steve-pagination-utils.ts
```diff
@@ -240,6 +240,8 @@ class StevePaginationUtils extends NamespaceProjectFilters {
     [POD]: [
       { field: 'spec.containers.image' },
       { field: 'spec.nodeName' },
+      { field: 'metadata.fields[3][0]' },
+      { field: 'metadata.fields[3][1]' },
```
codyrancher: Can we maybe do something like
```
const POD_RESTARTS_FIELD = `metadata.fields[3][0]`;
const POD_LAST_RESTART_FIELD=`metadata.fields[3][1 ]`;
```

And reference these both here and the explorer product.

By any chance is there a good way to view or see how these are being indexed in the backend? Baaed on what the front end receives these fields look wrong. At first glance I assumed we were referencing the 0th and 1st characters of the restart string.

Image

shell/config/product/explorer.js
```diff
@@ -428,7 +430,23 @@ export function init(store) {
...
         search: 'spec.containers.image'
-      }, 'Ready', 'Restarts', 'IP', {
+      },
+      'Ready',
+      {
+        name:     'pod-restart',
+        labelKey: 'tableHeaders.podRestarts',
+        search:   false,
+        sort:     ['metadata.fields[3][0]', 'metadata.fields[3][1]', 'metadata.name'],
+        value:    (row) => row.metadata.fields[3]?.match(POD_RESTART_REG_EX)?.[1] || '',
```
codyrancher: By any chance do we want to make a computed getter for the regex? As of right now we're running the regex 2 times per row.

shell/config/product/explorer.js
```diff
@@ -428,7 +430,23 @@ export function init(store) {
...
+      }, {
+        name:     'pod-last-restart',
+        labelKey: 'tableHeaders.podLastRestart',
+        value:    (row) => row.metadata.fields[3]?.match(POD_RESTART_REG_EX)?.[2] || '',
+        search:   false,
+        sort:     ['metadata.fields[3][1]', 'metadata.fields[3][0]', 'metadata.name'],
```
codyrancher: I don't know if this is your preference or not but when sorting in last restart descending order we're placing no restart above the others. Logically I would expect no restart to be below longer restart times but maybe as a part of UX we just want to disregard the restarts. 

I would assume this would require a backend change if we cared.

Image

richard-cox:
@codyrancher Good suggestions! I've made changes following them all. Some crossed over though so haven't updated each comment.

Changes can be found in last commit.

For the sorting one that makes sense. I couldn't fix browser side though so raised a r/r issue https://github.com/rancher/rancher/issues/54531

shell/utils/pagination-utils.ts
```diff
@@ -299,7 +299,7 @@ class PaginationUtils {
     return isEqual(aPrimitiveTypes, bPrimitiveTypes) &&
       this.paginationFiltersEqual(aFilter, bFilter) &&
       this.paginationFiltersEqual(aPN, bPN) &&
-      sameArrayObjects<PaginationSort>(aSort, bSort, true);
+      sameArrayObjects<PaginationSort>(aSort, bSort, false);
```
richard-cox, line 302: This is used to determine if pagination has changed, if so we kick off a http request.

When flicking between the two columns with the same entries but different order it did not trigger

sort:     [POD_RESTARTS_COUNT_FIELD, POD_RESTARTS_LAST_FIELD, 'metadata.name'],
sort:     [POD_RESTARTS_LAST_FIELD, POD_RESTARTS_COUNT_FIELD, 'metadata.name'],

the sort order is important, so ensured we set the right arg to take array order into account

Issue #16924, opened 2026-03-13
Filter box does not work for Machines and MachineSets in Cluster Provisioning when ui-sql-cache is enabled


khushalchandak17:
**Setup**
- Rancher version: `v2.13.1`
- Rancher UI Extensions: `None`
- Browser type & version: FireFox

**Describe the bug**
Filtering does not work for Cluster API `Machines` and `MachineSets` in:

`Local Cluster -> More Resources -> Cluster Provisioning -> Machines`
`Local Cluster -> More Resources -> Cluster Provisioning -> MachineSets`

When text is entered in the Filter box, the list does not filter by name/pattern.

This appears to happen when `ui-sql-cache` is enabled. In that mode, the UI performs server-side filtering and generates filter expressions that include fields not supported by the SQL cache layer for these resources, which can cause the backend request to fail and makes the UI filter appear non-functional.

**To Reproduce**
1. Deploy Rancher `v2.13.1`.
2. Deploy a Rancher-managed RKE2 cluster with multiple node pools so that multiple `MachineSets` are created.
3. Navigate to:
   - `Local Cluster -> More Resources -> Cluster Provisioning -> MachineSets`
   - or `Local Cluster -> More Resources -> Cluster Provisioning -> Machines`
4. Use the Filter box to search by name or pattern.
5. Observe that the list does not filter as expected.

Optional comparison:
6. Disable `ui-sql-cache`.
7. Repeat the same filter action.
8. Observe that filtering works in that mode.

**Result**
- The Filter box does not filter `MachineSets` or `Machines` by name/pattern.
- With `ui-sql-cache` enabled, the backing API request can fail with `HTTP 500`.
- No UI workaround is available.
- `kubectl` filtering is possible, but that is not a UI workaround.

**Expected Result**
The Filter box should filter `MachineSets` and `Machines` by name/pattern without backend errors.

**Screenshots**
N/A

**Additional context**
- Business impact: Low. Users can still query/filter via `kubectl`, but the UI behavior is a defect and customers reasonably expect filtering to work.
- Troubleshooting tried: No specific remediation beyond restart; issue appears UI-related.
- Scope: This appears to affect `Machines` and `MachineSets` specifically.
Investigation notes:

When `ui-sql-cache` is enabled, the UI generates a server-side filter request similar to:

```text
[...]

PR #16925, opened 2026-03-13
Fix Machine and MachineSet filtering with ui-sql-cache

khushalchandak17:
### Summary
Fixes: https://github.com/rancher/dashboard/issues/16924

Fix filtering for Cluster API `Machine` and `MachineSet` resources under `Local Cluster -> More Resources -> Cluster Provisioning` when `ui-sql-cache` is enabled.

- Added explicit explorer headers for `CAPI.MACHINE` and `CAPI.MACHINE_SET`.

### Technical notes summary
- When `ui-sql-cache` is enabled, explorer filtering uses server-side pagination/search.
- `Machine` and `MachineSet` did not previously define explicit pagination headers in explorer, so the UI generated a broader filter query from default columns/model values.
- For Cluster API resources, that query could include unsupported fields such as:
  - `$.status.nodeRef.name`
  - `$.spec.providerID`
  - `$.spec.version`
  - `$.status.phase`
- This change reuses the Steve pagination headers so search is constrained to stable metadata fields:
  - `metadata.state.name`
  - `metadata.name`
  - `metadata.namespace`
- `metadata.creationTimestamp` remains available for sorting through the age column, but it is not used as a search field.

- Browser used for local validation: FireFox.
- Validate with `ui-sql-cache` enabled.
- In `Local Cluster -> More Resources -> Cluster Provisioning -> Machines`, search by a unique Machine name fragment and confirm:
  - the list filters correctly
  - no error banner is shown
  - the backing `GET /v1/cluster.x-k8s.io.machines?...` request returns `200`
  - the generated filter does not include unsupported fields such as `$.status.nodeRef.name`, `$.spec.providerID`, `$.spec.version`, or `$.status.phase`
- Repeat the same validation for `MachineSets` using `GET /v1/cluster.x-k8s.io.machinesets?...`.

- Explorer list filtering for `Machine` and `MachineSet` resources.

### Screenshot/Video
- N/A for this change.

### Checklist
- [ ] The PR has a Milestone 


shell/config/product/explorer.js
```diff
@@ -497,6 +497,9 @@ export function init(store) {
     ]
   );
 
+  headers(CAPI.MACHINE, [STATE, NAME_COL, NAMESPACE_COL, AGE], [STEVE_STATE_COL, STEVE_NAME_COL, STEVE_NAMESPACE_COL, STEVE_AGE_COL]);
```
richard-cox, line 500: These remove all the useful columns from both tables.

They're automatically generated given information from schema/CRD columns.

It looks like those columns access paths that aren't indexed, and only indexed fields can be sorted/searched

We should
1. Make a request to get them indexed
2. To avoid this scenario in the future by updating the process to create columns from schema fields to check for valid paths. if not found we disable sort/search on them.

The short term fix would be to manually do step 2 for these tables (define the columns here but disable sort/search on them)

richard-cox: I've looked into this, and i think the core issue has been resolved. column paths should now point to metadata.field which are indexed. This is fixed in 2.14.0 and might have a backport. Just trying to find out details

richard-cox: Fix was implemented as part of two issues - https://github.com/rancher/rancher/issues/52121 and https://github.com/rancher/rancher/issues/52774

khushalchandak17: Thanks @richard-cox for looking into this and sharing the background. I checked it on r/r v2.14.0 and can confirm the issue looks fixed there already.
Since thats the case, this PR isnt needed anymore, so I will go ahead and close it.

PR #16932, opened 2026-03-13
Updating the asset imports to make them more vite friendly.

codyrancher:
### Summary
The next steps to slowly move us toward a world in which we can switch our build system to vite without causing any interruptions. We're also taking this a step at a time to reduce risk and review burden. The original pr was approximately 120 files of changes.

### Technical notes summary
 It will allow us to make one change to make all of these areas compatible with vite once we're ready. Vite has a different mechanism for dynamic imports.

**You'll notice two stylistic import changes:**
1.  We switch to an `import`
    - These are static paths that are known at compile time and therefor don't need a dynamic import (`require`)
        - There could also be cases where we want to delay the import to runtime but I don't think that's true for the cases here.
2. We switch from `require` to `requireAsset`
    - These are determined at runtime and need a dynamic import. 

Being exhaustive here doesn't seem necessary. I think we have some categories of changes and I'll list them here so we can test each one in order to get appropriate coverage.

1. **Brand switching** -  Go to **Global Settings > ui-brand**, set it to `suse`, save. Verify:
   - Header logo updates (burger menu + top nav)
   - Favicon changes in browser tab
   - Login page shows SUSE logo and background
   - Toggle dark/light theme - logos swap to correct variant
   - Reset `ui-brand` to - Rancher logo returns at full size
2. **Custom logo upload** - In **Global Settings**, set `ui-logo-light` and `ui-logo-dark` to custom image URLs. Verify the custom logo appears in header, and clears properly when removed.
3. **Cluster creation provider selection** - Click **Create** on the clusters page. Verify:
   - Provider cards show correct icons (RKE2, cloud providers)
   - Generic driver icon appears for enabled **Cluster Drivers**

See above

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/utils/require-asset.ts
+++ b/shell/utils/require-asset.ts
@@ -27,23 +27,18 @@ interface WebpackRequireContext {
   id: string;
 }
 
-declare global {
-  // eslint-disable-next-line no-unused-vars
-  interface NodeRequire {
-    context(directory: string, useSubdirectories?: boolean, regExp?: RegExp): WebpackRequireContext;
-  }
-}
-
 // --- Webpack: require.context (compile-time transform) ---
 
 let imgCtx: WebpackRequireContext | null = null;
 let jsonCtx: WebpackRequireContext | null = null;
 
 try {
+  // @ts-expect-error — require.context is a webpack compile-time transform, not visible to TypeScript
   imgCtx = require.context('@shell/assets', true, /\.(svg|png|jpe?g|gif|ico|webp)$/);
 } catch (e) {}
 
 try {
+  // @ts-expect-error — require.context is a webpack compile-time transform, not visible to TypeScript
   jsonCtx = require.context('@shell/assets', true, /\.json$/);
 } catch (e) {}
 


shell/components/LazyImage.vue
```diff
@@ -1,5 +1,6 @@
 <script>
 import { BLANK_IMAGE } from '@shell/utils/style';
+import genericCatalogSvg from '@shell/assets/images/generic-catalog.svg';
```
codyrancher, line 3: So the reason we switch these to import vs requireAsset is because the asset paths were static/determined at compile time so we don't need the dynamic import.

There's more examples in this pr.

shell/utils/require-asset.ts
```diff
@@ -0,0 +1,88 @@
...
+ *
+ *   const imgCtx = import.meta.glob(
+ *     '@shell/assets/**\/*.{svg,png,jpg,jpeg,gif,ico,webp}',
+ *     { eager: true, query: '?url', import: 'default' }
+ *   );
+ *
+ *   const jsonCtx = import.meta.glob(
+ *     '@shell/assets/**\/*.json',
+ *     { eager: true }
+ *   );
```
codyrancher, line 17: I think instead of replacing the code we'll need to switch based on webpack vs vite usage so we can still allow extensions to build using new versions of shell without migrating to vite. I'll tackle that when we're closer to switching to vite.

codyrancher:
> This looks good based on my initial pass. I do wonder if we need to address our `require.resolve()` and `require.context()` usage in preparation for the vite migration as well:
> 
>     * https://github.com/rancher/dashboard/blob/88e6e7f581e949c22aac4ba364079df16f5cfd82/shell/utils/dynamic-importer.js#L116-L142

You're right, it is something that needs to be addressed before we can switch to Vite. In the original unrefined PR it was touched here https://github.com/rancher/dashboard/pull/16710/changes#diff-d425bcbe2269e54c35340d01c97455a7ecdd1fa67caa6dfd2c5f9864f5a432d4.

If you'd like, I can add it to this PR as it fits conceptually. I'm just trying to keep the scope narrow for each of these smaller PRs.

PR #16937, opened 2026-03-14
Updated Container Shell to handle Safari better and upgraded xterm

### LINKED ISSUE #17273: Kubectl shell is not working in Safari with WebGL disabled
**Setup**
- Rancher version: 2.15
- Rancher UI Extensions: None
- Browser type & version: Safari

**Describe the bug**

When WebGL is disabled in Safari browser, commands cannot be run in kubectl shell

**To Reproduce**

Disable WebGL in the browser
Any cluster -> kubectl shell

**Result**
Shell shows as connected but it is empty and no command can be run

**Expected Result**

Shell should connect successfully and user should be able to run commands

**Screenshots**

**Additional context**
Came from https://github.com/rancher/dashboard/issues/16497#issuecomment-3895940522
Related to https://github.com/rancher/dashboard/issues/16371

eva-vashkevich:
### Summary
Fixes #17273

There seem to be some issues around WebGPU and webGL in Safari. With WebGPU disabled, we were getting into a state where all plugins were loaded successfully, but we couldn't print to the terminal, failing silently. To avoid that, we not avoid using webGL for Safari and rely on DOM rendering

Upgraded xterm library
Removed canvas fallback. Now the DOM rendering is a fallback, which removes an unnecessary dependency.
If Safari browser is detected, we fallback to avoid thinking that webgl was loaded successfully when it didn't.

Test that we are able to use shell console in all browsers.
Check that it loads successfully, command can be run sucessfully.
There should be no errors in the browser
 
We should validate that all browsers work inside a VM

### Checklist
- [x] The PR has a Milestone 


shell/package.json
rak-phillip, line 1: Ensure that the lockfile under the shell directory also gets updated.

shell/components/Window/ContainerShell.vue
```diff
@@ -243,13 +241,25 @@ export default {
...
+        if (isSafari) {
+          throw new Error('Safari WebGL support is unstable');
+        }
         this.webglAddon = new addons.webgl.WebglAddon();
         terminal.loadAddon(this.webglAddon);
       } catch (e) {
         // Some browsers (Safari) don't support the webgl renderer, so don't use it.
         this.webglAddon = null;
-        this.canvasAddon = new addons.canvas.CanvasAddon();
-        terminal.loadAddon(this.canvasAddon);
```
rak-phillip, line 252: Do we want to remove the canvas tier for situations where we reach this catch clause in browsers that are not Safari? The docs for `@xterm/addon-canvas` state: 

> The purpose of this addon is to be used as a fallback for the [webgl addon](https://www.npmjs.com/package/@xterm/addon-webgl) when better performance is desired over the default DOM renderer, but WebGL2 isn't supported or performant for some reason.

Now that the canvas renderer is gone entirely, all browsers now now silently drop to the slowest renderer if WebGL legitimately fails..

eva-vashkevich:
@a110605 Opened https://github.com/harvester/harvester/issues/10691 to update harvester-ui-extension

a110605:
> @a110605 Opened [harvester/harvester#10691](https://github.com/harvester/harvester/issues/10691) to update harvester-ui-extension

@eva-vashkevich , thanks for reminding. 

I will update harvester ui extension xterm after bumping the shell including this PR change.

rak-phillip (commented):
I think we need a rebase to address e2e test failures.

rak-phillip (approved):
LGTM

PR #16960, opened 2026-03-18
Fix cross-version $extension/$plugin compatibility for extension formatters

### LINKED ISSUE #16952: Fix $extension usage in older rancher systems
**Setup**
- Rancher version: `2.12.7`
- Rancher UI Extensions: `Kubewarden 4.1.0`
- Browser type & version: n/a

**Describe the bug**

Root cause for https://github.com/rancher/kubewarden-ui/issues/1444 is that `SortableTable` uses `$extension`:

```
const pluginFormatter = this.$extension?.getDynamic('formatters', c.formatter);
```

but there's no `this.$extension` in `2.12.7`

**To Reproduce**

Follow repro steps in https://github.com/rancher/kubewarden-ui/issues/1444

**Result**
Some table cols are not displayed (ones using a formatter)

**Expected Result**

All table cols should display fine

**Screenshots**

**Additional context**

aalves08:
### Summary
Fixes #16952

- Add shim to make sure older Rancher systems which still use `$plugin` will work also with `$extension`

### Technical notes summary

**Current status of problem on extensions**

This was the PR that brought in the problem: https://github.com/rancher/dashboard/pull/15894 -> milestone `2.14.0`
Shell versions with compatibility problem: `3.0.8` and `3.0.9`
Rancher systems where using Shell `3.0.8` and `3.0.9` we might see problems: `2.10`, `2.11`, `2.12` and `2.13`

**Problem**

Rancher's shell renamed the global Vue property that exposes the extension manager from `$plugin` (older versions) to `$extension` (newer versions). Shell components such as `SortableTable` were updated to look up extension-registered formatters via `this.$extension.getDynamic('formatters', ...)`. However, older Rancher hosts only inject `$plugin` onto `vueApp.config.globalProperties` — leaving `this.$extension` as `undefined` — so any formatter registered by an extension is silently invisible to the table.

The result: table columns that use custom extension formatters render as blank/raw values on affected Rancher versions, with no error surfaced to the user.

**Root cause**

The extension manager is injected into `vueApp.config.globalProperties` by Rancher's `inject()` mechanism inside `installInjectedPlugins`. 
Depending on the Rancher version, it is injected as either `$plugin `or `$extension`, but not necessarily both. The context object passed to an extension's init function carries a reference to the `Nuxt` app (not the Vue 3 app), so `context.app.config.globalProperties` is undefined — making it impossible for an extension to self-patch the missing property using only what Rancher passes in.

**Solution**

The fix is injected into the generated `importTypes()` function produced by `auto-import.js`, which every extension calls unconditionally at init time. Before registering any types, the generated function now runs a small compatibility shim:

```
var __vueApp = document.getElementById('app') && document.getElementById('app').__vue_app__;
if (__vueApp && __vueApp.config && __vueApp.config.globalProperties) {
  var __gp = __vueApp.config.globalProperties;
  if (!__gp.$extension && __gp.$plugin) { __gp.$extension = __gp.$plugin; }
  else if (!__gp.$plugin && __gp.$extension) { __gp.$plugin = __gp.$extension; }
}
```

Vue 3 always sets `__vue_app__` on the mounted root DOM element when app.mount('#app') is called (confirmed in `shell/initialize/entry-helpers.js`). This gives reliable, version-agnostic access to the real `vueApp.config.globalProperties` — exactly where the inject mechanism places `$plugin` or `$extension`.

- Provision Rancher `2.12.7`
- Get kubewarden repo https://github.com/rancher/kubewarden-ui , branch `main` to your local machine
- yarn link `@rancher/shell` from this PR into Kubewarden extension locally
- do a local build of the extension `yarn build-pkg kubewarden`
- serve the extension locally `yarn serve-pkgs`
- do a developer load of this built asset
- Go to Kubewarden UI  - `Policy Servers` - make sure all table cols appear fine
- Go to Kubewarden UI  - `Cluster Admission Policies` - make sure all table cols appear fine

### Screenshot/Video

Screenshot 2026-03-18 at 13 30 43
Screenshot 2026-03-18 at 13 30 46
Screenshot 2026-03-18 at 13 31 22

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/pkg/auto-import.js
+++ b/shell/pkg/auto-import.js
@@ -16,11 +16,35 @@ function replaceAll(str, find, replace) {
 // Vue globalProperties, regardless of which one the host injected. This makes all
 // extensions compatible across Rancher versions without any per-extension code changes.
 const COMPAT_SHIM = `  if (typeof document !== 'undefined') {
-    var __vueApp = document.getElementById('app') && document.getElementById('app').__vue_app__;
-    if (__vueApp && __vueApp.config && __vueApp.config.globalProperties) {
-      var __gp = __vueApp.config.globalProperties;
-      if (!__gp.$extension && __gp.$plugin) { __gp.$extension = __gp.$plugin; }
-      else if (!__gp.$plugin && __gp.$extension) { __gp.$plugin = __gp.$extension; }
+    var patchGlobalProps = function() {
+      var __vueApp = document.getElementById('app').__vue_app__;
+
+      if (!__vueApp) {
+        // no __vue_app__, vueApp.mount('#app') has not been called yet
+        return false;
+      }
+
+      if (__vueApp.config && __vueApp.config.globalProperties) {
+        var __gp = __vueApp.config.globalProperties;
+        if (!__gp.$extension && __gp.$plugin) { __gp.$extension = __gp.$plugin; }
+        else if (!__gp.$plugin && __gp.$extension) { __gp.$plugin = __gp.$extension; }
+        return true;
+      }
+
+      // Fallback to failure case
+      return false;
+    };
+
+    if (!patchGlobalProps()) {
+      // Could not patch, keep retrying until it works
+      var __retry = setInterval(function() {
+        if (patchGlobalProps()) {
+          clearInterval(__retry);
+        }
+      }, 100);
+
+      // Fallback: clear interval after 10 seconds just in case
+      setTimeout(function() { clearInterval(__retry); }, 10000);
     }
   }\n`;
 


aalves08:
CC @kravciak

shell/pkg/auto-import.js
```diff
@@ -79,6 +94,8 @@ function generateDynamicTypeImport(pkg, dir) {
   const template = fs.readFileSync(path.join(__dirname, 'import.js'), { encoding: 'utf8' });
   let content = 'export function importTypes($extension) { \n';
 
+  content += COMPAT_SHIM;
```
richard-cox, line 121: Are you sure this is needed? From what i can tell it only applies to built-in plugins which shouldn't have the issue.

shell/pkg/auto-import.js
```diff
@@ -11,6 +11,19 @@ function replaceAll(str, find, replace) {
   return str.split(find).join(replace);
 }
 
+// Injected at the top of every generated importTypes() function.
+// Ensures both $extension (newer Rancher) and $plugin (older Rancher) are available on
+// Vue globalProperties, regardless of which one the host injected. This makes all
+// extensions compatible across Rancher versions without any per-extension code changes.
+const COMPAT_SHIM = `  if (typeof document !== 'undefined') {
+    var __vueApp = document.getElementById('app') && document.getElementById('app').__vue_app__;
+    if (__vueApp && __vueApp.config && __vueApp.config.globalProperties) {
```
richard-cox: This definitely works? When i stepped through it for built-in pkgs (in theory that doesn't make a difference)  __vueApp is always undefined

richard-cox: I also didn't see this running after dev loading.

- running rancher 2.12.6 (no image for 2.12.7)
- clone kube warden ui repo, on `main` branch
- yarn link shell from this repo
- build pkg and server
- in 2.12.6 dev load extension
- refresh
- webpack://kubewarden-4.1.0/node_modules/@rancher/auto-import `__vueApp` is undefined

richard-cox: image

aalves08: DM'd videos showing it working on my machine @richard-cox 🙏

shell/pkg/auto-import.js
aalves08: when this solution didn't work for you, did you remove this line?

aalves08:
@richard-cox works in Chrome, Brave, Firefox, Edge (couldn't bypass localhost restrictions in Safari 🤷 )....

**Brave**
Screenshot 2026-03-19 at 15 16 00
Screenshot 2026-03-19 at 15 16 07

**Edge**
Screenshot 2026-03-19 at 15 17 06
Screenshot 2026-03-19 at 15 17 10

**Firefox**
Screenshot 2026-03-19 at 15 18 16
Screenshot 2026-03-19 at 15 18 18

I would say it looks good! Despite you doing the last commit, I don't think I can approve my own PR

shell/pkg/auto-import.js
richard-cox: resolved via https://github.com/rancher/dashboard/pull/16960/commits/fe96f024cfd0e1e769db41ba246a8c58b68044e9

cnotv:
Sorry to be late to the party, it works also for me.

Screenshot 2026-03-20 at 12 16 50

Screenshot 2026-03-20 at 12 16 30

PR #16962, opened 2026-03-18
Automation - Prime replicaset test fixes

izaac:
### Summary
Fixes rancher/qa-tasks#2266

Stabilizes the ReplicaSets Cypress test that validates rollback is unavailable by removing a flaky timing race caused by modal overlay transitions.

- Updated the ReplicaSets e2e spec to use page object waits before interacting with row actions.
- Added and used a list-page helper to wait until modal overlays are gone.
- Added and used a list readiness helper that waits for table visibility and loading completion.

E2E

E2E

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/cypress/e2e/po/pages/explorer/workloads-replicasets.po.ts
+++ b/cypress/e2e/po/pages/explorer/workloads-replicasets.po.ts
@@ -14,6 +14,11 @@ export class WorkloadsReplicasetsListPagePo extends BaseListPagePo {
   constructor(clusterId = 'local') {
     super(WorkloadsReplicasetsListPagePo.createPath(clusterId));
   }
+
+  waitForListReady(): void {
+    this.list().resourceTable().sortableTable().checkVisible();
+    this.list().resourceTable().sortableTable().checkLoadingIndicatorNotVisible();
+  }
 }
 
 export class WorkloadsReplicasetsEditPagePo extends BaseDetailPagePo {
--- a/cypress/e2e/po/prompts/ResourceSearchDialog.po.ts
+++ b/cypress/e2e/po/prompts/ResourceSearchDialog.po.ts
@@ -18,6 +18,10 @@ export default class ResourceSearchDialog extends ComponentPo {
     super('[data-testid="search-modal"]');
   }
 
+  waitForNoDialog(): Cypress.Chainable {
+    return this.self().should('not.exist');
+  }
+
   open() {
     cy.keyboardControls(CMD_K_KEY, 1);
   }
--- a/cypress/e2e/tests/pages/explorer2/workloads/replicasets.spec.ts
+++ b/cypress/e2e/tests/pages/explorer2/workloads/replicasets.spec.ts
@@ -13,6 +13,7 @@ describe('Cluster Explorer', { tags: ['@explorer2', '@adminUser'] }, () => {
       it('should not be able to rollback a replicaset', () => {
         // list view for replicasets
         const workloadsReplicasetsListPage = new WorkloadsReplicasetsListPagePo('local');
+        const resourceSearchDialog = new ResourceSearchDialog();
 
         workloadsReplicasetsListPage.goTo();
         workloadsReplicasetsListPage.waitForPage();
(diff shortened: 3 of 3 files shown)


yonasberhe23 (changes requested):
I dug into this and the issue is that we’re already on the ReplicaSets list after creating the resource (the UI navigates there automatically). So when we call ResourceSearchDialog.goToResource('ReplicaSets') and click the ReplicaSets link, we’re “navigating” to the same page, which has no effect and the modal never closes.

Fix would be to either:
- Remove the navigation – Drop the ResourceSearchDialog.goToResource('ReplicaSets') call and rely on the automatic redirect after create. This is the simpler fix (one liner) and removes the redundant step.
- Actually use the search – Navigate away from the ReplicaSets list first (e.g. to Cluster Dashboard or another resource), then use ResourceSearchDialog.goToResource('ReplicaSets') to come back. That way the modal is used for real navigation and should close reliably.

I’d lean toward the first one unless the goal was explicitly to assert the search -> ReplicaSets flow.

izaac:
@yonasberhe23 

That sounds good, I'll go with option 1. Dropping the `ResourceSearchDialog.goToResource('ReplicaSets')` call since the redirect is already handled automatically. I also added `waitForListReady()` to wait for the loading indicator to clear before asserting, does that seem valuable to keep, or is `waitForPage()` sufficient here?

cypress/e2e/po/pages/explorer/workloads-replicasets.po.ts
```diff
@@ -14,6 +14,16 @@ export class WorkloadsReplicasetsListPagePo extends BaseListPagePo {
   constructor(clusterId = 'local') {
     super(WorkloadsReplicasetsListPagePo.createPath(clusterId));
   }
+
+  waitForNoModalOverlay(): Cypress.Chainable {
```
IsaSih: I see that this function is going to be useful for any pages that have this racing condition regarding the modal. I'd like to request you to move this helper function to `@/cypress/e2e/po/pages/page.po`

IsaSih (changes requested):

cypress/e2e/po/pages/explorer/workloads-replicasets.po.ts
yonasberhe23: @IsaSih I agree that this should be moved so that it can be reused in other tests but I think it make more sense to move it `cypress/e2e/po/prompts/ResourceSearchDialog.po.ts` as it pertains to the resource search dialog.

To answer your question @izaac , I think it's valuable to keep but I think it can be simplified to the below. That will be sufficient since in the test itself we are waiting for the resource to be visible in the list https://github.com/izaac/dashboard/blob/787084681fe083cb85c0d5485fe739051c304452/cypress/e2e/tests/pages/explorer2/workloads/replicasets.spec.ts#L35
When we move it to `cypress/e2e/po/prompts/ResourceSearchDialog.po.ts`, it would just be:

```
 waitForNoDialog(): Cypress.Chainable {
    return this.self().should('not.exist');
  }
```

yonasberhe23:
> @yonasberhe23
> 
> That sounds good, I'll go with option 1. Dropping the `ResourceSearchDialog.goToResource('ReplicaSets')` call since the redirect is already handled automatically. I also added `waitForListReady()` to wait for the loading indicator to clear before asserting, does that seem valuable to keep, or is `waitForPage()` sufficient here?

@izaac see comment https://github.com/rancher/dashboard/pull/16962/changes#r2961640442

cypress/e2e/po/pages/explorer/workloads-replicasets.po.ts
yonasberhe23: What to you guys think?

izaac: I've implemented it, the move to `ResourceSearchDialog` makes sense as it's better scoped. But the idea of re-usability was on point.

yonasberhe23: @izaac I retract my statement about the simplified version being sufficient because the assertion that checks if the dialog exists happens on the replicaset create page, so what you have currently LGTM.

PR #16963, opened 2026-03-18
Fixes for ClusterRepo resources

### LINKED ISSUE #16852: Units mismatch for ClusterRepo refreshInterval
**Setup**
- Rancher version: 2.14 (and maybe older)

**Describe the bug**
Editing a `ClusterRepo` object (from the `catalog.cattle.io/v1` API group) through the UI shows "hours" units in the `Refresh Interval` input box, which then gets directly used as is in the `PUT` request to update the resource. However, the correct unit for this field is Seconds.

**To Reproduce**
Go to any Cluster's explorer -> Apps -> Repositories -> click any, e.g. "Rancher" (`rancher-charts`) -> "⋮" -> Edit Config -> (open Browser Dev Tools to inspect the network call) -> input "2" in the "Refresh interval" field and click Save.

**Result**
Inspect the request payload, and observe how ".spec.refreshInterval" is set to a value of "2" (not the equivalent in seconds for 2 hours).

**Expected Result**
Either the UI is changed to reflect the correct units, or it does the appropriate conversion to seconds.

**Screenshots**
Image
Image

**Additional context**
 - [Rancher code where the integer is interpreted as seconds](https://github.com/rancher/rancher/blob/main/pkg/controllers/dashboard/helm/repo.go#L118)

aruiz14:
### Summary

Fixes #16852

Misc fixes/improvements discovered while working with ClusterRepos.

- Fixes "Refresh Interval" units.
- Fixes default value for git-based resources (1 instead of 6 hours).
- Adds a new human-readable "Refresh Interval" to the details page.
- Adds support for disabling periodic updates (rancher/rancher#54124).

Displaying/editing ClusteRepos, for which I've added unit/e2e tests.

Editing ClusterRepo refreshInterval should really honour the value being submitted.

### Screenshot/Video

<details>
<summary>Details</summary>

-----

Default input box, shows correct unix and default value:

Screenshot 2026-03-17 at 16 18 02

-----

New tooltip:
Screenshot 2026-03-17 at 16 17 56

-----

New data in details page (blank value):
Screenshot 2026-03-18 at 17 14 07

-----

Shows disabled if `-1`:
Screenshot 2026-03-18 at 17 14 22

-----

</details>

### Checklist
- [X] The PR has a Milestone 
- [X] The PR has considered, and if applicable tested with, the three Global Roles `Admin`, `Standard User` and `User Base`

### DIFF
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -1441,7 +1441,6 @@ catalog:
     refreshInterval:
       label: Refresh Interval
       placeholder: 'default: {value}'
-      tooltip: Using any negative value disables periodic updates
     error:
       refresh: Error refreshing repository
   tools:
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -1440,7 +1440,7 @@ catalog:
           placeholder: 'default: 5'
     refreshInterval:
       label: Refresh Interval
-      placeholder: 'default: {value} {units}'
+      placeholder: 'default: {value}'
       tooltip: Using any negative value disables periodic updates
     error:
       refresh: Error refreshing repository
--- a/shell/assets/translations/en-us.yaml
+++ b/shell/assets/translations/en-us.yaml
@@ -1440,7 +1440,8 @@ catalog:
           placeholder: 'default: 5'
     refreshInterval:
       label: Refresh Interval
-      placeholder: 'default: {hours}'
+      placeholder: 'default: {value} {units}'
+      tooltip: Using any negative value disables periodic updates
     error:
       refresh: Error refreshing repository
   tools:
--- a/shell/models/catalog.cattle.io.clusterrepo.js
+++ b/shell/models/catalog.cattle.io.clusterrepo.js
@@ -1,6 +1,7 @@
 import { parse } from '@shell/utils/url';
 import { CATALOG } from '@shell/config/labels-annotations';
 import { insertAt } from '@shell/utils/array';
+import { formatDuration } from '@shell/utils/time';
 import { CLUSTER_REPO_APPCO_AUTH_GENERATE_NAME, CATALOG as CATALOG_TYPE } from '@shell/config/types';
 import { colorForState, stateDisplay } from '@shell/plugins/dashboard-store/resource-class';
 import { _CREATE } from '@shell/config/query-params';
@@ -218,6 +219,24 @@ export default class ClusterRepo extends SteveModel {
     return this.spec?.gitBranch || '(default)';
   }
 
+  get defaultRefreshIntervalHours() {
+    return this.isOciType ? 24 : 1;
+  }
+
+  get defaultRefreshInterval() {
+    return 60 * 60 * this.defaultRefreshIntervalHours;
+  }
+
+  get refreshInterval() {
+    const value = this.spec?.refreshInterval || this.defaultRefreshInterval;
+
+    if (value === -1) {
+      return 'Disabled';
+    }
+
+    return formatDuration(value);
+  }
+
   get details() {
     return [
       {
@@ -229,6 +248,10 @@ export default class ClusterRepo extends SteveModel {
         content:       this.status.downloadTime,
         formatter:     'LiveDate',
         formatterOpts: { addSuffix: true },
+      },
+      {
+        label:   'Refresh Interval',
+        content: this.refreshInterval,
       },
     ];
   }
--- a/shell/models/catalog.cattle.io.clusterrepo.js
+++ b/shell/models/catalog.cattle.io.clusterrepo.js
@@ -231,7 +231,7 @@ export default class ClusterRepo extends SteveModel {
     const value = this.spec?.refreshInterval || this.defaultRefreshInterval;
 
     if (value === -1) {
-      return 'Disabled';
+      return this.t('generic.disabled');
     }
 
     return formatDuration(value);
@@ -250,7 +250,7 @@ export default class ClusterRepo extends SteveModel {
         formatterOpts: { addSuffix: true },
       },
       {
-        label:   'Refresh Interval',
+        label:   this.t('catalog.repo.refreshInterval.label'),
         content: this.refreshInterval,
       },
     ];
--- a/cypress/e2e/po/edit/chart-repositories.po.ts
+++ b/cypress/e2e/po/edit/chart-repositories.po.ts
@@ -1,6 +1,7 @@
 import PagePo from '@/cypress/e2e/po/pages/page.po';
 import LabeledInputPo from '@/cypress/e2e/po/components/labeled-input.po';
 import UnitInputPo from '@/cypress/e2e/po/components/unit-input.po';
+import CheckboxInputPo from '@/cypress/e2e/po/components/checkbox-input.po';
 import AsyncButtonPo from '@/cypress/e2e/po/components/async-button.po';
 import LabeledSelectPo from '@/cypress/e2e/po/components/labeled-select.po';
 import NameNsDescription from '@/cypress/e2e/po/components/name-ns-description.po';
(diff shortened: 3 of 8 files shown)


aruiz14:
Merging this PR must be synced after:
- https://github.com/rancher/rancher/pull/54160

shell/models/catalog.cattle.io.clusterrepo.js
```diff
@@ -181,6 +181,24 @@ export default class ClusterRepo extends SteveModel {
     return this.spec?.gitBranch || '(default)';
   }
 
+  get defaultRefreshIntervalHours() {
+    return this.isOciType ? 24 : 1;
+  }
```
aruiz14, line 224: I made this to use it from `shell/edit/catalog.cattle.io.clusterrepo.vue`, as it was wrong there

shell/models/catalog.cattle.io.clusterrepo.js
```diff
@@ -193,6 +211,10 @@ export default class ClusterRepo extends SteveModel {
         formatter:     'LiveDate',
         formatterOpts: { addSuffix: true },
       },
+      {
+        label:   'Refresh Interval',
```
richard-cox:
```suggestion
        label:  this.t('catalog.repo.refreshInterval.label'),
```

The others should be translated as well, but it's oooold tech debt. Lets get this one correct though

shell/models/catalog.cattle.io.clusterrepo.js
```diff
@@ -181,6 +181,24 @@ export default class ClusterRepo extends SteveModel {
...
+
+  get defaultRefreshInterval() {
+    return 60 * 60 * this.defaultRefreshIntervalHours;
+  }
+
+  get refreshInterval() {
+    const value = this.spec?.refreshInterval || this.defaultRefreshInterval;
+
+    if (value === -1) {
+      return 'Disabled';
```
richard-cox:
```suggestion
      return this.t('generic.disabled');
```

shell/assets/translations/en-us.yaml
```diff
@@ -1438,7 +1438,8 @@ catalog:
           placeholder: 'default: 5'
     refreshInterval:
       label: Refresh Interval
-      placeholder: 'default: {hours}'
+      placeholder: 'default: {value} {units}'
+      tooltip: Using any negative value disables periodic updates
```
```suggestion
      tooltip: Using any negative value disables Refresh
```

whilst less grammatically correct, it's indicating that the refresh feature will be disabled

richard-cox (changes requested):
- In terms of disabling the feature by setting the refresh interval to a minus figure... ideally we should show a checkbox that enables / disables refresh (under the hood applying -1 or undefined), then only accept positive numbers in the input field.
- In terms of changing the unit of time from hours to seconds for the refresh interval... ideally we would provide a better component to select a duration. It feels like hours would be more helpful to users and can see them being frustrated at having to leave the browser to determine how many seconds are in 8 hours. However we would then become unstuck when editing a resource that has been set to seconds outside of the UI. 

@oboc-sts / @edenhernandez-suse Some of the reward / effort here is slim. will leave the final decision with UX, do we leave the PR as is for 2.150, do one of above, or both of above?

oboc-sts:
> * In terms of disabling the feature by setting the refresh interval to a minus figure... ideally we should show a checkbox that enables / disables refresh (under the hood applying -1 or undefined), then only accept positive numbers in the input field.
> * In terms of changing the unit of time from hours to seconds for the refresh interval... ideally we would provide a better component to select a duration. It feels like hours would be more helpful to users and can see them being frustrated at having to leave the browser to determine how many seconds are in 8 hours. However we would then become unstuck when editing a resource that has been set to seconds outside of the UI.
> 
> @oboc-sts / @edenhernandez-suse Some of the reward / effort here is slim. will leave the final decision with UX, do we leave the PR as is for 2.150, do one of above, or both of above?

Both please.

But, from a UX perspective, the enabling / disabling this has a much higher prio (we can't ask the user to do this kind of gymnastic = clicking on tooltips, giving negative values, etc.) - we need a straight forward way of enabling the feature and the checkbox seems like a reasonable way to achieve this. Or simply leaving the field empty...

The second issue implies having a combo (input + dropdown) if we can't find a proper unit of measurement that would satisfy the vast majority of the users. @edenhernandez-suse I know you've designed something for setting up cron jobs - wasn't that UI using something we could reuse here?

edenhernandez-suse:
> In terms of changing the unit of time from hours to seconds for the refresh interval... ideally we would provide a better component to select a duration. It feels like hours would be more helpful to users and can see them being frustrated at having to leave the browser to determine how many seconds are in 8 hours. However we would then become unstuck when editing a resource that has been set to seconds outside of the UI.

@richard-cox couldn't we set some kind of flag, for instance, `unit: "hours"`? So a resource without that flag in its YAML would be considered as created in seconds and could be converted. That way, we wouldn't need any unit selector, as less than 1 hour doesn't feel necessary in this context.

@oboc-sts 
> The second issue implies having a combo (input + dropdown) if we can't find a proper unit of measurement that would satisfy the vast majority of the users. edenhernandez-suse I know you've designed something for setting up cron jobs - wasn't that UI using something we could reuse here?

Allowing users to set the unit would make things even harder, unless my previous suggestion makes any sense. For the cron jobs, we have simple inputs with complex tooltips; it wouldn't be useful here. What we could use is something like this:

image

This component wasn't intended to be used with numbers and units, though.

aruiz14:
Thanks for the review everyone. I've updated this PR taking into account your discussion. See a new set of screenshots:
- Whole page, default value (empty). Units are blocked to show proper state combined with place holder (1 hour)
Screenshot 2026-05-06 at 16 20 14

- Units can be selected after typing a custom value
Screenshot 2026-05-06 at 16 20 27

- Clicking on disable disables the input.
Screenshot 2026-05-06 at 16 20 32

- Enabling back + changing units:
Screenshot 2026-05-06 at 16 20 55

- Typing an invalid value shows a warning
Screenshot 2026-05-06 at 16 24 07

Of course I tested that saving the different values correctly converts to seconds, delete the field (in case of empty or invalid value) or sets `-1` when disabled.

oboc-sts (commented):
@aruiz14 looking good. 1 minor change I'd like to see is replacing the "disabled" checkbox (unchecked by default) with an "enabled" checkbox (checked by default). From a semantic pov, it involves less cognitive power to process its meaning.

aruiz14:
> @aruiz14 looking good. 1 minor change I'd like to see is replacing the "disabled" checkbox (unchecked by default) with an "enabled" checkbox (checked by default). From a semantic pov, it involves less cognitive power to process its meaning.

Thanks for the input @oboc-sts ! Addressed, along with adding a new unit tests and adapting the e2e
New screenshots:
Screenshot 2026-05-07 at 09 54 01
Screenshot 2026-05-07 at 09 54 10

oboc-sts:
LGTM! 🚀

momesgin:
@oboc-sts @aruiz14 I think the new unit selector looks out of place specially when disabled:

image
<br/>

- the text color is different for `hours` when disabled compared to the repo name input
- the unit selector dropdown is overlapping the refresh interval input box, I think the unit selector should become detached

some other states:

image

image
<br/>

also the way that the unit selector gets disabled/enabled is a bit weird to me, I want to be able to change the unit if the refresh interval is enabled, and not be blocked by the empty input field:

https://github.com/user-attachments/assets/5fde710e-ab25-44c1-ad0b-0cb60b5515f1

oboc-sts:
> @oboc-sts @aruiz14 I think the new unit selector looks out of place specially when disabled:
> 
> image
> * the text color is different for `hours` when disabled compared to the repo name input
> * the unit selector dropdown is overlapping the refresh interval input box, I think the unit selector should become detached
> 
> some other states:
> 
> image image
> also the way that the unit selector gets disabled/enabled is a bit weird to me, I want to be able to change the unit if the refresh interval is enabled, and not be blocked by the empty input field:
> 
>  test.mov

Fully agree on all points - from the (static, light theme) screenshots attached previously, I couldn't discover issues - good catch @momesgin!, especially the last one ("I want to be able to change the unit if the refresh interval is enabled, and not be blocked by the empty input field").

Let's try to fix these please...

momesgin:
Closing this in favor of [the new PR](https://github.com/rancher/dashboard/pull/17924)

PR #16965, opened 2026-03-18
Using RBAC to hide UI elements the user can't interact with

### LINKED ISSUE #10094: [BUG] Create namespace button enabled for projects with read-only access
**Setup**
- Rancher version: 2.7.5
- Rancher UI Extensions:
- Browser type & version: Edge

**Describe the bug**

When user has  "project member" permission on a project, the "create namespace" button is present for all other projects

**To Reproduce**

1. Starting with a user with no permissions
2. Adding read-only to project A works as expected and no "create namespace" button is present.
3. Upon adding "project member" permission on project B the "create namespace" button becomes available for both projects

The permissions setup end result should be:

- Project A - Read-only
- Project B - Project Member

**Result**

When user has  "project member" permission on a project, the "create namespace" button is present for all other projects

**Expected Result**

Project A should not show the "create namespace" button

**Screenshots**

With "Project Member" on project B

image
image
image

With "Read-only" on project B

image
image
image

**Additional context**

We found a previous discussion on similar circumstances in #5120

codyrancher:
Fixes #10094

### Summary
Fixes #

### Checklist
- [ ] The PR has a Milestone 

### DIFF
--- a/shell/store/auth.js
+++ b/shell/store/auth.js
@@ -77,6 +77,71 @@ export const getters = {
 
   selfUser(state) {
     return state.selfUser;
+  },
+
+  /**
+   * Returns a function (roleTemplateMap, roleTemplateId, resource, verb) => boolean
+   * that walks the role template inheritance chain to check if a role template
+   * (or any of its ancestors) grants a specific verb on a resource.
+   */
+  roleTemplateGrantsAccess() {
+    return (roleTemplateMap, roleTemplateId, resource, verb) => {
+      const queue = [roleTemplateId];
+      const visited = new Set();
+
+      while (queue.length) {
+        const id = queue.pop();
+
+        if (visited.has(id)) {
+          continue;
+        }
+
+        visited.add(id);
+
+        const rt = roleTemplateMap[id];
+
+        if (!rt) {
+          continue;
+        }
+
+        const hasRule = (rt.rules || []).some((rule) => {
+          return (rule.resources || []).includes(resource) && ((rule.verbs || []).includes(verb) || (rule.verbs || []).includes('*'));
+        });
+
+        if (hasRule) {
+          return true;
+        }
+
+        queue.push(...(rt.roleTemplateNames || []));
+      }
+
+      return false;
+    };
+  },
+
+  /**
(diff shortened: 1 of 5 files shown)


shell/store/auth.js
```diff
@@ -77,6 +77,71 @@ export const getters = {
...
+      if (!principalId || !projectId) {
+        return false;
+      }
+
+      const bindings = rootGetters['management/all'](MANAGEMENT.PROJECT_ROLE_TEMPLATE_BINDING);
+      const roleTemplates = rootGetters['management/all'](MANAGEMENT.ROLE_TEMPLATE);
+      const roleTemplateMap = Object.fromEntries(roleTemplates.map((rt) => [rt.id, rt]));
+
+      const userBindings = bindings.filter((prtb) => {
+        return prtb.projectId === projectId && prtb.principalId === principalId;
```
richard-cox, line 140: Might need to rethink this. We've avoided manual rbac via bindings given it doesn't cover groups.

There's a new feature that we haven't wired in yet which should be able to resolve this, which works on the basis of can signed in user do x to y in context z.

Here's a link to an example of it https://github.com/rancher/rancher/issues/34731#issuecomment-2827937567, hopefully we can use it in some way (whenever we fetch projects determine additional per project permissions)

FWIW there's a ton of these type of issues - https://github.com/rancher/rancher/issues/45827

codyrancher: @richard-cox Good, I was going to ask you to review it anyway. I'll see how difficult doing some of the wiring is unless you already had ideas about it.

richard-cox: Not yet. It'll be tricky, would be good to avoid some kind of 'cache is populated with a response containing xyz permissions' but can only think of then bespoke requests.

Have a play with that param and see if we can get what we need from it, and if it works lets get together and have a think.

codyrancher: For the time being there appears to be a problem with the API but I'll poke a little to see if there's something I can fix on my end https://github.com/rancher/rancher/issues/34731#issuecomment-4086008845.

PR #16966, opened 2026-03-18
Retiring all of the references/usages of ember

### LINKED ISSUE #14005: Retire older UI (Ember) - Remove cluster driver support
Remove the last uses of Ember from the UI.

1. Remove support for cluster-driver based Ember plugins.
3. Update the backend to no longer include the Ember UI code. Accessing Ember routes should (ideally) redirect to the dashboard UI.
5. Fix up any remaining issues and validate UI

codyrancher:
## Notes before review:

- I think we'll have a fair amount of back and forth on this, we still need UX input but I wanted to get a first round of dev input first.
- The change looks huge but a large part of it is deleting testing "infrastructure"
    - It might help to watch the video first to get a better idea of what changes took place.

### Summary
Fully retires Ember UI pages from the Rancher Dashboard.

Fixes #14005

- Removed Ember iframe infrastructure (components, utils, etc.)
- Removed Ember page routes
- Removed ember route/url references
- Updated all related Cypress tests and removed Ember test page objects

*Up for discussion*
- Cluster create: kontainer drivers without an extension component are disabled with a tooltip
- Cluster edit: kontainer driver clusters show a deprecation banner
- Shows warning banner for drivers with Ember `uiUrl`

### Technical notes summary
- Removed/moved some of the emberlink computed props, reconsidering leaving some of it in and referencing it.

- Cluster create: EKS/AKS/GKE cards work; legacy drivers (Linode LKE, OTC CCE, Oracle OKE) appear disabled with tooltip
- Cluster edit: RKE2, custom, imported, and legacy kontainer driver clusters
- Cluster Drivers page: banner, available actions (deactivate/delete/viewInApi only)
- Logout flow

- Cluster create/edit flow — subType detection was refactored
- Navigation to previously Ember-hosted pages (`/g/`, `/n/`, `/k/` routes no longer proxy)

### Screenshot/Video

https://github.com/user-attachments/assets/fad3c2b2-fdbe-4dc3-8468-0a458094d3b0

#### Cluster Driver provisioned cluster edit page
- We now show a warning message in case the user has an existing cluster
 
image

#### Cluster Drivers edit page
- Removed `Custom UI Url` field

image

#### Create Cluster page
- Disable deprecated providers with tooltip

image

### Checklist
- [x] The PR has a Milestone 


shell/models/management.cattle.io.kontainerdriver.js
```diff
@@ -57,22 +48,6 @@ export default class KontainerDriver extends HybridModel {
     return this.showCreate && IMPORTABLE.includes(this.driverName);
   }
 
-  get emberCreatePath() {
```
codyrancher, line 60: I'm debating leaving this in and triggering some of the isEmber logic off this stuff instead of creating something new.

shell/edit/provisioning.cattle.io.cluster/index.vue
```diff
@@ -309,6 +249,24 @@ export default {
       return this.value.isRke2;
     },
 
+    isEmberKontainerDriver() {
```
codyrancher, line 269: Perhaps leave emberlink in instead and use that.

cypress/e2e/tests/pages/manager/kontainer-drivers.spec.ts
codyrancher, line 1: I've disabled a number of these tests either because they're no longer relevant or they're more of a pain to verify now that I've removed some of the actions we can perform on them. If any are still some you would like to have enabled let me know and I'll work on it more.

shell/edit/provisioning.cattle.io.cluster/index.vue
```diff
@@ -176,6 +160,30 @@ export default {
       subType = LOCAL;
     }
 
+    // Auto-detect subType for existing clusters being edited
+    if ( !subType && this.value?.id ) {
```
mantis-toboggan-md, line 181: Imported hosted clusters are loading the generic imported configuration page on edit now (tested w/ AKS) and I _think_ it ties into this line. I think the logic in this if block was previously running whether or not `subType` was already defined and we were reliant on that behavior. 

`this.value.isImported` will be `true` for imported hosted clusters but we need `subType` set to `this.value.provisioner.toLowerCase()` to load the right custom form.

mantis-toboggan-md (changes requested):
We should still support creating and activating kontainer drivers, just not the custom ui property nor any other ember-based UI for them (think LKE was technically baked into rancher/ui, not provided by the driver?). I apologize the github issue was not clearer on this point.

We have one customer in particular who want to build a dashboard extension for their kontainer driver offering. Down the road we want people using CAPI providers for hosted cluster solutions instead but we're a ways out from supporting that.

I've tested the create/import, edit, and detail pages for every category of cluster we currently support and hit an issue with hosted imported clusters only, commented inline, all others looked right to me.

shell/edit/provisioning.cattle.io.cluster/index.vue
codyrancher: You're correct about the cause of the problem. I made a small tweak to your suggestion. If all I did was replace 

```
else if (this.value.isImported) {
      subType = IMPORTED;
}
```

with 
```
else if (this.value.isImported) {
      subType = this.value.provisioner.toLowerCase();
}
```

Then we wouldn't show the edit page with the banner for the ember clusters. If we'd prefer to not show that page with the banner then we can go with your exact suggestion. (Assuming I interpreted it correctly)

I verified this on eks, but I suspect it would behave the same with aks.

codyrancher:
> We should still support creating and activating kontainer drivers, just not the custom ui property nor any other ember-based UI for them (think LKE was technically baked into rancher/ui, not provided by the driver?). I apologize the github issue was not clearer on this point.
> 
> We have one customer in particular who want to build a dashboard extension for their kontainer driver offering. Down the road we want people using CAPI providers for hosted cluster solutions instead but we're a ways out from supporting that.
> 
> I've tested the create/import, edit, and detail pages for every category of cluster we currently support and hit an issue with hosted imported clusters only, commented inline, all others looked right to me.

No problem, I figured that we may have wanted to keep them around.

shell/edit/provisioning.cattle.io.cluster/index.vue
mantis-toboggan-md: I think you've got the better solution, let's roll with it

mantis-toboggan-md (approved):
LGTM

PR #17000, opened 2026-03-24
Automation - Integrate Qase reporter, add test pre-filtering, and improve CI stability

izaac:
### Summary
Fixes rancher/qa-tasks#1866
Depends: https://github.com/rancherlabs/corral-packages/pull/78

Switches our test reporting to use the Qase Cypress reporter directly instead of relying on a custom Python script. It also adds a new script to filter tests before they run and cleans up the CI environment teardown process so things run more smoothly.

- Qase Integration: Brought in `cypress-qase-reporter` and started wrapping tests (like in `version.spec.ts`) to link them directly to Qase cases.
- Removed Old Scripts: Deleted the old junit_to_qase.py and transform-junit.sh files since we don't need them anymore.
- Pre-Filtering: Added `grep-filter.js` to filter tests by tags upfront, speeding up Cypress startup.
- Pipeline Tweaks: Updated the Jenkinsfile and shell scripts (`cypress.sh`, `init.sh`) to support the new reporter and pass the Qase token.
- Cleaner Teardown: Added a clean_corral function in utils.sh to make sure Jenkins test environments are properly deleted, even if tests fail.

### Technical notes summary
  
- Switched to cypress-multi-reporters to generate Mochawesome, JUnit, and Qase reports all at the same time.
- The Qase reporter now automatically turns itself on or off depending on whether the `QASE_AUTOMATION_TOKEN` is present in the environment.
 - Moved the grepTags calculation into the shell/JS layer. By figuring out the tags before Cypress boots, we avoid unnecessary overhead from skipped tests.
- Made the environment cleanup script more reliable to prevent orphaned test clusters from hanging around in Jenkins.

E2E / CI (Jenkins/Github Actions)

E2E / CI (Jenkins/Github Actions)

### Checklist
- [x] The PR has a Milestone 


izaac:
Decoration examples:

```
App Bar Version Number
    ✓ app bar shows version number (Qase ID: 3579)
    ✓ app bar shows short version number (Qase ID: 3578)
```

cypress/jenkins/cypress.config.jenkins.ts
```diff
@@ -74,15 +81,15 @@ export default defineConfig({
...
-    grepFilterSpecs:     true,
+    grepFilterSpecs:     false,
     grepOmitFiltered:    true,
     baseUrl,
     api:                 apiUrl,
     username,
     password:            process.env.CATTLE_BOOTSTRAP_PASSWORD || process.env.TEST_PASSWORD,
     bootstrapPassword:   process.env.CATTLE_BOOTSTRAP_PASSWORD,
-    // Jenkins runs against Vai-enabled Rancher; exclude @noVai tests.
-    grepTags:            `${ process.env.GREP_TAGS }+-@noVai`,
```
yonasberhe23, line 85: I'm not certain, but i think`+-@noVai` would need to be added in the `cypress.sh`, otherwise Vai will be disabled in our test environments. That tag was added here to prevent that from happening.

izaac: Of course, all the tagging configuration got moved out from the cypress lib configuration to use `--spec` directly. So the `@noVai` logic is moved to `init.sh` 

One of the reasons was to allow the proper tests isolation. The old cypress tags lib was not playing along with the "advanced tagging" we are doing, it was not sure how to interpret it properly and delegating the task to the browser at run time.

That's the reason Cypress was running what matched and then everything else with zero tests making the execution longer than needed.

yonasberhe23: You’re right, I totally forgot init.sh already adds `+-@noVai` to CYPRESS_TAGS. That addresses my concern. Thank you!

cypress/jenkins/init.sh
```diff
@@ -42,285 +45,271 @@ YARN_VERSION="${YARN_VERSION:-1.22.19}"
...
+    CYPRESS_TAGS=$(clean_tags "${CYPRESS_TAGS}")
   else
-    CYPRESS_TAGS="${CYPRESS_TAGS}+-@noVai+-@prime"
+    # Automatically exclude @noPrime tests on Prime/Alpha/Latest environments
+    if [[ "${RANCHER_HELM_REPO}" == "rancher-prime" || "${RANCHER_HELM_REPO}" == "rancher-latest" || "${RANCHER_HELM_REPO}" == "rancher-alpha" ]]; then
+      if [[ ! "${CYPRESS_TAGS}" =~ "@noPrime" ]]; then
+        CYPRESS_TAGS="${CYPRESS_TAGS}+-@noPrime"
+      fi
+    # Automatically exclude @prime tests on Community/RC environments
+    elif [[ "${RANCHER_HELM_REPO}" == "rancher-com-rc" ]]; then
```
yonasberhe23: `+-@prime` is only applied for `rancher-com-rc` here; `rancher-community` and `rancher-com-alpha` never get it, so Prime-only tests will run on those builds. Can we extend this condition or use else to match the previous non-Prime behavior?

izaac: Good catch, I simply changed it to an `else` so any build not using `rancher-prime`, `rancher-latest`, `rancher-alpha` will exclude them.

PR #17002, opened 2026-03-24
Pin GH Actions to commit sha

richard-cox:
### Summary
- Pin GH Actions to commit sha
  - Only inserts sha, does not change action repo/name
  - Example
    - `docker/login-action@v4` -->
    - `docker/login-action@b45d80f862d83dbcd57f89517bcf500b2ab88fb2 # v4`
- Created using the security team's automated tool
  - file updates resulted in some trailing blank lines being removed
- For more info see Gary's thread in our slack channel

Some of note

- build-extension-charts.yml
  - `uses: rancher/dashboard/.github/workflows/build-extension-charts.yml@master`
  - `uses: rancher/dashboard/.github/workflows/build-extension-charts.yml@9eb70a732e9be146722e1dbab431338366c2afc6 # creators-pkg-v3.0.10`

- There's a lot here, we're just going to have to react to issues instead of explicitly testing everything

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/.github/workflows/test-extension-workflows-master.yml
+++ b/.github/workflows/test-extension-workflows-master.yml
@@ -17,7 +17,7 @@ jobs:
       test-extensions-creators-and-update-paths-job-status: ${{ job.status }}
     steps:
       - name: Checkout code
-        uses: actions/checkout@v6
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
 
       - name: Retrieve Shell and Creators versions
         run: .github/workflows/scripts/test-extensions-creators-build.sh
@@ -32,7 +32,7 @@ jobs:
         branch: [master, release-2.9, release-2.8]
     steps:
       - name: Checkout
-        uses: actions/checkout@v6
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
 
       - name: Validate shipped workflows with creator for branch ${{ matrix.branch }}
         shell: bash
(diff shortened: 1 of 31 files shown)


.github/workflows/test-extension-workflows-master.yml
```diff
@@ -61,7 +61,7 @@ jobs:
 
   test-build-extension-charts-master:
     needs: retrieve-tags-master
-    uses: rancher/dashboard/.github/workflows/build-extension-charts.yml@master
+    uses: rancher/dashboard/.github/workflows/build-extension-charts.yml@9eb70a732e9be146722e1dbab431338366c2afc6 # creators-pkg-v3.0.10
```
codyrancher, line 64: We have a few that appear to be referencing this commit https://github.com/rancher/dashboard/commit/9eb70a732e9be146722e1dbab431338366c2afc6

I wonder if there will be any problems.

richard-cox: i was wondering about this too. We don't make r/d releases explicitly for this workflow, just general releases. 

We also hit the issue where we would need extensions to update their workflows every time we want to fix/update our workflow (as they'll all start to be pinned).

Not sure if there's a good solution here. We could stick with `master` but that might cause that org level change to flag us. We could create a specific release name for this workflow, but that would still flag. We can leave as is in the pr and dependabot will bump extension workflows whenever we do any r/d release. We could drop the `#` and then dependabot would leave alone and users would need to manually bump.

Very open to ideas here @codyrancher & @aalves08

codyrancher: I think we just leave it for now and see what comes out of it.

codyrancher: Are we also going to have to audit all of the dependencies since the pinning doesn't ensure downstream dependencies or images are immutable?

richard-cox: That was brought up in the RATs channel, they're currently discussing steps

PR #17008, opened 2026-03-24
Jenkins: Skip chart e2e tests when charts are filtered by catalog (2.14)

yonasberhe23:
### Summary
Fixes #https://github.com/rancher/qa-tasks/issues/2255, https://github.com/rancher/qa-tasks/issues/2256, https://github.com/rancher/qa-tasks/issues/2257, https://github.com/rancher/qa-tasks/issues/2258, https://github.com/rancher/qa-tasks/issues/2259, https://github.com/rancher/qa-tasks/issues/2260, https://github.com/rancher/qa-tasks/issues/2261, https://github.com/rancher/qa-tasks/issues/2264
 

Stabilizes chart-related E2E tests when the catalog hides charts by design (e.g. version / annotation filtering) and when only pre-release chart versions exist.

Problem: Some charts are deliberately hidden on the Charts page because of **chart annotations** (and related catalog rules). Others disappear from the list when the user has not enabled `Include Prerelease Versions` (show-pre-release): the raw Helm index can still include those charts, but the UI only shows a filtered catalog. Our tests used to assume every chart in the index would appear on Charts, which led to false failures when a chart was correctly hidden (by annotations or by that preference).

Fixes:
1. Adds custom command `runTestWhenChartAvailable` which skips the test when the chart exists in the unfiltered index but not in the filtered one (intentional catalog filtering). If the chart appears in the filtered index, the test runs as usual.
2. Enables `Include Prerelease Versions` user preference so that chart-related specs set show-pre-release before opening the Charts catalog, so charts that only have pre-release versions are not hidden by default UI behavior.
3. Extends setUserPreference with an optional verify flag: after the PUT, it retries GETs until the stored preferences match the requested values; if they still do not match after the retries, it logs a failure message and continues the test (helps surface timing/sync issues without failing the run on verification alone).
4. Resolves flakiness in various Charts tests

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/cypress/e2e/tests/pages/charts/compliance.spec.ts
+++ b/cypress/e2e/tests/pages/charts/compliance.spec.ts
@@ -5,10 +5,12 @@ import { MEDIUM_TIMEOUT_OPT, LONG_TIMEOUT_OPT } from '@/cypress/support/utils/ti
 import Kubectl from '@/cypress/e2e/po/components/kubectl.po';
 import { CompliancePo, ComplianceListPo } from '~/cypress/e2e/po/other-products/compliance.po';
 import ProductNavPo from '@/cypress/e2e/po/side-bars/product-side-nav.po';
+import { runTestWhenChartAvailable } from '@/cypress/support/commands/rancher-api-commands';
 
 describe('Charts', { testIsolation: 'off', tags: ['@charts', '@adminUser'] }, () => {
   before(() => {
     cy.login();
+    cy.setUserPreference({ 'show-pre-release': true }, true); // Show pre-release versions so charts with only -rc versions appear on Charts page
     HomePagePo.goTo();
   });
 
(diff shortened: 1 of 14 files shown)


cypress/e2e/tests/pages/charts/compliance.spec.ts
```diff
@@ -17,78 +19,80 @@ describe('Charts', { testIsolation: 'off', tags: ['@charts', '@adminUser'] }, ()
...
-            expect(elementRect.bottom).to.eq(pageHeight);
-            expect(elementRect.bottom).to.eq(viewportHeight);
+        it('Footer controls should sticky to bottom', function() {
+          runTestWhenChartAvailable('rancher-charts', 'rancher-compliance', this, () => {
+            ChartPage.navTo(null, 'Rancher Compliance');
+            chartPage.waitForChartHeader('Rancher Compliance', MEDIUM_TIMEOUT_OPT);
+            chartPage.goToInstall();
+            installChartPage.nextPage().editYaml();
+
+            cy.get('#wizard-footer-controls').should('be.visible');
```
izaac: Can this be encapsulated in something like:

```javascript
 footerControls(): Cypress.Chainable {
   return cy.get('#wizard-footer-controls');
 }
// ...
// Then
// ...
 installChartPage.footerControls().should('be.visible');
 
 installChartPage.footerControls().then(($el) => {
   const elementRect = $el[0].getBoundingClientRect();
   // ...
 });
```

cypress/e2e/tests/pages/charts/opa-gatekeeper.spec.ts
```diff
@@ -48,27 +52,31 @@ describe('Charts', { testIsolation: 'off', tags: ['@charts', '@adminUser'] }, ()
...
+        it('Footer controls should sticky to bottom', function() {
+          runTestWhenChartAvailable('rancher-charts', 'rancher-gatekeeper', this, () => {
+            ChartPage.navTo(null, 'OPA Gatekeeper');
+            chartPage.waitForPage('repo-type=cluster&repo=rancher-charts&chart=rancher-gatekeeper');
+            chartPage.goToInstall();
+            installChartPage.nextPage().editYaml();
+
+            cy.get('#wizard-footer-controls').should('be.visible');
+
+            cy.get('#wizard-footer-controls').then(($el) => {
```
izaac: Same about the footer encapsulation.

cypress/support/commands/rancher-api-commands.ts
```diff
@@ -1233,7 +1234,29 @@ Cypress.Commands.add('setUserPreference', (prefs: any) => {
...
+
+          return Object.entries(prefs).every(([key, value]) => {
+            const actual = responseData[key];
+
+            return actual === value || String(actual) === String(value);
+          });
+        }, retries)
+          .then((success) => {
+            if (!success) {
+              return cy.log(`setUserPreference: Failed to verify preferences ${ JSON.stringify(prefs) } after ${ retries } retries (non-fatal)`).then(() => putResp);
```
izaac: Shouldn't this be fatal for tests using this, if the intention is to change the setting and is not changed.

cypress/support/commands/rancher-api-commands.ts
```diff
@@ -1233,7 +1234,29 @@ Cypress.Commands.add('setUserPreference', (prefs: any) => {
...
+          return putResp;
+        }
+
+        return cy.waitForRancherResource('v1', 'userpreferences', update.id, (getResp: any) => {
+          const responseData = getResp?.body?.data ?? getResp?.body ?? {};
+
+          return Object.entries(prefs).every(([key, value]) => {
+            const actual = responseData[key];
+
+            return actual === value || String(actual) === String(value);
```
izaac: Are the value types returned from the API known? Or can we just use `String()` here.

cypress/support/commands/rancher-api-commands.ts
```diff
@@ -1233,7 +1234,29 @@ Cypress.Commands.add('setUserPreference', (prefs: any) => {
...
+          });
+        }, retries)
```
izaac, line 1251: Could this be simplified?

Literal logic in your code:

```javascript
     }, retries)
       .then((success) => {
         const msg = success
           ? `Successfully verified preferences ${ JSON.stringify(prefs) }`
           : `Failed to verify preferences ${ JSON.stringify(prefs) } after ${ retries } retries (non-fatal)`;
 
         return cy.log(`setUserPreference: ${ msg }`).then(() => putResp);
       });
   });
```
With fatal failure:
```javascript
     }, retries)
       .then((success) => {
         expect(success, `setUserPreference: preferences ${ JSON.stringify(prefs) } verified after polling`).to.be.true;
 
         return putResp;
       });
```

izaac (changes requested):
A few improvement candidates to review. Looking solid.

cypress/support/commands/rancher-api-commands.ts
yonasberhe23: The verify step is an extra read-back to confirm the value stuck. I had that failing the test at first, but I don’t want CI blocked on a false negative when the read-back doesn’t match in time (timing issue, etc.), even though the preference update was accepted.
So if verify doesn’t match after retries, we log it and continue. If the preference really wasn’t applied, the test will fail on something visible later (e.g., expected chart missing).

cypress/e2e/tests/pages/charts/opa-gatekeeper.spec.ts
yonasberhe23: good call, updated!

cypress/e2e/tests/pages/charts/compliance.spec.ts
yonasberhe23: good call, updated!

cypress/support/commands/rancher-api-commands.ts
yonasberhe23: Agreed, we can use String() here. Normalizing with String() on both sides keeps the check consistent.

cypress/support/commands/rancher-api-commands.ts
yonasberhe23: yessir!

izaac (approved):
LGTM :shipit:

PR #17018, opened 2026-03-25
Adding private registry to cloud provider imported clusters

### LINKED ISSUE #14916: Add private registry support to hosted imported providers
JIRA: SURE-10031

We should add private registry support to imported hosted clusters.
In 2.11 we added it to generic imported clusters, however, we are being asked to add it to hosted imported clusters as well

codyrancher:
### Summary
We already had the private registry for generic import clusters, this generalizes the component and uses it in the imported cloud provider clusters.

Fixes #14916

Create a new component, used it in each of these cluster types.

### Technical notes summary
- Assuming same `spec` structure for each cluster.

- I verified in generic and eks, I don't have access to gke or aks yet but once I do I'll verify there as well.
- Verify private registry component when creating and editing the following imported clusters:
    - GKE
    - AKS
    - EKS
    - Generic

- See above

### Screenshot/Video

https://github.com/user-attachments/assets/f2b2616a-abf1-45ab-86ac-8793577ac500

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/pkg/aks/components/CruAks.vue
+++ b/pkg/aks/components/CruAks.vue
@@ -265,7 +265,7 @@ export default defineComponent({
     },
 
     isImportedCluster() {
-      return this.isImport || !!this.config?.imported;
+      return this.isImport || this.value.isImported;
     },
     /**
      * fv mixin accepts a rootObject in rules but doesn't seem to like that the norman cluster isn't yet defined when the rule set is defined so we're ignoring that and passing in the key we want validated here
--- a/pkg/aks/components/CruAks.vue
+++ b/pkg/aks/components/CruAks.vue
@@ -22,6 +22,7 @@ import Loading from '@shell/components/Loading.vue';
 import Config from './Config.vue';
 import Import from './Import.vue';
 
+import PrivateRegistry from '@shell/components/form/PrivateRegistry.vue';
 import ClusterMembershipEditor, { canViewClusterMembershipEditor } from '@shell/components/form/Members/ClusterMembershipEditor.vue';
 import type { AKSDiskType, AKSNodePool, AKSPoolMode, AKSConfig } from '../types/index';
 import {
@@ -89,6 +90,7 @@ const defaultCluster = {
   labels:                  {},
   annotations:             {},
   windowsPreferedCluster:  false,
+  importedConfig:          { privateRegistryURL: null },
 };
 
 export const NETWORKING_AUTH_MODES = {
@@ -110,6 +112,7 @@ export default defineComponent({
     Labels,
     Accordion,
     Banner,
+    PrivateRegistry,
     Loading,
     Config,
     Import
@@ -176,6 +179,9 @@ export default defineComponent({
         this.normanCluster.annotations[CREATOR_PRINCIPAL_ID] = this.$store.getters['auth/principalId'];
       }
     }
+    if (!this.normanCluster.importedConfig) {
+      this.normanCluster.importedConfig = {};
+    }
     if (this.isImport) {
       this.normanCluster.aksConfig = cloneDeep(importedDefaultAksConfig);
       this.config = this.normanCluster.aksConfig;
@@ -204,7 +210,7 @@ export default defineComponent({
 
     return {
       NETWORKING_AUTH_MODES,
-      normanCluster:    { name: '' } as any,
+      normanCluster:    { name: '', importedConfig: { privateRegistryURL: null } } as any,
       nodePools:        [] as AKSNodePool[],
       config:           { } as AKSConfig,
       membershipUpdate: {} as any,
@@ -225,10 +231,18 @@ export default defineComponent({
       {
         path:  'clusterName',
         rules: ['importedName']
+      },
+      {
+        path:  'normanCluster.importedConfig.privateRegistryURL',
+        rules: ['registryUrl']
       }
       ] : [{
         path:  'name',
         rules: ['nameRequired', 'clusterNameChars', 'clusterNameStartEnd', 'clusterNameLength'],
+      },
+      {
+        path:  'normanCluster.importedConfig.privateRegistryURL',
+        rules: ['registryUrl']
       }],
     };
   },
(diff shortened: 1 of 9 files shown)


shell/components/form/PrivateRegistry.vue
```diff
@@ -0,0 +1,86 @@
+<script lang="ts">
```
richard-cox: Comment only.

Having a generic component here will really help us if the discussion around authenticated registries turns into a feature. We'd need to ensure all cluster types support registries... and also provide a secret.

pkg/eks/components/CruEKS.vue
```diff
@@ -330,6 +346,10 @@ export default defineComponent({
       return this.$fetchState;
     },
 
+    isImportedCluster(): boolean {
+      return this.isImport || !!this.config?.imported;
```
codyrancher: Maybe there's a better way to do this, ultimately we want to show the private registry component on imported clusters both when creating/editing the cluster. this.isImport is only true when creating an imported cluster and not when editing.

eva-vashkevich: I think instead of !!this.config?.imported; we could use this.value.isImported but I don't think it makes a big difference?

codyrancher: I think that's valid, looks like it still works even though it's more involved. I switched each over to use `isImported`.

shell/components/form/PrivateRegistry.vue
```diff
@@ -0,0 +1,61 @@
...
+  />
+  <LabeledInput
+    v-if="showInput"
+    :value="value as string"
+    :mode="mode"
+    :rules="rules"
+    label-key="catalog.chart.registry.custom.inputLabel"
+    :data-testid="inputTestId"
+    :placeholder="t('catalog.chart.registry.custom.placeholder')"
+    @update:value="onInput"
```
eva-vashkevich: preference: if it's just an emit, why not put it here?

pkg/aks/components/CruAks.vue
```diff
@@ -176,6 +179,9 @@ export default defineComponent({
         this.normanCluster.annotations[CREATOR_PRINCIPAL_ID] = this.$store.getters['auth/principalId'];
       }
     }
+    if (!this.normanCluster.importedConfig) {
```
eva-vashkevich: shouldn't we only do in case of an import/editing imported cluster?
same for other providers

pkg/eks/components/__tests__/CruEKS.test.ts
```diff
@@ -235,6 +235,52 @@ describe('eKS provisioning form', () => {
...
+      ...requiredSetup()
+    });
+
+    await setCredential(wrapper, { ...DEFAULT_EKS_CONFIG, imported: true } as EKSConfig);
+    const registriesAccordion = wrapper.find('[data-testid="registries-accordion"]');
+
+    expect(registriesAccordion.exists()).toBe(true);
+  });
+
+  it('should not show the registries accordion for a non-imported cluster', async() => {
```
eva-vashkevich, line 272: I think it would be good to add an edit case as well

eva-vashkevich (changes requested):
Shouldn't registry be required if enabled? If I enable it, but don't provide the registry, it switches to disabled on edit. I think this behavior is confusing.
other than that, tested:
- Generic imported cluster: Import and edit
- EKS: Creation(doesn't provision fully, but doesn't seem related), edit, import(doesn't import fully, but doesn't seem related), edit imported
- AKS: Creation, edit, import(doesn't become available with fake registry), edit imported (doesn't become available with fake registry)
- GKE: Creation, edit, import (doesn't become available with fake registry), edit imported (doesn't become available with fake registry)
Image

codyrancher:
> Shouldn't registry be required if enabled? If I enable it, but don't provide the registry, it switches to disabled on edit. I think this behavior is confusing. other than that, tested:
> 
>     * Generic imported cluster: Import and edit
> 
>     * EKS: Creation(doesn't provision fully, but doesn't seem related), edit, import(doesn't import fully, but doesn't seem related), edit imported
> 
>     * AKS: Creation, edit, import(doesn't become available with fake registry), edit imported (doesn't become available with fake registry)
> 
>     * GKE: Creation, edit, import (doesn't become available with fake registry), edit imported (doesn't become available with fake registry)
> 
> 
> Image

I'm not sure, from the issue it sounded like we just wanted to carry the existing behavior over. 

I'll check with UX and see what their opinion is.

_Edit: I changed my mind and included the registry url as a required field._

pkg/aks/components/CruAks.vue
codyrancher: Yeah, I think that valid.

One thing to note, the CruAks files seem to define isEdit as edit || create for some reason. Because CruImported already used value.id to check if it existed I'm doing the same here but I'd be happy to check the mode manually if that's your preference.

pkg/eks/components/__tests__/CruEKS.test.ts
codyrancher: Done

shell/components/form/PrivateRegistry.vue
codyrancher: I've gone ahead and made the change in case it's a general preference for provisioning but the reason I don't do that by default is because I want as little logic/code inside templates as possible.

It muddies the templates and makes it more difficult to add breakpoints when debugging.

pkg/imported/components/CruImported.vue
```diff
@@ -407,35 +409,24 @@ export default defineComponent({
...
-        this.normanCluster.importedConfig.privateRegistryURL = null;
-      }
-    }
   }
-
 });
 </script>
 
 <template>
+  <Loading v-if="$fetchState.pending" />
```
codyrancher, line 417: Made this consistent with the other providers. Without this the provider will momentarily flash an error banner for the name requirement.

We could also solve that by conditionally passing in the fvUnreportedValidationErrors but I preferred this direction.

When running locally the loading happens to fast to even show the loading component but this is true of all the other providers as well.

eva-vashkevich (approved):
lgtm

PR #17022, opened 2026-03-26
Add instructions for agentic workflow gh issue and pr label format

richard-cox:
### Summary
Add instructions for agentic workflow gh issue and pr label format

- discussed in team channel
- i've kept this in the agents.md instead of more generally. we might be lucky and bots automatically do it (if we don't hardcode it)?

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/docs/agents.md/contributors/4_agentic-bot.md
+++ b/docs/agents.md/contributors/4_agentic-bot.md
@@ -2,7 +2,9 @@
 
 ## Labels
 
-- When an agentic workflow creates a github issues or pull requests it should add a label in the following format: `bot/<bot name>`
+Agentic workflows may update github issue and pull requests labels. In addition to any they manage they should also adhere to the following
+
+- When an agentic workflow creates a github issue or pull request it should add a label in the following format: `bot/<bot name>`
   - e.g. bot name is `issue grooming`, label is `bot/issue-grooming`
-- An agentic workflow can receive instructions or state via github issue or pull request labels with the following format: `bot/<bot name>/<instruction>`
+- An agentic workflow can receive instructions or state from itself or other sources via github issue or pull request labels with the following format: `bot/<bot name>/<instruction>`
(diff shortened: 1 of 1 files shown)


docs/agents.md/contributors/4_agentic-bot.md
```diff
@@ -0,0 +1,8 @@
+## Agentic Workflows
+
+## Labels
+
+- When an agentic workflow creates a github issues or pull requests it should add a label in the following format: `bot/<bot name>`
+  - e.g. bot name is `issue grooming`, label is `bot/issue-grooming`
+- An agentic workflow can receive instructions or state via github issue or pull request labels with the following format: `bot/<bot name>/<instruction>`
+  - e.g. bot name is `issue grooming`, instruction is `ready for triage`, label is `bot/issue-grooming/ready-for-triage`
```
marcelofukumoto, line 10: I think it would be better to clarify that those bot/<bot name> labels are only for the ones used for the bot internal usage (or between bots)
A bot can still use a normal label used on the normal process, I believe. So only for the specific ones used for the internal process it should use the `bot/`

richard-cox: I got the text wrong, but not quite in that way.

- `When an agentic workflow creates a github` --> `When an agentic workflow creates or updates`. This makes it easy to find all their activity
- `An agentic workflow can receive instructions or state` --> `An agentic workflow can receive instructions or state from itself or other sources`. This means we ourselves can do things like `bot/issue-grooming/ignore` or `bot/issue-grooming/force`

marcelofukumoto: For me the text is explaining how the bot should use its labels.
But what I mean is something like this:

> Note: In addition to these bot-specific tracking labels, agentic workflows can apply standard process labels (e.g., kind/bug) just like a human contributor.

Just to be clear that it can use standard labels as well.

richard-cox: Added blurb at start covering this. i've stuck to 'any they manage' given they would manage kind/bug, etc

PR #17024, opened 2026-03-26
Prevent secret values from leaking into the DOM when concealed

### LINKED ISSUE #17023: Concealed secret values can be copied as plaintext via text selection
**Setup**
- Rancher version: v2.13.3 (also verified on master)
- Browser type & version: Chrome (all versions affected)

**Describe the bug**
Secret values displayed in the Kubernetes Secret detail view can be copied as plaintext by selecting text on the page (Ctrl+A, mouse drag, or any text selection method). The concealment mechanism uses a custom dot font (`dotsfont`) that renders characters as dots visually, but the actual secret values remain as plaintext in the DOM.

**To Reproduce**
1. Navigate to Cluster Explorer → Storage → Secrets
2. Click on any Opaque secret to view its details
3. On the Data tab, observe the masked values (dots)
4. Select text on the page using Ctrl+A or mouse drag
5. Copy with Ctrl+C and paste into a text editor

**Result**
The pasted content contains the actual secret values in plaintext, not the masked dots. This occurs because the `.conceal` CSS class applies a custom font (`dotsfont`) that renders characters as dots, but the real text content remains in the DOM and is copied on selection.

**Expected Result**
Selecting and copying concealed secret values should not expose the actual plaintext values. The DOM should not contain real secret values when they are masked.

**Screenshots**
The screenshot below shows the Data tab with masked secret values. Selecting any of these values and copying exposes the real secret in plaintext.

**Additional context**
The issue affects two components:
- `shell/components/form/KeyValue.vue` — uses `.conceal` CSS class with `dotsfont` on `TextAreaAutoGrow`, leaving real values in DOM
- `shell/components/DetailText.vue` — same dot font approach on `<span>` and `CodeMirror` elements

The root cause is in `shell/assets/styles/fonts/_dots.scss` where `.conceal` is defined as a font-family swap rather than actual content replacement.

adifsgaid:
### Summary
Fixes #17023

- Secret values concealed with the `dotsfont` CSS approach were still present as plaintext in the DOM, allowing text selection + copy to expose actual secret values
- `KeyValue.vue`: replaced the `TextAreaAutoGrow` with `.conceal` class with a static placeholder `<div>` that renders bullet characters and uses `user-select: none` when `valueConcealed` is true
- `DetailText.vue`: when `concealed`, renders a static bullet placeholder instead of the real value. JSON values skip `CodeMirror` entirely when concealed

### Technical notes summary
- The fix ensures real secret values are never rendered in the DOM when concealed — the `dotsfont` visual masking is replaced with a static placeholder that contains only bullet characters
- The `CopyToClipboard` button still copies the real value (this is intentional — it requires an explicit user click)
- The `input type="password"` fallback path is unchanged as it already uses native browser masking

- Browser used for local testing: Chrome
- View an Opaque secret with the "Hide Sensitive Data" preference enabled (default)
- Select all text on the page (Ctrl+A) and copy — should NOT contain real values
- Mouse-select text on the value area and copy — should NOT contain real values
- Click the eye icon to reveal values — should show the real values
- Click the Copy button — should copy the real value to clipboard
- Toggle the global "Hide Sensitive Data" preference off and verify values show normally
- Test with JSON-formatted secret values
- Test in both light and dark mode

- Secret detail view (Data tab) for all secret types (Opaque, Basic Auth, SSH, TLS, Registry)
- `KeyValue` component used elsewhere with `valueConcealed` prop
- `DetailText` component used for displaying concealed values

### Checklist
- [x] The PR has a Milestone


shell/components/form/KeyValue.vue
```diff
@@ -941,6 +948,12 @@ export default {
         padding: 10px 10px 10px 10px;
       }
 
+      .concealed-value {
+        padding: 10px;
+        min-height: 40px;
+        user-select: none;
+      }
```
rak-phillip, line 956: nit: I think that removing the bullets from the template and defining them in CSS content will make this change purely presentational:

```suggestion
      .concealed-value {
        padding: 10px;
        min-height: 40px;
        user-select: none;

        &::before {
          content: '••••••••••••••••••••';
        }
      }
```

Doing something like this will ensure that the text doesn't render at all in the DOM tree, which also means that this text shouldn't be able to be selected by default.

shell/components/DetailText.vue
```diff
@@ -199,9 +198,16 @@ export default {
       :class="{'conceal-wrapper': concealed}"
     >
       <span
+        v-if="concealed"
+        data-testid="detail-top_html"
+        class="conceal"
+        aria-live="polite"
+      >&#8226;&#8226;&#8226;&#8226;&#8226;&#8226;&#8226;&#8226;&#8226;&#8226;&#8226;&#8226;&#8226;&#8226;&#8226;&#8226;&#8226;&#8226;&#8226;&#8226;&#8226;&#8226;&#8226;&#8226;</span>
```
rak-phillip: nit: we could also perform the same in for the `conceal` class below: https://github.com/rancher/dashboard/pull/17024/changes#r3054565234

rak-phillip (commented):
@adifsgaid thanks for raising this PR and the associated issue. We might also be able to remove the dots font with this change now that we no longer render every character as a dot by swapping the font family. I think we can do the following now: 

- remove `shell/assets/styles/fonts/_dots.scss`
- delete the dots import in `shell/assets/styles/app.scss`

shell/components/form/KeyValue.vue
adifsgaid: Done, moved the bullets into ::before content and emptied the template. Pushed in e01e2c9.

shell/components/DetailText.vue
adifsgaid: Done, same treatment here, ::before on the scoped .conceal rule.

adifsgaid:
@rak-phillip applied both nits, bullets are now in CSS ::before and the templates are empty.

Also went ahead and removed the dots font. LabeledInput in rancher-components was the only other consumer of the .conceal class for multiline-password masking (used in googleoauth.vue). Replaced it with -webkit-text-security: disc which does the same thing natively without the custom font, if anything else needs to be done feel free to add further comments !

rak-phillip (approved):
@adifsgaid excellent, thanks for the follow up and for contributing to Rancher!

PR #17029, opened 2026-03-26
Rancher Backups two new monitoring options

### LINKED ISSUE #13513: [RFE] Update UI to allow enabling metrics for Backup&Restore Operator
The O&B team has implemented metrics for the Backup&Restore Operator. These can currently be enabled via Helm Charts by setting these values to true:

```
monitoring:
  metrics:
    enabled: false

  serviceMonitor:
    enabled: false
```
We also want users to be able to set these via UI in cluster tools --> rancher-backup --> step 2 (the screen shown in the image):
Image

Ideally we'd see another section below the existing "Default Storage Location" named "Monitoring". In this new section we should have the ability to turn on the metrics (Values.monitoring.metrics.enabled) and the serviceMonitor 
 (Values.monitoring.serviceMonitor.enabled), each independently of the other. Also, the user should only be allowed to enable the serviceMonitor via UI if rancher-monitoring is installed.

**Additional context**
https://github.com/rancher/backup-restore-operator/pull/648
https://github.com/rancher/backup-restore-operator/issues/352

momesgin:
### Summary
Fixes #13513

We now expose two new options when installing Rancher Backups chart.
### Technical notes summary

- added a new "Monitoring" section and two new checkboxes to allow setting `value.monitoring.metrics.enabled` and `value.monitoring.serviceMonitor.enabled`.
- using the existing `monitoringStatus` helper function to detect if the monitoring is installed, and accordingly disable/enable the `serviceMonitor` option.
- unit tests.

- Make sure when you install Rancher Backups chart you see the new options and you can interact with them
- Make sure having the `rancher-monitoring` chart installed changes how the `serviceMonitor` option gets enabled/disabled

- This change should only affect the Rancher Backups installation. Make sure different configurations of this chart will end up with a successful installation.

### Screenshot/Video

image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/chart/__tests__/rancher-backup-index.test.ts
+++ b/shell/chart/__tests__/rancher-backup-index.test.ts
@@ -2,9 +2,6 @@ import { shallowMount } from '@vue/test-utils';
 import RancherBackup from '@shell/chart/rancher-backup/index.vue';
 import { set } from '@shell/utils/object';
 
-// Mock the monitoringStatus utility
-jest.mock('@shell/utils/monitoring', () => ({ monitoringStatus: jest.fn(() => ({ monitoringStatus: () => ({ installed: false }) })) }));
-
 describe('rancher-backup: index', () => {
   const defaultMocks = {
     $store: {
--- a/shell/chart/rancher-backup/index.vue
+++ b/shell/chart/rancher-backup/index.vue
@@ -42,6 +42,16 @@ export default {
     }
   },
 
+  created() {
+    // Initialize monitoring before template renders to avoid undefined access
+    if (!this.value.monitoring) {
+      set(this.value, 'monitoring', {
+        metrics:        { enabled: false },
+        serviceMonitor: { enabled: false }
+      });
+    }
+  },
+
   async  fetch() {
     const hash = await allHash({
       storageClasses:    this.$store.dispatch('cluster/findAll', { type: STORAGE_CLASS }),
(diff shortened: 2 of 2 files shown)


shell/chart/__tests__/rancher-backup-index.test.ts
```diff
@@ -0,0 +1,251 @@
+import { shallowMount } from '@vue/test-utils';
+import RancherBackup from '@shell/chart/rancher-backup/index.vue';
+import { set } from '@shell/utils/object';
+
+// Mock the monitoringStatus utility
+jest.mock('@shell/utils/monitoring', () => ({ monitoringStatus: jest.fn(() => ({ monitoringStatus: () => ({ installed: false }) })) }));
```
codyrancher: It looks like this isn't ever actually exercised. I removed it locally and the tests still passed.

shell/chart/rancher-backup/index.vue
```diff
@@ -49,6 +52,13 @@ export default {
     this.persistentVolumes = hash.persistentVolumes;
 
     this.storageSource = this.getStorageSource(this.value) || 'none';
+
+    if (!this.value.monitoring) {
+      set(this.value, 'monitoring', {
+        metrics:        { enabled: false },
+        serviceMonitor: { enabled: false }
+      });
+    }
```
codyrancher: I didn't notice an actual issue but this seems to imply that `value.monitoring` may not exist but it appears to be used in the template i,e. `value.monitoring.serviceMonitor.enabled`. 

This reference in the template could cause problems during the initial render prior to fetch() being executed.

If there's a concern I think we'd want to move the initialization code. But again, I didn't see an issue in practive.

shell/chart/__tests__/rancher-backup-index.test.ts
momesgin: removed

shell/chart/rancher-backup/index.vue
momesgin: good point, moved to created block

PR #17031, opened 2026-03-26
Handle multiple sources for one extension

### LINKED ISSUE #16228: Wrong extension version installed when extension is present from two sources
**Setup**
- Rancher version: 2.13
- Rancher UI Extensions: ali-ui

**Describe the bug**

When extension has two cards, coming from different repos, the wrong extension might get installed

**To Reproduce**

Have 2 repos: 
1. from github using gh-pages with version 1.0.1-rc (has to be higher than the other one)
2. one from ui-plugin catalog with version 1.0.0 (lower than the other)
3.  click on the card with lower version

**Result**
RC extension gets installed

**Expected Result**

The selected extension should be installed

**Screenshots**

https://github.com/user-attachments/assets/9f595eb7-9c55-413b-bd48-3af651da5d45

**Additional context**

### LINKED ISSUE #16935: Display the source repo on extension cards

Extensions can come from different repositories. This can lead into scenarios where you have multiple identical cards. We should address this issue by displaying a repository icon and the repo name next to it, like what we do on the cards in Explorer > Apps > Charts.

We should also display this new repository info on the right-side drawer after you click on an extension card.

Image

We only display the repository info if it's available.

momesgin:
### Summary
Fixes #16935 
Fixes #16228 

- Display repository name on extension cards even when the original repository is removed
- When an extension is installed from one repository and the user tries to install the same extension from a different repository, a new modal is now shown prompting them to uninstall the existing version first
- Created new `UninstallExistingExtensionDialog.vue` component for handling this scenario

### Technical notes summary

- The `available` computed property in `shell/pages/c/_cluster/uiplugins/index.vue` now creates standalone items for installed plugins when the original repository is removed, rather than marking a fallback chart from another repo as "installed"
- Original repository name is extracted from Helm app annotations (catalog.cattle.io/ui-source-repo or catalog.cattle.io/cluster-repo-name) and stored in `originalRepoName`
- `showInstallDialog` method now checks if the same extension is already installed from a different source and shows `UninstallExistingExtensionDialog` instead of `InstallExtensionDialog`
- Added shared extension-dialog mixin in `shell/assets/styles/base/_mixins.scss` used by `InstallExtensionDialog`, `UninstallExtensionDialog`, and `UninstallExistingExtensionDialog`
- Added unit tests

1. Install an extension from Repository A, then remove Repository A - verify the extension card still shows the original repo name and displays correctly in the Installed tab
2. Add two repositories that provide the same extension, install from one, then try to install from the other - verify the new "already installed from another source" modal appears
3. In the new modal, click "Uninstall existing version" - verify the extension is uninstalled correctly
4. Verify extensions from other repositories still appear in the Available tab after installing an extension with the same name from a different repo
5. Test upgrade/downgrade flows still work correctly for installed extensions

- Extensions page - particularly the Available and Installed tabs filtering logic
- Extension installation, upgrade, and downgrade flows
- Extension uninstallation flow

### Screenshot/Video

https://github.com/user-attachments/assets/59b4145a-35ec-4a5f-8bb3-357ba7755f55

UPDATE:
addressing the wrong capitalization styling for repo names
image

### Checklist
- [x] The PR has a Milestone 


shell/pages/c/_cluster/apps/charts/AppChartCardFooter.vue
```diff
@@ -162,6 +162,7 @@ function getTooltip(key: string, fallback?: string): string | undefined {
       overflow: hidden;
       text-overflow: ellipsis;
       white-space: nowrap;
+      text-transform: capitalize;
       min-width: 0;
```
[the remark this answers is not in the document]
momesgin: that's an intentional design decision.

shell/pages/c/_cluster/uiplugins/index.vue
```diff
@@ -384,20 +404,42 @@ export default {
...
+            installedVersion:    p.version,
             installing:          false,
             builtin:             false,
             uiplugin:            p,
+            primeOnly:           appAnnotations[CATALOG_ANNOTATIONS.PRIME_ONLY] === 'true',
+            experimental:        appAnnotations[CATALOG_ANNOTATIONS.EXPERIMENTAL] === 'true',
+            certified:           appAnnotations[CATALOG_ANNOTATIONS.CERTIFIED] === CATALOG_ANNOTATIONS._RANCHER,
+            // Store the original repo name from the Helm app so we can display it even after the repo is removed
+            originalRepoName:    originalRepoDisplayName,
           };
```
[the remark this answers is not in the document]
momesgin: done

shell/pages/c/_cluster/uiplugins/index.vue
```diff
@@ -595,6 +663,36 @@ export default {
...
+              closed: (res) => {
+                if (res?.uninstalled) {
+                  this.didUninstall(res.plugin);
+                }
+              }
+            }
+          });
+
+          return;
+        }
```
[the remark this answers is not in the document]
momesgin: done

shell/pages/c/_cluster/uiplugins/index.vue
```diff
@@ -100,6 +101,12 @@ export default {
       hash.helmOps = this.$store.dispatch('management/findAll', { type: CATALOG.OPERATION });
     }
 
+    // Load CATALOG.APP to cross-reference installed plugins with their source repositories
+    // This allows us to determine from which repo an extension was installed if the same extension name exists in multiple repos
+    if (this.$store.getters['management/schemaFor'](CATALOG.APP)) {
+      hash.apps = this.$store.dispatch('management/findAll', { type: CATALOG.APP });
```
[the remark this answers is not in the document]
momesgin: done

shell/dialog/UninstallExistingExtensionDialog.vue
```diff
@@ -0,0 +1,145 @@
...
+        await plugin.uiplugin.remove();
+      }
+
+      // Find the app for this plugin
+      const apps = await this.$store.dispatch('management/findAll', { type: CATALOG.APP });
+
+      const pluginApp = apps.find((app) => {
+        return app.namespace === UI_PLUGIN_NAMESPACE && app.name === plugin.name;
+      });
+
```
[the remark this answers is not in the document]
momesgin: done

shell/pages/c/_cluster/uiplugins/index.vue
```diff
@@ -595,6 +657,35 @@ export default {
...
 
+      // Check if a plugin with the same name is already installed from a different repo
+      if (action === 'install') {
+        const installedPlugin = this.available.find((p) => p.name === plugin.name && p.installed);
+        const isInstalledFromDifferentSource = !!installedPlugin && installedPlugin.id !== plugin.id;
+
+        if (isInstalledFromDifferentSource) {
+          // Show a different dialog that prompts the user to uninstall the existing version first
+          this.$store.dispatch('management/promptModal', {
+            component:                            'UninstallExistingExtensionDialog',
```
aalves08, line 669: can we add this modal to the promptModal tests in https://github.com/rancher/dashboard/blob/master/shell/components/__tests__/PromptModal.test.ts please?

shell/dialog/UninstallExtensionDialog.vue
```diff
@@ -65,7 +65,7 @@ export default {
 
       const plugin = this.plugin;
 
-      this.updateStatus(plugin.name, 'uninstall');
+      this.updateStatus(plugin.id, 'uninstall');
```
aalves08, line 68: what's different between `id` and `name`?

shell/pages/c/_cluster/uiplugins/index.vue
```diff
@@ -100,6 +100,14 @@ export default {
       hash.helmOps = this.$store.dispatch('management/findAll', { type: CATALOG.OPERATION });
     }
 
+    // Load apps in UI_PLUGIN_NAMESPACE to determine which repo an extension was installed from
+    if (this.$store.getters['management/schemaFor'](CATALOG.APP)) {
```
aalves08, line 105: How's the resilience of the code/feature if the user doesn't have permissions to fetch `CATALOG.APP`? Did you test it?

shell/pages/c/_cluster/uiplugins/index.vue
```diff
@@ -595,6 +657,35 @@ export default {
       ev?.preventDefault?.();
       ev?.stopPropagation?.();
 
+      // Check if a plugin with the same name is already installed from a different repo
+      if (action === 'install') {
+        const installedPlugin = this.available.find((p) => p.name === plugin.name && p.installed);
```
aalves08, line 663: Shouldn't we be matching here from `id`?

oboc-sts:
Thanks for pointing these @aalves08 ... I'm trying to go through them but I'm having some issues:

> In the case of `Rancher Extensions`, uppercase looks good to me, but for `Elemental-Repo` and others it seems weird... Just a personal opinion but I would prefer to see `elemental-repo` or `ui-plugin-catalog-charts`. This last one could be "translated" to `Image Catalog` or something and fallback to `ui-plugin-catalog-charts`.

I... am not aware of having any case-sensitive formatting applied there, or I don't recall spec'ing something like this 😁 . Not for the extension, nor for the chart cards. @momesgin am I missing something here?

> I got the extension installed, but these two cards don't display anything (warning?) in regards to being an extension with the same ID already installed... Don't know if you guys covered this scenario or not but would love to hear your opinion on it 🙏

I see the screenshot but I have no idea which extensions you're talking about and what is the warning message you're missing? In other words, I need a bit of context / more details around this. Are you sure this is a matter related to the component (the card) itself and not to the installation process that should've checked if there is an extension already installed with that ID?

aalves08:
> Thanks for pointing these @aalves08 ... I'm trying to go through them but I'm having some issues:
> 
> > In the case of `Rancher Extensions`, uppercase looks good to me, but for `Elemental-Repo` and others it seems weird... Just a personal opinion but I would prefer to see `elemental-repo` or `ui-plugin-catalog-charts`. This last one could be "translated" to `Image Catalog` or something and fallback to `ui-plugin-catalog-charts`.
> 
> I... am not aware of having any case-sensitive formatting applied there, or I don't recall spec'ing something like this 😁 . Not for the extension, nor for the chart cards. @momesgin am I missing something here?
> 
> > I got the extension installed, but these two cards don't display anything (warning?) in regards to being an extension with the same ID already installed... Don't know if you guys covered this scenario or not but would love to hear your opinion on it 🙏
> 
> I see the screenshot but I have no idea which extensions you're talking about and what is the warning message you're missing? In other words, I need a bit of context / more details around this. Are you sure this is a matter related to the component (the card) itself and not to the installation process that should've checked if there is an extension already installed with that ID?

@oboc-sts sorry for the lack of context, I tested with `Elemental` extension. The cards are duplicated (even triplicated, but that third one cannot be seen because it's that third one is in the `installed` tab)

shell/dialog/UninstallExtensionDialog.vue
momesgin: The `id` is what we return in `available()` computed property in `shell/pages/c/_cluster/uiplugins/index.vue`.

If you check the logic inside `available()` you'll see in different cases we return different formats of ids. For example, if an extension is chart based, we get the id from `chart.id`(e.g. 'repo-type/repo-name/chart-name') which is generated by `generateKey` in `shell/store/catalog.js`.

In another case for example, on line 423 in `shell/pages/c/_cluster/uiplugins/index.vue` we format the id like this:
`id: `${ p.name }-${ p.version }`.

This inconsistency for the ids is totally fine with my duplicate detection logic, but we could try to standardize them a bit to make it less confusing.

At the end of the day the `id` is unique but the `name` is not.

shell/pages/c/_cluster/uiplugins/index.vue
momesgin: good point, no I haven't yet.

shell/pages/c/_cluster/uiplugins/index.vue
momesgin: no, we're intentionally looking for **any** installed extension with the **same name** regardless of which repo it came from. The goal is to find if an extension is already installed (from any source), so we can then check if it was installed from a different source.

The next line does the id comparison:
```
const isInstalledFromDifferentSource = !!installedPlugin && installedPlugin.id !== plugin.id;
```
so basically the logic is:
1. Find any installed plugin with the same name
2. Check if its id differs from the one the user is trying to install

oboc-sts:
They seem duplicated, but I see they come from different repos. So they are not in fact identical. Or are they?

I'm trying to better define the problem we're trying to solve here but it's still not clear to me.

Maybe it's something I'm missing but having conflicts (in general, not necessarily for extensions) of this type (having extensions with the same ID, installing extensions from the same source, etc) is something that is (should be) always treated during the process (of adding, installing, etc) and doesn't have anything to do with the components being used in the end result.

In other words, the warnings, errors and other conflictual use cases should be accounted for at "design time" (e.g. before the extension actually gets installed / added / etc), not "runtime" (after the extension is already installed / added / etc). The micro version of this would be:  act proactively (try to prevent having identical extensions installed), not retroactively (passively inform the use there are identical extensions installed).

Maybe having a call with you and @momesgin  would clarify the whole case a bit more.

momesgin:
> I... am not aware of having any case-sensitive formatting applied there, or I don't recall spec'ing something like this 😁 . Not for the extension, nor for the chart cards. @momesgin am I missing something here?

@oboc-sts I thought it was by design, but I might've misunderstood that. Will address that along with this PR.

Re the other question from @aalves08, I have the same understanding as you do.

shell/dialog/UninstallExtensionDialog.vue
aalves08: Let's keep the `id`. Thanks for clarifying it 🙏

aalves08:
@oboc-sts 

> They seem duplicated, but I see they come from different repos. So they are not in fact identical. Or are they?
> 
> I'm trying to better define the problem we're trying to solve here but it's still not clear to me.
> 
> Maybe it's something I'm missing but having conflicts (in general, not necessarily for extensions) of this type (having extensions with the same ID, installing extensions from the same source, etc) is something that is (should be) always treated during the process (of adding, installing, etc) and doesn't have anything to do with the components being used in the end result.

> 
> In other words, the warnings, errors and other conflictual use cases should be accounted for at "design time" (e.g. before the extension actually gets installed / added / etc), not "runtime" (after the extension is already installed / added / etc). The micro version of this would be: act proactively (try to prevent having identical extensions installed), not retroactively (passively inform the use there are identical extensions installed).
> 
> Maybe having a call with you and @momesgin would clarify the whole case a bit more.

I am fine with that path @oboc-sts. Just wanted to raise it so that we would consider all angles, that's it. You are the UX man so I completely trust your opinion. As long as the experience is aligned with your expert analysis, I am fine with it 🙏

oboc-sts:
Thanks, Alex. I'll take it with @momesgin

oboc-sts:
Just had a call with @momesgin and:
1. the case sensitive thingie used on the extension cards will be fixed
2. the warnings / flags / etc on the card BEFORE the installation are NOT needed.

aalves08:
> Just had a call with @momesgin and:
> 
> 1. the case sensitive thingie used on the extension cards will be fixed
> 2. the warnings / flags / etc on the card BEFORE the installation are NOT needed.

Totally in agreement 🙏  Thanks for looking into this

shell/pages/c/_cluster/uiplugins/index.vue
momesgin: hmm.. the logic for detecting duplicates heavily relies on this permission. Without this we would still choose the wrong extension to install as mentioned in the issue's description.

I'm seeing the same approach for other cases:
```
if (this.$store.getters['management/schemaFor'](UI_PLUGIN)) {...
```
```
if (this.$store.getters['management/schemaFor'](MANAGEMENT.CLUSTER)) {...
```
```
if (this.$store.getters['management/schemaFor'](CATALOG.OPERATION)) {...
```
@aalves08 I need your help to understand how likely it is to have users without a permission to `CATALOG.APP` or the other ones?

momesgin: @richard-cox I'm tagging you hoping you might be able to help since Alex will be on vacation

shell/pages/c/_cluster/uiplugins/index.vue
momesgin: done

shell/pages/c/_cluster/uiplugins/index.vue
aalves08: @momesgin I've just checked that the product registration for the extensions install/uninstall page is gated to `admin` as per https://github.com/rancher/dashboard/blob/master/shell/config/product/uiplugins.js#L10, unless someone is given very specific permissions, but I wouldn't worry about that for now imho.

aalves08:
@momesgin is this ready for approval?

shell/pages/c/_cluster/uiplugins/index.vue
momesgin: sounds good, thanks Alex!

momesgin:
> @momesgin is this ready for approval?

yes

UPDATE: no! I forgot to address the wrong styling for the repo names, but now it's fixed. So now... YES!

PR #17032, opened 2026-03-26
Automation: Stabilize OIDC provider tests

yonasberhe23:
### Summary
Fixes #https://github.com/rancher/qa-tasks/issues/2231

Updates the OIDC provider tests so they navigate to the OIDC Apps list page and navigate into the detail page from the resource table instead of using goTo() on the detail URL. Deep-linking to the detail page was tied to flaky behavior for the full-secret copy row (load-order / snapshot timing).

### Screenshot/Video
image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/cypress/e2e/po/pages/users-and-auth/oidc-client.po.ts
+++ b/cypress/e2e/po/pages/users-and-auth/oidc-client.po.ts
@@ -22,7 +22,7 @@ export default class OidcClientsPagePo extends PagePo {
     const sideNav = new ProductNavPo();
 
     BurgerMenuPo.burgerMenuNavToMenubyLabel('Users & Authentication');
-    sideNav.navToSideMenuGroupByLabel('Client Application');
+    sideNav.navToSideMenuEntryByLabel('OIDC Apps');
   }
 
   createOidcClient() {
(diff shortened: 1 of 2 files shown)


izaac:
Can we start adding the `qase()` wrapper mapping to the test id in our test case management? 
Example: https://github.com/rancher/dashboard/pull/17000/changes#diff-9bc5a7fefd85c9ae12c963ee728f5d8bfd586f806e09e88be8bb674af996c76cR22

yonasberhe23:
> Can we start adding the `qase()` wrapper mapping to the test id in our test case management? Example: https://github.com/rancher/dashboard/pull/17000/changes#diff-9bc5a7fefd85c9ae12c963ee728f5d8bfd586f806e09e88be8bb674af996c76cR22

@izaac yes, done. For the qase(<id>, it(...)) wrappers in Cypress: do the numeric ids need to match real case ids in our Qase project, or can I use any sequence (e.g., continuing from the version.spec.ts example)? I’m trying to understand what breaks or misreports if the id doesn’t exist or doesn’t match the test in Qase.

izaac:
@yonasberhe23 yes they have to match an existing TC, the behavior right now won't create new ones if they don't have an ID, I didn't dig in that specific use case where the ID doesn't match any TC in the Qase documentation to check how the Qase plugin behave, it might just ignore it as it first decorates the title and then process it in the backend.

izaac (dismissed):
👍🏽

PR #17051, opened 2026-03-27
fix issue with `getUIConfig` and missing extension points in older Rancher versions

### LINKED ISSUE #17047: Fix singleton `getUIConfig` issue
**Setup**
- Rancher version: reproduced with `2.13.2`
- Rancher UI Extensions: Extension with shell `3.0.10` and a custom list view
- Browser type & version: n/a

**Describe the bug**

In Dashboard we introduced ExtensionPoint.TABLE = 'Table' in @rancher/shell `3.0.9`, but the Rancher `2.13` dashboard's extension manager singleton was never rebuilt with that enum value.

Dashboard creates the singleton via `createExtensionManager()` in `extension-manager-impl.js`. At [initialisation](https://github.com/rancher/dashboard/blob/master/shell/core/extension-manager-impl.js#L29-L31), it loops over `ExtensionPoint` and creates `uiConfig` keys — but the `2.13` Dashboard's `ExtensionPoint` enum doesn't include 'Table'.

The extension bundles @rancher/shell `3.0.10` which does have `ExtensionPoint.TABLE`. When Dashboard components call `getApplicableExtensionEnhancements()` from `plugin-helpers.ts`, it calls `$extension.getUIConfig('Table', uiArea)` on the singleton at [`extension-manager-impl.js`](https://github.com/rancher/dashboard/blob/master/shell/core/extension-manager-impl.js#L430-L432)

```
return uiConfig[type][uiArea] || [];

which throws the errror: uiConfig['Table'] is undefined
```

Biggest problem is that even though `ExtensionPoint.TABLE` exists in this new shell, the Dashboard uses its own bundled singleton, not the extension's/shell copy, so it doesn't recognise it and therefore throws the error.

This brings increased difficulty in resolving this matter Shell side

**To Reproduce**

- provision a Rancher 2.13 system
- build an extension with a custom list view with shell `3.0.10` locally
- do `yarn serve-pkgs` and do a developer load
- go to the custom list view page of the extension and observe the browser console

**Result**
Throws a console error

**Expected Result**

Should not throw a console error

**Screenshots**

Image

**Additional context**

aalves08:
### Summary
Fixes #17047 

- fix issue with `getUIConfig`
- add unit tests 
- bump shell version

### Technical notes summary

As per described on the issue:

In Dashboard we introduced ExtensionPoint.TABLE = 'Table' in @rancher/shell `3.0.9`, but the Rancher `2.13` dashboard's extension manager singleton was never rebuilt with that enum value.

Dashboard creates the singleton via `createExtensionManager()` in `extension-manager-impl.js`. At [initialisation](https://github.com/rancher/dashboard/blob/master/shell/core/extension-manager-impl.js#L29-L31), it loops over `ExtensionPoint` and creates `uiConfig` keys — but the `2.13` Dashboard's `ExtensionPoint` enum doesn't include 'Table'.

The extension bundles @rancher/shell `3.0.10` which does have `ExtensionPoint.TABLE`. When Dashboard components call `getApplicableExtensionEnhancements()` from `plugin-helpers.ts`, it calls `$extension.getUIConfig('Table', uiArea)` on the singleton at [`extension-manager-impl.js`](https://github.com/rancher/dashboard/blob/master/shell/core/extension-manager-impl.js#L430-L432)

Biggest problem is that even though `ExtensionPoint.TABLE` exists in this new shell, the Dashboard uses its own bundled singleton, not the extension's/shell copy, so it doesn't recognise the extension point and therefore throws the error.

The solution is to patch the uiConfig with the first intervention of shell when trying to interpret an extension point, so that it won't throw any error for an extension point not available in a given Rancher version. With this we ensure that extension points added in the future won't break extensions in older Rancher systems.

Also fool-proofed the method with the issue with the missing optional chaining and added unit tests.

- Provision Rancher 2.13.2
- Get a copy of the `supportability-review-app` extension in local machine https://github.com/rancher/supportability-review-app
- yarn link the shell from this PR with `supportability-review-app` extension, branch `main`
- do a local build `yarn build-pkg supportability-review-app`
- serve it locally `yarn serve-pkgs`
- do a developer load on Rancher 2.13.2
- Go to Supportability top-level-product -> Review Bundles
- Make sure there are no errors on the console

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/core/plugin-helpers.ts
+++ b/shell/core/plugin-helpers.ts
@@ -137,35 +137,27 @@ function checkExtensionRouteBinding($route: any, locationConfig: any, context: a
   return res;
 }
 
-// Backfill any ExtensionPoint keys missing from the extension manager's uiConfig.
-// getAllUIConfig() returns the singleton's closure variable by reference, so mutating
-// it here permanently patches the singleton for ALL callers (extensions AND dashboard).
+// Track which ExtensionPoint keys are missing from the extension manager's uiConfig.
 // This handles forwards-compatibility when extensions ship a newer shell that defines
 // ExtensionPoint values the running dashboard doesn't know about (e.g. 'Table' on 2.13).
-let _uiConfigPatched = false;
+const _uiConfigPatched: { [point: string]: boolean } = {};
 
 function ensureUIConfigCompat(extensionManager: any) {
-  if (_uiConfigPatched) {
-    return;
-  }
-
   const uiConfig = extensionManager.getAllUIConfig?.();
 
   if (uiConfig) {
-    const missingPoints: string[] = [];
+    const missingPoints: { [point: string]: boolean } = {};
 
     Object.values(ExtensionPoint).forEach((ep) => {
-      if (!uiConfig[ep]) {
-        uiConfig[ep] = {};
-        missingPoints.push(ep);
+      if (!uiConfig[ep] && !_uiConfigPatched[ep]) {
+        missingPoints[ep] = true;
+        _uiConfigPatched[ep] = true;
       }
     });
 
-    if (missingPoints.length) {
-      console.warn(`[plugin-helpers] These ExtensionPoints aren't available for usage in this Rancher version: ${ missingPoints.join(', ') }`); // eslint-disable-line no-console
+    if (Object.keys(missingPoints).length) {
+      console.warn(`[plugin-helpers] These ExtensionPoints aren't available for usage in this Rancher version: ${ Object.keys(missingPoints).join(', ') }`); // eslint-disable-line no-console
     }
-
-    _uiConfigPatched = true;
   }
 }
 
@@ -181,6 +173,11 @@ export function getApplicableExtensionEnhancements<T>(
   // gate it so that we prevent errors on older versions of dashboard
   if (pluginCtx.$extension?.getUIConfig) {
     ensureUIConfigCompat(pluginCtx.$extension);
+
+    // Exit early if actionType doesn't exist in the extension manager's uiConfig
+    if (_uiConfigPatched[actionType]) {
+      return [];
+    }
 
     const actions = pluginCtx.$extension.getUIConfig(actionType, uiArea);
 
--- a/shell/core/__tests__/extension-manager-impl.test.js
+++ b/shell/core/__tests__/extension-manager-impl.test.js
@@ -488,8 +488,8 @@ describe('extension Manager', () => {
   });
 
   describe('ensureUIConfigCompat (via plugin-helpers)', () => {
-    // These tests verify that getApplicableExtensionEnhancements backfills
-    // missing ExtensionPoint keys into the singleton's uiConfig via getAllUIConfig().
+    // These tests verify that getApplicableExtensionEnhancements tracks missing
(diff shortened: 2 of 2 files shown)


shell/core/plugin-helpers.ts
```diff
@@ -148,6 +180,8 @@ export function getApplicableExtensionEnhancements<T>(
 
   // gate it so that we prevent errors on older versions of dashboard
   if (pluginCtx.$extension?.getUIConfig) {
+    ensureUIConfigCompat(pluginCtx.$extension);
```
rak-phillip, line 175: My major concern about this approach is that we are patching a singleton inside of a "get" function, introducing side-effects that might not be anticipated. The impact of this change looks very low to me, but I would like for us to consider if this could be done in `shell/core/extension-manager-impl.js`. Conceptually, this is where we manage the component initialization, but I could be missing some key lifecycle detail where this wouldn't work. 

I think that this won't be possible because of this line from the PR description:

> Biggest problem is that even though ExtensionPoint.TABLE exists in this new shell, the Dashboard uses its own bundled singleton, not the extension's/shell copy, so it doesn't recognise the extension point and therefore throws the error.

What do you think?

Otherwise, this change looks good.

aalves08: It's not possible because old Dashboard's don't have `extension-manager-impl`, which was only brought in with Rancher `2.13`. I also tried it after your suggestion and I confirmed that it doesn't work.

I also know for a fact that extensions use their own singleton, so possibly the only entry point that comes from a shell version is `plugin-helpers`, which is needed in order to fix this problem without patching older Rancher systems.

We are pretty hands-tied with this one, imho.

Let's wait for @richard-cox feedback

richard-cox: agree it would be better in `extension-manager-impl.js`, agree that wouldn't fix the core issue.

i think we can improve this though. if `actionType` doesn't exist in `extensionManager.getAllUIConfig?.()`, couldn't we just exit `getApplicableExtensionEnhancements` early?

- change _uiConfigPatched to `{ [point: string]: missing }`
- skip `uiConfig[ep] = {}`
- change `missingPoints.push(ep)` to `missingPoints[ep] = true`
- after call to `ensureUIConfigCompat` `if missingPoints[actionType]` exit early

aalves08: `ensureUIConfigCompat(pluginCtx.$extension);` runs and the very next line is getting the actions applicable:

`const actions = pluginCtx.$extension.getUIConfig(actionType, uiArea);` 

which is going to a couple of nested keys in an object, accessing it directly. No more checks here, no iterations.

If it's trying to get the type that we just patched, then the actions is always an empty array, therefore not iterating through all of that code from L187 and onwards.

Is it really worth changing the PR again for a 1 extra line that it runs without iterations or any other "expensive" operations?

richard-cox: it cuts out the mutation

aalves08: Now I get it... Wasn't seeing that. Good point

rak-phillip (approved):
The enhancement looks good and the issue remains resolved with the latest changes.

PR #17056, opened 2026-03-27
Automation: Skip chart e2e tests when charts are filtered by catalog

yonasberhe23:
### Summary
Fixes #https://github.com/rancher/qa-tasks/issues/2255, https://github.com/rancher/qa-tasks/issues/2256, https://github.com/rancher/qa-tasks/issues/2257, https://github.com/rancher/qa-tasks/issues/2258, https://github.com/rancher/qa-tasks/issues/2259, https://github.com/rancher/qa-tasks/issues/2260, https://github.com/rancher/qa-tasks/issues/2261, https://github.com/rancher/qa-tasks/issues/2264

Forward port of https://github.com/rancher/dashboard/pull/17008

#### Sample results:

Community build (head image):
image

Prime build (2.15.0-alpha1 image):
image

### Checklist
- [x] The PR has a Milestone 


cypress/support/commands/rancher-api-commands.ts
```diff
@@ -1445,3 +1466,107 @@ Cypress.Commands.add('applyDefaultTestTheme', () => {
...
+) {
+  cy.checkChartPresence(repo, chartKey).then(({ inFiltered, inUnfiltered }) => {
+    if (!inUnfiltered) {
+      throw new Error(`Chart '${ chartKey }' is missing from the unfiltered catalog index. This looks like a publishing/sync issue, not due to version compatibility filtering.`);
+    }
+
+    if (!inFiltered) {
+      cy.log(`Skipping: chart '${ chartKey }' is intentionally hidden from the filtered catalog index`);
+
+      return mochaContext.skip();
```
richard-cox: Won't this mean we silently skip the functionality we need to test if the chart is missing?

cypress/support/commands/rancher-api-commands.ts
```diff
@@ -1445,3 +1466,107 @@ Cypress.Commands.add('applyDefaultTestTheme', () => {
...
+    }
+
+    callback();
```
richard-cox, line 1600: Not quite sure it's worth the refactor, but generally the callback pattern was replaced with promise chains a while back

cypress/support/commands/rancher-api-commands.ts
yonasberhe23: We only skip when the chart is present in the unfiltered index (skipFilter=true) but not in the filtered index - i.e. Rancher is intentionally hiding it from the catalog the UI uses (e.g. chart annotations / compatibility rules). In that case the UI wouldn’t show the chart either, so running the test would be a false failure, not a real regression.

If the chart is missing from the unfiltered index (publish/sync/environment issue), we throw an error and the test fails - we do not skip that case.

So we are not skipping because “the chart is missing”; we skip because “the chart exists in the raw index but is filtered out for this cluster/Rancher version,” which is the scenario we wanted to stop failing in CI.

cypress/support/commands/rancher-api-commands.ts
yonasberhe23: Good call on promise chains being the more common pattern elsewhere. Happy to refactor this helper for consistency in a follow-up: https://github.com/rancher/qa-tasks/issues/2292

cypress/support/commands/rancher-api-commands.ts
richard-cox: There's a few weird angles

My Q: we have functionality we want to test using a chart, if that chart isn't there we won't be testing that functionality. if the functionality is specifically about that chart that we can silently skip, but if not it should still fail (because we won't be testing something we need to test).

For example cypress/e2e/tests/pages/explorer2/cluster-tools.spec.ts tests the install/edit/delete functionality of installing charts in the cluster/tools page, however if rancher-alerting-drivers chart is missing we'll silently skip and not test that feature.

There's also a caveat as well - `head` builds will normally return all charts, which i think is the same as the `skipFilter` filter. i don't think that changes this though

yonasberhe23: You’re right — that's the tradeoff. In the cluster-tools.spec.ts example, those tests use `rancher-alerting-drivers` to exercise install/edit/delete. If that chart isn’t shown in the filtered catalog for that environment the test used to fail and we’d spend time chasing what looked like regressions but were really catalog filtering / compatibility rules. This change is meant to avoid those false failures, so QA isn’t burning time on investigations that aren’t real dashboard bugs.

For reference, the backend behavior around filtered vs unfiltered index is in [content.go](https://github.com/rancher/rancher/blob/1f31eaa3cea9cd1223debc430cc01ea0b2522280/pkg/catalogv2/content/content.go#L457)

yonasberhe23: I considered swapping to another chart, but any chart can be filtered out in a future release when catalog rules change, so that doesn’t feel like a stable long-term fix either. The skip path is meant to handle “filtered by design” without treating it as a dashboard failure; I’m open to other ideas if you have a better solution.

richard-cox: Can you have a flag that only silently skips the test when in QA world?

yonasberhe23: I like that idea, yes, we can with an env flag. I'll work on that

yonasberhe23: https://github.com/rancher/dashboard/pull/17056/commits/09eae56c0620dc1c987536735ffa6c654261fa15

yonasberhe23: @richard-cox when you have a moment, could you take another look?

izaac (approved):
:shipit:

a-arias (approved):

PR #17064, opened 2026-03-30
[16714][fleet] Change to only show workspace selector on list screen

### LINKED ISSUE #16714: Make Fleet cluster context more obvious

I was confused when a Git Repo I configured was not visible under the Resources sub menu, until I finally noticed that there was a Workspace context drop down menu at the top (with another Workspace selected). It is not obvious that the menu context may be different from the workspace currently being viewed.

Having the Workspace context automatically change to the currently viewed Workspace might help. Having the Git Repo, Helm Ops, Clusters, and Cluster Groups tiles clickable might also help (linking to their corresponding Resource view in the right Workspace context).

marcelofukumoto:
### Summary
Fixes #16714

- Changed the workspace selectors to be disabled on create/edit/view pages
- It only can be changed on list screens
- On the workspace it doesnt show
- Added a click on the card to navigate to the proper page and interact better with workspace selector

### Technical notes summary

- The check is done by the name id

- Fleet screens
- Views/Create/Edit should not show
- List should show
- List for Workspaces should not show

### Screenshot/Video
image
image
image
image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/nav/Header.vue
+++ b/shell/components/nav/Header.vue
@@ -1,7 +1,7 @@
 <script>
 import { mapGetters } from 'vuex';
 import debounce from 'lodash/debounce';
-import { FLEET, MANAGEMENT, NORMAN, STEVE } from '@shell/config/types';
+import { MANAGEMENT, NORMAN, STEVE } from '@shell/config/types';
 import { HARVESTER_NAME as HARVESTER } from '@shell/config/features';
 import { ucFirst } from '@shell/utils/string';
 import { isAlternate, isMac } from '@shell/utils/platform';
@@ -190,17 +190,9 @@ export default {
       // Don't show if the header is in 'simple' mode
       const notSimple = !this.simple;
       // One of these must be enabled, otherwise there's no component to show
-      const showWorkspace = this.currentProduct?.showWorkspaceSwitcher && !this.isWorkspacesPage;
-      const validFilterSettings = this.currentProduct?.showNamespaceFilter || showWorkspace;
+      const validFilterSettings = this.currentProduct?.showNamespaceFilter || this.showWorkspaceSwitcher;
 
       return validClusterOrProduct && notSimple && validFilterSettings;
-    },
-
-    /**
-     * Whether the current route is the Workspaces list page.
-     */
-    isWorkspacesPage() {
-      return this.$route?.params?.resource === FLEET.WORKSPACE;
     },
 
     /**
@@ -583,7 +575,7 @@ export default {
       >
         <NamespaceFilter v-if="clusterReady && currentProduct && (currentProduct.showNamespaceFilter || isExplorer)" />
         <WorkspaceSwitcher
-          v-else-if="clusterReady && currentProduct && currentProduct.showWorkspaceSwitcher && showWorkspaceSwitcher"
+          v-else-if="clusterReady && showWorkspaceSwitcher"
           :disabled="disableWorkspaceSwitcher"
         />
       </div>
--- a/shell/components/nav/Header.vue
+++ b/shell/components/nav/Header.vue
@@ -1,7 +1,7 @@
 <script>
 import { mapGetters } from 'vuex';
 import debounce from 'lodash/debounce';
-import { MANAGEMENT, NORMAN, STEVE } from '@shell/config/types';
+import { FLEET, MANAGEMENT, NORMAN, STEVE } from '@shell/config/types';
 import { HARVESTER_NAME as HARVESTER } from '@shell/config/features';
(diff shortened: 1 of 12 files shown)


shell/components/nav/Header.vue
```diff
@@ -556,7 +582,10 @@ export default {
         class="top"
       >
         <NamespaceFilter v-if="clusterReady && currentProduct && (currentProduct.showNamespaceFilter || isExplorer)" />
-        <WorkspaceSwitcher v-else-if="clusterReady && currentProduct && currentProduct.showWorkspaceSwitcher && showWorkspaceSwitcher" />
+        <WorkspaceSwitcher
+          v-else-if="clusterReady && currentProduct && currentProduct.showWorkspaceSwitcher && showWorkspaceSwitcher"
```
torchiaf: Not part of this change - we can probably remove the `currentProduct && currentProduct.showWorkspaceSwitcher` condition as it is included in the getter:

```
  showWorkspaceSwitcher(state, getters) {
    const product = getters['currentProduct'];

    if (!product) {
      return false;
    }

    return product.showWorkspaceSwitcher && state.showWorkspaceSwitcher;
  },
```

shell/components/nav/Header.vue
```diff
@@ -189,12 +189,38 @@ export default {
...
+     * Only list pages should allow changing the workspace.
+     */
+    disableWorkspaceSwitcher() {
+      // Disable on detail/edit pages (route has an id param)
+      if (this.$route?.params?.id) {
+        return true;
+      }
+
+      // Disable on create pages (route names end with '-create')
+      if (this.$route?.name?.endsWith('-create')) {
```
torchiaf, line 209: This is a very nit thing so you can ignore it: this condition depends on how developers define the creation pages - not sure if there is a way to enforce this constraint

shell/components/nav/Header.vue
```diff
@@ -189,12 +189,38 @@ export default {
...
+      const validFilterSettings = this.currentProduct?.showNamespaceFilter || showWorkspace;
 
       return validClusterOrProduct && notSimple && validFilterSettings;
     },
 
+    /**
+     * Whether the current route is the Workspaces list page.
+     */
+    isWorkspacesPage() {
+      return this.$route?.params?.resource === FLEET.WORKSPACE;
```
torchiaf: Nit: in the fleet dashboard we are controlling the workspace switcher visibilty using the store:

```
  created() {
    this.$store.dispatch('showWorkspaceSwitcher', false);
    ...
```

Maybe we can use the same strategy in the workspace page and remove `isWorkspacesPage`.

torchiaf (dismissed):
LGTM - I added some comments about minor things - feel to ignore them and merge.

edenhernandez-suse:
Not 100% sure about this one. The behaviour of the current filter is messed up, but hiding it during the App Bundle creation means users won't know in which workspace they're creating the app bundle. They need to remember which one was selected before clicking "Create".

@kwwii any thoughts on this?

marcelofukumoto:
> Not 100% sure about this one. The behaviour of the current filter is messed up, but hiding it during the App Bundle creation means users won't know in which workspace they're creating the app bundle. They need to remember which one was selected before clicking "Create".
> 
> @kwwii any thoughts on this?

It is not hidden during AppBundle creation. It is just disabled.

It is only hidden on the WORKSPACE context.

shell/components/nav/Header.vue
marcelofukumoto: Applied the changes! Thanks!

shell/components/nav/Header.vue
marcelofukumoto: Applied the changes! Thanks!

marcelofukumoto:
@torchiaf could you review again?

kwwii:
I think it would be better to make the selector inactive in these situations rather than to remove it.

marcelofukumoto:
> I think it would be better to make the selector inactive in these situations rather than to remove it.

A yes. So now it is disabled or inactive.
image

Like on this one.

torchiaf (approved):
LGTM

marcelofukumoto:
@kwwii @edenhernandez-suse 

Can we get an approval? Let me know if there is any additional question.

Currently the flow is:
- Show | Disabled if it is on CREATE | EDIT | VIEW, any specific resource.
- Hide if it is on screen that doesnt have workspaces (only WORKSPACE and DASHBOARD)
  - on DASHBOARD is already not there, only on WORKSPACE it changed
- Show | Enabled on LIST screens

edenhernandez-suse (approved):

PR #17071, opened 2026-03-30
[16870][fleet] Change the targetClusters logic to consider nameDisplay as 2n…

### LINKED ISSUE #16870: [Fleet] Cluster selection by display name still not performing as expected
**Setup**
- Rancher version: 2.13.3
- Rancher UI Extensions: continuous integration, elemental, harvester
- Browser type & version: Chrome latest

**Describe the bug**
Looking for clarification on the resolution of this issue: https://github.com/rancher/dashboard/issues/16370
It does seem as though the originating issue for the backend selection was resolved.
However, now the UI does not properly reflect the backend logic.

In the GitRepo view, this example is shown as targeting 5 clusters via clusterName, where the clusterName is the display name. (ignore the error message, the 5 clusters is the important bit.)
Image

When inspecting the yaml, this is the list of 5 targets:
Image

When I navigate to the editor, the `Selected Clusters` does not match
Image

In this example, the `platform-services` cluster is a kubernetes cluster managed by the Harvester cluster provider. krum-development is an AKS cluster. krum-prod is an EKS cluster. harvester-lab is an actual Harvester cluster. dev-desktop is an imported cluster.

If I remove dev-desktop from the list, then re-add dev-desktop, it will appear in the Selected Clusters list:
Image

But, when inspecting the yaml, it is not using the clusterName as the selector, it is using the Rancher-generated cluster ID: 
Image

This indicated inconsistency of the matching logic between the (correct) cluster count versus the (incorrect) list of selected clusters.

(I believe the inconsistent use of the cluster metadata name between cluster providers to be a separate issue)

**To Reproduce**

1. Clickops create an Imported cluster with a name `foo`
2. Apply a GitRepo yaml with target configuration of 
  ```
  targets:
    - clusterName: foo
  ```
3. View GitRepo table, observe 1/1 clusters selected in the table
4. Edit/View GitRepo configuration with UI - observe `foo` is omitted from Selected Clusters

**Result**

**Expected Result**

ClusterName should select what appears to the user to be the Cluster Name.

**Screenshots**
in description

**Additional context**
n/a

marcelofukumoto:
…d option

### Summary
Fixes #16870

3 changed logics:

1. **`clustersOptions` uses `nameDisplay` as the value** - The cluster selector options now use `nameDisplay` for both `label` and `value`, so the UI consistently works with display names.
2. **`displayName` is resolved from `metadata.name` on load** - When loading existing targets from the YAML (via `fromTargets`), the `allClusters` watcher resolves any `metadata.name` values (e.g., `c-p2bmg`) in `selectedClusters` to their corresponding `nameDisplay` (e.g., `test-k3s`), so the `LabeledSelect` can match them against the options.
3. **Model `targetClusters` getters use `nameDisplay` lookup** - In `fleet-application.js`, the `findBy` uses a fallback: `findBy(clusters, 'metadata.name', tgt.clusterName) || findBy(clusters, 'nameDisplay', tgt.clusterName)`. In `fleet.cattle.io.bundle.js`, it uses `findBy(clusters, 'nameDisplay', tgt.clusterName)`. This ensures clusters are found regardless of whether the stored `clusterName` is a `metadata.name` or a `nameDisplay`.

### Technical notes summary

- displayName already considers not having the label for displayName and use the metadata.name or the id instead.

- Target clusters on AppBundle creation — select clusters, verify `nameDisplay` appears in the selector tags
- Target clusters on AppBundle edit — existing targets with `metadata.name` in the YAML should resolve to `nameDisplay` in the UI
- AppBundle resources list — the "targeted clusters" column should show the correct clusters
- Clusters with no `CLUSTER_DISPLAY_NAME` label — `nameDisplay` falls back to `metadata.name`, should still match correctly
- Clusters with a `CLUSTER_DISPLAY_NAME` label different from `metadata.name` — the display name should appear in the UI, and the resource should still resolve correctly

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/models/fleet-application.js
+++ b/shell/models/fleet-application.js
@@ -105,7 +105,7 @@ export default class FleetApplication extends SteveModel {
 
     for (const tgt of this.spec.targets) {
       if (tgt.clusterName) {
-        const cluster = findBy(clusters, 'metadata.name', tgt.clusterName);
+        const cluster = findBy(clusters, 'metadata.name', tgt.clusterName) || findBy(clusters, 'nameDisplay', tgt.clusterName);
 
         if (cluster) {
           addObject(out, cluster);
--- a/shell/models/fleet.cattle.io.bundle.js
+++ b/shell/models/fleet.cattle.io.bundle.js
@@ -47,7 +47,7 @@ export default class FleetBundle extends SteveModel {
 
     for (const tgt of this.spec.targets) {
       if (tgt.clusterName) {
-        const cluster = findBy(clusters, 'metadata.name', tgt.clusterName);
+        const cluster = findBy(clusters, 'metadata.name', tgt.clusterName) || findBy(clusters, 'nameDisplay', tgt.clusterName);
 
         if (cluster) {
           addObject(out, cluster);
--- a/shell/components/fleet/FleetClusterTargets/index.vue
+++ b/shell/components/fleet/FleetClusterTargets/index.vue
@@ -159,7 +159,7 @@ export default {
     clustersOptions() {
       return this.allClusters
         .filter((x) => x.metadata.namespace === this.namespace && !isHarvesterCluster(x))
-        .map((x) => ({ label: x.nameDisplay, value: x.metadata.name }));
+        .map((x) => ({ label: x.nameDisplay, value: x.nameDisplay }));
     },
 
     clusterGroupsOptions() {
--- a/shell/components/fleet/FleetClusterTargets/index.vue
+++ b/shell/components/fleet/FleetClusterTargets/index.vue
@@ -124,6 +124,15 @@ export default {
         this.update();
       }
     },
+
+    allClusters(clusters: any[]) {
+      if (clusters.length) {
+        // Resolve metadata.name values to nameDisplay for UI display
+        this.selectedClusters = this.selectedClusters.map(
+          (name) => this.resolveClusterDisplayName(name)
+        );
+      }
+    },
   },
 
   computed: {
@@ -350,6 +359,14 @@ export default {
       }
 
       return undefined;
+    },
+
+    resolveClusterDisplayName(name: string): string {
+      const cluster = this.allClusters.find(
+        (c: any) => c.metadata.namespace === this.namespace && c.metadata.name === name
+      );
+
+      return cluster ? cluster.nameDisplay : name;
     },
 
     reset() {


shell/models/fleet-application.js
```diff
@@ -105,7 +105,7 @@ export default class FleetApplication extends SteveModel {
 
     for (const tgt of this.spec.targets) {
       if (tgt.clusterName) {
-        const cluster = findBy(clusters, 'metadata.name', tgt.clusterName);
+        const cluster = findBy(clusters, 'metadata.name', tgt.clusterName) || findBy(clusters, 'nameDisplay', tgt.clusterName);
```
torchiaf, line 108: Should we just use `findBy(clusters, 'nameDisplay', tgt.clusterName)` here?
The `nameDisplay` member already include the metadata.name:
```
  get nameDisplay() {
    return this.metadata?.labels?.[FLEET_LABELS.CLUSTER_DISPLAY_NAME] || this.metadata?.name || this.id;
  }
```

shell/models/fleet.cattle.io.bundle.js
```diff
@@ -47,7 +47,7 @@ export default class FleetBundle extends SteveModel {
 
     for (const tgt of this.spec.targets) {
       if (tgt.clusterName) {
-        const cluster = findBy(clusters, 'metadata.name', tgt.clusterName);
+        const cluster = findBy(clusters, 'metadata.name', tgt.clusterName) || findBy(clusters, 'nameDisplay', tgt.clusterName);
```
torchiaf, line 50: Same here, see comment above.

torchiaf (commented):
LGTM, I left 1 comment mostly to clarify my understanding of this change

edenhernandez-suse (dismissed):

shell/models/fleet-application.js
marcelofukumoto: For regression cases we want the metadata.name to work.
So in the end what I did is that it keeps checking for metadata.name which is the current release status.
But if doesn't find because it would use the displayName instead and maybe not find anything.

The example is if you use the ID : c-xxxx than that is the same as the metadata.name: c-xxxx.
Previously it was c-xxxx and it used to work, so that is what is on the YAML. 

But that made me realize that the correct fix is to force the displayName on the SELECT! So that will work better. But even so, we need to consider the cases with the c-xxxx and show properly on the UI.
It just means that metadata.name is priority compared to displayName. and the logic for the get nameDisplay is to use displayName first.

marcelofukumoto: Seems complicated, let me know if you need a call for it, I can only understand testing the case that I have. Make me think if I am missing other cases but this is clearly one.

torchiaf: All clear, thanks!

torchiaf (approved):
LGTM

Issue #17081, opened 2026-03-31
Validate checksum of downloaded files


richard-cox:
Jenkins Based
- https://github.com/rancher/dashboard/blob/master/cypress/jenkins/Dockerfile.ci#L6
- https://github.com/rancher/dashboard/blob/master/cypress/jenkins/Dockerfile.ci#L11
- https://github.com/rancher/dashboard/blob/master/cypress/jenkins/init.sh#L57
- https://github.com/rancher/dashboard/blob/master/cypress/jenkins/init.sh#L60
- https://github.com/rancher/dashboard/blob/master/cypress/jenkins/init.sh#L63

core dashboard based (https://github.com/rancher/dashboard/pull/17083)
- https://github.com/rancher/dashboard/blob/master/.github/actions/install-chrome/action.yaml#L16
- https://github.com/rancher/dashboard/blob/master/.github/workflows/scripts/install-codecov.sh#L3
- Remove scripts/build-embedded EMBED_PKG usage
- https://github.com/rancher/dashboard/blob/master/scripts/e2e-k3s-start#L52
- https://github.com/rancher/dashboard/blob/master/scripts/e2e-k3s-start#L59

richard-cox:
@izaac can you run through the jenkins usages of curl and update each to validate the checksum of the downloaded file matches a validated checksum?

I found gemini pretty good for for updating the scripts (see pr linked to other locations in description)

For more info see https://github.com/rancher/rancher-security/issues/1762 + https://github.com/rancher/security-team/wiki/how-to-act#31-insecure-usage-of-curl-and-wget

izaac:
@richard-cox sure, will do. We're transitioning the orchestration to Ansible and deprecating corral moving the majority of the pipeline logic to a different repo. I will be taking care of what's going to stay in our dashboard repo

richard-cox:
@izaac would you be able to update with an ETA?

izaac:
All the Jenkins based checksum validations listed in the description were moved to a QA infra repository. 

https://github.com/rancher/dashboard/pull/17135
https://github.com/rancher/qa-infra-automation/pull/84

PR #17083, opened 2026-03-31
Ensure we check checksum for certain downloads

richard-cox:
### Summary
Contributes to 
- https://github.com/rancher/rancher-security/issues/1762
- https://github.com/rancher/dashboard/issues/17081

- removed embedding of legacy harvester cluster (and curl that fetched it)
- valdiate checksum of downloads for k3s, helm, code cov, install chrome (added process, but we don't currently use them)

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/.github/actions/install-chrome/action.yaml
+++ b/.github/actions/install-chrome/action.yaml
@@ -4,6 +4,9 @@ inputs:
   chrome-version:
     description: 'The version of Google Chrome to target. Versions are formatted as `144.0.7559.96-1`.'
     required: true
+  chrome-version-checksum:
+    description: 'The checksum for the target Google Chrome version DEB file.'
+    required: true
 runs:
   using: "composite"
   steps:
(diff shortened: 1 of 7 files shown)


.github/workflows/build-and-upload.yaml
```diff
@@ -84,8 +84,6 @@ jobs:
       - id: build-embedded
         name: Build Embedded
         run: ./scripts/build-embedded
-        env:
```
nwmac, line 87: This seems unrelated to the PR?

.github/actions/install-chrome/action.yaml
```diff
@@ -4,6 +4,9 @@ inputs:
   chrome-version:
     description: 'The version of Google Chrome to target. Versions are formatted as `144.0.7559.96-1`.'
     required: true
+  chrome-version-checksum:
```
nwmac, line 7: So, is this action not used anywhere?

.github/workflows/build-and-upload.yaml
richard-cox: This relates to the change in `scripts/build-embedded`, we no longer use the env var

.github/actions/install-chrome/action.yaml
richard-cox: it's commented out in test.yaml, but we've hit times in the past where we need to pin chrome to avoid blocking gates. this is untested, but means we won't need to re-implement and remember to do the checksum stuff

PR #17101, opened 2026-04-01
Add `--frozen-lockfile` to yarn install commands in CI

rak-phillip:
### Summary
This updates yarn install usage to enforce `--frozen-lockfile` across all `yarn` and `yarn install` commands.

- Add `--frozen-lockfile` to yarn install commands in CI
- Add `--frozen-lockfile` to bare `yarn` commands
- Change `--pure-lockfile` to `--frozen-lockfile`

### Technical notes summary

`--frozen-lockfile` makes installs deterministic by ensuring that yarn installs the exact dependency versions already defined in `yarn.lock`. If the lockfile is out of sync with `package.json`, the install fails immediately instead of silently re-resolving dependencies or mutating the lockfile. 

CI should pass without failure.

There are some scripts that have been altered, mostly related to extension builds and tooling. I'm not sure if we need to change these files, but if they are run in CI in any way, the answer is yes: the aim of this PR is to prevent accidental lockfile updates during CI runs and to ensure that every pipeline execution installs the same dependency graph.

### Screenshot/Video

NA

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/scripts/build-e2e
+++ b/scripts/build-e2e
@@ -28,7 +28,7 @@ OUTPUT_DIR="${E2E_BUILD_DIST_DIR:-dist}"
 
 if [ "$BUILD_DASHBOARD" == "true" ]; then
     echo "Building production build for e2e ..."
-    yarn --pure-lockfile install
+    yarn install --frozen-lockfile
 
     source scripts/version
     echo "BRANCH: ${COMMIT_BRANCH:-<none>}"
--- a/scripts/build-embedded
+++ b/scripts/build-embedded
@@ -11,7 +11,7 @@ fi
 cd $(dirname $0)/..
 
 echo "Bootstrapping..."
-yarn --pure-lockfile install
+yarn install --frozen-lockfile
 
 source scripts/version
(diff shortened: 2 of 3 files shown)


cypress/scripts/build.sh
```diff
@@ -80,7 +80,7 @@ else
   echo ""
   echo "To test with yarn link:"
   echo "  cd dist"
-  echo "  yarn install && yarn link"
+  echo "  yarn install --frozen-lockfile && yarn link"
```
rak-phillip: This change isn't needed. It's just a command for developers.

richard-cox (approved):
Looking at the docs  https://classic.yarnpkg.com/en/docs/cli/install#toc-yarn-install... they're bad.

From what i can tell neither modify the lock file (https://github.com/yarnpkg/yarn/issues/5847#issuecomment-537521943), however pure-lockfile will still install package.json ranges whilst frozen-lockfile exits with a failure.

In both cases we should avoid the package.json overriding the lock, so frozen-lockfile seems a good choice (even if it will lead to lots of failures)

PR #17138, opened 2026-04-06
Remove `defu` dependency from Dashboard and Shell

rak-phillip:
### Summary
This removes the `defu` dependency from Rancher Dashboard and Shell and replaces its usage with Lodash's `defaultsDeep`.

- Remove `defu` dependency
- Replace `defu` usage with lodash's `defaultsDeep`

### Technical notes summary

I noted that `defu` is only used in a single location in Dashboard for configuring axios while reviewing the `defu` bumps raised by dependabot:

- https://github.com/rancher/dashboard/pull/17126
- https://github.com/rancher/dashboard/pull/17127

Since lodash already provides what appears to be the same functionality, I'm opting to remove defu to reduce the overall dependency footprint in Dashboard and Shell.

Network requests. Ultimately, we should be able to verify that the axios defaults contain the expected values after this change - they should remain unchanged.

Network requests - axios defaults and options merging behavior could be changed.

### Screenshot/Video

NA

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/utils/axios.js
+++ b/shell/utils/axios.js
@@ -1,4 +1,3 @@
-import { defaultsDeep } from 'lodash';
 import Axios from 'axios';
 
 import axiosRetry from 'axios-retry';
@@ -38,9 +37,6 @@ const axiosExtra = {
   onError(fn) {
     this.onRequestError(fn);
     this.onResponseError(fn);
-  },
-  create(options) {
-    return createAxiosInstance(defaultsDeep({}, this.defaults, options));
   }
 };
 


rak-phillip:
> I don't the the create() method is ever getting called. It might be possible to completely remove it.
>
> Tracing through the code I don't think create is used, I never ran into it with a breakpoint or console debugging. I think the create method is used when creating an axios instance and we're using the axiosExtras to extend the instance that was created in this file.

Indeed.. `extendAxiosInstance()` is invoked after `Axios.create()`, which makes sense. This begs the question, what was the original intent for this extending axios in the first place? Based on some investigation, it looks like this was originally a Nuxt implementation detail that we preserved before the Vue3 migration.

rak-phillip:
@codyrancher I removed the unused create method in e4063dde3ee329b80241099c2a80680410433b57. This is ready for another review

PR #17145, opened 2026-04-07
[14627][17943] fleet appco

### LINKED ISSUE #14627: Fleet: AppCo integration
AppCo integration

Users should be able to select the AppCo registry as Chart source in Fleet.
Fleet should be able to create imagePullSecrets in target clusters to deploy docker images from AppCo charts
The UI should warn the users that any changes in AppCo charts (managed by Rancher) will be not reflected on HelmApps in Fleet

Related issue: https://github.com/rancher/fleet/issues/3900

### LINKED ISSUE #17943: Update link to Fleet docs in AppCo integration
**Context**
In the new AppCo integration in Fleet, we've included a link to the `Automated propagation of resources to downstream clusters` article in Fleet's documentation.

**Problem**
This feature is currently experimental and the article is under the Experimental section. This will be changed in 2.16 and the URL will change as well. We need to update the link to the new URL as soon as they make the change in the docs.

Image

Image

Please sync with Tan Néau @weyfonk to be aligned when this change happens in Fleet's docs.

marcelofukumoto:
### Summary
Fixes #14627
Fixes #17943

- Added AppCo App Bundle creation to Fleet

### Technical notes summary

- Dashboard should still treat as HelmOps

- Regression for HelmOps and GitRepo AppBundle creation
- List of AppBundles
- 

### Screenshot/Video

- 
image
image
image
image
image
image
image
image
image
image
image

VIEW:
image
image
image
image
image

EDIT:
image

ALL CLUSTERS:
image

ERROR STATE:
image

CHARTS EMPTY STATE:
image

### Checklist
- [x] The PR has a Milestone 


torchiaf (changes requested):
I leave some comments as first round of review:

- We should consider to move out of the shell the `suse-app-collection-dark-mode.svg` icon, since we already experienced the shell build's size increased by adding new icons

marcelofukumoto:
> I leave some comments as first round of review:
> 
> * We should consider to move out of the shell the `suse-app-collection-dark-mode.svg` icon, since we already experienced the shell build's size increased by adding new icons

I actually now have no idea why I added a different file since I added it before for the AppCo on Charts. Thank you for noting! Just fixed to use the existing one.

pkg/rancher-components/src/components/Form/Radio/RadioButton.vue
```diff
@@ -189,7 +197,7 @@ export default defineComponent({
     />
     <div class="labeling">
       <label
-        :class="[ muteLabel ? 'text-muted' : '', 'radio-label', 'm-0']"
+        :class="['radio-label', 'm-0', { 'text-muted': muteLabel, 'body-text-color': useBodyTextColor }]"
```
torchiaf: Nit, for readability:

```suggestion

        class="radio-label m-0"
        :class="{
          'text-muted': muteLabel,
          'body-text-color': useBodyTextColor
        }"

```

shell/assets/translations/en-us.yaml
```diff
@@ -3333,6 +3347,24 @@ fleet:
       label: Workspace
       addWorkspace: Create a workspace
     helmRepoURLRegex: Helm Repos (URL Regex)
+  appCo:
+    credentials:
+      title: Create an App Bundle from SUSE Application Collection
+      subtitle: Set up your credentials
+      infoText: An Application Collection access token is required to connect to the Helm Charts catalogue.
```
torchiaf: Probably we should use American-english `catalog` -  there are 4 occurences.

shell/config/types.js
```diff
@@ -420,6 +421,18 @@ export const CLUSTER_REPO_APPCO_AUTH_GENERATE_NAME = 'clusterrepo-appco-auth-';
  */
 export const CLUSTER_REPO_AUTH_GENERATE_NAME = 'clusterrepo-auth-';
 
+/**
+ * The `generateName` prefix used when creating Fleet Helm authentication secrets
+ * for SUSE App Collection.
+ */
+export const FLEET_APPCO_AUTH_GENERATE_NAME = 'fleet-appco-auth-';
```
torchiaf: Nit: I think we should move this on in fleet directory to keep it separated from the shell

shell/components/nav/__tests__/Header.test.ts
```diff
@@ -157,6 +157,21 @@ describe('component: Header', () => {
 
       expect((wrapper.vm as any).disableWorkspaceSwitcher).toBe(true);
     });
+
+    it.each([
+      ['c-cluster-fleet-application-appco-credentials', '/c/local/fleet/application/suse-app-collection/credentials'],
+      ['c-cluster-fleet-application-appco-charts', '/c/local/fleet/application/suse-app-collection/charts'],
+      ['c-cluster-fleet-application-appco-chart', '/c/local/fleet/application/suse-app-collection/chart'],
+    ])('should return true on AppCo page %s (via route meta)', (name, path) => {
```
torchiaf:
```suggestion
    ])('should disable Workspace Switcher on AppCo page %s (via route meta)', (name, path) => {
```

shell/components/fleet/AppCoChartGrid.vue
```diff
@@ -0,0 +1,401 @@
...
+const store = useStore();
+const { t } = useI18n(store);
+
+const searchQuery = ref('');
+const searchInput = ref<HTMLInputElement | null>(null);
+const selectedSortOption = ref(CATALOG_SORT_OPTIONS.ALPHABETICAL_ASC);
+
+const sortOptions = [
+  { kind: 'group', label: t('catalog.charts.sort.prefix') },
+  { value: CATALOG_SORT_OPTIONS.LAST_UPDATED_DESC, label: t('catalog.charts.sort.lastUpdatedDesc') },
```
torchiaf, line 59: I think the `created` date is the creation date of the repo, so it will be the same for all charts, maybe it's worth verifying it. BTW it doesn't hurt, so we can keep.

shell/utils/fleet-appco.ts
```diff
@@ -0,0 +1,264 @@
...
+}
+
+export async function createAppCoAuthSecret(store: VuexStore, credentials: AuthCredentials, namespace: string) {
+  const {
+    selected,
+    publicKey,
+    privateKey,
+  } = credentials;
+
+  if (![AUTH_TYPE._SSH, AUTH_TYPE._BASIC, AUTH_TYPE._S3].includes(selected)) {
```
torchiaf: Question: Why do we need ssh, s3 and basic types check ?

shell/utils/fleet-appco.ts
```diff
@@ -0,0 +1,264 @@
...
+
+        throw err;
+      }
+    }
+  }
+
+  return repoName;
+}
+
+async function waitForRepoReady(
```
torchiaf, line 225: Nit: I think we could reuse the `waitForTestFn` from `shell/plugins/dashboard-store/resource-class.js`, please take a look. We have an example in Fleet as well: `waitForFleetClusterLastRevision`

shell/assets/translations/en-us.yaml
```diff
@@ -3186,6 +3197,9 @@ fleet:
         description: GitRepo describes a git repository that is watched by Fleet. The resource contains the necessary information to deploy the repo, or parts of it, to target clusters.
       fleet.cattle.io.helmop:
         description: HelmOp is a simplified way of creating bundles by directly pointing to a Helm repository or to an OCI registry, without needing to set up a git repository.
+      suse-application-collection:
+        label: SUSE Application Collection
+        description: A SUSE Application Collection is a curated set of applications from SUSE for enterprise usage.
```
torchiaf: This is describing what SUSE AppCo is. I think it should describe what is a Fleet HelmOp created from the SUSE AppCo catalog in this case.

shell/models/fleet.cattle.io.helmop.js
```diff
@@ -4,10 +4,29 @@ import { set } from '@shell/utils/object';
 import { SOURCE_TYPE } from '@shell/config/product/fleet';
 import FleetUtils from '@shell/utils/fleet';
 import { FLEET } from '@shell/config/types';
-import { FLEET as FLEET_ANNOTATIONS } from '@shell/config/labels-annotations';
+import { CATALOG, FLEET as FLEET_ANNOTATIONS } from '@shell/config/labels-annotations';
 import FleetApplication from '@shell/models/fleet-application';
+import { SUSE_APP_COLLECTION_REPO_URL } from '@shell/utils/fleet-appco';
 
 export default class HelmOp extends FleetApplication {
+  get isSuseAppCollectionFromUI() {
```
torchiaf, line 12: Nit - could we create a new AppCoHelmOp class extending HelmOp ? Not sure if it brings any or minimal benefits, but i think it would simplify things in `fleet.cattle.io.helmop.vue`.
Maybe we could quickly evaluate that and postpone this improvements in the next release, just in case...

shell/edit/fleet.cattle.io.helmop.vue
```diff
@@ -110,12 +136,59 @@ export default {
...
+
+        if (queryVersion) {
+          set(this.value, 'spec.helm.version', queryVersion);
+        }
+
+        if (querySecret) {
+          const ns = this.value.metadata.namespace;
+
+          this.updateAuth(`${ ns }/${ querySecret }`, 'helmSecretName');
+          this.addAppCoImagePullSecretToSpec(`${ querySecret }-image-pull-secret`);
```
torchiaf: Can we move this suffix  `-image-pull-secret` to a const variable ?

shell/components/form/LabeledSelect.vue
```diff
@@ -513,6 +527,35 @@ export default {
...
+            :title="`Deselect ${ getOptionLabel(selectedOption) }`"
+            :aria-label="`Deselect ${ getOptionLabel(selectedOption) }`"
+            @click="deselect(selectedOption)"
+          >
+            <svg
+              xmlns="http://www.w3.org/2000/svg"
+              width="10"
+              height="10"
+            >
+              <path d="M6.895455 5l2.842897-2.842898c.348864-.348863.348864-.914488 0-1.263636L9.106534.261648c-.348864-.348864-.914489-.348864-1.263636 0L5 3.104545 2.157102.261648c-.348863-.348864-.914488-.348864-1.263636 0L.261648.893466c-.348864.348864-.348864.914489 0 1.263636L3.104545 5 .261648 7.842898c-.348864.348863-.348864.914488 0 1.263636l.631818.631818c.348864.348864.914773.348864 1.263636 0L5 6.895455l2.842898 2.842897c.348863.348864.914772.348864 1.263636 0l.631818-.631818c.348864-.348864.348864-.914489 0-1.263636L6.895455 5z" />
```
torchiaf: Should we move this to a svg file ?

shell/models/fleet.cattle.io.helmop.js
```diff
@@ -4,10 +4,29 @@ import { set } from '@shell/utils/object';
...
+
+  get isSuseAppCollection() {
+    // Annotation set by the UI on create, or fallback to URL check for older resources
+    return this.isSuseAppCollectionFromUI ||
+      (this.spec?.helm?.repo || '').startsWith(SUSE_APP_COLLECTION_REPO_URL);
+  }
+
+  get applicationType() {
+    if (this.isSuseAppCollectionFromUI) {
+      return 'SUSE AppCo';
```
torchiaf: This should be a const var ?

torchiaf (changes requested):
I tested this feature and reviewed most of the files - compontents in shell/components/fleet/ postponed to the next round of reivew.

I left some comments, mostly technical and found some issues, please take a look:

### Details

1- If I manually remove the repository, it gives me connection error -> it should be possible to verify that the repo exists and ask the user if they want to add it again: Something like "Missing repository: <repo name>. Do you want to add it again? (Y/N)"

2- When I switch from a workspace to another, and then I try to create a new AppCo HelmOp, the AppCo display again the Setup page even it's already setup (in a different workspace). Should we clone the existing config from one workspace where it's already configured ? For sure there is no need to create the AppCo repo again.

3- Should we mark the AppBundles created from AppCo in the dashboard ? The only way to know it is by seeing the OCI registry URL: if repo is "oci://dp.apps.rancher.io" -> AppCo. Maybe we can add a label or a tag to identify them.

4- Not related to this PR, but I noticed that it's not possible to clone a repo/bundle/AppCo chart from 1 workspace to another, since the workspace switcher is disabled in edit mode. Same for edit -> not possible to edit the workspace of an existing HelmOp/GitRepo bundle. Is this a limitation or is it by design ?

shell/models/fleet.cattle.io.helmop.js
marcelofukumoto: Initially the AppCo flow and the HELMOP flow were quite similar.
Than later it changed to the multiple pages.
I kept the last screen the same since it uses the same resources.

But the main thing is that ultimately we want the HELMOP op and the APPCO flow to be the same. So the HELMOP flow should change to be same as the APPCO, so I think it is when the complexity will be removed.

Ideally we could have created another one... you are right. I think that would work better even for refactoring later. But I got biased to how I started.

I believe we will merge them next release so I think we can improve next release or when this new change comes.

pkg/rancher-components/src/components/Form/Radio/RadioButton.vue
marcelofukumoto: Done!

shell/assets/translations/en-us.yaml
marcelofukumoto: Changed.

shell/config/types.js
marcelofukumoto: Moved to shell/utils/fleet-appco.ts.

shell/components/nav/__tests__/Header.test.ts
marcelofukumoto: Changed!

shell/components/fleet/AppCoChartGrid.vue
marcelofukumoto: That is true... after I commit the other changes I will check this one.

shell/utils/fleet-appco.ts
marcelofukumoto: Right, I simplified. It is copied from the Auth Select.

shell/utils/fleet-appco.ts
marcelofukumoto: I did the change. Quite a big change but I think it is ok from what I tested. I will do more tests on this new logic.

shell/assets/translations/en-us.yaml
marcelofukumoto: I just checked and the text is actually different on the figma! Good catch.
The new text is:
`Explore our maintained and secure Application Collection, select a Helm chart and deploy it to your clusters.`

Fixed.

shell/edit/fleet.cattle.io.helmop.vue
marcelofukumoto: Moved.

shell/components/form/LabeledSelect.vue
marcelofukumoto: That was leftover. I copied the X from the vue-select.
I changed now to not create the slot from scratch.

shell/models/fleet.cattle.io.helmop.js
marcelofukumoto: Moved.

marcelofukumoto:
> I tested this feature and reviewed most of the files - compontents in shell/components/fleet/ postponed to the next round of reivew.
> 
> I left some comments, mostly technical and found some issues, please take a look:
> 
> ### Details
> 1- If I manually remove the repository, it gives me connection error -> it should be possible to verify that the repo exists and ask the user if they want to add it again: Something like "Missing repository: . Do you want to add it again? (Y/N)"
> 
> 2- When I switch from a workspace to another, and then I try to create a new AppCo HelmOp, the AppCo display again the Setup page even it's already setup (in a different workspace). Should we clone the existing config from one workspace where it's already configured ? For sure there is no need to create the AppCo repo again.
> 
> 3- Should we mark the AppBundles created from AppCo in the dashboard ? The only way to know it is by seeing the OCI registry URL: if repo is "oci://dp.apps.rancher.io" -> AppCo. Maybe we can add a label or a tag to identify them.
> 
> 4- Not related to this PR, but I noticed that it's not possible to clone a repo/bundle/AppCo chart from 1 workspace to another, since the workspace switcher is disabled in edit mode. Same for edit -> not possible to edit the workspace of an existing HelmOp/GitRepo bundle. Is this a limitation or is it by design ?

1 - That one is something we discussed before. There is a logic already on the select to recreate it. I will reuse when opening the page. If there is no REPO it will create automatically.

2 - It is a great idea. But we need a proper UX for that. Specially for maintenance of the data. This is something I have been thinking on how to make sure the secrets/image-pull-secret are fully linked. If we change one it doesnt apply to the other ones. I believe that the best is to create a "smart" UI that can show and be able to clone and check the secrets state. Kind of a "recover" system. I think if we have that it will start making more sense to do it automatically as well. 
Low-bar we could a button to clone the secret from other namespace. High-bar a full recover system + the button or auto clone.
A good reminder that the PRIME is 2.15.1 so maybe we can add the low-bar, let's recheck when @edenhernandez-suse is back.
But for this iteration, I think we can keep the way it is.

3 - Hmmm I think I can do this change! Since we should just change the ICON which we already have. Let me try this one.

4 - Right, we removed the ability to change workspaces on EDIT. So it breaks the clone. We can think about recovering it. We found many bugs on this interaction so we removed. Let's create a ticket on it, maybe allow the namespace change on some circumstances instead of completely blocking it on EDIT. But yes, not for this ticket.

Tldr.:
1 - I will implement it
2 - Let's discuss for 2.15.1 or 2.16.0
3 - I will do this change.
4 - Let's add this to 2.16.0 and discuss the workspace switcher.

marcelofukumoto:
Implemented 1 and 3.
Added to number 3 the isSuseAppCollectionFromUI so it should show same information that we see on the list.
image

To number 1 I added a logic to reuse the logic on the auth selection.
Added a new fallback check for the image-pull-secret on the list of the data as well to make sure it is there for the case where only the image-pull-secret is deleted.

Updated for the #17943 fix as well for the link on the downstream cluster downstream resource.

shell/utils/fleet-appco.ts
```diff
@@ -0,0 +1,323 @@
...
+ */
+export function getDownstreamResourcesDocsUrl(rancherVersion?: string): string {
+  // Harcoded to 2.X.0, it is fragile because if the version changes it will break.
+  // Ideally it would require a correlation between versions, but we have the fallback in place.
+  const match = /^v?2\.(\d+)/.exec(rancherVersion || '');
+  const minor = match ? parseInt(match[1], 10) : NaN;
+
+  // It should only exists after version 0.15.0, which will be fleet 0.16.0.
+  if (!isNaN(minor) && minor >= 15) {
+    return `https://fleet.rancher.io/0.${ minor + 1 }/downstream-resources`;
```
torchiaf, line 30: Why minor + 1 ?

shell/components/fleet/HelmOpAppCoResourcesSection.vue
```diff
@@ -0,0 +1,162 @@
...
+
+          <div
+            class="row"
+          >
+            <div class="col span-6">
+              <FleetSecretSelector
+                :value="downstreamSecretsList"
+                :namespace="value.metadata.namespace"
+                :mode="mode"
+                :locked-options="lockedSecrets"
```
torchiaf, line 119: I found a small issue here: it's not possible to remove a secret after adding it.

https://github.com/user-attachments/assets/9467a6c2-c2b8-4ce4-ac99-eaedde897adf

Could you verify? I think it depends on `locked-options` prop.

```
const updateSecrets = (list: string[]) => {
  const newList = [...list];

  for (const locked of toValue(lockedSecrets)) {
    if (!newList.includes(locked)) {
      newList.push(locked);  // ← Always adds locked secrets back
    }
  }

  emit('update:downstream-resources', { kind: 'Secret', list: newList });
};
```

Also I noticed a small slowness when opening/closing the secret selector, maybe it's depends on some data re-fetch, not a big issue though.

shell/components/fleet/HelmOpValuesTab.vue
```diff
@@ -70,71 +63,83 @@ const updateYamlForm = () => {
...
-    </div>
-
-    <div class="mb-20">
-      <FleetValuesFrom
-        :value="value.spec.helm.valuesFrom"
-        :namespace="value.metadata.namespace"
-        :mode="realMode"
-      />
+      <div class="mb-20">
+        <FleetValuesFrom
```
torchiaf, line 139: Small issue: the ValuesFrom data is not being saved:

https://github.com/user-attachments/assets/14d371f9-df77-4531-a383-e3281890f2bd

The event's handler is missing: `@update:value`. I think this bug is also on master branch

torchiaf (changes requested):
I tested again 1,3 points from above, now it looks good, the UI is able to recover from missing repository and the SUSE appco icon is nice! Well done.

I double checked the new components in `shell/components/fleet/` and found a couple of issues (1 is not from thie PR). Could you take a look ?
See comments below

shell/utils/fleet-appco.ts
marcelofukumoto: Fleet is always one minor up.
This is from Tan's comment: https://github.com/rancher/dashboard/issues/17943#issuecomment-4692454203 So 

So fleet is 0.16.0 for rancher 2.15.0

shell/components/fleet/HelmOpValuesTab.vue
marcelofukumoto: That is a good catch, let me check on it. I saw it was working on 2.13 and 2.14 it stopped working, let me fix this on this PR.

shell/components/fleet/HelmOpAppCoResourcesSection.vue
marcelofukumoto: Perfect on this catch. Is when the name is the same. I will fix this.

torchiaf (approved):
LGTM

PR #17150, opened 2026-04-07
Chart Install UI improvements

### LINKED ISSUE #17073: Fix visual issues on Chart Install Wizard page
'Linux Only' looks awful - I don't know why it is shown like this:

Image

We have two banners that should be removed or otherwise shown another way.

I don't care about the banner at the top - remove it.

The text wraps for the chart name and the steps.

Maybe we could show a number in a circle before the step name rather than 'Install: Step 1'. The steps then take loads of space that they don't use well.

Discuss with UX to determine if we should make small, incremental changes here, or consider a better design.

If latter, I would still like to make small improvements in quick order if we can.

momesgin:
### Summary
Fixes #17073

Removed the type description banner and "Linux Only" badge from the chart installation wizard. Replaced the generic wizard step header with a new structured header layout that prominently displays the chart's logo, the action being performed (e.g., `<chart-name>: Install`), the chart version (including suffixes like "Curren" and "Linux-only"), and allows navigating back to the chart details page by clicking the chart's name.

- Removed the type description banner ("All charts have at least...") and its corresponding translation keys.
- Removed "Linux Only" badge and `windowsIncompatible` warning logic.
- Changed the title format to `<chart-name>: <process>` (e.g., "CloudCasa: Install" or "CloudCasa: Upgrade").
- Made the chart name clickable to redirect to the chart details page.
- Added a version icon next to the chart version subtext.
- Added suffixes ("Current", "Current, Linux-only" etc) to the version in the header and in the version selector.
- Added a `showStepHeader` prop to the `Wizard` component to allow hiding the default active step banner and applied it to the chart installation flow.

### Technical notes summary
- Introduced a `showStepHeader` boolean prop to `shell/components/Wizard.vue` (defaulting to `true`). When set to `false`, it prevents the rendering of the subtext header that displays the current step (e.g., ": Step N").
- Passed `:show-step-header="false"` to the `Wizard` component within `shell/pages/c/_cluster/apps/charts/install.vue`.
- Replaced the previous `bannerTitleImage` and subtitle implementation within `install.vue` with custom structured HTML targeting the `bannerTitle` slot. This layout incorporates the logo, the version subtext with a new `icon-version-alt` icon, and the chart name as a `router-link` targeting `chartLocation()`.
- Removed the `<TypeDescription resource="chart" />` tag at the top of the install wizard and the translation strings from `en-us.yaml` and `zh-hans.yaml`.
- Removed the `windowsIncompatible` computed property and corresponding warning label from `install.vue`.
- Updated `cypress/e2e/po/pages/explorer/charts/install-charts.po.ts` to expose the new `data-testid` and added assertions in `cypress/e2e/tests/pages/charts/chart-install-wizard.spec.ts` to ensure the chart name renders correctly and contains the expected `href`.

- **Chart Installation Flow:** 
  1. Navigate to Apps -> Charts and select any chart.
  2. Click "Install" and verify the header UI.
  3. Confirm the type description banner at the top of the page is absent.
  4. Confirm the step banner (e.g. "Install: Step 1") is removed, and the title correctly reads `<chart-name>: Install`.
  5. Verify the version number has a version icon next to it.
  6. Click the hyperlinked chart name and verify it redirects back to the chart's details page.
- **Chart Upgrade Flow:** Navigate to an installed app and initiate an upgrade. Verify the title reads `<chart-name>: Upgrade` without the step banner, and the chart name correctly links back.
- **Windows Incompatible Charts:** Attempt to install a chart that previously displayed a "Linux Only" / Windows incompatible badge. Ensure the badge no longer appears but the install wizard functions normally.

- Fleet's Git Repo or HelmOp process also uses the Wizard component, make sure nothing's changed over there.

### Screenshot/Video

image
image

### Checklist
- [x] The PR has a Milestone 


shell/pages/c/_cluster/apps/charts/install.vue
```diff
@@ -762,21 +760,6 @@ export default {
...
-      if (this.versionInfo) {
-        const incompatibleVersion = !(this.versionInfo?.chart?.annotations?.[CATALOG_ANNOTATIONS.PERMITTED_OS] || LINUX).includes('windows');
-
-        if (incompatibleVersion && !this.chart.windowsIncompatible) {
-          return this.t('catalog.charts.versionWindowsIncompatible');
-        }
-      }
-
-      return null;
-    },
```
rak-phillip, line 784: hmmm... do we have any concerns with removing this "warning"? Does a user have any indication that they are installing a windows incompatible chart with this removal? I can understand that this might have caused layout issues, but I would think that our goal would be to redesign the behavior rather than outright remove it.

rak-phillip (commented):
I think this change looks good overall, but I have some concerns about the removal of the `windowsIncompatible` computed. Can we confirm that this is an intended change? I don't see an indication that this should be removed in the issue or its comments #17073

shell/pages/c/_cluster/apps/charts/install.vue
momesgin: Removing the "Linux Only" badge has been one of the UX/UI requirements based on this comment:
https://github.com/rancher/dashboard/issues/17073#issuecomment-4182337532

I'm also working on another issue that is related to the os compatibility on the install page, the gist of it is that we should only apply OS restrictions to the charts coming from our official repos:
https://github.com/rancher/dashboard/issues/11670

We currently disable incompatible versions in the versions dropdown. All that said, I have no preferences on this. If you want we can add Ovi as a reviewer to have a second look?

momesgin: oh wait, you're only asking about the warning banner, I misread your comment. Let me double-check that

rak-phillip: Yes, specifically the `windowsIncompatible` messaging. 

> If you want we can add Ovi as a reviewer to have a second look?

I think that's a good idea, overall. It might be a simple answer with regards on how to deal with this. @oboc-sts FYI

momesgin: I thought there's a warning banner that I missed, but this is the same "Linux only"/"Linux only version" badge(should we call it warning badge? lol) I mentioned earlier, so my previous comment is still valid.

momesgin: just for making it clearer, we would display either "Linux only" or  "Linux only version", I manually forced rendering the "Linux only version" below:

Image

rak-phillip: Indeed, that seems to match the behavior described in the discussion, thanks for double checking this.

oboc-sts:
Looking good @momesgin .

2 tiny things we need to fix styling wise as per the screenshot below:
- left = your implementation of the install page (I took it from this ticket)
- right = the implementation of the chart details page
- all 4 blocks, green and red, have a height of 24px
    - the green ones indicate the correct spacing used for the page header (margin to top and margin to horizontal diviver)
    - the red ones indicate incorrect spacing we need to fix

image

--

Fixing this achieves a nice effect when you transition from the chart details page to the install / flow page, namely the logo should feel like it's staying in place and the rest changes...

shell/pages/c/_cluster/apps/charts/install.vue
oboc-sts: the yellow "badges" / warnings with "linux only" and similar have been converted to tags on the chart... that's why we need to remove them

momesgin: @oboc-sts you're absolutely right about the "Linux only" but "Linux only version" is slightly different. It's specific to this page, and we display it if both of these conditions are met:
 - the entire chart is not marked as linux only
 - the currently selected version IS a linux only version

oboc-sts: In this case, especially because it's version dependent (your second point), we should display an info text... but somewhere in the proximity of the version selector.

In other words, if the selected version IS a linux only version AND the entire chart is not marked as "linux only", we display a gray text right under the version dropdown selector: "Linux only version" - it should look the same as it looks on the AI configuration page (see below)

image

momesgin: @rak-phillip I've applied the latest changes and updated the description.

momesgin: @oboc-sts I wanted to mention this scenario before asking Phil to do a final review.

On a fresh install, we don't display the version selector at all. If a user lands directly on the install page, there's no indicator letting them know that the chart or the specific version they are installing is Linux-only.

oboc-sts: > On a fresh install, we don't display the version selector at all...

Why not?

Can you attach 2 screenshots next to each other please with these 2 cases (fresh install vs upgrade / downgrade)? I'm curious what the difference is...

momesgin: Historically, we only displayed the version selector under two conditions:
```
      // Allow the user to choose a version if:
      // - the app exists (editing/upgrading)
      // - OR they've come from tools
```

I believe the reasoning for them was routing-based: if a user came from the Installed Apps or Cluster Tools page, they didn't have a chance to pick a version yet, so we showed the selector. However, if they came from the Chart Detail page, they had already selected a version, so we hid it on the install page.

we recently added the third case for the multiple installation of a chart:
```
    // - OR they're installing a new instance of an already-installed chart
```

here are some screenshots:

Fresh install:
image

Edits:
image

Fresh install from cluster tools:
image

Install a new instance:
image

oboc-sts: Thanks @momesgin 🙌 

Is it possible to always show the version selector and simply select the right version by default, allowing the user to change it, without the current gymnastics around it (= without all the exceptions and conditionals)?

momesgin: @oboc-sts Unfortunately this isn't as simple a change as I initially thought. It turns out a lot of related code would need to be modified. Given the scope, it might be better to address this as part of the upcoming install page redesign.

oboc-sts: Alright. Can you make ticket for 2.16 please?

Now, moving fwd with the current situation, I have 2 questions and 1 proposal considering the screenshot below (it's in one of your comments above)

image

1. Why do we show that version format (`current version` > `new version`) in the header and when did we decide to do it? It was never designed like this, that's why I'm asking...
2. The version we show in the header is present at all times, right? (including during a fresh install)
3. The version we show in the header updates in real time if the version selector is present and you change to another version than the one you started with, right?

Now, assuming the answer to the last 2 questions above is affirmative (it should be), then my suggestion is to:
- stick to the version format in the header that was designed in the first place (NOT a transition of versions as I've pointed in the 1st question, but a single version... the one that you're about to install, edit, downgrade or upgrade to) - and this is because... (see the next point)
- if that's a linux only version, we append this as a suffix = " (linux only)" to the version in the header. If we wouldn't consider doing the previous point, there wouldn't be enough real estate to display this (think about both version having the linux only suffix)

Based on these 2 changes ^, the version in the header from the screenshot above should change:
- from this: "109.3.1+up1.11.2 > 109.2.0tup1.10.2"
- to this: "109.2.0tup1.10.2 (linux only)"

Also, the linux only versions in the version selector below should have the same prefix = " (linux only)"

momesgin: > Why do we show that version format (current version > new version) in the header and when did we decide to do it? It was never designed like this, that's why I'm asking...

This is a 5 year old code with a beautiful design using ">" character only! we just didn't really notice it until now.

> The version we show in the header is present at all times, right? (including during a fresh install)

correct

> The version we show in the header updates in real time if the version selector is present and you change to another version than the one you started with, right?

correct


I think the new solution works great! during the implementation I realized that we won't need the "Linux only version" footnote under the version selector anymore since it'd become the third time we'd say linux only lol, please check the screenshot below:

Screenshot 2026-06-10 at 8 33 45 PM

I just need to mention that now we don't have a "Linux only" badge(the yellow thingy) that we were previously displaying **if the whole chart was linux only**, not sure if that's important to you.

oboc-sts: > This is a 5 year old code with a beautiful design using ">" character only! we just didn't really notice it until now.

I meant we shouldn't display the old and the new versions there, we should only display the version you're about to install, edit, upgrade to or downgrade to. Please read again my previous comment.

> I think the new solution works great! during the implementation I realized that we won't need the "Linux only version" footnote under the version selector anymore since it'd become the third time we'd say linux only

Yes, that is correct, the current solution replaces the previous one with the footnote under the version selector so this should, indeed, be removed.

> I just need to mention that now we don't have a "Linux only" badge(the yellow thingy) that we were previously displaying **if the whole chart was linux only**,

I'd say this is no longer necessary anyway since we're now treating this at version level.

momesgin: > I meant we shouldn't display the old and the new versions there, we should only display the version you're about to install, edit, upgrade to or downgrade to. Please read again my previous comment.

I know, I think my joke didn't land 😅 if you check the screenshot in my message it's the downgrade mode while displaying only the selected version in the header.

shell/assets/translations/zh-hans.yaml
```diff
@@ -981,6 +981,8 @@ catalog:
     version: 版本
     versions:
       current: '{ver}（当前）'
+      current_linux: '{ver}（当前，仅限 Linux）'
+      current_windows: '{ver}（当前，仅限 Windows）'
```
momesgin: @rak-phillip These are translated by AI, let me know if you think we should not add them ourselves.

momesgin:
@rak-phillip Sorry for the delay in getting this ready for your re-review! I've updated the PR description and screenshots to reflect the latest UI/UX decisions.

shell/assets/translations/zh-hans.yaml
rak-phillip: I think that it would be safer to defer to the default fallback behavior if we can't independently verify the accuracy of this translation.

rak-phillip (dismissed):

rak-phillip (approved):

PR #17170, opened 2026-04-08
Fix login background preview cross-fallback between light/dark modes

### LINKED ISSUE #17085: Custom login background image preview appears under both light/dark branding settings but only applies to the one it was uploaded to
**Setup**
- Rancher version: 2.13.3
- Rancher UI Extensions: None
- Browser type & version: Chrome 146, Safari (macOS Sequoia) (tested on two diff Macs for repro)

**Describe the bug**
When uploading a custom background image to either the Dark or Light login background setting under Global Settings → Branding, the image thumbnail preview renders under both the Dark and Light buttons in the UI — regardless of which one it was actually uploaded to. This falsely implies the image has been applied to both modes.

**To Reproduce**
1. Navigate to Global Settings → Branding (no background logo selected/unchecked)
2. Upload a custom image to the **Dark** login background setting only
3. Observe the branding UI — the uploaded image preview appears under both Dark and Light buttons
4. Open the Rancher login page in a browser set to **light mode**

**Result**
The default Rancher ship background is displayed on the login page. The branding UI shows no indication that the light mode setting is still unconfigured.

**Expected Result**
The UI should clearly distinguish which mode has a custom image applied vs. which is still using the default. Applying the same image to both by default wouldn't make sense as the whole point of option between light/dark background is for customer logos etc to present aesthetically. There should be a clear indication of when each setting is using the default Rancher ship branding.

**Screenshots**
Image

**Additional context**
The silent fallback to the default background on light mode is particularly confusing because there is zero visual difference in the branding settings screen before and after uploading the light mode image — making it very difficult to diagnose. Users will reasonably assume the image is applied universally based on the preview behavior.

codyrancher:
### Summary
Fixes #17085

Login background image preview in the branding settings page no longer shows a cross-fallback between light and dark modes. Each preview now only appears when its own image is set, accurately reflecting the actual login page behavior.

### Technical notes summary
- Removed `||` conditions and ternary fallbacks in the login background preview `v-if` and `:src` bindings in `brand.vue`
- Split the single Cypress `Login Background` test into two separate tests (`Dark` and `Light`), each uploading only one image and asserting that the other preview does not exist

- Upload only a dark login background image and verify the light preview does not appear
- Upload only a light login background image and verify the dark preview does not appear
- Upload both and verify both previews appear correctly
- Verify the actual login page displays the correct background for each theme

- Branding settings page login background preview section

### Screenshot/Video
image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/pages/c/_cluster/settings/brand.vue
+++ b/shell/pages/c/_cluster/settings/brand.vue
@@ -440,14 +440,14 @@ export default {
             />
           </div>
           <SimpleBox
-            v-if="uiLoginBackgroundLight || uiLoginBackgroundDark"
+            v-if="uiLoginBackgroundLight"
             class="theme-light mb-10"
           >
             <label class="text-muted">{{ t('branding.loginBackground.lightPreview') }}</label>
             <img
               class="img-preview"
               data-testid="branding-login-background-light-preview"
-              :src="uiLoginBackgroundLight ? uiLoginBackgroundLight : uiLoginBackgroundDark"
+              :src="uiLoginBackgroundLight"
             >
           </SimpleBox>
         </div>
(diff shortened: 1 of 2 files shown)


shell/pages/c/_cluster/settings/brand.vue
```diff
@@ -440,14 +440,14 @@ export default {
...
-            v-if="uiLoginBackgroundLight || uiLoginBackgroundDark"
+            v-if="uiLoginBackgroundLight"
             class="theme-light mb-10"
           >
             <label class="text-muted">{{ t('branding.loginBackground.lightPreview') }}</label>
             <img
               class="img-preview"
               data-testid="branding-login-background-light-preview"
-              :src="uiLoginBackgroundLight ? uiLoginBackgroundLight : uiLoginBackgroundDark"
+              :src="uiLoginBackgroundLight"
```
codyrancher, line 450: It looks like we were doing this because we would show the same image for both themes if only one was configured, however we're not doing that, we fall back to the rancher default.

PR #17178, opened 2026-04-08
Fix incorrect button sizes introduced by RcButton changes

### LINKED ISSUE #16755: Fix incorrect button sizes introduced by changes to RcButton
Verify the design system RcButton to make sure that it's correct and then ensure that it's consistent throughout the app.

Highlights: Show Configuration, Edit and Cancel buttons.

Image

Image

codyrancher:
### Summary
Fixes #16755

- Fixing bugs and the most glaring visual inconsistencies caused by changes in RcButton.

### Technical notes summary
- **RcButton**: Override global style to fix spacing.
- **RcDropdownTrigger**: Only render `#before`/`#after` when provided to resolve spacing issue.
- **SortableTable**: Converted buttons to RcButton on bulk's size.
- **Masthead**: Converted Create button to RcButton
- **CruResourceFooter**: Converted Cancel button to RcButton
- **TitleBar**: Removed .icon-document` margin styles
- **ActionDropdownShell**: Removed `ml-10` class from chevron icon

- Header icon buttons (bell, page actions, user menu) sizing and spacing
- Table action buttons
- Create button on list pages
- Show Configuration button and action menu on resource detail pages
- Label/annotation tags on resource detail pages
- Cancel button in create/edit forms

- Any component using RcButton
- Navigation via Create button (now uses `$router.push` instead of `router-link`)

### Checklist
- [x] The PR has a Milestone

### DIFF
--- a/pkg/rancher-components/src/components/RcButton/RcButton.vue
+++ b/pkg/rancher-components/src/components/RcButton/RcButton.vue
@@ -33,7 +33,7 @@ const buttonSizesNew: { size: ButtonSize, className: string }[] = [
   { size: 'large', className: 'btn-large' },
 ];
 
-const props = withDefaults(defineProps<ButtonVariantProps & ButtonSizeProps & ButtonVariantNewProps & ButtonSizeNewProps & IconProps>(), { size: 'medium' });
+const props = withDefaults(defineProps<ButtonVariantProps & ButtonSizeProps & ButtonVariantNewProps & ButtonSizeNewProps & IconProps>(), { size: 'large' });
 
 const activeVariantClassName = computed(() => {
   if (props.variant === 'multiAction' || props.multiAction) {
@@ -132,6 +132,12 @@ button {
   display: inline-flex;
   align-items: center;
   justify-content: center;
+
+  // Override global .btn > .icon:not(:only-child) { margin-right: 6px } from _button.scss.
+  // RcButton uses flex gap for spacing instead. :deep() is needed to target slotted content.
+  &.rc-button > :deep(.icon) {
+    margin-right: 0;
+  }
 
   // Much of the styling in here came from _button.scss. Because we're making changes from role to variant and we don't want to impact existing use cases we're pulling in some of these styles. We should in the long run deprecate that file.
   // Variant styles
--- a/pkg/rancher-components/src/components/RcButton/RcButton.vue
+++ b/pkg/rancher-components/src/components/RcButton/RcButton.vue
@@ -53,6 +53,7 @@ const props = withDefaults(
 );
 
 const tag = computed(() => (props.to ? resolveComponent('RouterLink') : 'button'));
+const role = computed(() => (props.to ? 'link' : 'button'));
 
 const activeVariantClassName = computed(() => {
   if (props.variant === 'multiAction' || props.multiAction) {
@@ -116,6 +117,7 @@ defineExpose({ focus });
   <component
     :is="tag"
     ref="RcFocusTarget"
+    :role="role"
     :to="to"
     :class="{ ...buttonClass }"
   >
--- a/pkg/rancher-components/src/components/RcButton/RcButton.vue
+++ b/pkg/rancher-components/src/components/RcButton/RcButton.vue
@@ -33,7 +33,7 @@ const buttonSizesNew: { size: ButtonSize, className: string }[] = [
   { size: 'large', className: 'btn-large' },
 ];
 
-const props = withDefaults(defineProps<ButtonVariantProps & ButtonSizeProps & ButtonVariantNewProps & ButtonSizeNewProps & IconProps>(), { size: 'large' });
+const props = withDefaults(defineProps<ButtonVariantProps & ButtonSizeProps & ButtonVariantNewProps & ButtonSizeNewProps & IconProps>(), { size: 'medium' });
 
 const activeVariantClassName = computed(() => {
   if (props.variant === 'multiAction' || props.multiAction) {
--- a/pkg/rancher-components/src/components/RcButton/RcButton.vue
+++ b/pkg/rancher-components/src/components/RcButton/RcButton.vue
@@ -156,8 +156,10 @@ defineExpose({ focus });
   justify-content: center;
 
   // Override global .btn > .icon:not(:only-child) { margin-right: 6px } from _button.scss.
-  // RcButton uses flex gap for spacing instead. :deep() is needed to target slotted content.
-  & > :deep(.icon) {
+  // RcButton uses flex gap for spacing instead. The :not(:only-child) clause matches the global
+  // selector so we win on specificity regardless of stylesheet load order; :deep() is needed to
+  // target slotted content.
+  & > :deep(.icon:not(:only-child)) {
     margin-right: 0;
   }
 
--- a/pkg/rancher-components/src/components/RcButton/RcButton.vue
+++ b/pkg/rancher-components/src/components/RcButton/RcButton.vue
@@ -7,10 +7,15 @@
  *
  * <rc-button variant="primary" @click="doAction">Perform an Action</rc-button>
  */
-import { computed, ref } from 'vue';
+import { computed, ref, resolveComponent } from 'vue';
+import type { RouteLocationRaw } from 'vue-router';
 import {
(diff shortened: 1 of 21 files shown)


pkg/rancher-components/src/components/RcDropdown/RcDropdownTrigger.vue
```diff
@@ -35,18 +35,20 @@ defineExpose({ focus });
...
-    <template #before>
-      <slot name="before">
-        <!-- Empty Content -->
-      </slot>
+    <template
+      v-if="$slots.before"
+      #before
+    >
+      <slot name="before" />
     </template>
```
codyrancher, line 43: This was causing some extra space, if I remember correctly it was the whitespace. We only render these now if they're defined.

pkg/rancher-components/src/components/RcButton/RcButton.vue
```diff
@@ -33,7 +33,7 @@ const buttonSizesNew: { size: ButtonSize, className: string }[] = [
   { size: 'large', className: 'btn-large' },
 ];
 
-const props = withDefaults(defineProps<ButtonVariantProps & ButtonSizeProps & ButtonVariantNewProps & ButtonSizeNewProps & IconProps>(), { size: 'medium' });
+const props = withDefaults(defineProps<ButtonVariantProps & ButtonSizeProps & ButtonVariantNewProps & ButtonSizeNewProps & IconProps>(), { size: 'large' });
```
codyrancher: We switched to large because the majority of the existing buttons were supposed to be large by default.

shell/components/SortableTable/index.vue
```diff
@@ -1116,17 +1118,17 @@ export default {
         >
           <slot name="header-left">
             <template v-if="tableActions">
-              <button
+              <RcButton
```
codyrancher, line 1121: Our actions dropdown buttons  was an RcButton, I changed these so the style was consistent for all of the table actions.

shell/components/CruResourceFooter.vue
```diff
@@ -84,16 +87,16 @@ export default {
   <div class="cru-resource-footer">
     <slot name="footer-prefix" />
     <slot name="cancel">
-      <button
+      <RcButton
```
codyrancher, line 90: We can get rid of this but I did it just to get more consistent styling. On pages that don't use RcButton there will be inconsistency.

shell/components/Wizard.vue
```diff
@@ -444,28 +442,28 @@ export default {
             name="cancel"
             :cancel="cancel"
           >
-            <button
+            <RcButton
```
codyrancher, line 445: Same as the footer. I can remove this if we'd like to wait.

shell/components/ActionDropdownShell.vue
```diff
@@ -35,14 +35,15 @@ const setBulkActionOfInterest = (act: HiddenAction | null, event: 'mouseover' |
...
+      size="medium"
       :disabled="disabled"
     >
       <template #before>
         <i class="icon icon-gear" />
       </template>
       <span>{{ t('sortableTable.bulkActions.collapsed.label') }}</span>
       <template #after>
-        <i class="ml-10 icon icon-chevron-down" />
+        <i class="icon icon-chevron-down" />
```
codyrancher, line 46: The button is handling the spacing.

shell/components/Wizard.vue
```diff
@@ -185,9 +186,6 @@ export default {
       return this.steps.filter((step) => !step.hidden);
     },
 
-    nextButtonStyle() {
-      return this.activeStep.nextButton?.style || `btn role-primary`;
-    },
```
codyrancher, line 190: I searched through the codebase and extensions to verify that this isn't used. Specifically looked for the usage of nextButton and didn't find it.

I looked through all of these:
https://github.com/search?q=%27%40shell%2Fcomponents%2FWizard%27&type=code&p=4

codyrancher:
I'll still need to verify things with Ovi but I wanted to get a first round review on this.

pkg/rancher-components/src/components/RcButton/RcButton.vue
rak-phillip: That could be right, but I imagine that this could have an effect on any extensions that might be using RcButton. We might want to warn extension authors that the default is changing here.

shell/components/ResourceList/Masthead.vue
```diff
@@ -198,22 +200,22 @@ export default {
           <slot name="extraActions" />
 
           <slot name="createButton">
-            <router-link
+            <RcButton
```
rak-phillip, line 226: This is good - router-links styled as buttons has been an accessibility issue for a while that required quite a bit of ceremony to ensure that we got it correct. 

My only concern is that we might lose some affordances with a screen reader in switching from links to actual buttons. I'll dive into this and follow up with more details.

shell/components/CruResourceFooter.vue
rak-phillip: I think this is good. I say keep it.

shell/components/Wizard.vue
rak-phillip: It looks like dead code to me. I think this makes sense to remove.

rak-phillip (commented):
I think this looks good. I reviewed the output for affected areas and the buttons look good compared to before.

There's a few points that we will need to follow up on before moving forward.

pkg/rancher-components/src/components/RcButton/RcButton.vue
codyrancher: That's a good point. I think I should actually revert then and just manually set the size where relevant.

codyrancher: Changed.

shell/components/ResourceList/Masthead.vue
codyrancher: Updated per our conversation.

pkg/rancher-components/src/components/RcButton/RcButton.vue
```diff
@@ -94,9 +113,10 @@ defineExpose({ focus });
 </script>
 
 <template>
-  <button
+  <component
+    :is="tag"
     ref="RcFocusTarget"
-    role="button"
```
rak-phillip, line 99: It's not required when using proper semantic html, but can we reintroduce the role for both button and link? Our initial goal was to keep in line with the WAI-ARIA Roles, States, and Properties defined in the APG[^1][^2] for these base patterns.

[^1]: https://www.w3.org/WAI/ARIA/apg/patterns/button/
[^2]: https://www.w3.org/WAI/ARIA/apg/patterns/link/

storybook/src/stories/Components/RcButton.stories.ts
```diff
@@ -14,7 +14,7 @@ const meta: Meta<typeof RcButton> = {
     size: {
       options:     ['small', 'medium', 'large'] as ButtonSize[],
       control:     { type: 'select' },
-      description: 'Determines the size of the button. Medium is the default size for most use cases.'
+      description: 'Determines the size of the button. Large is the default size, matching the standard 40px button height.'
```
rak-phillip: I think there's two more items that we need to address in storybook before we merge:

1. Revert the default back to medium to match the change in the component
2. Can we add some stories to highlight and demonstrate the link button behavior

codyrancher: - Updated the description of `size` and added one for `to`.
- Added a section to highlight the `<router-link>` version
- Fixed it so `<router-link>` renders  `<a>` in storybook.
- Made it so we see the code in the examples

image


image

rak-phillip (dismissed):
LGTM

nwmac:
Can be merged now @codyrancher

codyrancher:
Adding for quick comparisons:

compare-01-masthead-create
compare-02-sortable-bulk-buttons
compare-03-sortable-bulk-overflow
compare-04-cru-footer-cancel
compare-05-wizard-footer
compare-06-chart-install

oboc-sts:
Looking awesome, @codyrancher ! 🚀

rak-phillip (approved):

PR #17219, opened 2026-04-12
Charts multiple installation

### LINKED ISSUE #16907: [BUG] No option to create new release for same chart
SURE-11434

**Rancher Server Setup**
- Rancher version: 2.13
   - Helm Chart
   - RKE2 v1.34.1+rke2r1
- Proxy/Cert Details:

- Kubernetes version: RKE2 v1.34.1+rke2r1
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): Hosted

**User Information**
Admin

**Describe the bug**

Unable to create new helm chart, that has already been installed. Rancher suggests to edit existing without an option to create another release.
**To Reproduce**

Go Apps, Charts, select already installed chart
**Result**

**Expected Result**

Expected button to create a new chart, but there are none
**Screenshots**

Image

<br/>

**Additional context**

Note: Not all charts can have multiple installed instances. If a chart has a target namespace defined, meaning it has these two annotations: `catalog.cattle.io/namespace` and `catalog.cattle.io/release-name` (e.g. Harvester CSI Driver chart) it can only be installed once. For these charts you should see a warning banner on the chart detail page:

Image

While testing you should not see any differences for these charts. But for the charts that can have multiple installed instances we'll have these changes:

On Chart detail page(Apps > Charts > click on an app card):
Once you've installed an app, you should see the new split button:

Image

clicking on the arrow button should open a dropdown with an option "Install as a new instance":

Image

clicking on "Install as a new instance" should redirect you to the install wizard page, and allow you to select a version, and add a new name. Top section should include "Install" in its title:

Note: if a chart has this annotation `catalog.cattle.io/release-name` the name field becomes disabled(not a new change though).

Image

Once you have installed more than one instance of an app you should see a new selector on the chart detail page, the formatting of each installed instance's label is `<namespace>/<app-name>`:

Image

After selecting an installed instance from the dropdown the page should reflect the data for that instance, e.g. statuses in the header section, the top right button and the installed version on the right side of page. Also if an installed instance is upgradeable, the "upgradeable" suffix should be added to the label of that instance in the selector:

Image

There are a few changes on Apps > Charts page too:
If an app has more than one installed instances, the tooltip for the installed status(checkmark icon) should be "Installed (multiple)":

Image

If at least one of the installed instances of a chart is upgradeable we should display the upgradeable status(arrow up icon) like in the previous screenshot.

### LINKED ISSUE #17211: [BUG] RKE2 Cluster Template reuses previous cluster state instead of creating a new cluster
**Rancher Server Setup**
- Rancher version: v2.13.2 / v2.13.4 (Tested on both versions, 2.13.2 on community and prime, v2.13.4 tested on prime)
- Installation option (Docker install/Helm Chart): Helm Chart
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE2
- Proxy/Cert Details: letsencrypt no proxy

- Kubernetes version: v1.34.4+rke2r1
- Cluster Type (Local/Downstream): Local

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom) Admin

**Describe the bug**
When using Rancher RKE2 Cluster Templates, creating a second cluster from the same template does not behave as expected. Instead of initializing a new cluster creation flow, Rancher reuses the state of the previously created cluster.

**To Reproduce**
1) Navigate to Cluster Management in Rancher
2) Click “Create Cluster”
3) Select an existing RKE2 Cluster Template (based on the official template)
4) Fill out the required template questions
5) Create the first cluster (Cluster 1) – this works as expected
6) Repeat the process to create a second cluster (Cluster 2) using the same template

**Result**
- The form is pre-filled with values from Cluster 1
- The UI displays the Helm release name of Cluster 1 in the title
- Instead of showing an “Install” button, the UI shows “Save Changes” (as if editing an existing cluster)
- Modifying values and proceeding causes Cluster 1 to be deleted and replaced with Cluster 2

**Expected Result**
A fresh, empty form should be displayed when creating a new cluster from the template.
Clicking “Install” should create a new, independent cluster (Cluster 2) without modifying or deleting Cluster 1.

**Screenshots**
Example of how the Cluster Template Installation page should look like (Install)

Image

Example of broken Cluster Template Installation page (Save changes instead of Install)
Note the wrong title (release of Cluster 1 instead of Chart Name and the wrong button at the end)

Image

**Additional context**
- In the template we use vSphere as provider type
- The template is based on the official RKE2 Cluster Template and has not been recently modified
- The issue occurs consistently across multiple environments
- This setup has worked correctly in the past
- The issue can be temporarily resolved by changing the chart name or chart version in the cluster template
After such a change, one new cluster can be created successfully before the issue reoccurs

momesgin:
### Summary
Fixes #16907 
Fixes #17211 

- Added support for managing and installing multiple instances of the same Helm chart directly from the chart details page.
- Replaced the single-action button with a split button (`RcButtonSplit`) that allows users to either manage the selected instance or install a brand-new instance.
- Introduced a dropdown selector (`LabeledSelect`) on the chart details page to choose between different installed instances of the same chart.
- Updated the `Chart` model and mixin logic to correctly calculate properties like `isInstalled`, `upgradeable`, and `installedCount` when multiple instances exist.
- Modified the install wizard to ensure the version selection step is shown when creating a new instance of an already-installed chart.

### Technical notes summary
- Added a `NEW_APP_INSTANCE` query parameter to distinguish between upgrading/editing an existing app and installing a new instance.
- Refactored the `Chart` model: `isInstalled` now checks if instances `>= 1`, and `upgradeable` checks if *any* instance has an upgrade available. The card tooltip was also updated to omit the version string when multiple instances of varying versions exist.
- Updated `chart.vue` to compute app statuses (e.g., installed version, upgrade availability) dynamically based on the specific instance selected in the dropdown.
- Updated the Cypress E2E tests to rely on the new `data-testid="btn-chart-install"` selector since the DOM structure for the button changed.
- Maintained protections for "targeted" charts (those with fixed namespace/name annotations); these correctly block the "install new instance" flow by setting `canInstallNew = false`.

For testing #16907  please check the **additional context** section on the issue's description.
For testing #17211  please check the issue's description.

- Chart installation, upgrade, and edit workflows.
- Status badge calculations (installed, upgradeable, deprecated) on the chart details page and catalog grid.
- App selections for "Targeted" charts (e.g., special charts with predefined namespaces/names) that should strictly prohibit multiple instances e.g. Harvester CSI Driver.

### Screenshot/Video

https://github.com/user-attachments/assets/4d0abb41-1cb3-4ad4-8125-4cf5b328e02e

### Checklist
- [x] The PR has a Milestone 


shell/models/chart.js
```diff
@@ -89,12 +89,12 @@ export default class Chart extends SteveModel {
...
   /**
-   * Determines if the chart is installed by checking if exactly one matching installed app is found.
+   * Determines if the chart is installed by checking if at least one matching installed app is found.
    *
    * @returns {boolean} `true` if the chart is currently installed.
    */
   get isInstalled() {
-    return this.matchingInstalledApps.length === 1;
+    return this.matchingInstalledApps.length >= 1;
   }
```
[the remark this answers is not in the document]
momesgin: this needs the design team's review

shell/mixins/chart.js
```diff
@@ -410,19 +427,23 @@ export default {
...
-          if (matching.length === 1) {
-            this.existing = matching[0];
-            this.mode = _EDIT;
+          if (isChartDetailPage) {
+            const matching = this.chart?.matchingInstalledApps;
+
+            if (matching?.length >= 1) {
+              // Populate the instance selector and default to the first installed app.
+              this.installedInstances = matching;
+              this.existing = matching[0];
```
bpedersen2: To me this looks still broken , the fork of the logic needs to happen earlier:
 - The endpoint has the following params:
 `...apps/charts/install?namespace=cert-manager&name=cert-manager&version=v1.20.1&repo=jetstack&repo-type=cluster&chart=cert-manager`

If trying a fresh install name will be emtpy.
On edit only a single install can ever match ( name + namespace must be unique).

richard-cox: @bpedersen2 thank you for taking a look, but the PR is still in draft and not ready for review.

shell/models/chart.js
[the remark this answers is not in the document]
momesgin: done

shell/mixins/chart.js
momesgin: @bpedersen2 thanks for checking, could you please confirm that the issue you're describing doesn't happen in 2.12(excluding 2.12.7 & 2.12.8)?

shell/pages/c/_cluster/apps/charts/chart.vue
```diff
@@ -153,22 +218,109 @@ export default {
...
       });
     },
+
+    /**
+     * Handle installed app selection from the dropdown.
+     * Updates selectedInstalledAppId and this.existing so the page reflects the current installed app and displays the correct information for the selected instance.
+     */
+    handleInstalledAppSelect(id) {
+      this.selectedInstalledAppId = id;
+      this.existing = this.installedInstances?.find((app) => app.id === id) || null;
```
rak-phillip: nit: Do we intend to use the nullish coalescing operator instead?

```suggestion
      this.existing = this.installedInstances?.find((app) => app.id === id) ?? null;
```

I don't think there's any change in behavior between vs the other, but `??` more effectively communicates that we want to short-circuit when there's no result vs. a falsy value when using `||`.

shell/pages/c/_cluster/apps/charts/chart.vue
```diff
@@ -138,6 +152,57 @@ export default {
...
+    },
+
+    /**
+     * Returns options for the installed apps selector dropdown.
+     * Adds "(Upgradeable)" suffix to apps that have an upgrade available.
+     */
+    installedAppOptions() {
+      return (this.installedInstances || []).map((app) => {
+        const isUpgradeable = app.upgradeAvailable === APP_UPGRADE_STATUS.SINGLE_UPGRADE;
+        const baseName = `${ app.metadata.namespace }/${ app.metadata.name }`;
```
rak-phillip: nit: consider using optional chaining for `app.metadata.namespace` & `app.metadata.name`. These should always be present, but I think it's better to be defensive in these areas.

rak-phillip (dismissed):
I confirmed the behavior with regards to the following issues:

- #16907
  - The new pattern for installing a new instance resolves the issue and is an improvement from the old behavior
  - There are some instances where the "Install as a new instance" options should not be shown and the failure modes can be confusing. I think we need to follow-up in a separate issue to enhance the UX in this area
  - The first install and the "Install as a new instance" workflows differ slightly. I don't think that this is an issue, but we should confirm the UX:
     - The first install of an app will not provide any option for setting the version in install step 1, the version is locked to the version selected on the previous page
     - The "Install as a new instance" introduces a dropdown for changing the version without needing to cancel the form to select a different version
- #17211: Confirmed that the old behavior is fixed and that the form starts from "empty", renders the "Install" button, and creates a new cluster without modifying the original

shell/pages/c/_cluster/apps/charts/chart.vue
momesgin: done

shell/pages/c/_cluster/apps/charts/chart.vue
momesgin: done

shell/mixins/chart.js
bpedersen2: Correct, this started to happen with 2.13

rak-phillip (approved):
The latest changes look good.

shell/mixins/chart.js
```diff
@@ -375,6 +386,9 @@ export default {
         // Use those values to check for a catalog app resource.
         // If found, set the form to edit mode. If not, set the
         // form to create mode.
+        // This is a hard blocker - installing a new instance is NOT allowed.
+
```
richard-cox, line 390: blank line can be removed, probably needs to go above the comment?

richard-cox (commented):
- in the chart detail page when switching between an app that has an upgrade and not, the new drop down component wiggles left/right
- in the chart detail page when installing a second chart with a lower version (or i guess higher), choosing 'Install as a new instance' does not select the version

@momesgin can you address the above points?

I gave the code as good an eyeball as i could spare and only spotted one very minor nit

momesgin:
> * in the chart detail page when switching between an app that has an upgrade and not, the new drop down component wiggles left/right
> * in the chart detail page when installing a second chart with a lower version (or i guess higher), choosing 'Install as a new instance' does not select the version
> 
> @momesgin can you address the above points?
> 
> I gave the code as good an eyeball as i could spare and only spotted one very minor nit

@richard-cox thanks for the review, for the two points you mentioned I'll discuss it with Ovi and address them along with the extra blank line you found in the coce 👍

PR #17228, opened 2026-04-13
Update usages of PCIC, reduce fetching all clusters

richard-cox:
### Summary
Contributes to #16602

- Remove places where we fetch all of provisioning.cattle.io.cluster (PCIC) or management.cattle.io.cluster (MCIC)
- Where possible migrate to using MCIC instead of PCIC
- Update places where we iterate through the whole store to find a resource with an id... to just use the byId getter

Additionally
- Buffed up the message we use to track waiting for clusters when creating them
- Removed more rke1 stuff
- Fix machine, machineset and machinesetdepolyment grouping (use common code for this)

- Import a harvester cluster and then create a downstream cluster within it. Enable `	
harvester-baremetal-container-workload` feature flag. Navigate to the downstream cluster. Below `Cluster Dashboard` click on Provider: Harvester --> it should take you to the harvester's host cluster in the harvester product
- create imported cluster works (return to cluster detail page)
- create rke2 custom cluster works (return to cluster detail page)
- Create an rke2 cluster
  - --> create/edit rke2 page --> loads without console errors
  - --> detail page --> loads without console errors
  - --> detail page --> create a snapshot --> wait for cluster to be active again --> restore snapshot
  - --> advanced --> machines --> group by cluster --> cluster groups should show name of cluster correctly
  - --> advanced --> machinesets --> group by cluster --> cluster groups should show name of cluster correctly
  - --> advanced --> machinedeployments --> group by cluster --> cluster groups should show name of cluster correctly
- Create some workloads successfully (checking general process)
- Create a Deployment --> ensure that the container tab's Networking `Add Port or Service` button can be pressed with no console errors
- Download a kubeconfig for any cluster --> Advanced --> Kubeconfigs --> click on the name of the cluster for the new kubeconfig (should take you through to cluster detail page
- user avatar --> Preferences --> Login Landing Page 'Take me to cluster' feature works (can select a cluster, on login taken to that cluser)
- Create an RKE2 cluster in a Harvester cluster --> find it in the side bar and click on it --> should go to the cluster but in the harvester product NOT explorer (wtf?!)
- About --> Diagnostics --> page correctly loads all clusters and when expanding `Resource Counts By Cluster` sections correctly show `Count`s
  - Note - Generate Response Times does work, however i've found requests to downstream clusters can get stuck in pending so the result only shows when they time out

Covered by above

### Checklist
- [x] The PR has a Milestone 


shell/components/LandingPagePreference.vue
```diff
@@ -90,13 +86,25 @@ export default {
   },
 
   methods: {
-    updateLoginRoute(neu) {
+    async updateLoginRoute(neu) {
       if (neu) {
         this.afterLoginRoute = neu;
       } else {
+        this.clusters = await this.$store.dispatch('management/findAll', { type: MANAGEMENT.CLUSTER });
```
richard-cox, line 93: This is kinda temporary. What we should be doing is using a paginated select component but... i think the plan is to remove all of this, so take the simple approach for the moment.

simpler approach is to still get all clusters, but only if the user explicitly wants to set one as the landing page (so we won't fetch all on every visit)

shell/plugins/steve/steve-pagination-utils.ts
```diff
@@ -258,14 +258,21 @@ class StevePaginationUtils extends NamespaceProjectFilters {
       { field: 'spec.displayName' },
       { field: `status.provider` },
       { field: `status.connected` },
+      { field: `status.info.machineProvider` },
+      { field: `status.driver` },
+      { field: `status.info.kubernetesVersion` },
+      { field: `spec.fleetWorkspaceName` },
```
richard-cox, line 265: this was missing from the previous PR that switched lists from PCIC to MCIC

shell/models/cluster.js
```diff
@@ -65,10 +65,10 @@ export default class NormanCluster extends NormanModel {
   }
 
   waitForProvisioning(timeout = 60000, interval) {
```
richard-cox, line 67: without the changes in shell/edit/provisioning.cattle.io.cluster/index.vue this part would trigger a failure.

i thought it worth while to buff up the message

richard-cox:
@codyrancher this is the final new functionality for single source of truth stuff for 2.15. Around half the lines changed are test related

shell/pages/diagnostic.vue
```diff
@@ -12,28 +12,25 @@ export default {
...
-
+    clusterForCounts.forEach((mgmtCluster, i) => {
       finalCounts.push({
-        id:             cluster.id,
-        name:           mgmtCluster?.spec?.displayName || cluster.metadata?.name,
-        namespace:      cluster.metadata?.namespace,
-        capiId:         cluster.mgmt?.id,
+        id:             mgmtCluster.id,
+        name:           mgmtCluster.spec.displayName,
+        namespace:      mgmtCluster.provClusterNamespace,
```
codyrancher: Lost the `|| cluster.metadata?.name` fallback. A cluster with empty/missing `spec.displayName` will render `undefined` in the diagnostic table.

cypress/e2e/po/pages/cluster-manager/cluster-manager-list.po.ts
```diff
@@ -34,24 +34,31 @@ export default class ClusterManagerListPagePo extends BaseListPagePo {
...
+    cy.intercept('GET', `/v1/management.cattle.io.clusters/${ mgmtClusterList.data[0].id }?*`, (req) => {
+      req.reply({
+        statusCode: 200,
+        body:       mgmtClusterList.data[0],
+      });
+    }).as('mgmtCluster');
+
+    // Covers both side bar and cluster list
+    cy.intercept('GET', '/v1/management.cattle.io.clusters?page=1&pagesize=10*', (req) => {
       req.reply({
```
codyrancher: `?page=1&pagesize=10*` is brittle. If pagination defaults shift (different page, page size, extra leading params) the intercept silently misses and tests hit the real backend. Prefer `{ pathname, query: { pagesize: '*' } }` like the [new prov-cluster intercept](https://github.com/rancher/dashboard/blob/3f55a6e94d1a2001d2131196c03a2d5816306fa1/cypress/e2e/blueprints/nav/fake-cluster.ts#L2528-L2531).

shell/assets/translations/en-us.yaml
```diff
@@ -2396,7 +2396,7 @@ cluster:
       label: Auto Replace
       toolTip: If greater than 0, nodes that are unreachable for this duration will be automatically deleted and replaced.
       unit: "Seconds"
-  managementTimeout: The cluster to become available. It's possible the cluster was created. We suggest checking the clusters page before trying to create another.
+  managementTimeout: The {type} {name} to become available. It's possible the cluster was created. We suggest checking the clusters page before trying to create another.
```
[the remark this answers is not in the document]
richard-cox: there's already a prefix, just not part of this string

shell/assets/translations/en-us.yaml
```diff
@@ -2396,7 +2396,7 @@ cluster:
...
   memberRoles:
```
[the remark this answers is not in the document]
codyrancher: I don't think it needs to be fixed but I did notice that we're using the clusterId in place of the name.

Image

codyrancher (approved):
I really didn't find any issues, solid pr.

PR #17254, opened 2026-04-15
Fix a few low-hanging a11y issues from the a11y test report

nwmac:
### Summary
Contributes towards #17279

Addresses a number of accessibility (a11y) violations identified by the existing accessibility e2e test suite.

With these changes, total violation count is reduced from 222 to 116.

The goal was to tackle the low-hanging groups of violations and remove whole groups, regardless of severity. 

- **Missing/empty dialog titles** – `AppModal` now accepts a `title` prop that is applied as `aria-label` on the dialog container. `PromptModal` derives a human-readable title from the component name and passes it through; `PromptRemove` passes an explicit translated label.
- **Empty headings in Settings** – `Setting.vue` now guards the `<h2>` description element so it is only rendered when a translation key resolves to a non-empty string, preventing empty heading violations.
- **Screen reader header for the Actions column** – `SortableTable/THead` adds a visually-hidden `<span class="sr-only">` inside the actions `<th>` so the column has an accessible name.
- **Invalid `aria-controls` references** – `CruResource.vue` now sets an `id` on each step container (`step-container-<name>`) so that `aria-controls` in `Wizard.vue` resolves to a real element. `HorizontalPanel` likewise scopes `aria-controls` with the panel position to avoid ID collisions.
- **Incorrect ARIA roles in the Wizard** – Step list items were using `role="presentation"`, which suppressed them from the accessibility tree; changed to `role="listitem"`.
- **Duplicate aria-labels in nav groups** – Nested `Group` components (e.g. items under "More Resources") now receive their parent's label as a `parentLabel` prop so each header announces as `"<Parent>, <Child>"`, making labels unique and contextually meaningful.
- **Incorrect role on the pin toggle** – `Pinned.vue` used `role="button"` on a toggle that represents a checked/unchecked state; corrected to `role="checkbox"`.
- **Missing landmark label on chart page** – The chart detail page `<main>` was replaced with a labelled `<region>` and the chart info `<aside>` was given an `aria-label`.
- **OCI logo alt text** – The alt text key for the OCI repository logo was corrected to `catalog.repo.target.oci.logo` so it describes the image rather than repeating the section title.
- **Cypress a11y test fix** – The kubectl-explain slide-in panel test now waits for animations and expands content before running the axe scan - this avoids an issue with the tooltip remaining visible and causing a11y issues.

### Technical notes summary

- The `AppModal` `title` prop is intentionally optional (defaults to `''`) to avoid breaking existing usages; callers should supply it for any dialog that lacks a visible heading.
- `PromptModal`'s `modalTitle` computed property derives a title only when `modalData.component` is a string (a registered component name). Object-based component definitions fall through and produce an empty string, which is safe — those modals are expected to carry their own visible heading.
- `role="navigation"` on the collapsible nav group header is the correct landmark for a nav section toggle.

Covered by the existing Cypress accessibility test suite (`cypress/e2e/tests/accessibility/shell.spec.ts`). No additional manual test steps required.

- Navigation sidebar collapsing/expanding behaviour
- Wizard / multi-step form navigation
- SortableTable column headers and sorting
- `PromptRemove` dialog (confirm delete flow)
- App/Helm chart detail page layout
- kubectl-explain slide-in panel

### Screenshot/Video

N/A — changes are non-visual (ARIA attributes, roles, and screen-reader-only text).

### A11y

After these changes, the a11y test report looks like this:

#### Accessibility Violations Summary

| Severity | Count |
|----------|-------|
| 🔴 Critical | 62 |
| 🟠 Serious | 37 |
| 🔵 Moderate | 17 |
| ⚪ Minor | 0 |
| **Total** | **116** |

#### Violations Detail

| Level | Violation | Count | Description |
|-------|-----------|-------|-------------|
| 🟠 Serious | `color-contrast` | 2 | Ensure the contrast between foreground and background colors meets WCAG 2 AA minimum contrast ratio thresholds |
| 🔵 Moderate | `heading-order` | 17 | Ensure the order of headings is semantically correct |
| 🔴 Critical | `aria-required-parent` | 51 | Ensure elements with an ARIA role that require parent roles are contained by them |
| 🔴 Critical | `aria-required-children` | 11 | Ensure elements with an ARIA role that require child roles contain them |
| 🟠 Serious | `listitem` | 34 | Ensure &lt;li&gt; elements are used semantically |
| 🟠 Serious | `nested-interactive` | 1 | Ensure interactive controls are not nested as they are not always announced by screen readers or can cause focus problems for assistive technologies |

### Checklist
- [x] The PR has a Milestone

### DIFF
--- a/shell/components/CruResource.vue
+++ b/shell/components/CruResource.vue
@@ -685,10 +685,10 @@ export default {
               >
                 <div
                   v-if="step.name === activeStep.name || step.hidden"
+                  :id="'step-container-' + step.name"
                   :key="step.name"
                   class="step-container__step"
                   :class="{'hide': step.name !== activeStep.name && step.hidden}"
-                  :id="'step-container-' + step.name"
                 >
                   <slot
                     :step="step"
--- a/shell/components/CruResource.vue
+++ b/shell/components/CruResource.vue
@@ -688,6 +688,7 @@ export default {
                   :key="step.name"
                   class="step-container__step"
                   :class="{'hide': step.name !== activeStep.name && step.hidden}"
+                  :id="'step-container-' + step.name"
                 >
                   <slot
                     :step="step"
--- a/shell/components/PromptModal.vue
+++ b/shell/components/PromptModal.vue
@@ -71,6 +71,18 @@ export default {
     },
     modalName() {
       return this.modalData?.modalName;
+    },
+
+    /**
+     * When the component is a string, we can generate a title from the name of the component which we can then use for the aria-label attribute.
+     * For example, if the component is "MyComponentDialog", we generate a title of "My Component Dialog"
+     */
+    modalTitle() {
+      if (typeof this.modalData?.component === 'string') {
+        let name = this.modalData.component;
+
+        return name.replace(/([A-Z])/g, ' $1').trim()
+      }
     }
   },
 
@@ -124,6 +136,7 @@ export default {
     :custom-class="customClass"
     :styles="styles"
     :height="height"
+    :title="modalTitle"
     :trigger-focus-trap="true"
     :return-focus-selector="returnFocusSelector"
     :return-focus-first-iterable-node-selector="returnFocusFirstIterableNodeSelector"
--- a/shell/components/PromptModal.vue
+++ b/shell/components/PromptModal.vue
@@ -79,10 +79,11 @@ export default {
      */
     modalTitle() {
       if (typeof this.modalData?.component === 'string') {
-        let name = this.modalData.component;
-
-        return name.replace(/([A-Z])/g, ' $1').trim()
+        // Insert a space before all caps and trim any leading/trailing whitespace
+        return this.modalData.component.replace(/([A-Z])/g, ' $1').trim();
       }
+
+      return '';
     }
   },
 
--- a/shell/components/Wizard.vue
+++ b/shell/components/Wizard.vue
@@ -358,7 +358,7 @@ export default {
 
                     :id="step.name"
                     :class="{step: true, active: step.name === activeStep.name, disabled: !isAvailable(step)}"
-                    role="presentation"
+                    role="listitem"
                   >
                     <span
                       :aria-controls="'step-container-' + step.name"
--- a/shell/edit/catalog.cattle.io.clusterrepo.vue
+++ b/shell/edit/catalog.cattle.io.clusterrepo.vue
@@ -67,7 +67,7 @@ export default {
       {
         id:      CLUSTER_REPO_TYPES.OCI_URL,
         header:  { title: { key: 'catalog.repo.target.oci.title' } },
-        image:   { src: requireAsset('@shell/assets/images/providers/oci-open-containers.svg'), alt: { key: 'catalog.repo.target.oci.title' } },
+        image:   { src: requireAsset('@shell/assets/images/providers/oci-open-containers.svg'), alt: { key: 'catalog.repo.target.oci.title.logo' } },
         content: { key: 'catalog.repo.target.oci.description' },
       },
     ];
--- a/shell/pages/c/_cluster/apps/charts/chart.vue
+++ b/shell/pages/c/_cluster/apps/charts/chart.vue
@@ -227,7 +227,7 @@ export default {
 
 <template>
   <Loading v-if="$fetchState.pending" />
-  <main v-else>
+  <div v-else>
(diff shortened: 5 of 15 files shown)


shell/components/PromptModal.vue
```diff
@@ -71,6 +71,19 @@ export default {
...
+    },
+
+    /**
+     * When the component is a string, we can generate a title from the name of the component which we can then use for the aria-label attribute.
+     * For example, if the component is "MyComponentDialog", we generate a title of "My Component Dialog"
+     */
+    modalTitle() {
+      if (typeof this.modalData?.component === 'string') {
+        // Insert a space before all caps and trim any leading/trailing whitespace
+        return this.modalData.component.replace(/([A-Z])/g, ' $1').trim();
```
rak-phillip, line 83: nit: Lodash provides a `startCase()` function which is more readable at a glance.

```suggestion
        return _.startCase(this.modalData.component).trim();
```

shell/pages/c/_cluster/apps/charts/chart.vue
```diff
@@ -227,7 +227,10 @@ export default {
 
 <template>
   <Loading v-if="$fetchState.pending" />
-  <main v-else>
+  <region
+    v-else
+    :aria-label="chart.chartNameDisplay"
+  >
```
rak-phillip, line 233: hmmm... I don't think that the region tag is right and I don't believe that it's a valid html element - what does exist is a region role[^1] that can be applied to a div or section tag.

I think that a div would fit best for this page and will be consistent with the design of other pages used throughout Dashboard. I'm also thinking that the additional `aria-label` will be redundant, causing screen readers to announce the region along with the h1 on line 253

[^1]: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/region_role#description

shell/edit/catalog.cattle.io.clusterrepo.vue
```diff
@@ -67,7 +67,7 @@ export default {
       {
         id:      CLUSTER_REPO_TYPES.OCI_URL,
         header:  { title: { key: 'catalog.repo.target.oci.title' } },
-        image:   { src: requireAsset('@shell/assets/images/providers/oci-open-containers.svg'), alt: { key: 'catalog.repo.target.oci.title' } },
+        image:   { src: requireAsset('@shell/assets/images/providers/oci-open-containers.svg'), alt: { key: 'catalog.repo.target.oci.title.logo' } },
```
rak-phillip, line 70: This doesn't match the new key added to `shell/assets/translations/en-us.yaml`:

```suggestion
        image:   { src: requireAsset('@shell/assets/images/providers/oci-open-containers.svg'), alt: { key: 'catalog.repo.target.oci.logo' } },
```

shell/components/Wizard.vue
```diff
@@ -358,7 +358,7 @@ export default {
 
                     :id="step.name"
                     :class="{step: true, active: step.name === activeStep.name, disabled: !isAvailable(step)}"
-                    role="presentation"
+                    role="listitem"
```
rak-phillip, line 361: Which issues is this change addressing? Consider that the usage of the `presentation`[^1] might be intentional and exists to to strip the conflicting list semantics from the tab-like design of the wizard. 

I think that updating the role of the `ul` container to `role="tablist"`[^2] might be the more appropriate choice, given the context.

[^2]: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/tablist_role
[^1]: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/presentation_role

shell/components/CruResource.vue
```diff
@@ -685,6 +685,7 @@ export default {
               >
                 <div
                   v-if="step.name === activeStep.name || step.hidden"
+                  :id="'step-container-' + step.name"
```
rak-phillip, line 688: This change looks unexpected given the rest of the context for this PR. What issue is this addressing?

rak-phillip: I think that I understand - this is overriding the entire slot and the consumer is responsible for specify the ids and match the naming convention enforced by `aria-controls` in the wizard..

I think this fix makes sense, but the pattern, as implemented, requires the consumer to know about internal implementation details of the Wizard and is prone to error.

nwmac:
Closing out - covered by other work

PR #17274, opened 2026-04-15
Ignore dependabot bumps to agentic workflow dependencies

rak-phillip:
### Summary
This adds `gh-aw` dependencies to the list of dependencies for dependabot to ignore.

- Ignore dependabot bumps to agentic workflow dependencies

### Technical notes summary

This adds `github/gh-aw-actions` and `github/gh-aw` to the list of ignored dependencies[^1] for dependabot bumps. Managing these dependencies should be done via the `gh aw` cli.

[^1]: https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#ignore--

Monitor dependabot bumps to confirm that PRs aren't raised against these dependencies. 

NA

### Screenshot/Video

NA

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,6 +10,9 @@ updates:
       default-days: 7
     open-pull-requests-limit: 5
     labels: ["component/dependencies"]
+    ignore:
+      - dependency-name: "github/gh-aw-actions"
+      - dependency-name: "github/gh-aw"
 
   # Maintain dependencies for dashboard, shell, and packages
   - package-ecosystem: "npm"


marcelofukumoto (approved):
It is good idea to remove the noise for now of those actions.
We just need to think what process we will do to bump those gh aw in the future.

I believe later we can let the dependabot check those, and we go and run a gh aw instead of accepting the PR. But without a plan in mind would be better to just remove it since if we merge there are big risks that will stop working.

LGTM

rak-phillip:
> It is good idea to remove the noise for now of those actions. We just need to think what process we will do to bump those gh aw in the future.
> 
> I believe later we can let the dependabot check those, and we go and run a gh aw instead of accepting the PR. But without a plan in mind would be better to just remove it since if we merge there are big risks that will stop working.
> 
> LGTM

Yes, I agree. Let's work on developing a process for dependencies that require manual/additional actions to properly bump. I think that we need to clearly communicate what needs to be done so that all maintainers will understand how to manage agentic workflow deps, as well as core frameworks like Vue.

Issue #17159, opened 2026-04-08
Harvester ui extension auto install sometimes failed in first run


a110605:
**Setup**
- Rancher version: 2.10.0 ~ 2.14.0
- Rancher UI Extensions: harveter ui ext
- Browser type & version: Chrome

**Describe the bug**
Installing the harvester-ui-extension via the single click button in the Rancher UI always fails on first run for me. Espescilly for fresh installed Rancher.

But it's success for the second or third tries.

**To Reproduce**
1. Setup a fresh installed Rancher via docker or helm.
2. Go to Virtualization Management page
3. Click install button

**Result**
Auto install failed. 

<img width="1467" height="802" alt="Image" src="https://github.com/user-attachments/assets/16b2078a-86b0-4af9-aada-a9efcd2cbf65" />

**Expected Result**

Should successfully installed if no network or resource race condition.

**Screenshots**

**Additional context**

yonasberhe23:
@a110605 can this be backported to 2.14?

a110605:
> [@a110605](https://github.com/a110605) can this be backported to 2.14?

Let's put in 2.15.0 as 2.14.1 is code freeze.

yonasberhe23:
@a110605 I understand 2.14.1 is code frozen, so this request is for the next 2.14.x patch release (for example, 2.14.2), not 2.14.1.
The same auto-install issue exists on 2.14, so backporting the fix would help keep supported versions aligned.

yonasberhe23:
Tested on 
- Rancher v2.15-ad16989777846bbbab7e2d71379bdcf6909a251a-head
- Dashboard master `d70abc9`

Harvester auto-install is successful on first and subsequent attempts. 
Up to 10 retries have been seen in console log.
Console error thrown after 10th retry attempt.
UI messaging reflects installation failure as expected.


https://github.com/user-attachments/assets/37c1fe63-115b-4d7c-b990-3e688daaf215

PR #17301, opened 2026-04-17
Automation: Update harvester tests for prime

yonasberhe23:
### Summary
Fixes #https://github.com/rancher/qa-tasks/issues/2282

1. Drive catalog setup, intercepts, waits, and cleanup from cy.getRancherVersion() so the spec works on both Community and Rancher Prime builds (Prime uses the rancher / ui-plugin-charts catalog instead of harvester / harvester-ui-extension).
2. Removed the install retry now that [rancher/dashboard#17159](https://github.com/rancher/dashboard/issues/17159) fixed the underlying flakiness; the spec now just asserts the install flow without the 500-retry path.

### Screenshot/Video
image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/cypress/e2e/tests/pages/virtualization-mgmt/harvester.spec.ts
+++ b/cypress/e2e/tests/pages/virtualization-mgmt/harvester.spec.ts
@@ -3,6 +3,7 @@ import { HarvesterClusterDetailsPo, HarvesterClusterPagePo } from '@/cypress/e2e
 import RepositoriesPagePo from '@/cypress/e2e/po/pages/chart-repositories.po';
 import { LONG_TIMEOUT_OPT, MEDIUM_TIMEOUT_OPT } from '@/cypress/support/utils/timeouts';
 import { CLUSTER_REPOS_BASE_URL } from '@/cypress/support/utils/api-endpoints';
+import { qase } from '~/cypress/support/qase';
 
 const extensionsPo = new ExtensionsPagePo();
 const harvesterPo = new HarvesterClusterPagePo();
@@ -30,6 +31,27 @@ function harvesterExtensionCatalog(version: Cypress.RancherVersion) {
 }
 
 describe('Harvester', { tags: ['@virtualizationMgmt', '@adminUser'] }, () => {
+  before(() => {
+    cy.login();
+
+    // Clean up stale Harvester extension app and extension source repos to prevent test collisions
+    cy.createRancherResource('v1', 'catalog.cattle.io.apps/cattle-ui-plugin-system/harvester?action=uninstall', {}, false);
+
+    cy.getRancherResource('v1', 'catalog.cattle.io.clusterrepos').then((resp: Cypress.Response<any>) => {
+      const extensionRepoUrls = ['rancher/ui-plugin-charts', 'harvester/harvester-ui-extension'];
+      const matchingRepoIds = (resp.body?.data || [])
+        .filter((repo: any) => extensionRepoUrls.some((url) => repo?.spec?.gitRepo?.includes(url)))
+        .map((repo: any) => repo?.id)
+        .filter((id: string | undefined) => !!id);
+
+      cy.wrap(matchingRepoIds, { log: false }).each((repoId: string) => {
+        cy.deleteRancherResource('v1', 'catalog.cattle.io.clusterrepos', repoId);
+        cy.waitForRancherResource('v1', 'catalog.cattle.io.clusterrepos', repoId, (resourceResp) => resourceResp.status === 404, 10, { failOnStatusCode: false })
+          .should('eq', true);
+      });
+    });
+  });
+
   beforeEach(() => {
     cy.login();
     cy.getRancherVersion().then((version) => {
@@ -40,14 +62,7 @@ describe('Harvester', { tags: ['@virtualizationMgmt', '@adminUser'] }, () => {
     });
   });
 
-  /**
-   * Assumes that Harvester Extension is NOT installed
-   *
-   * Harvester Extension will also be removed after all tests run
-   *
-   * (pattern needs fixing)
-   */
-  it('can auto install harvester and begin process of importing a harvester cluster', () => {
+  qase(7020, it('can auto install harvester and begin process of importing a harvester cluster', () => {
     cy.get<Cypress.RancherVersion>('@rancherVersion').then((version) => {
       const catalog = harvesterExtensionCatalog(version);
       const chartRepo = catalog.repo;
@@ -117,9 +132,9 @@ describe('Harvester', { tags: ['@virtualizationMgmt', '@adminUser'] }, () => {
         cy.deleteRancherResource('v1', 'provisioning.cattle.io.clusters', `fleet-default/${ harvesterClusterId }`);
       });
     });
-  });
+  }));
 
-  it('missing repo message should display when repo does NOT exist', () => {
+  qase(7021, it('missing repo message should display when repo does NOT exist', () => {
     cy.get<Cypress.RancherVersion>('@rancherVersion').then((version) => {
       const catalog = harvesterExtensionCatalog(version);
       const chartRepo = catalog.repo;
@@ -193,9 +208,9 @@ describe('Harvester', { tags: ['@virtualizationMgmt', '@adminUser'] }, () => {
       harvesterPo.updateOrInstallButton().checkVisible();
       harvesterPo.extensionWarning().should('have.text', 'The Harvester UI Extension is not installed');
     });
-  });
+  }));
 
-  it('able to update harvester extension version', () => {
+  qase(7022, it('able to update harvester extension version', () => {
     cy.get<Cypress.RancherVersion>('@rancherVersion').then((version) => {
       const catalog = harvesterExtensionCatalog(version);
       const chartRepo = catalog.repo;
@@ -280,12 +295,16 @@ describe('Harvester', { tags: ['@virtualizationMgmt', '@adminUser'] }, () => {
         extensionsPo.extensionCardHeaderStatusTooltip(harvesterTitle, 0).waitForTooltipWithText(`Installed (${ versions[0] })`);
       });
     });
-  });
+  }));
 
   afterEach(() => {
+    // Ensure Harvester extension is uninstalled and repo is deleted
     cy.createRancherResource('v1', 'catalog.cattle.io.apps/cattle-ui-plugin-system/harvester?action=uninstall', {}, false);
     cy.get<Cypress.RancherVersion>('@rancherVersion').then((version) => {
       cy.deleteRancherResource('v1', 'catalog.cattle.io.clusterrepos', harvesterExtensionCatalog(version).repo, false);
+      // Verify deletion completed before proceeding (with retries)
+      cy.waitForRancherResource('v1', 'catalog.cattle.io.clusterrepos', harvesterExtensionCatalog(version).repo, (resp) => resp.status === 404, 10, { failOnStatusCode: false })
+        .should('eq', true);
     });
   });
 });
--- a/cypress/e2e/tests/pages/virtualization-mgmt/harvester.spec.ts
+++ b/cypress/e2e/tests/pages/virtualization-mgmt/harvester.spec.ts
@@ -9,30 +9,24 @@ const harvesterPo = new HarvesterClusterPagePo();
 const appRepoList = new RepositoriesPagePo(undefined, 'manager');
 
 let harvesterClusterName = '';
-const harvesterGitRepoName = 'harvester';
 const harvesterTitle = 'Harvester';
-const branchName = 'gh-pages';
-const harvesterGitRepoUrl = 'https://github.com/harvester/harvester-ui-extension.git';
 
-/**
- * Conditional retry for Harvester extension installation based on UI warning message
- * Checks for the "Warning, Harvester UI extension automatic installation failed" message
- * and reloads the page to retry if detected
- * Need conditional retry until this is resolved https://github.com/rancher/dashboard/issues/13093
- */
-function conditionalRetryHarvesterInstallation() {
-  // Check if the installation failed warning is displayed
-  harvesterPo.extensionWarning().then(($warning) => {
-    if ($warning.text().includes('Warning, Harvester UI extension automatic installation failed')) {
-      cy.log('Detected Harvester installation failed warning, reloading page and retrying...');
-      cy.reload();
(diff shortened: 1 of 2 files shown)


a-arias (dismissed):
LGTM

cypress/e2e/tests/pages/virtualization-mgmt/harvester.spec.ts
```diff
@@ -51,237 +45,247 @@ describe('Harvester', { tags: ['@virtualizationMgmt', '@adminUser'] }, () => {
...
-
-    // add harvester repo
-    cy.createRancherResource('v1', 'catalog.cattle.io.clusterrepos', {
-      type:     'catalog.cattle.io.clusterrepo',
-      metadata: { name: harvesterGitRepoName },
-      spec:     {
-        clientSecret: null, gitRepo: harvesterGitRepoUrl, gitBranch: branchName
+      // install harvester extension
+      harvesterPo.updateOrInstallButton().click();
+      if (!isPrime) {
```
IsaSih: Why do you cy.wait the POST only for community?

The test always intercept POST CLUSTER_REPOS_BASE_URL as @createChart, so if Prime never does the POST, but if it still does, the alias will capture requests that are never asserted on. 

Is it expected that this POST does not happen on Prime? Shouldn't be also be asserting that?

cypress/e2e/tests/pages/virtualization-mgmt/harvester.spec.ts
```diff
@@ -51,237 +45,247 @@ describe('Harvester', { tags: ['@virtualizationMgmt', '@adminUser'] }, () => {
...
-
-      const harvesterClusterId = response.body.id;
-      const harvesterDetails = new HarvesterClusterDetailsPo(undefined, undefined, harvesterClusterId);
-
-      harvesterDetails.waitForPage(null, 'registration');
-      harvesterDetails.title().should('contain', harvesterClusterName);
-
-      // navigate to harvester list page and verify the logo and tagline do not display after cluster created
-      HarvesterClusterPagePo.navTo();
+    cy.getRancherVersion().then((version) => {
```
IsaSih: if anything triggers network traffic before the version callback runs, the test will fail. test maintenance can be trickier if there are navigation refactors (API/VAI refactors).
As all the test code relies on this callback, what do you think about moving the 
`cy.getRancherVersion().as('version')` to a beforeEach, then in the test read it and set up intercepts before navigation/clicks?

cypress/e2e/tests/pages/virtualization-mgmt/harvester.spec.ts
```diff
@@ -9,30 +9,24 @@ const harvesterPo = new HarvesterClusterPagePo();
...
+
+// Cluster chart repository that supplies the Harvester UI extension (repo id, Git URL, branch)—differs for Community vs Prime.
+const HARVESTER_EXTENSION_CATALOG = {
+  community: {
+    repo:      'harvester',
+    gitRepo:   'https://github.com/harvester/harvester-ui-extension.git',
+    gitBranch: 'gh-pages',
+  },
+  prime: {
+    repo:      'rancher',
```
IsaSih, line 23: If Rancher Prime environments sometimes already have a cluster repo named rancher for other reasons, this test could become conflictive. I'm unsure if that's gonna be the case but naming the repo 'rancher' could lead to this risk? what do you think?

Maybe ensure that there's no other `catalog.cattle.io.clusterrepos/rancher `or make sure that they're guaranteed to be the plugin charts catalog? 

Same with the community chart repo

cypress/e2e/tests/pages/virtualization-mgmt/harvester.spec.ts
yonasberhe23: thats a good call out. the POST is expected in both Prime and Community. [updated](https://github.com/rancher/dashboard/compare/dca42ac32edfd69be273d65f223e5aa5ea37f77e..f64cfa0f940463fc3f13929e6595be6484a61b24)

cypress/e2e/tests/pages/virtualization-mgmt/harvester.spec.ts
yonasberhe23: great call, [update](https://github.com/rancher/dashboard/pull/17301/changes/c776ec7cde6c8866a7a20add5e812d6a7959c9cd)

cypress/e2e/tests/pages/virtualization-mgmt/harvester.spec.ts
yonasberhe23: good catch. addressed by adding a before hook that cleans up any stale Harvester extension and chart repos before the suite runs, [updated](https://github.com/rancher/dashboard/pull/17301/commits/a1c1cf66ec39eacba86a034d9ef970e17c6c3de2)

cypress/e2e/tests/pages/virtualization-mgmt/harvester.spec.ts
```diff
@@ -9,279 +9,287 @@ const harvesterPo = new HarvesterClusterPagePo();
...
 }
 
 describe('Harvester', { tags: ['@virtualizationMgmt', '@adminUser'] }, () => {
+  before(() => {
+    cy.login();
+    cy.getRancherVersion().then((version) => {
+      const catalog = harvesterExtensionCatalog(version);
+
+      // Clean up any stale Harvester extension and chart repos to prevent test collisions
+      cy.createRancherResource('v1', 'catalog.cattle.io.apps/cattle-ui-plugin-system/harvester?action=uninstall', {}, false);
```
IsaSih: After careful analysis, I'm wondering if we should have an assertion confirming that the chart repos come pre-installed in rancher prime. By doing this cleanup, we're missing this validation, as rancher prime is expected to have these repos installed.

cypress/e2e/tests/pages/virtualization-mgmt/harvester.spec.ts
```diff
@@ -9,279 +9,287 @@ const harvesterPo = new HarvesterClusterPagePo();
...
+      cy.deleteRancherResource('v1', 'catalog.cattle.io.clusterrepos', catalog.repo, false);
```
IsaSih: This cleanup structure might introduce regressions for rancher prime. Harvester in prime, is part of the rancher chart repo, which contains other relevant charts. Deleting it from the environment may cause unintended side effects. Maybe, a safer approach could be to verify if the rancher repo exists first. And intercepting `cy.intercept('POST', CLUSTER_REPOS_BASE_URL).as('createChart');` for prime only if we need to add the rancher chart repo.

IsaSih: In before() we're deleting the repo, but not wait/assert that it’s actually gone.
Even if deletion is “requested”, the repo could still be present for a short time, and the UI might decide “repo exists” and skip create. 
So we should verify the repo is actually gone (via API) like:
`cy.getRancherResource('v1','catalog.cattle.io.clusterrepos', chartRepo, 404) `
But again, this is safer for community env, as the harvester extension comes from its own cluster repo.

IsaSih (dismissed):

cypress/globals.d.ts
```diff
@@ -73,6 +73,10 @@ export interface CreateResourceNameOptions {
 
 declare global {
   namespace Cypress {
+    interface RancherVersion {
+      RancherPrime?: string;
```
IsaSih, line 77: this could be addressed in a separate PR if needed. I just want to understand a bit more why we are treating this as a string and not a boolean

cypress/e2e/tests/pages/virtualization-mgmt/harvester.spec.ts
yonasberhe23: That is not what I'm seeing in Prime. On a fresh Rancher Prime install, the preinstalled repos are Partners, Rancher Prime, and RKE2. The rancher repo in this test refers to the UI-plugin charts source, which is added by the Harvester auto-install flow, not shipped by default.

The before() cleanup is necessary because this repo is not guaranteed to be clean between test runs. If a previous test failed, it may have left the repo behind. So the cleanup is the right approach to make the test deterministic—not a side effect we should avoid.

Your point about deletion verification is valid though. We should wait to confirm the repo is actually gone before the tests run, not just request the deletion. I can add a check after the deletion to assert it's gone.

yonasberhe23: see https://github.com/rancher/dashboard/pull/17301#discussion_r3141523281

cypress/e2e/tests/pages/virtualization-mgmt/harvester.spec.ts
yonasberhe23: see https://github.com/rancher/dashboard/pull/17301#discussion_r3141523281

cypress/globals.d.ts
yonasberhe23: RancherPrime is intentionally typed as string because /rancherversion returns it as a string ("true"/"false")

izaac:
let's squash it, it's looking good

cypress/e2e/tests/pages/virtualization-mgmt/harvester.spec.ts
IsaSih: I see what you mean. I apologize I got confused as I'm working on the elemental test fixes and it's the elemental chart that comes in a repo that is pre-intalled in prime. However, by analyzing the installation of the repo that contains the harvester extension, I identified that it can have another name if installed via the extensions page in prime:
Screenshot 2026-04-27 at 6 52 56 PM

IsaSih: given that, deleting the repo named rancher is not going to do a fully cleanup if there's an existing "Rancher Extensions" repo installed via the extensions page. if Rancher Extensions exists, the harvester UI extension seems to be installed from rancher repo with no issues, but maybe we should modify the cleanup step to delete all existing repositories with the `ui-plugin-charts` url instead of by the repo object from the test

cypress/e2e/tests/pages/virtualization-mgmt/harvester.spec.ts
IsaSih: I agree. Please see https://github.com/rancher/dashboard/pull/17301#discussion_r3163762472

IsaSih (changes requested):

cypress/e2e/tests/pages/virtualization-mgmt/harvester.spec.ts
IsaSih: https://github.com/user-attachments/assets/0b347900-6650-4acc-9327-06ad02f2cc79

cypress/e2e/tests/pages/virtualization-mgmt/harvester.spec.ts
```diff
@@ -3,285 +3,298 @@ import { HarvesterClusterDetailsPo, HarvesterClusterPagePo } from '@/cypress/e2e
...
+      });
     });
-  });
+  }));
 
   afterEach(() => {
     cy.createRancherResource('v1', 'catalog.cattle.io.apps/cattle-ui-plugin-system/harvester?action=uninstall', {}, false);
-    cy.deleteRancherResource('v1', 'catalog.cattle.io.clusterrepos', harvesterGitRepoName, false);
+    cy.get<Cypress.RancherVersion>('@rancherVersion').then((version) => {
+      cy.deleteRancherResource('v1', 'catalog.cattle.io.clusterrepos', harvesterExtensionCatalog(version).repo, false);
```
izaac: It might be worth tossing in a `waitFor` here to poll that `clusterrepo` until it hits a `404`. It’s a safer way to make sure the cleanup actually finished before we move on to the next test.

cypress/e2e/tests/pages/virtualization-mgmt/harvester.spec.ts
yonasberhe23: Ah I see, I wasn't aware of that. Going to update the code based on this information you've shared. Thanks!

izaac (approved):
:shipit:

IsaSih (approved):
Approved. I took a look at all our specs that could be aftected by removing the cluster repo in an after each () to see if there could be any regressions when it comes to tests that would assume that this repo already exists. All of them are currently adding the repo before each so, we should be good:
https://github.com/search?q=repo%3Arancher%2Fdashboard+path%3Acypress+ui-plugin-charts&type=code

cypress/e2e/tests/pages/virtualization-mgmt/harvester.spec.ts
```diff
@@ -3,285 +3,308 @@ import { HarvesterClusterDetailsPo, HarvesterClusterPagePo } from '@/cypress/e2e
...
+
+  qase(7021, it('missing repo message should display when repo does NOT exist', () => {
+    cy.get<Cypress.RancherVersion>('@rancherVersion').then((version) => {
+      const catalog = harvesterExtensionCatalog(version);
+      const chartRepo = catalog.repo;
+
+      cy.intercept('POST', `${ CLUSTER_REPOS_BASE_URL }/${ chartRepo }?action=install`).as('installHarvesterExtension');
+      cy.intercept('PUT', `${ CLUSTER_REPOS_BASE_URL }/${ chartRepo }`).as('updateHarvesterChart');
+
+      cy.createRancherResource('v1', 'catalog.cattle.io.clusterrepos', {
```
IsaSih, line 145: One non-blocking suggestion:
Keep the API creation, but standardize via a small helper local to the spec, e.g. createHarvesterClusterRepo(catalog) so tests 2 and 3 don’t duplicate the payload.

IsaSih (approved):

IsaSih:
I also verified if the extension repo removal in the `aftereach()` could cause any regressions or fail other tests that could potentially rely on the `ui-plugin-charts` repo, as in https://github.com/search?q=repo%3Arancher%2Fdashboard+path%3Acypress+ui-plugin-charts&type=code. 
But they seem to be also assuming the chart is not pre-installed so we should be good :)

cypress/e2e/tests/pages/virtualization-mgmt/harvester.spec.ts
yonasberhe23: Good suggestion. Since it’s not urgent and the duplication is minor, I’ll address it in a separate PR.

PR #17329, opened 2026-04-20
Improve styling of helm apps upgrade badge used in the apps list

### LINKED ISSUE #17328: Minor formatting issue on the upgrade badge for helm apps list
See below:

Image

For long version numbers, the icon is lost at the end.

We should fix this and improve the styling slightly.

nwmac:
### Summary
Fixes #17328

Improves the styling of the upgrade-available badge shown in the Helm apps list so that long version strings are handled gracefully and the badge renders consistently with other badges in the UI.

- Updated the upgrade badge in [shell/list/catalog.cattle.io.app.vue](shell/list/catalog.cattle.io.app.vue) used for the `catalog.cattle.io.app` list.
- Switched the badge element from a `<span>` to a `<div>` wrapping the version label and the upload icon, so flex layout and text truncation work correctly.
- Added a tooltip (`v-clean-tooltip`) that displays the full `upgradeAvailableVersion` on hover, so the full version is accessible even when the visible text is truncated.
- Added a new `app-upgrade-badge` class with scoped SCSS rules:
  - `display: inline-flex` with `align-items: center` to vertically align the version text and icon.
  - `border-radius: var(--border-radius)` and `padding: 2px 5px` for consistent badge styling.
  - Inner `<div>` uses `overflow: hidden`, `text-overflow: ellipsis`, `white-space: nowrap`, and `min-width: 0` so long version strings truncate with an ellipsis instead of breaking the row layout.
  - Icon uses `flex-shrink: 0` to ensure it is never squashed when the container narrows.
- Converted the `<style>` block from plain CSS to `lang="scss"` to allow the nested selectors.

### Screenshot/Video

With this change, now renders like this:

image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/list/catalog.cattle.io.app.vue
+++ b/shell/list/catalog.cattle.io.app.vue
@@ -40,14 +40,15 @@ export default {
     data-testid="installed-app-catalog-list"
   >
     <template #cell:upgrade="{row}">
-      <span
+      <div
         v-if="row.upgradeAvailable === APP_UPGRADE_STATUS.SINGLE_UPGRADE"
-        class="badge-state bg-warning hand"
+        v-clean-tooltip="row.upgradeAvailableVersion"
+        class="badge-state bg-warning hand app-upgrade-badge"
         @click="row.goToUpgrade(row.upgradeAvailableVersion)"
       >
-        {{ row.upgradeAvailableVersion }}
+        <div>{{ row.upgradeAvailableVersion }}</div>
         <i class="icon icon-upload" />
-      </span>
+      </div>
       <span
         v-else-if="row.upgradeAvailable === APP_UPGRADE_STATUS.NOT_APPLICABLE"
         v-t="'catalog.app.managed'"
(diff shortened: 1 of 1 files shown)


shell/list/catalog.cattle.io.app.vue
```diff
@@ -69,8 +70,27 @@ export default {
...
+<style scoped lang="scss">
 .apps :deep() .state-description{
   color: var(--error)
 }
+
+.badge-state.app-upgrade-badge {
+  display: inline-flex;
+  align-items: center;
+  border-radius: var(--border-radius);
+  padding: 2px 5px;
```
momesgin: could we change the `5px` to `4px`? for spacing, we should use numbers multiplied by 2, e.g. 2, 4, 8, 16,...

momesgin: our rule could be 2px, 4px, **6px**, 8px, **10px**,... but will confirm with Ovi

momesgin: got the answer from Ovi:

> it’s 4, 8, 12, 16, 24, 32, 40, 48, 56, 64,, …
In other words, it the classic 8 point grid system with an added, smaller granularity for small sizing (that’s why you get the 4px and 12px in there

so we should go with 4px

momesgin (dismissed):
Only one comment. If the check isn't necessary feel free to merge.

shell/list/catalog.cattle.io.app.vue
nwmac: Okay - will update

momesgin (approved):

PR #17330, opened 2026-04-20
Prod reg part 2: implement remanining DSL methods

### LINKED ISSUE #16834: Feature: new product registration, part 2 - covering remaining of the DSL methods

This tracks the continuation of the work done in https://github.com/rancher/dashboard/issues/16046 with the following points:

- add support for `headers`
- cover `mapGroup` DSL method **->** regExp/ID to remap group display names in the side-menu  (keep as internal)
- cover `ignoreGroup` DSL method **->** don't display a given group nav  (keep as internal)
- cover `mapType` DSL method **->** rename a specific type to a given name/label
- cover `ignoreType` DSL method **->** don't display a specific type on the side nav
- cover `hideBulkActions` DSL method **->** bool to control the display of bulk actions for a given resource page
- cover `moveType` DSL method **->** move a resource-type page to a given group  (keep as internal)

**Additional context**

### LINKED ISSUE #17426: Bug: `hideBulkActions` won't work on pages with custom list view where `table-actions` is set to `true` on `ResourceTable`
**Setup**
- Rancher version: v2.14 / master (v2.15 development)
- Rancher UI Extensions: Product Registration API (experimental)
- Browser type & version: N/A (code-level issue)

**Describe the bug**

The DSL `hideBulkActions` method correctly sets `state.hideBulkActions[type] = true` in the `type-map` store, and the `ResourceTable` component's `_showBulkActions` computed correctly checks this via `type-map/hideBulkActionsFor`. However, some resource types have **custom list components** (under `shell/list/`) that render their own `<ResourceTable>` with a hardcoded `:table-actions="true"` prop. This prop takes precedence over the store-based check, making the `hideBulkActions` DSL method ineffective for those resource types.

**To Reproduce**

1. In an extension, register a resource page for `provisioning.cattle.io.cluster` with `hideBulkActions: true` (either via the new product registration API or the legacy DSL)
2. Navigate to the resource list page
3. Observe that bulk action checkboxes are still visible

**Result**

Bulk action checkboxes are still shown in the table despite `hideBulkActions` being set to `true` in the store.

The `_showBulkActions` computed in `ResourceTable.vue` exits early at the `tableActions !== null` check because the custom list component at `shell/list/provisioning.cattle.io.cluster.vue` (line 241) passes `:table-actions="true"` directly. This means the store-based `hideBulkActionsFor` getter is never consulted.

**Expected Result**

When `hideBulkActions(type, true)` is called via the DSL, bulk action checkboxes should be hidden for that resource type, regardless of whether the resource type has a custom list component.

**Screenshots**
N/A

**Additional context**

The affected file is `shell/list/provisioning.cattle.io.cluster.vue`, but this pattern likely exists in other custom list components as well. A quick search shows multiple custom list components using hardcoded `:table-actions` values.

Possible fixes:
- **Option A:** In custom list components that hardcode `:table-actions="true"`, replace it with a computed that also checks `type-map/hideBulkActionsFor` — respecting the DSL override while preserving the default behavior
- **Option B:** In `ResourceTable.vue`, change `_showBulkActions` so that the store-based `hideBulkActionsFor` check takes precedence over the `tableActions` prop when the store explicitly sets it to hidden
- **Option C:** Document that `hideBulkActions` only works for resource types that do NOT have a custom list component in `shell/list/`

aalves08:
### Summary
Fixes #16834 
Fixes #17426 (added caveat to `hideBulkActions` docs)

- add support for `headers`
- cover `mapGroup` DSL method **->** regExp/ID to remap group display names in the side-menu(**kept as internal** since it renames a group of resources based on `schema.attributes.group`)
- cover `ignoreGroup` DSL method **->** don't display a given group nav (**kept as internal** only since it targets `schema.attributes.group` so this is only useful in `explorer` world mostly)
- cover `mapType` DSL method **->** rename a specific type to a given name/label
- cover `ignoreType` DSL method **->** don't display a specific type on the side nav
- cover `hideBulkActions` DSL method **->** bool to control the display of bulk actions for a given resource page
- cover `moveType` DSL method **->** move a resource-type page to a given group (**kept as internal**)

_Did not cover these DSL methods_
- `componentForType` -> this defines a new component for a specific type that we wish to override. Doesn't make sense to me given the new pattern, since you'll always need to specify a component. Just change the prod registration
- `setGroupDefaultType` -> this is used to set the a different default page for a group. Same as the above, just change the prod registration (the topmost child of the definition of a group is always the "entry" page)
- adds documentation to docusaurus to cover the topics above
- `spoofedType` - we've decided to drop implementation in the new product registration

### Checklist
- [x] The PR has a Milestone 


shell/core/plugin-products-base.ts
```diff
@@ -227,6 +236,20 @@ export abstract class BasePluginProduct {
       basicType(names);
     }
 
+    // process mapToGroup entries
+    if (this.product?.mapToGroup?.length) {
+      this.product.mapToGroup.forEach((mapping) => {
+        mapGroup(mapping.condition, mapping.group);
```
codyrancher: By any chance do we want to update the type of mapping so we can support the entire `mapGroup` arguments interface/function signature https://github.com/rancher/dashboard/blob/master/shell/store/type-map.js#L346-L350

shell/core/plugin-products-base.ts
```diff
@@ -242,13 +265,51 @@ export abstract class BasePluginProduct {
...
+      spoofedTypeConfig.route = pluginProductsHelpers.generateConfigureTypeRoute(parentName, spoofedTypeConfig, { extendProduct: !this.isNewProduct });
+
+      // generate schema for the spoofed type if not provided - this is required to have the type show up in the UI
+      if (!spoofedTypeConfig.schemas || spoofedTypeConfig.schemas.length === 0) {
+        spoofedTypeConfig.schemas = [pluginProductsHelpers.generateSchemaForSpoofedType({
+          type:              spoofedTypeConfig.name,
+          collectionMethods: spoofedTypeConfig.collectionMethods || ['GET'],
+          namespaced:        spoofedTypeConfig.namespaced || false,
+          resourceFields:    spoofedTypeConfig.resourceFields || {},
+          attributes:        spoofedTypeConfig.attributes || { namespaced: spoofedTypeConfig.namespaced || false },
```
codyrancher: Minor: 

`{ namespaced: spoofedTypeConfig.namespaced || false }`  is both on this line and on line 291. If we always wanted these to be consistent we may want to create a const and use it in both places.

Not necessary though.

shell/core/types.ts
```diff
@@ -595,18 +613,27 @@ export interface DSLReturnType {
...
+   */
+  ignoreType: (regexOrString: string | RegExp) => void;
+
+  /**
+   * Never show the specified group or any types in it
+   * @param regexOrString String or regex to match against group names
+   * @param cb Optional callback
+   * @returns {@link void}
+   */
+  ignoreGroup: (regexOrString: string | RegExp, cb?: Function) => void;
```
codyrancher: Based on https://github.com/rancher/dashboard/pull/17330/changes#diff-cd39b66fa1a942a9464e17446c7001abb9ac8f327f576f7ec418b73421494d0bR248

and https://github.com/rancher/dashboard/pull/17330/changes#diff-9b9138695e3103571222831ec0f4cccb2f4156362f4d120ce531e13373e9dc6dR300

It seems that `cb` should have a more specific type.

shell/core/plugin-products-base.ts
```diff
@@ -242,13 +265,51 @@ export abstract class BasePluginProduct {
...
+
+      if (spoofedTypeConfig.overrideListResourceName) {
+        mapType(spoofedTypeConfig.type, spoofedTypeConfig.overrideListResourceName);
+      }
+
+      if (spoofedTypeConfig.hideBulkActions) {
+        hideBulkActions(spoofedTypeConfig.type, true);
+      }
+
+      spoofedType(spoofedTypeConfig);
```
codyrancher: It looks like we're omitting configuring `weight` for spoofed types. Am I missing where that's handled? Do we just want the consumer to handle that after creating the spoofed type?

codyrancher (commented):
I kind of wish we weren't exposing more of this, in particular spoofedType. By doing so I think we're going to solidify the inclusion of these concepts into rancher.

I would much rather we expose simpler APIs to just add items to navigation and use the already existing route apis.

With that said I really only found a few minor issues.

shell/core/plugin-products-base.ts
aalves08: I did notice that when connecting `mapGroup` to the new product registration API. My decision to "simplify" the method mapping was because all usages in Dashboard only uses the two required arguments, as per:

Screenshot 2026-04-27 at 10 22 18

Do you think we should expand this OR just keep it simple @codyrancher ?

shell/core/plugin-products-base.ts
aalves08: Still waiting to confirm if I should remove entirely the support for `spoofedType` on the new product registration or not... If we keep it, I'll fix this

shell/core/plugin-products-base.ts
aalves08: Waiting on confirmation on wether we should drop `spoofedType` to address this... Thanks for bringing this up 🙏

shell/core/types.ts
aalves08: Updated. Both have exactly the same signature and cb is optional, not mandatory. Updated docs as well.

shell/core/plugin-products-base.ts
aalves08: `spoofedType` support removed

shell/core/plugin-products-base.ts
aalves08: `spoofedType` support removed

shell/core/plugin-products-base.ts
codyrancher: I'm okay with keeping it simple.

codyrancher (dismissed):

shell/core/plugin-types.ts
```diff
@@ -218,6 +226,14 @@ export type ProductMetadata = Omit<ProductOptions, 'name' | 'label' | 'labelKey'
    * Product name (unique identifier)
    */
   name: string;
+  /** Array of mapping resources to groups based on conditions/regEx */
```
richard-cox: Suggested change
      
    
    
      
          
            
              /** Array of mapping resources to groups based on conditions/regEx */
          
          
            
              /** Associate kubernetes resource groups or specific resources to side-menu groups.*/

shell/core/plugin-types.ts
```diff
@@ -218,6 +226,14 @@ export type ProductMetadata = Omit<ProductOptions, 'name' | 'label' | 'labelKey'
...
+  mapToGroup?: { condition: RegExp | string; group: string }[];
+  /** Array of groups to ignore based on conditions */
```
richard-cox:
```suggestion
  /** Array of groups to exclude (based on conditions) from the side-menu  */
```

shell/core/plugin-types.ts
```diff
@@ -218,6 +226,14 @@ export type ProductMetadata = Omit<ProductOptions, 'name' | 'label' | 'labelKey'
...
+  ignoreGroups?: {
+    /** String or regex to match against group names */
+    regexOrString: string | RegExp;
+    /** Optional conditional function that accepts getters and returns true if the group should be ignored */
```
richard-cox:
```suggestion
    /** Optional conditional function that accepts the root Dashboard Vuex store getters and returns true if the group should be ignored */
```

shell/core/plugin-types.ts
```diff
@@ -185,6 +185,14 @@ export type ProductChildResourcePage = {
...
   /** Ordering weight for this page among its siblings */
   weight?: number;
+  /** Use this to override the resource name used in the list view for this type */
+  overrideListResourceName?: string;
+  /** Whether to hide this resource from the side navigation entirely */
+  hideFromNav?: boolean;
+  /** Whether to hide bulk actions for this resource */
+  hideBulkActions?: boolean;
+  /** Table headers for this resource type */
+  headers?: any[];
```
richard-cox: the `headers` signature is `(type, headers, paginationHeaders = [])`

so we need to accommodate paginationHeaders as well.

we can probably split this into two, and update type-map headers fn to only apply the ones that are supplied

```
headers?: any[];
paginationHeaders?: any[];
```

richard-cox: these can be typed as well `TableColumn`

it's not widely used yet, but is publicly - plugin.addTableColumn

shell/core/types.ts
```diff
@@ -562,12 +561,24 @@ export interface DSLReturnType {
   product: (options: ProductOptions) => void;
 
   /**
-   * Create and label a group. The group will show up in navigation
-   * @param groupNane Name of the group
-   * @param label Label in navigation
+   * Remap a group name to a display name
+   * @param match String or regex to match against group names
```
richard-cox: it gets a bit narly, but match can either be

1. strict string to match
2. a regex string, which will be converted into a regex object
3. a regex object

So i'd vote
```suggestion
   * @param match String, string for a regex or a regex object to match against group names
```

richard-cox: this also applies to  mapType, ignoreType, ignoreGroup

shell/core/types.ts
```diff
@@ -595,18 +606,27 @@ export interface DSLReturnType {
...
+  /**
+   * Never show the specified group or any types in it
+   * @param regexOrString String or regex to match against group names
+   * @param fn Conditional function that accepts getters and returns true if the group should be ignored
+   * @returns {@link void}
+   */
+  ignoreGroup: (regexOrString: string | RegExp, fn?: (getters: any) => boolean) => void;
+
+  /**
+   * Hide bulk actions for a specific resource type
```
richard-cox:
```suggestion
   * Hide list bulk actions for a specific resource type
```

shell/core/types.ts
```diff
@@ -595,18 +606,27 @@ export interface DSLReturnType {
...
-  // groupBy: (type: string, field: string)
-  // hideBulkActions: (type: string, field)
-  // ignoreGroup: (regexOrString)
-  // ignoreType: (regexOrString)
-  //
-  // mapType: (match, replace)
-  // moveType: (match, group)
-  // setGroupDefaultType: (input, defaultType)
-  // spoofedType: (obj)
+  hideBulkActions: (type: string, field: boolean) => void;
```
richard-cox: not sure why the original code uses `field`, but it's just a `truthy`. 
```
        const hideTableActions = this.$store.getters['type-map/hideBulkActionsFor'](this.schema);

        return !hideTableActions;
```

so lets switch `field` to `show: boolean` and update the fn description to something a little less ambiguous

shell/core/types.ts
```diff
@@ -663,8 +683,6 @@ export interface IExtension {
    * @param name
    * @param config
    */
-  addProduct(product: ProductMetadata, config: ProductChildGroup[]): void;
```
richard-cox, line 667: we added these and the ones for `extendProduct` to improve IDE auto-complete

shell/core/plugin-products-base.ts
richard-cox: tricky one.

the two missing props weight and continueOnMatch are about applying the new rule before or after others, or to add additional rules to match.

they would allow extensions to override or hide existing groups. which might be something they want to do, but feels scary.

let's wait until someone really needs them

shell/core/plugin-types.ts
```diff
@@ -218,6 +226,14 @@ export type ProductMetadata = Omit<ProductOptions, 'name' | 'label' | 'labelKey'
    * Product name (unique identifier)
    */
   name: string;
+  /** Array of mapping resources to groups based on conditions/regEx */
+  mapToGroup?: { condition: RegExp | string; group: string }[];
```
richard-cox: `condition` should be `regexOrString` inline with all the others

docusaurus/docs/extensions/api/nav/product-registration.md
```diff
@@ -316,17 +321,21 @@ import {
...
-Extends `ProductMetadata` with:
+A product that renders a single full-page Vue component with no side-menu. Extends `ProductMetadata` with:
 
 | Property | Type | Required | Description |
 | --- | --- | --- | --- |
 | `component` | `RouteComponent` | Yes | Vue component to render for the entire product |
 
 ### `ProductChildCustomPage`
 
+A page inside a product that renders a Vue component you provide. Equivalent to a `virtualType` in the DSL approach.
```
richard-cox:
```suggestion
A page inside a product that renders a Vue component you provide. Equivalent to a `virtualType` in the legacy DSL approach.
```

(same for other DSL references)

It's worth mentioning DSL for those migrating, but we should make it clear that it's not now an option

docusaurus/docs/extensions/api/nav/product-registration.md
```diff
@@ -354,6 +372,210 @@ Extends `ProductMetadata` with:
...
+
+```ts
+type ProductChild = ProductChildGroup | ProductChildPage;
+```
+
+## Advanced Configuration
+
+### Table headers (`headers`)
+
+Resource pages support custom table column headers via the `headers` property. Headers control which columns appear in the resource list view, what data they display, and how they behave (sorting, searching, formatting).
```
richard-cox:
```suggestion
Resource lists use Headers to control which columns appear in the resource list view, what data they display, and how they behave (sorting, searching, formatting).

By default these mostly come from the resource's definition and it's additionalPrinterColumns field.

When you set `headers` on a resource page, these columns **replace** the default columns.
```

shell/core/plugin-types.ts
richard-cox: i tried the following but it didn't move pods into the group

```
  const product: ProductMetadata = {
    name:       'product-1',
    label:      'Product One',
    mapToGroup: [{
      condition: 'pod',
      group:     'product-1-childPage'
    }]
  };

  const childResourcePage: ProductChildResourcePage = { type: 'pod' };

  plugin.addProduct(product, [childResourcePage]);

  const childCustomPage: ProductChildCustomPage = {
    name:      'product-1-childPage',
    label:     'product-1-childPage',
    component: () => import('./components/empty-component.vue')
  };

  plugin.addProduct(product, [childCustomPage]);
```

shell/core/plugin-types.ts
```diff
@@ -218,6 +226,14 @@ export type ProductMetadata = Omit<ProductOptions, 'name' | 'label' | 'labelKey'
...
+  /** Array of groups to ignore based on conditions */
+  ignoreGroups?: {
```
richard-cox, line 326: similar to https://github.com/rancher/dashboard/pull/17330/changes#r3166496219 couldn't get this to work either
```
  const product: ProductMetadata = {
    name:         'product-1',
    label:        'Product One',
    ignoreGroups: [{ regexOrString: 'pods' }]
  };

  const childResourcePage: ProductChildResourcePage = { type: 'pod' };

  plugin.addProduct(product, [childResourcePage]);

  const childCustomPage: ProductChildCustomPage = {
    name:      'product-1-childPage',
    label:     'product-1-childPage',
    component: () => import('./components/empty-component.vue')
  };

  plugin.addProduct(product, [childCustomPage]);
```

also tried with, and pods/pod

console.error wasn't triggered

```
ignoreGroups: [{
      regexOrString: 'pods',
      fn:            () => {
        console.error('!!!!');

        return false;
      }
    }]
```

shell/core/plugin-types.ts
```diff
@@ -185,6 +185,14 @@ export type ProductChildResourcePage = {
   config?: ConfigureTypeConfiguration;
   /** Ordering weight for this page among its siblings */
   weight?: number;
+  /** Use this to override the resource name used in the list view for this type */
+  overrideListResourceName?: string;
+  /** Whether to hide this resource from the side navigation entirely */
```
richard-cox:
```suggestion
  /** Whether to hide this resource from the side-menu entirely */
```

for consistency with docs

shell/core/plugin-types.ts
aalves08: updated

shell/core/plugin-types.ts
aalves08: updated

shell/core/plugin-types.ts
aalves08: updated

shell/core/types.ts
aalves08: updated

shell/core/types.ts
aalves08: updated

shell/core/types.ts
aalves08: It's not `show` but `hide`... It's the other way around that it works. Updated

shell/core/types.ts
aalves08: Gotcha. With spoofedType, there was another signature overload, but for the IDE it matters. Forgot about that. Updated

docusaurus/docs/extensions/api/nav/product-registration.md
aalves08: updated

docusaurus/docs/extensions/api/nav/product-registration.md
aalves08: updated

shell/core/plugin-types.ts
aalves08: updated

shell/core/plugin-types.ts
aalves08: updated

shell/core/plugin-types.ts
aalves08: `mapToGroup` -> DSL `mapGroup` is and "advanced" DSL operation that renames a group of resources based on `schema.attributes.group` property).

After syncing on this matter, we decided to keep it as internal.

Also, doing `addProduct` twice on the same product is an anti-pattern. Under the hood we calling different class instances even though the whole DSL part is just state/store based. We should allow a "double call" on the `addProduct`. Will gate it.

shell/core/plugin-types.ts
aalves08: This is pretty similar to `mapGroup`...

`ignoreGroups` -> DSL `ignoreGroup` is and "advanced" DSL operation that ignores/hides group of resources based on `schema.attributes.group` property). We are keeping it as internal only

shell/core/plugin-types.ts
aalves08: added support for paginationHeaders. Called it `sspHeaders`

shell/store/type-map.js
```diff
@@ -269,20 +269,25 @@ export function DSL(store, product, module = 'type-map') {
       store.commit(`${ module }/groupBy`, { type, field });
     },
 
-    headers(type, headers, paginationHeaders = []) {
-      headers.forEach((header) => {
+    headers(type, headers = [], paginationHeaders = []) {
```
richard-cox, line 273: i'm not sure we've run with config where there's only the pagination headers. 

might be worth a test

lets exclude the 2.15 scenario where users can still turn off ssp

shell/core/plugin-types.ts
```diff
@@ -218,6 +228,106 @@ export type ProductMetadata = Omit<ProductOptions, 'name' | 'label' | 'labelKey'
...
+   *   label: 'My App',
+   *   renameGroups: [
+   *     // Rename a group with an ugly internal ID to a friendlier display name
+   *     { regexOrString: 'cert-manager.io', group: 'Certificates' },
+   *     // Use a regex to rename all groups matching a pattern
+   *     { regexOrString: /^networking\./, group: 'Networking' },
+   *   ],
+   * };
+   */
+  renameGroups?: {
```
richard-cox, line 248: i had to really digest the comments to work out / remember what these do.
which means we could improve the prop names 


renameGroups: { regexOrString; group }
moveToGroup: { id; group; weigh; }
ignoreGroups: { regexOrString; fn }

so maybe something like.... ?

renameGroups: { groupSelector, newName }
moveToGroup: { entryId; groupName, weight; }
ignoreGroups: { groupSelector; condition }

shell/core/plugin-types.ts
```diff
@@ -185,6 +185,16 @@ export type ProductChildResourcePage = {
...
+  /** Use this to override the resource name used in the list view for this type */
+  overrideListResourceName?: string;
+  /** Whether to hide this resource from the side-menu entirely */
+  hideFromNav?: boolean;
+  /** Whether to hide bulk actions for this resource */
+  hideBulkActions?: boolean;
+  /** Table headers for this resource type (client-side pagination) */
+  headers?: HeaderOptions[];
+  /** Table headers for this resource type (server-side pagination) */
+  sspHeaders?: PaginationHeaderOptions[];
```
richard-cox, line 197: i played with this but it doesn't quite work as expected

```
 type: 'pod',
 sspHeaders: [
      STEVE_STATE_COL,
      STEVE_NAME_COL,
      STEVE_AGE_COL,
    ]
```

the default product inStore value is `management` (inStore:             'management',)

so ssp for the type has to be enabled via 

```
  plugin.enableServerSidePagination({ management: { resources: { enableSome: { enabled: ['pod'] } } } });
```
(note that the `management` key is the inStore value)

atm because we default to management and don't provide a way to change inStore we can only support people having generic resource pages for that store. this applies to both ssp enabled and disabled

if we also expose the product `inStore` prop we can then support people who wish to use cluster'ed resources. they would still need to have a route the cluster in (atm everything is the blank cluster `_`)

shell/core/plugin-types.ts
richard-cox, line 248: from out conversation i think we discovered that renameGroups, moveToGroup and ignoreGroups were all really only applicable to the explorer view.. and can be used when we convert that to use the product api

I've just been playing around with them

- renameGroups
  - this only makes sense in explorer view, otherwise people can just configure the groups a different way when addProduct
  - trying to use this on groups defined via addProduct does not work
  - tagged as internal, all good
- moveToGroup
  - same as renameGroups, does not make sense when they can just put them in the right group via addProduct
  - interestingly though this does work on groups provided via addProduct
  - tagged as internal though, all good
- ignoreGroups
  - in theory same as both above, shouldn't make sense when they can just not include the groups via addProduct...however.. it has a usefull conditional fn which could be of use 
  - currently tagged as internal
  - so question is.... could this be made to work? lots of effort --> leave as is, not a lot of effort --> go for it and un 'internal' it

richard-cox: comment only

fun point, i managed to test all three in the end by forcing the side nav to render the more resources section.

shell/components/SideNav.vue --> getProductsGroups --> move `modes.push(TYPE_MODES.USED);` out of the if block to just below

weirdly this brings in the typemap definitions for explorer, but at least renders the section ok

i then used

```
renameGroups: [{
      regexOrString: 'turtles-capi.cattle.io',
      group:         'abc',
    }], // works, but only for groups not already renamed (given they're all weight 5, explorer already has one) `apiregistration.k8s.io`
```

```
ignoreGroups: [{ regexOrString: 'turtles-capi.cattle.io', fn: () => true }] // working
```

docusaurus/docs/extensions/api/nav/product-registration.md
```diff
@@ -354,11 +371,223 @@ Extends `ProductMetadata` with:
...
+| `formatter` | `string` | Name of a built-in formatter component from `@shell/components/formatter/` or a custom formatter defined in your extension |
+| `formatterOpts` | `any` | Configuration options passed to the formatter component |
+| `width` | `number` | Fixed column width in pixels. If not set, the column auto-sizes |
+| `getValue` | `(row: any) => string \| number \| null` | Custom function to extract or compute the display value from a row. Takes precedence over `value` for display, but `value` is still used for sorting/searching |
+
+### Server-side pagination headers (`sspHeaders`)
+
+When server-side pagination is enabled for a resource type, Rancher Dashboard uses a separate set of headers optimized for paginated data. These headers follow the same format as `headers` but **do not support `getValue`** — since rows are fetched page-by-page from the server, computed client-side values are not available.
+
+Use `sspHeaders` alongside `headers` to define columns for both pagination modes:
```
richard-cox, line 515: We need info somewhere here that links to https://extensions.rancher.io/extensions/next/performance/scaling/lists#choosing-fields-to-display-sort-and-filter-on, which contains important information on what can be sorted / searched on. Some maybe something like

`Property paths used to display, sort and search on must exist natively on the resource and not in the Dashboard's model for the resource type. For more information please see [here](https://extensions.rancher.io/extensions/next/performance/scaling/lists#choosing-fields-to-display-sort-and-filter-on)`

richard-cox (commented):
~is there anyway~ moved to comment

shell/store/type-map.js
aalves08: by doing `headers = []` we are protecting the scenario IF we only pass `paginationHeaders`.

I've also added some unit tests:

- Only pagination headers provided (undefined regular headers) — commits only `paginationHeaders`
- Empty array regular headers with pagination headers — commits only `paginationHeaders`
- Both provided — commits both `headers` and `paginationHeaders`
- Only regular headers (no pagination) — commits only `headers`
- Both empty — commits nothing

I think this suffices, right @richard-cox ?

shell/core/plugin-types.ts
aalves08: the renaming was important as the DSL methods naming aren't real clear on what they do (learned that the hard way on the last review cycle) and they mean nothing to developers, especially outside rancher/dashboard context.

Updated the params naming in docs, tests, implementation, etc

aalves08: `moveToGroup` works on groups provided via `addProduct` because I employed a mechanism for it to work, as per https://github.com/aalves08/dashboard/blob/0c1b34892c4bdc4071a3ff546812feb53fec1727/shell/core/plugin-products-base.ts#L295-L320.  

The underlying  DSL `moveType` method logic depends on `schema.attributes.group` 😛 I thought it would be important at that point in time to cover it, but then we realised that these are internal methods and shouldn't be surfaced in terms of docs for external devs, but I kept the logic anyway.

For me `moveToGroup` is the same as `ignoreGroups`: In the scope of the new product registration, why do you need move a page to a group? just move it when defining it. Same for ignoring groups, just don't define it.

As far as implementation goes, I think that it's not worth it to try and make this work because the DSL store state props that command those aren't keyed and that a totally different level of changes and testing to be done.

Let me know your thoughts 🙏

docusaurus/docs/extensions/api/nav/product-registration.md
aalves08: updated. Thanks for the heads up! 🙏

shell/core/plugin-types.ts
aalves08: I've added `plugin.enableServerSidePagination` to the plugin logic + added caveat to the docs page. We need to chat about this because I'll need your help documenting this into an issue with a defined solution 💦 💦 💦 

I think that for now it will suffice. You agree?

shell/core/plugin-products-base.ts
```diff
@@ -107,6 +122,27 @@ export abstract class BasePluginProduct {
...
       }
     });
+
+    // Auto-enable SSP for resource pages that define sspHeaders
+    if (this.product) {
+      const sspTypes = this.collectSspTypes(this.config);
+
+      if (sspTypes.length) {
+        sspTypes.forEach((type: string) => {
+          if (!store.getters['management/schemaFor'](type)) {
```
richard-cox: The store is also hardcoded to 'management', the same reference should be used when the product is registered, when the schema is checked for and when ssp is set below.

Whether we go ahead with this is TBD, see other comments

shell/store/type-map.js
richard-cox: Test as in, what happens to rancher functionally when a type only has the paginated headers. probably only related to lists though

shell/core/plugin-types.ts
richard-cox: The code snippets above were an example of how this would need to be done in the wild and the comment revolved around configuring the store of the product. In slack i think we agreed to document the inStore limitation of management as a current limitation and address in a different PR (PR contains the update for docs 👍).

The change to automatically call `enableServerSidePagination` though introduces issues.

`enableServerSidePagination` was designed to be called once per extension. the code in the PR calls this and would overwrite / be overwritten given the parent extension calling it. For example if i wanted to enable SSP for resourceA and resourceB in the management store...  in my extension i would do ...

```
        plugin.enableServerSidePagination({ management: { resources: { enableSome: { enabled: ['resourceA', 'resourceB'] } } } });
```

if i then passed through custom headers for deployments the new code would then call

```
        plugin.enableServerSidePagination({ management: { resources: { enableSome: { enabled: sspTypes } } } });
```

the second call value would overwrite the first call's value.

we could either
1. back this change out
2. handle via, and documente a merge process

shell/core/plugin-types.ts
richard-cox: The thing i was most interested in with ignoreGroups was the conditional side. For instance if i had a page for showing settings i might want to conditionally hide that for some users. the page config would be added and the `conditional` param used. moveGroups isn't as important (not sure why someone would want to conditionally move a group, and also we don't support it ourselves).

if it's a lot of work to support we can leave this though and wait for feedback

docusaurus/docs/extensions/api/nav/product-registration.md
```diff
@@ -354,11 +371,229 @@ Extends `ProductMetadata` with:
...
+
+> **Important:** Property paths used to display, sort and search on must exist natively on the resource and not in the Dashboard's model for the resource type. For more information please see [Choosing fields to display, sort and filter on](https://extensions.rancher.io/extensions/next/performance/scaling/lists#choosing-fields-to-display-sort-and-filter-on).
+
+When `sspHeaders` is defined on a resource page, server-side pagination is automatically enabled for that resource type — no additional configuration is needed.
+
+> **Important:** Server-side pagination via `sspHeaders` is currently supported for global-level resources (e.g. `provisioning.cattle.io.cluster`). Cluster-scoped resources like `pod` or `service` are not yet supported by `addProduct`. A console warning will be shown if `sspHeaders` is set on a resource type that is not available at the global level.
```
richard-cox: this isn't quite right. global level resources are resources in the upstream / local cluster, which can contain pods and services. the current implementation uses pods, which are the local cluster pods.

```suggestion
> **Important:** Server-side pagination via `sspHeaders` is currently supported for global-level resources (e.g. `provisioning.cattle.io.cluster`) in the Rancher `local` cluster. Resources in downstream clusters are not yet supported by `addProduct`. A console warning will be shown if `sspHeaders` is set on a resource type that is not available at the global level.
```

shell/core/plugin-types.ts
aalves08: But `ignoreGroups` is connected to the new prod reg and works, we aren't just surfacing it (marked as internal, but we will use definitely in our internal dashboard products). The work involved to make this work outside of `schema.attributes.group` in order to surface it is not worth it at this point and frankly there's no valid usecase that I can think of. Resolving this thread

aalves08: ALEX TODO: create issue for support ignoreGroups where the conditional part is the important bit... this should cover individual pages and groups (can be either RBAC or other conditions)

shell/core/plugin-products-base.ts
aalves08: TODO: nuke it alex

shell/core/plugin-types.ts
aalves08: TODO: nuke it alex (enabling on new prod reg). Send them to docs appropriately

shell/store/type-map.js
aalves08: TODO: alex, test ONLY sspHeaders (no normal headers) on `secret` (on mgmt store) and see if anything weird comes from it (create, update, delete, sort, filtering, check if headers are there, etc)

shell/core/plugin-products-base.ts
aalves08: updated

shell/core/plugin-types.ts
aalves08: updated

shell/store/type-map.js
aalves08: Screenshot 2026-05-19 at 17 02 18

If I try to add the `secrets`  to a new top-level product (via new prod reg) the only side-effect is that it has two mastheads, independently of passing `sspHeaders` or not…. I haven’t found any side-effect of only passing `sspHeaders`, tbh. Filtering, sorting, creating, editing, deleting, all looks good.

the double masthead is unrelated to `sspHeaders` 

I think this means that the `sspHeaders` can be passed in isolation, but I recommend you give it a quick test if you wish to do so.

The double masthead is coming from `ResourceList` and the `custom list view for secrets`

I think this happens because of https://github.com/rancher/dashboard/pull/17561, where the `typeOptions` is now keyed by product, meaning that we need to declare options per-product like https://github.com/rancher/dashboard/blob/master/shell/config/product/explorer.js#L200

Are we ok on this front then?

shell/core/plugin-types.ts
aalves08: Issue created @richard-cox https://github.com/rancher/dashboard/issues/17732. Resolving

shell/store/type-map.js
richard-cox: Sounds good. I also had a play with secrets, and could nicely use the product api to disable

```
 const childResourcePage: ProductChildResourcePage = {
    type:   'secret',
    config: { showListMasthead: false },
}
```

richard-cox (approved):

PR #17343, opened 2026-04-21
Add vee-validate support to LabeledInput

### LINKED ISSUE #15366: Investigate use of composition API for wiring veevalidate as form validation
# Description

Vee-validate library recommends composition API, which is also the architecture we are adopting.

rak-phillip:
### Summary
This adds vee-validate support to LabeledInput and replaces the custom validation for the User Retention form with vee-validate.

Fixes #15366

- Add vee-validate to labeled input
- Replace validation of user retention form with vee-validate
- Resolve typescript warnings in user retention form

### Technical notes summary

> 🗒️ NOTE
>
> d6715862d91646053a0d86491aa0bb25f675eca6 contains typescript changes that aren't required for this fix. We can split this out into a different PR if it is easier for review

The vee-validate support added to LabeledInput keeps the existing validation method in place so that we can move forward with both validation methods until vee-validate becomes the preferred/default approach.

The idiomatic pattern for vee-validate promotes centralizing form-level validations via the `validationSchema`, which is what the update to the User Retention form demonstrates. With that in mind, we can still omit the validation schema and provide a collection of validators to the `rules` prop, like we do with validation today. I think that supporting both patterns will help to ease the migration of existing forms over to vee-validate in the future. I anticipate that this pattern will be demonstrated in future PRs that handle validation for more complex forms. 

The user retention form should perform validations in the same way that it has before.

Existing forms should validate inputs in the same way as they do today.

Form validation in existing forms could possibly change with modification made to LabeledInput.

User Retention validation. 

### Screenshot/Video

image

image

image

### Checklist
- [x] The PR has a Milestone 


pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.test.ts
```diff
@@ -105,4 +107,119 @@ describe('component: LabeledInput', () => {
...
+
+      const wrapper = mount(TestWrapper, { global: { mocks: { $store: { getters: { 'i18n/t': jest.fn() } } } } });
+
+      await triggerFormValidation();
+      await flushPromises();
+
+      const labeledInput = wrapper.findComponent(LabeledInput);
+
+      expect(labeledInput.vm.validationMessage).toStrictEqual(errorMessage);
+    });
```
aalves08, line 223: Could we add some tests to check the following:
- error clearing when a previously invalid value becomes valid
- dual `name` + `rules` path — what happens when both are provided?

shell/package.json
```diff
@@ -32,39 +32,40 @@
...
     "@novnc/novnc": "1.2.0",
     "@popperjs/core": "2.11.8",
     "@rancher/icons": "2.0.55",
+    "@smithy/fetch-http-handler": "5.1.1",
     "@types/is-url": "1.2.30",
     "@types/node": "25.3.3",
     "@types/semver": "^7.5.8",
     "@typescript-eslint/eslint-plugin": "5.62.0",
     "@typescript-eslint/parser": "5.62.0",
+    "@vee-validate/zod": "^4.15.1",
```
aalves08: I can only find the import of `vee-validate` and not `vee-validate/zod`. Are you sure this is being used or did I miss something?

shell/pages/c/_cluster/auth/user.retention/index.vue
```diff
@@ -111,7 +135,11 @@ const fetchSetting = async(id: string) => {
   return await store.dispatch('management/find', { type: MANAGEMENT.SETTING, id });
 };
 
-const ids = Object.keys(userRetentionSettings);
+const getKeys = <T extends object>(obj: T): Array<keyof T> => {
```
aalves08: Can't we just do:

`const ids = Object.keys(userRetentionSettings) as UserRetentionSettingId[];` ?

Seems over-engineered this bit...

aalves08 (changes requested):
Left some comments.

Also, shouldn't we be bumping and then releasing `pkg/rancher-components/package.json`?

shell/package.json
rak-phillip: We will eventually need `@vee-validate/zod` when we start using zod to write validation schemas in pages & component - this will most likely happen in new components or adding validation to components that currently don't have any validators specified.

I'm noting that I missed the `zod` dependency at this point.. I think that we can handle this in one of two ways:

- Add all of the dependencies now so that we can be ready write validators when vee-validate is in place
- Add the zod dependencies later when we need them

shell/pages/c/_cluster/auth/user.retention/index.vue
rak-phillip: hmmm.. Would we be more amenable to this function if we implemented it elsewhere as a utility? 

I was considering if we had a `shell/utils/object.ts`, but we only have a javascript file. I think that I can just do the casting for now, but I find the utility function with the generics to be mighty convenient in these cases.

pkg/rancher-components/src/components/Form/LabeledInput/LabeledInput.test.ts
rak-phillip: I updated the tests to include these cases.

shell/package.json
rak-phillip: I'm opting for adding the zod dependencies later when we need them

aalves08 (approved):
LGTM

PR #17348, opened 2026-04-22
Preserve URL hash when clicking 'Mark all as read'

### LINKED ISSUE #16923: Bug: "Mark all as read" button in Notification Center wipes URL hash fragment
**Setup**
- Rancher version: 2.13.2
- Rancher UI Extensions: n/a

**Describe the bug**

Clicking **"Mark all as read"** in the Notification Center clears any existing URL hash fragment (e.g. `#pod` → `#`). This is caused by the default browser navigation behavior of the anchor element in `NotificationHeader.vue`, which uses `href="#"` without preventing the default click event:

```html
<!-- shell/components/nav/NotificationCenter/NotificationHeader.vue -->
<a
  ref="markAllReadButton"
  role="button"
  tabindex="-1"
  href="#"
  @click="markAllRead(false)"
>
```

When clicked, the browser navigates to `#`, replacing whatever fragment was in the URL.

This also impacts extensions that use the `LocationConfig` `hash` field (e.g. `plugin.addTableColumn`) to scope injections to a specific URL fragment — when the fragment is silently cleared, the extension's injected columns/components disappear until the page is refreshed.

**To Reproduce**

1. Navigate to a workload detail page (by default URL ends with e.g. `#pod`).
3. Open the Notification Center.
4. Click **"Mark all as read"**.

**Result**

The URL fragment is cleared to `#` and any extension columns/components scoped to the previous fragment disappear.

**Expected Result**

Clicking "Mark all as read" should **not** modify the current URL. The button's only side effect should be dispatching the `notifications/markAllRead` action.

**Screenshots**

https://github.com/user-attachments/assets/593cd995-9e2f-45d4-aa27-1285d02358e4

**Additional context**

Suggested fix — add `@click.prevent` to the anchor (or replace it with a `<button>`) so the browser's default navigation is suppressed:

```html
<!-- Option A: prevent default on the anchor -->
<a
  ref="markAllReadButton"
  role="button"
  tabindex="-1"
  href="#"
  @click.prevent="markAllRead(false)"
>

<!-- Option B: use a button element instead -->
<button
  ref="markAllReadButton"
  tabindex="-1"
  @keydown.enter.space.stop="markAllRead(true)"
  @click="markAllRead(false)"
>
```

codyrancher:
### Summary
Fixes #16923

Clicking **Mark all as read** in the Notification Center no longer strips the URL hash fragment. The anchor used `href="#"` without preventing the default click, so the browser was navigating to `#` and wiping any existing fragment (e.g. `#pod`). This also broke extensions using `LocationConfig.hash` to scope injected columns/components to a specific fragment.

### Technical notes summary
- Added `.prevent` to the `@click` handler on the `Mark all as read` anchor in `shell/components/nav/NotificationCenter/NotificationHeader.vue` so the default anchor navigation no longer occurs.
- Added unit tests in `shell/components/nav/NotificationCenter/__tests__/NotificationHeader.test.ts` covering render visibility, dispatch of `notifications/markAllRead`, and a regression test asserting `event.defaultPrevented` is true on click.

- Navigate to a page with a hash fragment (e.g. `/dashboard/c/local/explorer/pod#pod`), open the Notification Center, and click **Mark all as read** — the URL hash should be preserved.
- Verify clicking **Mark all as read** still actually marks notifications as read (dispatches `notifications/markAllRead`).
- Verify keyboard activation (Enter/Space) still works.
- Verify extensions that register `LocationConfig` with a `hash` continue to match after the click.

- Anywhere relying on the prior default-navigation behavior of the Notification Center's **Mark all as read** anchor (none expected — the anchor target was `#`).

### Screenshot/Video

**Before fix** — clicking "Mark all as read" strips the `#pod` hash:

Before fix: URL hash is stripped

**After fix** — the `#pod` hash is preserved after clicking:

After fix: URL hash is preserved

Higher-quality webm captures: [before-fix.webm](https://github.com/codyrancher/dashboard/raw/issue-16923-media/before-fix.webm) · [after-fix.webm](https://github.com/codyrancher/dashboard/raw/issue-16923-media/after-fix.webm)

### Checklist
- [x] The PR has a Milestone

### DIFF
--- a/shell/components/nav/NotificationCenter/NotificationHeader.vue
+++ b/shell/components/nav/NotificationCenter/NotificationHeader.vue
@@ -69,7 +69,7 @@ const gotFocus = (e: Event) => {
       <div v-if="unreadCount !== 0">
         <RcButton
           ref="markAllReadButton"
-          variant="link"
+          variant="ghost"
           size="small"
           tabindex="-1"
           class="mark-all-read"
@@ -107,9 +107,21 @@ const gotFocus = (e: Event) => {
         flex: 1;
       }
 
+      // Preserve the original <a> appearance on top of RcButton ghost variant.
+      // Inherit the header's font metrics and mirror the global anchor hover
+      // from _basic.scss so there is no visible change from <a href="#">.
       .mark-all-read {
         padding: 0;
         min-height: auto;
+        font-size: inherit;
+        line-height: inherit;
+        gap: 0;
+        color: var(--link);
+
+        &:hover, &._hover {
+          color: var(--body-text);
+          text-decoration: underline;
+        }
       }
     }
   }
--- a/shell/components/nav/NotificationCenter/NotificationHeader.vue
+++ b/shell/components/nav/NotificationCenter/NotificationHeader.vue
@@ -107,18 +107,14 @@ const gotFocus = (e: Event) => {
         flex: 1;
       }
 
-      // Preserve the original <a> appearance on top of RcButton ghost variant.
-      // Inherit the header's font metrics and mirror the global anchor hover
-      // from _basic.scss so there is no visible change from <a href="#">.
       .mark-all-read {
         padding: 0;
         min-height: auto;
         font-size: inherit;
         line-height: inherit;
-        gap: 0;
         color: var(--link);
 
-        &:hover, &._hover {
+        &:hover {
           color: var(--body-text);
           text-decoration: underline;
         }
--- a/shell/components/nav/NotificationCenter/NotificationHeader.vue
+++ b/shell/components/nav/NotificationCenter/NotificationHeader.vue
@@ -2,17 +2,19 @@
 import { useStore } from 'vuex';
 import { computed, inject, ref } from 'vue';
 import { DropdownContext, defaultContext } from '@components/RcDropdown/types';
+import RcButton from '@components/RcButton/RcButton.vue';
+import { RcButtonType } from '@components/RcButton/types';
 
 const { dropdownItems } = inject<DropdownContext>('dropdownContext') || defaultContext;
 const store = useStore();
 const unreadCount = computed<number>(() => store.getters['notifications/unreadCount']);
-const markAllReadButton = ref<HTMLElement>();
+const markAllReadButton = ref<RcButtonType | null>(null);
 
 const markAllRead = (keyboard: boolean) => {
   store.dispatch('notifications/markAllRead');
 
-  // If we have focus, then move to the next item if activated by the keyboard
-  if (keyboard && document.activeElement === markAllReadButton?.value) {
+  // If activated via keyboard, move focus to the next dropdown item
+  if (keyboard) {
     moveFocus(true);
   }
 };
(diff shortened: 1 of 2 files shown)


codyrancher:
Everything up to the first commit in this pr was generated with AI, including selecting the issue, fixing it, recording the videos, making the commit and posting the pr.

After reviewing the comments in the issue I realized the original fix was a little off by not switching to a button and I've had to update it from there.

shell/components/nav/NotificationCenter/NotificationHeader.vue
```diff
@@ -104,8 +107,17 @@ const gotFocus = (e: Event) => {
         flex: 1;
       }
 
-      A {
+      .mark-all-read {
```
codyrancher, line 110: I don't love this but the original `link` variant looks nothing like the original link primarily due to the spacing and hover.

shell/components/nav/NotificationCenter/NotificationHeader.vue
```diff
@@ -2,17 +2,19 @@
...
-const markAllReadButton = ref<HTMLElement>();
+const markAllReadButton = ref<RcButtonType | null>(null);
 
 const markAllRead = (keyboard: boolean) => {
   store.dispatch('notifications/markAllRead');
 
-  // If we have focus, then move to the next item if activated by the keyboard
-  if (keyboard && document.activeElement === markAllReadButton?.value) {
+  // If activated via keyboard, move focus to the next dropdown item
+  if (keyboard) {
```
nwmac, line 17: Was this change intentional? This was done like this to prevent focus moving when the button is clicked using the mouse vs using the keyboard.

codyrancher: It appeared to be redundant.

- `markAllRead` is only invoked from the `.mark-all-read` button.
- When the mouse is used `keyboard` is `false` so we exit early
- When the keyboard is used it could've only come from the `.mark-all-read` button since that's the only location we invoke the `markAllRead` method from.

As far as I can tell we still have the same behavior as `master`. I can revert though since it doesn't address the immediate issue we're trying to resolve. Just let me know.

Demo:

https://github.com/user-attachments/assets/f36bf14b-891b-45f1-a4fb-45de0bf4126e

shell/components/nav/NotificationCenter/NotificationHeader.vue
rak-phillip: It looks like the `link` variant is limited usage. Do you think that it warrants a design pass to confirm that this variant is designed correctly? If I recall correctly, it was implemented using existing styles before the design system came to light.

codyrancher: Yeah, we can it looks like the usage is fairly limited so it shouldn't be too bad to tackle.. 

I'll open a separate issue to take care of that though.

rak-phillip (approved):
This looks good from my perspective. I know that we still have one outstanding question to resolve, so I'm adding my approval in the event that this gets cleared.

shell/components/nav/NotificationCenter/NotificationHeader.vue
codyrancher: Created the issue https://github.com/rancher/dashboard/issues/17454

PR #17355, opened 2026-04-23
Migrate components off the LabeledFormElement mixin

### LINKED ISSUE #17354: Find a way to migrate existing forms with validation to vee-validate
Many forms used throughout Rancher Dashboard utilize the custom form validation mixin and will need to be migrated over to vee-validate. In order to reduce the amount of friction involved in this transition, we will need a way to easily transition these forms with existing validation. 

## Proposed solution

Create a `useFormValidation()` composable to act as a bridge between the existing form validation mixin and vee-validate. Existing validation rule arrays should resolve into `vee-validate` validator functions.

## Acceptance Criteria
- [ ] `LabeledInput` and `LabeledSelect` correctly display errors when registered with a `vee-validate` form
- [ ] Secret forms prevent submission if required fields are missing or invalid
- [ ] Validation errors appear immediately on blur, matching existing behavior of the form validation mixin
- [ ] Form validation mixin is removed from the migrated components
- [ ] Unit tests for `LabeledInput` and SSH secrets are updated to support the new validation flow

rak-phillip:
### Summary
This is part of the broader migration to vee-validate in #17354 and migrates migrates the remaining components that relied on the LabeledFormElement to the `useLabeledFormElement()` and `useLabeledSelect()` composables.

Related to #17354

- Migrate LabeledSelect away from LabeledFormElement mixin
- Migrate remaining components away from LabeledFormElement mixin

### Technical notes summary

The primary goal for this change is to reduce reliance on mixed mixin/composition API patterns in these components and makes it easier to reuse common vee-validate behavior through the `useForm` composable. 

Different selects used throughout Rancher Dashboard should behave in the same way that they did before.

I think that LabeledSelect.vue and Select.vue are at the highest risk for regressions, especially surrounding filtering and pagination. 

### Screenshot/Video

NA

### Checklist
- [x] The PR has a Milestone 


shell/components/form/InputWithSelect.vue
```diff
@@ -1,18 +1,22 @@
...
 import { LabeledInput } from '@components/Form/LabeledInput';
 import LabeledSelect from '@shell/components/form/LabeledSelect';
 import Select from '@shell/components/form/Select';
+import { computed } from 'vue';
+import { _VIEW } from '@shell/config/query-params';
+
 export default {
   name:       'InputWithSelect',
+
+  inheritAttrs: false,
```
rak-phillip, line 11: `inheritAttrs: false` used to automatically be applied via the mixin.

shell/components/form/labeled-select-utils/useLabeledSelectPagination.ts
```diff
@@ -0,0 +1,130 @@
+import {
```
richard-cox, line 4: can we keep all the comments?

rak-phillip: Absolutely! They have been reintroduced via cccc9675a3eebf9604bbba5928b027dccf1996d3

shell/components/form/InputWithSelect.vue
```diff
@@ -84,23 +88,27 @@ export default {
     selectRules: {
       default: () => [],
       type:    Array,
+    },
+    mode: {
+      type:    String,
+      default: 'edit'
```
codyrancher: I think we should use the `_EDIT` constant from `@shell/config/query-params` .

shell/composables/useLabeledFormElement.ts
```diff
@@ -82,6 +84,10 @@ export const useLabeledFormElement = (props: LabeledFormElementProps, emit: Emit
     return props.required || props.rules?.some((rule: any) => rule?.name === 'required');
   });
 
+  const empty = computed(() => {
+    return !!`${ props.value }`;
```
codyrancher, line 88: I might be missing something.

It looks like this implementation may give inconsistent results based on the type of value.

Image

Image

Image

Image

shell/components/form/labeled-select-utils/useLabeledSelectPagination.ts
```diff
@@ -0,0 +1,140 @@
...
+  const pageSize = ref(10);
+  const pages = ref(0);
+
+  // External
+  const page = ref<any[]>([]);
+  const totalResults = ref(0);
+  const paginating = ref(false);
+
+  const canPaginate = computed(() => {
+    if (!store) return false;
```
codyrancher: Is there a scenario where `store` could be undefined? The return type doesn't indicate that it could.

shell/components/form/labeled-select-utils/useLabeledSelectPagination.ts
```diff
@@ -0,0 +1,140 @@
...
+      totalCount: totalResults.value
+    });
+  });
+
+  const requestPagination = async(resetPage = false) => {
+    paginating.value = true;
+
+    const { page: p, pages: pg, total } = await (props.paginate as LabelSelectPaginateFn)({
+      resetPage,
+      pageContent: page.value || [],
```
codyrancher, line 95: `|| []` based on how `page` is initialized it doesn't look like it could ever be falsey.

codyrancher (commented):
Only found a few items.

shell/components/form/Select.vue
```diff
@@ -252,40 +303,6 @@ export default {
     },
   },
   computed: {
-    requiredField() {
-      // using "any" for a type on "rule" here is dirty but the use of the optional chaining operator makes it safe for what we're doing here.
-      return (this.required || this.rules.some((rule) => rule?.name === 'required'));
-    },
-    validationMessage() {
```
codyrancher, line 259: Where are these coming from now? Will this break the existing implementation?

shell/composables/useLabeledFormElement.ts
rak-phillip: I'll have to dive into the history to better understand the intent. The logic is lifted directly from the original mixin:

https://github.com/rancher/dashboard/blob/c20230bb19695b09d46517e396e7c290bd97dabd/shell/mixins/labeled-form-element.ts#L112-L114

shell/components/form/labeled-select-utils/useLabeledSelectPagination.ts
rak-phillip: I think we're safe to remove this guard. Dashboard and all extensions should have a valid store configured where we won't need it.

shell/components/form/labeled-select-utils/useLabeledSelectPagination.ts
rak-phillip: I think that you are correct in your assessment. My preference is to keep this portion as-is for the time being to keep the logic consistent with the mixin:

https://github.com/rancher/dashboard/blob/c20230bb19695b09d46517e396e7c290bd97dabd/shell/components/form/labeled-select-utils/labeled-select-pagination.ts#L97-L107

shell/components/form/Select.vue
rak-phillip: These are coming from the `useLabeledFormElement()` composable now. I initially removed these to reduce the duplication, but I can reintroduce them to reduce the chance of breakage and to maintain consistency with the existing implementation:

https://github.com/rak-phillip/dashboard/blob/cccc9675a3eebf9604bbba5928b027dccf1996d3/shell/components/form/Select.vue#L123-L125

codyrancher (dismissed):

codyrancher (approved):

PR #17371, opened 2026-04-26
Fix window manager

### LINKED ISSUE #15848: Move WindowManager to App root
**Is your change request related to a problem? Please describe.**

We have multiple instances of `WindowManager` in 2.13 .

- They are mounted when navigating to pages that are build on top of one of [default.vue, home.vue, plain.vue] layouts.
- They are un-mounted when moving out.

This means that components hosted in the WindowManager panels have to handle (via store) their internal status so that it's saved across layouts navigation and restored when the WindowManager is mounted.

We should use 1 global instance of `WindowManager`, mounted in `shell/public/index.html`, and remove the others.

torchiaf:
### Summary
Fixes https://github.com/rancher/dashboard/issues/15848

### Technical notes summary

We are removing the WindowManager compontent from the template pages and creating only 1 at root level, in the App.vue component.
This allows the web app to keep only 1 instance of the window alive , even when navigating through templates.

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/templates/home.vue
+++ b/shell/components/templates/home.vue
@@ -11,7 +11,6 @@ import BrowserTabVisibility from '@shell/mixins/browser-tab-visibility';
 import Inactivity from '@shell/components/Inactivity';
 import { mapState, mapGetters } from 'vuex';
 import PromptModal from '@shell/components/PromptModal';
-import WindowManager from '@shell/components/nav/WindowManager';
 import { Layout } from '@shell/types/window-manager';
 
 export default {
@@ -25,14 +24,14 @@ export default {
     AwsComplianceBanner,
     Inactivity,
     PromptModal,
-    WindowManager
   },
 
   mixins: [Brand, BrowserTabVisibility],
 
+  inject: ['notifyWmContainerReady'],
+
   data() {
     return {
-      layout:           Layout.home,
       // Assume home pages have routes where the name is the key to use for string lookup
       name:             this.$route.name,
       noLocaleShortcut: process.env.dev || false,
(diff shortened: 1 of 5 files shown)


shell/components/templates/home.vue
```diff
@@ -82,7 +85,12 @@ export default {
           class="outlet"
         />
       </main>
-      <WindowManager :layout="layout" />
+      <!-- Teleport target for WindowManager (unique per layout) -->
```
codyrancher, line 88: The scoped `.wm` / `.wm-vr` / `.wm-vl` rules below are dead — panels render under App's scope hash. Only working via duplicates in `_layout.scss`. Delete the scoped block.

shell/components/templates/plain.vue
```diff
@@ -98,7 +101,12 @@ export default {
           @shortkey="toggleNoneLocale()"
         />
       </main>
-      <WindowManager :layout="layout" />
+      <!-- Teleport target for WindowManager (unique per layout) -->
```
codyrancher, line 104: Same dead scoped `.wm*` rules as in `home.vue` — delete.

codyrancher (commented):
I really only noticed some dead styling code.

I have a little concern with how we trigger teleports via the mounted hook because I think we could end up with it not triggering in some edge cases but I wasn't able to break it in practice.

torchiaf:
@codyrancher I removed the `.wm*` classes, thanks!

I couldn't replicate the edge cases either - consider that the only scenario where we keep the side windows open when navigating from 1 template to another is the Liz chat, so I think the dashboard is safe - I'll monitor the Chat on Liz.

codyrancher (approved):

PR #17394, opened 2026-04-28
/meta/proxy and ProxyEndpoint extension api

### LINKED ISSUE #16490: Support new `/meta/proxy` process
RFD: https://github.com/SUSE/rancher-architecture/pull/41

Notes
- The existing node driver + /meta/proxy headers will continue to work

Tasks
- [ ] Ensure https://github.com/rancher/rancher/issues/43315 is covered
  - It's likely some parts like CA / trust unsigned won't yet be included. If so we need to make this an epic and spin out some issue
- [ ] Create a library / helper to exercise `/meta/proxy` endpoint
  - Given this will be used in r/d it shouldn't just be extension based
  - Includes replacing / updating docs at https://extensions.rancher.io/extensions/next/provisioning/node-driver/proxying
- [ ] Create a library / helper to manage ProxyEndpoint CRDs
  - More orientated to make it easy for extensions to bootstrap env specific domains in crd
  - TBD - they could just use existing / future store extension library
- [ ] Use new process for current usages of /meta/proxy
  - TBD
  - if not, deprecate r/d code with comments
- [ ] Use new process for new usages of /meta/proxy
  - Ensure we have docs / process in place
- [ ] Document library used to support meta/proxy world
  - This is the supportable layer, not the root api requests, so we need docs on extension site
  - Add link to new docs, replacing existing content, at https://extensions.rancher.io/internal/code-base-works/machine-drivers#api-calls

mantis-toboggan-md:
### Summary
Fixes #16490

This PR adds a new extension API to interface with the /meta/proxy endpoint and associated ProxyEndpoint custom resources. aws, digitalocean, and linode provider stores have been updated to use the new api instead of referencing /meta/proxy directly.

### Technical notes summary
The pr adds a ProxyApi interface (and associated class) that centralizes all /meta/proxy request logic previously scattered across store modules. 
- URL construction (`/meta/proxy/<host>/<path>`)
- X-API-Auth-Header (plain-token auth) header assembly
- X-Api-CattleAuth-Header (Cloud Credential auth) header assembly
- optional automatic de-pagination
- ProxyEndpoint CR management 

Some unit tests have been added. Existing e2e tests should also provide coverage.

Creating cloud credentials and pre-populating provisioning forms for digital ocean, aws (ec2 and eks), and linode clusters.

### Checklist
- [x] The PR has a Milestone 


richard-cox:
We might need to sub divide ProxyRequestOptions, let me know when PRs ready for a quick first review

mantis-toboggan-md:
The method to add a ProxyEndpoint is blocked til https://github.com/rancher/rancher/pull/54828 merges. If you're interested in reviewing the other aspects of this PR that is most welcome, otherwise I'll ping when the r/r pr is in

mantis-toboggan-md:
This PR is ready for review now.

shell/apis/intf/proxy.ts
```diff
@@ -0,0 +1,220 @@
...
+ * - To use a different port: `/meta/proxy/example.com:1234/<path>`.
+ * - For plain HTTP (if absolutely required):
+ *   `/meta/proxy/http:/example.com:1234/<path>` (note one slash after `http:`).
+ *
+ * Notes from docs:
+ * - `X-API-Auth-Header` is forwarded as `Authorization` to the upstream.
+ * - `X-Api-CattleAuth-Header` can instruct Rancher to sign/authenticate
+ *   upstream requests with a Cloud Credential
+ */
+export interface ProxyRequestOptions {
```
richard-cox: This is the surface layer that developers proxying to any and all weird and wonderful endpoints might use, so should be clear, simple and generic.

With that in mind i think we should simplify this to...

```
method?: 'GET' | 'POST' | 'PUT' | 'PATCH' | 'DELETE';
url: Url
headers?: Record<string, string>;
accept?: string;

data?: any;

authentication?: ProxyAuthCloudCredential | ProxyAuthToken
```

where cloud credential and token are

```
ProxyAuthCloudCredential =  {
    id: string,
    authSigner?: 'bearer' | 'awsv4' | 'basic' | 'digest' | 'arbitrary';
    usernameField?: string,
    passwordField?: string,
  }
ProxyAuthToken =  {
    authSigner?: 'bearer' | 'awsv4' | 'basic' | 'digest' | 'arbitrary';
    token: string;
  }
```

so the following are removed

(url based, given we get them all for free in the URL object)
endpoint (now part of URL. very use case specific)
command  (now part of URL. very use case specific)
params (now part of URL. we get this for free there)

(pagination based, given they're all very api specific)
perPage: number
dePaginate?: boolean;
nextUrlPath?: string;
mergeKey?: string;
out?: any;

richard-cox: to keep the API surface clean from the specific API dePaginate stuff we pass through a fn that would post-process things.

so something like

```
postProcess: (res) => Promise
```

which would basically be the `if ( options.dePaginate` part of `request`) as a general util function somewhere that's passed in

shell/apis/intf/shell.ts
```diff
@@ -33,4 +35,9 @@ export interface ShellApi {
    * Provides access to the system API which providers information about the current system
    */
   get system(): SystemApi;
+
+  /**
+   * Provides access to the proxy API
```
richard-cox:
```suggestion
   * Provides access to the proxy API which can be used to make http request via Rancher
```

docusaurus/docs/internal/code-base-works/machine-drivers.md
```diff
@@ -71,15 +71,76 @@ export default class MyDriverMachineTemplate extends MachineTemplate {
...
+Rancher routes UI requests to third-party domains through its server-side `/meta/proxy` endpoint. Use the **`shell.proxy` API** (`this.$shell.proxy`) to make these requests — it handles URL construction, credential-based auth header signing, and de-pagination for you.
 
-TLS and port 443 are assumed.  Add a port after the hostname to change the port (`example.com:1234`).  For plain HTTP, first stop and consider the chain of life decisions which have led you to this point. Then if you still think you need that, use `/meta/proxy/http:/example.com:1234` (note one slash after `http:`, not two).  The hostname must be included in the whitelist defined in global settings, or in the configuration for an active node driver.  If if isn't your request will be denied.  (This prevents a malicious (non-admin) user from abusing the Rancher server as an arbitrary HTTP proxy or reach internal IPs/names that the server can reach directly but the user can't from the outside.)
```
richard-cox, line 76: Comment only.

we might consider supporting this if we get the demand. by using the `URL` class it should be straight forward

shell/apis/intf/proxy.ts
mantis-toboggan-md: All sensible suggestions. I've updated the interface and added a test case to make sure params are preserved when a URL object is passed in.

I kept a generic depaginator function in the proxy file - if this is too much noise I'm happy to move it elsewhere and remove it from the extension docs

docusaurus/docs/internal/code-base-works/machine-drivers.md
mantis-toboggan-md: Why not! I've added support for http urls and added a test case.

shell/store/pnap.js
```diff
@@ -68,6 +68,7 @@ export const actions = {
     return out;
   },
 
+  // DEPRECATED: accessing /meta/proxy directly is deprecated
```
mantis-toboggan-md, line 71: bit more admin work this way but I would like to update pnap independently, after confirming an existing issue is resolved https://github.com/rancher/dashboard/issues/17410

shell/apis/shell/proxy.ts
```diff
@@ -0,0 +1,260 @@
+import { Store } from 'vuex';
+import { ProxyApi, ProxyAuthCloudCredential, ProxyAuthToken, ProxyRequestOptions } from '@shell/apis/intf/shell';
+import { Metadata } from '@shell/types/resources/settings';
+import { MANAGEMENT } from '@shell/config/types';
+
+/** Traverses a dot-separated path on an object, returning `undefined` if any segment is missing. */
+function getByPath(obj: any, path: string): any {
```
richard-cox: if this is a copy and paste from what was there before, no change required.

if this is net new though probably worth using `shell/utils/object.js` `get` (we can then also say we support JSON Path as well)

richard-cox (commented):
Just one minor thing left, otherwise looks good

i gave it a test by manually creating a proxy

```
kind: ProxyEndpoint
metadata:
  name: mine
spec:
  routes:
    - domain: <my test rancher domain>
```

and calling it

```
await this.$shell.proxy.request({ url: new URL('https://<my domain>/v1') })
```

and got the desired result.

haven't gone through the providers (DO, etc), but will leave that with you and QA

shell/apis/shell/proxy.ts
mantis-toboggan-md: fair point re: JSONpath. I've updated this function to use the pre-existing get object utility

richard-cox (approved):
I tested again and the call to `proxy.request` kept failing with a 502 error

Looking the logs it prints

```
http: proxy error: tls: failed to verify certificate: x509: certificate signed by unknown authority
```

i had used a different rancher instance with a dodgy cert. i think at some point the proxy endpoint (we call) might need to support unsigned certs, but not right now. Also the endpoint should pass that kind of error back. We can ping them separately about that side

mantis-toboggan-md:
Are unsigned certs the same as self-signed and would that be covered by this RFE? https://github.com/rancher/rancher/issues/53667

PR #17402, opened 2026-04-29
Declare `diff` as a dependency for Dashboard and Shell

### LINKED ISSUE #17401: Declare `diff` as a direct dependency
`FileDiff.vue` imports `createPatch` from `diff`, but `diff` is a transitive dependency. In the past, we've experienced breakages when transitive dependencies changed without notice. We also want to keep our dependency policy explicit for improved security and maintenance.

## Proposed solution

Add `diff` as an explicit dependency in the `package.json` for Dashboard root and `shell/`.

## Acceptance criteria

- [ ] `diff` is declared as a dependency for both the root `package.json` and in `shell/`
- [ ] `FileDiff.vue` continues to work as expected

rak-phillip:
### Summary
This explicitly declares `diff` as a dependency of Rancher Dashboard and Shell.

Fixes #17401 

- Declare `diff` as a dependency for Dashboard and Shell

### Technical notes summary

This is being done so that we can be explicit with dependencies that we consume in our code. While we may import from transitive dependencies, they can change without our knowledge, forcing us to react to changes that might come in.

I find that the compare changes button in Local Cluster Explorer => Apps => Charts works for a good test.

The version of diff we are targeting is the latest version, several versions ahead of other dependencies. It's difficult to assert which one we were actually using, but it appears to behave the same as before.

### Screenshot/Video

image

image

### Checklist
- [x] The PR has a Milestone 


codyrancher (approved):

package.json
```diff
@@ -93,6 +93,7 @@
     "d3": "7.3.0",
     "d3-selection": "3.0.0",
     "dayjs": "1.11.18",
+    "diff": "9.0.0",
```
cnotv, line 96: Nitpicking: I don't think this is needed if we have it inside the shell.

cnotv (approved):
LGTM

PR #17441, opened 2026-04-30
Create form validation parser

### LINKED ISSUE #17354: Find a way to migrate existing forms with validation to vee-validate

rak-phillip:
### Summary
This creates a `useFormValidation()` composable that acts as a bridge between existing validation rules defined in the form validation mixin and vee-validate. This also updates the secret forms to demonstrate replacing existing form validation logic with the `useFormValidation()` composable.

Fixes #17354

- Update LabeledSelect and NameNsDescription for vee-validate
- Create form validation parser
- Migrate Secret form to vee-validate
- Add new validation rules for SSH secrets
- Add validation for custom secrets
- Add validation for remaining secret forms
- Create `useVeeValidateField()` composable
- Fix failing test by updating auth requirements

### Technical notes summary

This change intends to introduce a "bridge" that will help us to transition away from the form validation mixin over to vee-validate with a minimal amount of refactoring. By introducing a parser that translates form validation rules into a format usable by vee-validate, most of our future effort will involve identifying existing components with validators already defined and introducing a setup option that invokes `useFormValidation()` with the same rules applied. Most of this work could be automated if this approach works reliably enough. 

This PR also demonstrates how we could use the composable to easily expand validation rules to existing components via `useFromRules()`. `useFromRules()` relies on some root vee-validate form implementation, typically applied by `useFormValidation()` in these examples.  

Cluster Explorer => Storage => Secrets. Ensure that all secrets properly pass/fail validation and can be created/updated when validation passes.

Users & Authentication => User retention settings. Validation should behave as it did before with the minor changes to the vee-validate implementation. 

#### Validation fails when missing required inputs

image

#### Validation passes

image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/composables/useFormValidation.ts
+++ b/shell/composables/useFormValidation.ts
@@ -1,0 +1,81 @@
+import { computed, provide, ref } from 'vue';
+import { useForm } from 'vee-validate';
+import formRulesGenerator from '@shell/utils/validators/formRules/index';
+import type { Validator } from '@shell/utils/validators/formRules/index';
+import type { Translation } from '@shell/types/t';
+
+export interface RuleSet {
+  path: string;
+  rules: string[];
+  translationKey?: string;
+}
+
+const nullValidator: Validator = () => undefined;
+
+function createRuleResolver(
+  t: Translation,
+  ruleSets: RuleSet[],
+  extraRules: Record<string, Validator>
+) {
+  return (path: string): Validator[] => {
+    const set = ruleSets.find((s) => s.path === path);
+
+    if (!set) return [];
+
+    const key = set.translationKey ? t(set.translationKey) : 'Value';
+    const allRules = {
+      ...formRulesGenerator(t, { key }),
+      ...extraRules,
+    };
+
+    return set.rules.map((r) => (allRules[r] as Validator) || nullValidator);
+  };
+}
+
+/**
+ * For root/parent form components. Creates a vee-validate form context and
+ * resolves fvFormRuleSets-style rule string arrays into validator functions.
+ * Child components rendered within this form should use useFormRules instead.
+ */
+export function useFormValidation(
+  t: Translation,
+  ruleSets: RuleSet[],
+  extraRules: Record<string, Validator> = {}
+) {
+  const { errors, validate } = useForm();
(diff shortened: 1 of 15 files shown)


shell/composables/useFormValidation.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    if (!set) return [];
+
+    const key = set.translationKey ? t(set.translationKey) : 'Value';
+    const allRules = {
+      ...formRulesGenerator(t, { key }),
+      ...extraRules,
+    };
+
+    return set.rules.map((r) => (allRules[r] as Validator) || nullValidator);
```
codyrancher: When a rule name doesn't match any key in `allRules`, the fallback to `nullValidator` silently passes validation. A typo like `'reqired'` instead of `'required'` would be invisible at runtime.

Maybe we should improve the type system here or do runtime errors in dev to help catch it?

rak-phillip: ea1d3a28b6be548821e0d73606a8f148a292a83a introduces a change so that we throw when an invalid validation rule is encountered in environments outside of production. 

I think that we can follow up to enhance the typescript experience. I'm realizing again that our spotty typescript coverage in components makes this less useful, but it is something that we can benefit from in components that target typescript.

codyrancher (approved):

PR #17450, opened 2026-05-01
Fix runAsUser and fsGroup validation error when clearing fields

### LINKED ISSUE #9601: Removing
**Setup**
- Rancher version: 2.7.5
- Rancher UI Extensions: none
- Browser type & version: 116.0.5845.111 (Official Build) (64-bit) (Windows 11)

**Describe the bug**
removing Security Context "User ID" causes error in UI

**To Reproduce**
1. open cluster
2. select workloads
3. Select Create
4. choose Deployment
5. on container-0 provide image (e.g nginx:latest)
6. select Security Context
7. Enter 33 in "Run as User ID"
8. Save
9. Edit Config for workload just created
10. Under Security Context remove 33 from "Run as User ID"

**Result**
Deployment in version "v1" cannot be handled as a Deployment: json: cannot unmarshal string into Go struct field SecurityContext.spec.template.spec.containers.securityContext.runAsUser of type int64

**Expected Result**
Config change should be saved

Note:
When removing the number single quotes are left behind causing the failure.
If I edit yaml and remove ID then it saves correctly

Tested on 2.7.5 and 2.7.7-rc4 - both have issue

codyrancher:
### Summary
Fixes #9601

Clearing `runAsUser` or `fsGroup` inputs sent `""` or `null` instead of omitting the key.

### Technical notes summary
- Strip `null`, `""` and `undefined` values for `fsGroup`/`runAsUser` in `Security.vue`'s `update()` before emit
- Add `type="number" min="0"` to `runAsUser` input (was missing); add `min="0"` to `fsGroup` input

- Edit a Deployment with `runAsUser` set, clear the field, save - should succeed
- Same for `fsGroup` on the pod security context tab

- Workload forms using the `Security` component

### Screenshot/Video

**Before fix** - clearing Run as User ID and saving produces `cannot unmarshal string into Go struct field of type int64`:

https://github.com/user-attachments/assets/5ac47a58-7ea3-47c1-801e-cc2f25c1a7ae

**After fix** - clearing Run as User ID and saving succeeds, redirects to deployment list:

https://github.com/user-attachments/assets/0324970f-fec0-4815-8601-24159f387229

### Checklist
- [x] The PR has a Milestone

### DIFF
--- a/shell/components/form/Security.vue
+++ b/shell/components/form/Security.vue
@@ -140,11 +140,12 @@ export default {
         this.afterPrivilegedTickedMessage = this.t('workload.container.security.privileged.afterTick.false');
       }
 
-      if (securityContext.fsGroup === '') {
+      // Drop empty values so we don't send a string for int64 fields.
+      if (securityContext.fsGroup === '' || securityContext.fsGroup === null) {
         delete securityContext.fsGroup;
       }
 
-      if (securityContext.runAsUser === '') {
+      if (securityContext.runAsUser === '' || securityContext.runAsUser === null) {
         delete securityContext.runAsUser;
       }
 
@@ -175,6 +176,7 @@ export default {
             ref="firstFocusable"
             v-model:value.number="securityContext.fsGroup"
             type="number"
+            min="0"
             :mode="mode"
             :label="t('workload.container.security.fsGroup')"
             @update:value="update"
@@ -283,6 +285,8 @@ export default {
           </legend>
           <LabeledInput
             v-model:value.number="securityContext.runAsUser"
+            type="number"
+            min="0"
             :label="t('workload.container.security.runAsUser.label')"
             :mode="mode"
             @update:value="update"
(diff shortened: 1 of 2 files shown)


shell/components/form/Security.vue
```diff
@@ -140,11 +140,12 @@ export default {
         this.afterPrivilegedTickedMessage = this.t('workload.container.security.privileged.afterTick.false');
       }
 
-      if (securityContext.fsGroup === '') {
+      // Drop empty values so we don't send a string for int64 fields.
+      if (securityContext.fsGroup === '' || securityContext.fsGroup === null) {
```
torchiaf: could this also be undefined by any chance?

codyrancher: Yeah, it looks like it could be. I went ahead and put an explicit check in for undefined as well. We could combine the checks into a simpler expression but I thought being explicit would make the intentions more clear.

torchiaf (approved):
LGTM

PR #17512, opened 2026-05-06
Validation for duplicate pool names

### LINKED ISSUE #5211: Duplicate Machine Pool Names are not caught when saving a v2prov cluster
**Setup**
- Rancher version: v2.6-head

While reviewing a Rancher [PR](https://github.com/rancher/rancher/pull/36614#discussion_r814420338) and testing functionality changes for the PR, we found a bug. The rancher backend doesn't allow duplicate machine pool names but the UI let's an end-user save a cluster config that contains duplicate machine pool names and then the cluster immediately goes into an error state.

The user should not be able to save the cluster config when it will immediately cause their cluster to enter an error state. The backend checks for the duplicate names and the UI needs to do the same. The machine pool isn't shown in the UI after saving but the configuration is present in both the form and yaml config options.

image

image

image

```yml
    machinePools:
    - controlPlaneRole: true
      etcdRole: true
      machineConfigRef:
        kind: VmwarevsphereConfig
        name: nc-ross46-rke2-linux-4kmwq
      machineOS: linux
      name: rke2-linux
      quantity: 3
      unhealthyNodeTimeout: 0s
      workerRole: true
    - controlPlaneRole: false
      etcdRole: false
      labels: {}
      machineConfigRef:
        kind: VmwarevsphereConfig
        name: nc-ross46-rke2win-x9jlq
      machineOS: windows
      name: rke2win
      quantity: 1
      unhealthyNodeTimeout: 0m
      workerRole: true
    - controlPlaneRole: false
      etcdRole: false
      labels: {}
      machineConfigRef:
        kind: VmwarevsphereConfig
        name: nc-ross46-rke2-linux-drs7h
      name: rke2-linux
      quantity: 1
      unhealthyNodeTimeout: 0m
      workerRole: true
```

momesgin:
### Summary
Fixes #5211

The Rancher backend rejects duplicate machine pool names, but the UI did not validate for this. Users could save a cluster config with duplicate pool names, only to get a backend error. This PR adds client-side validation to catch duplicates before submission.

- Added a `uniquePoolName` validation rule to the machine pool name field. When a duplicate name is detected (case-insensitive), the field shows a warning icon with the tooltip "Pool names should be unique" and the Create/Save button is disabled.

### Technical notes summary
- The validation uses the existing `fvExtraRules` mechanism from the `FormValidation` mixin.
- The comparison is case-insensitive to match backend behavior.
- Pools marked for removal are excluded from the duplicate check.
- Changed the v-for key on the machine pool template from array index (idx) to `obj.id`. The index was causing the error icon not going away after removing the pool with duplicate name.
- Replaced the hardcoded '(Not Named)' string in the machine pool tab label with the translation key `cluster.machinePool.name.notNamed`.

- Create a cluster with multiple machine pools using the same name — the validation error should appear and the Create button should be disabled.
- Create a cluster with pools that differ only by case (e.g. `Pool1` and `pool1`) — should be flagged as duplicates.
- Remove a pool that causes a duplicate — the validation error on the remaining pool should clear.
- Existing validations (empty name, quantity) should still work as before.

- Machine pool creation and editing flow.
- The Create/Save button enabled state, since `machinePoolValidation` feeds into `validationPassed`.

### Screenshot/Video

https://github.com/user-attachments/assets/2a04176d-aff6-4a49-bf47-6f313057c4b5

### Checklist
- [x] The PR has a Milestone 


rak-phillip (commented):
This looks good, but I have one nit. When I enter a duplicate pool name and switch back to the first tab, the "Pool Name" field does not relay an error until the input gains focus. This might be a quirk of our existing validation mixin.

https://github.com/user-attachments/assets/bc4b202e-b701-4a7b-89b9-ea77ff1f0937

momesgin:
> This looks good, but I have one nit. When I enter a duplicate pool name and switch back to the first tab, the "Pool Name" field does not relay an error until the input gains focus. This might be a quirk of our existing validation mixin.
> 

Good catch! passing `false` value for `requireDirty` prop to the name field's `LabeledInput` did the trick. It bypasses the need to have the input being focused for running the validation.

rak-phillip (approved):
LGTM

PR #17558, opened 2026-05-08
Fix percent rollout inputs

### LINKED ISSUE #6179: Can't type value for inputs with % units
**Setup**
- Rancher version: 2.6.5
- Browser type & version: all

**Describe the bug**
In input with units % it's not possible to type the value.

**To Reproduce**
- Open any Cluster -> Workload -> Deployment form
- Open Scaling and Upgrade Policy tab
- Type any number

**Result**
Value is cancelled and any typed key is removed as well.

**Expected Result**
It's possible to type as using the unit `Pods`

**Screenshots**

https://user-images.githubusercontent.com/5009481/173865055-2e483c53-9b6f-4203-a9df-5680d757faad.mov

**Additional context**
Code not changed since 22 months, however the bug seems to appear after 2.6.4.
https://github.com/rancher/dashboard/blob/master/shell/edit/workload/Upgrading.vue#L174-L179

vvlisn:
### Summary
Fixes #6179

Fixes percentage-based rollout inputs in the Workload Deployment form so users can type values when the unit is `%`.

- Keeps `maxSurge` and `maxUnavailable` form state as numeric input values while users edit percentage-based rollout settings.
- Appends `%` only when writing the rolling update values back to the workload spec.
- Enables regression coverage for percentage max surge and max unavailable values.

### Technical notes summary
- `InputWithSelect` is rendered with `type=\"number\"`, so storing values like `123%` in the input-bound state causes the browser to reject the value and clear the input.
- The fix preserves the numeric display value for the input while still persisting Kubernetes rollout values as strings with `%` when the selected unit is `%`.
- The `%` check now stringifies the value before calling `includes`, which also handles numeric values when switching units.

- Workload Deployment create/edit form > Scaling and Upgrade Policy > Max Surge with `%` selected.
- Workload Deployment create/edit form > Scaling and Upgrade Policy > Max Unavailable with `%` selected.
- Switch Max Surge and Max Unavailable between `Pods` and `%`, then type values and save.
- Confirm saved workload spec stores percentage values with `%` and pod-count values without `%`.
- Local automated test run:
  - `source \"$HOME/.nvm/nvm.sh\" && nvm use 24 && yarn test:ci shell/edit/workload/__tests__/Upgrading.test.ts`

- Workload Deployment rolling update fields: Max Surge and Max Unavailable.
- Workload DaemonSet rolling update Max Unavailable field.
- Unit switching behavior between `Pods` and `%` in workload upgrade policy inputs.

### Screenshot/Video
N/A. This is a form input behavior fix covered by unit tests.

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/edit/workload/__tests__/Upgrading.test.ts
+++ b/shell/edit/workload/__tests__/Upgrading.test.ts
@@ -26,9 +26,9 @@ describe('component: Upgrading', () => {
   });
 
   it.each([
-    ['surge', 'maxSurge', '%'],
-    ['unavailable', 'maxUnavailable', '%'],
-  ])('should set typed value in %p into %p and unit', (_field, key, unit) => {
+    ['maxSurge', '%'],
+    ['maxUnavailable', '%'],
+  ])('should set typed value in %p with %p unit', (key, unit) => {
     const wrapper = mount(Upgrading);
     const newValue = 123;
     const expectation = `${ newValue }${ unit }`;
--- a/shell/edit/workload/__tests__/Upgrading.test.ts
+++ b/shell/edit/workload/__tests__/Upgrading.test.ts
@@ -25,20 +25,17 @@ describe('component: Upgrading', () => {
     expect(wrapper.props('value')?.[key]).toBe(newValue);
   });
 
-  // TODO: #6179: Integrate test with component fix, as the scope is not to check the value of the input
-  // eslint-disable-next-line jest/no-disabled-tests
-  it.skip.each([
+  it.each([
     ['surge', 'maxSurge', '%'],
     ['unavailable', 'maxUnavailable', '%'],
-  ])('should set typed value in %p into %p and unit', (field, key, unit) => {
+  ])('should set typed value in %p into %p and unit', (_field, key, unit) => {
     const wrapper = mount(Upgrading);
-    const input = wrapper.find(`[data-testid="input-policy-${ field }"]`).find('input');
     const newValue = 123;
     const expectation = `${ newValue }${ unit }`;
 
-    input.setValue(newValue);
-    input.trigger('blur');
+    wrapper.vm.updateWithUnits({ selected: unit, text: newValue }, key);
 
+    expect(wrapper.vm[key]).toBe(newValue);
     expect(wrapper.props('value')?.strategy.rollingUpdate[key]).toBe(expectation);
   });
 
--- a/shell/edit/workload/Upgrading.vue
+++ b/shell/edit/workload/Upgrading.vue
@@ -133,10 +133,10 @@ export default {
       } = this;
       let { maxSurge, maxUnavailable } = this;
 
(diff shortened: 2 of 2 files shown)


richard-cox:
@vvlisn can you update the description with the PR description template?

vvlisn:
Updated the PR description with the requested template. Thanks!

shell/edit/workload/__tests__/Upgrading.test.ts
```diff
@@ -25,20 +25,17 @@ describe('component: Upgrading', () => {
...
   });
 
-  // TODO: #6179: Integrate test with component fix, as the scope is not to check the value of the input
-  // eslint-disable-next-line jest/no-disabled-tests
-  it.skip.each([
+  it.each([
     ['surge', 'maxSurge', '%'],
     ['unavailable', 'maxUnavailable', '%'],
-  ])('should set typed value in %p into %p and unit', (field, key, unit) => {
+  ])('should set typed value in %p into %p and unit', (_field, key, unit) => {
```
codyrancher: If we're not going to use the `_field` it seems like we should just remove that column from the test data.

codyrancher (changes requested):
I just have the one change otherwise this looks good and seems to work well. Thanks for opening the PR!

Verified we can have percentages or pods and switch between the two while editing.

### Before (bug)

Typing a value in Max Surge or Max Unavailable with `%` selected clears the input on blur.

https://github.com/user-attachments/assets/0545f72e-6838-4579-b71c-e9967ded63f2

### After (fix)

Values are retained after blur. The `%` is appended only when writing to the workload spec.

https://github.com/user-attachments/assets/2148c712-962b-4e13-8ad3-bad3127a6a01

vvlisn:
Updated the test data to remove the unused field column and verified the focused test passes

shell/edit/workload/__tests__/Upgrading.test.ts
```diff
@@ -25,20 +25,17 @@ describe('component: Upgrading', () => {
...
+    ['maxSurge', '%'],
+    ['maxUnavailable', '%'],
+  ])('should set typed value in %p with %p unit', (key, unit) => {
     const wrapper = mount(Upgrading);
-    const input = wrapper.find(`[data-testid="input-policy-${ field }"]`).find('input');
     const newValue = 123;
     const expectation = `${ newValue }${ unit }`;
 
-    input.setValue(newValue);
-    input.trigger('blur');
```
codyrancher, line 40: Just as a note if anyone else reviews this, I decided switching this over to `updateWithUnits` was more inline with unit testing.

Doing the `setValue` and `trigger` is closer to an integration test.

codyrancher (approved):

codyrancher:
I'll go ahead and merge since it's an external contributor.

@vvlisn thanks for putting in the effort to open the pr and respond to feedback!

PR #17561, opened 2026-05-08
fix: scope configureType by product to prevent global route overrides

### LINKED ISSUE #17400: Bug: Menu items added by ProductChildResourcePage break explorer links
**Setup**
- Rancher version: v2.15.0 (dev of)

**Describe the bug**
- create an extension with a product with a ProductChildResourcePage pointing to a resource, say pods
- the explorer product's pods link points to the extension product's pod list
- clicking on it takes the user from the explorer view to the extension view
- resource links should retain the current product

**To Reproduce**
```
  const product: ProductMetadata = { name: 'product-1', label: 'Product One' };
  const childResourcePage: ProductChildResourcePage = { type: 'pod' };

  plugin.addProduct(product, [childResourcePage]);
```

- Navigate to any cluster --> Workloads --> Pods

**Result**
- User taken to the extension product's pods page

**Expected Result**
- User stay in the explorer product and sees the pods list

**Screenshots**

Image

Image

aalves08:
### Summary
Fixes #17400 

**Root Cause**

The `state.typeOptions` array was flat and global, with no product scoping. When `configureType` was called with `customRoute`, it stored the route globally. The `optionsFor` getter returned the first match without filtering by product, causing routes from one product to leak into others.

**Solution**

Scope `typeOptions` by product, following the existing architectural pattern used by `virtualTypes`, `spoofedTypes`, and `basicTypes`. This ensures each product's type configuration is isolated and only applies within that product's context.

**Changes**

`shell/store/type-map.js`

1. **State initialization (line 420)**
   - Changed `typeOptions: []` to `typeOptions: {}`
   - Now stores options as `{ [product]: [options] }` instead of flat array

2. **DSL method (line 298)**
   - Updated `configureType` to pass `product` parameter from closure to mutation
   - No changes needed to extension code - automatically works correctly

3. **optionsFor getter (lines 567-568)**
   - Added `const currentProduct = rootGetters['productId']`
   - Filter `typeOptions` by current product: `state.typeOptions[currentProduct] || []`
   - Fallback to empty array if product not configured

4. **Remove mutation (lines 1534-1536)**
   - Added cleanup of `state.typeOptions[product]` when product is removed
   - Prevents memory leaks when extensions are uninstalled

5. **configureType mutation (lines 1800-1828)**
   - Extract and validate `product` parameter
   - Initialize product-scoped array if needed: `state.typeOptions[product] = []`
   - Store options in product-scoped array instead of flat array
   - Maintain existing merge logic for multiple configures of same type

`shell/store/__tests__/type-map.test.ts`

Added 9 new unit tests covering:
- Product scoping isolation
- Merging multiple configures in same product
- Custom data object merging
- Error handling for missing product parameter
- Multiple types in same product
- optionsFor filtering by product
- Default behavior when type not configured
- Missing product graceful handling
- Product isolation verification

- Edit `pkg/kubectl-explain/index.ts` and add the following code:
```
// Add a product with a child resource page (new product registration)
  plugin.addProduct({
    name:  'product-1',
    label: 'Product One',
  }, [{ type: 'pod' }]);
```
- a new top-level extension should appear on the sidebar
- Navigate to any cluster --> Workloads --> Pods
- Pods list view page should retain the context of the cluster
- test links for `create`, `edit`, `detail` and make sure links are consistent with the cluster explorer page
- Navigate to the new top-level-product
- Pods list view page on the new product registered should follow `<!--PRODUCT-->/c/_/pod`
- test links for `create`, `edit`, `detail` and make sure links are consistent with the url structure from the previous point

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/config/product/manager.js
+++ b/shell/config/product/manager.js
@@ -73,6 +73,7 @@ export function init(store) {
   ]);
 
   configureType(SNAPSHOT, { depaginate: true });
+  configureType(CATALOG.CLUSTER_REPO, { listCreateButtonLabelKey: 'catalog.repo.add' });
 
   configureType(CAPI.RANCHER_CLUSTER, {
     showListMasthead: false, namespaced: false, alias: [HCI.CLUSTER]
--- a/shell/store/type-map.js
+++ b/shell/store/type-map.js
@@ -551,7 +551,7 @@ export const getters = {
       subTypes:                 [],
     };
 
-    return (schemaOrType, pagination) => {
+    return (schemaOrType, pagination, product) => {
       // Note - This can run a LOT so needs to be performant
 
       if (!schemaOrType) {
@@ -559,8 +559,8 @@ export const getters = {
       }
 
       const type = (typeof schemaOrType === 'object' ? schemaOrType.id : schemaOrType);
-      const currentProduct = rootGetters['productId'];
-      const productTypeOptions = state.typeOptions[currentProduct] || [];
+      const productToUse = product || rootGetters['productId'];
+      const productTypeOptions = state.typeOptions[productToUse] || [];
       const found = productTypeOptions.find((entry) => {
         const re = stringToRegex(entry.match);
 
@@ -948,7 +948,7 @@ export const getters = {
         });
 
         const attrs = schema.attributes || {};
-        const typeOptions = getters['optionsFor'](schema);
+        const typeOptions = getters['optionsFor'](schema, undefined, product);
 
         schemaModes[TYPE_MODES.BASIC] = schemaModes[TYPE_MODES.BASIC] && getters.groupForBasicType(product, schema.id);
 
@@ -1859,8 +1859,10 @@ export const actions = {
     dispatch('prefs/set', { key: EXPANDED_GROUPS, value: groups }, { root: true });
   },
 
-  configureType({ commit }, options) {
-    commit('configureType', options);
+  configureType({ commit, rootGetters }, options) {
+    const product = options.product || rootGetters['productId'];
+
+    commit('configureType', { ...options, product });
   }
 };
 
--- a/shell/store/type-map.js
+++ b/shell/store/type-map.js
@@ -20,7 +20,7 @@
 // importList(type)                 Returns a promise that resolves to the list component for type
 // importDetail(type[,subType])     Returns a promise that resolves to the detail component for type
 // importEdit(type[,subType])       Returns a promise that resolves to the edit component for type
-// optionsFor(schemaOrType)         Return the configured options for a type (from configureType)
+// optionsFor(schemaOrType, pagination(bool), product(string)) Return the configured options for a type (from configureType) - additional product param can be passed if "optionsFor" needed isn't the current product
 //
 // 3) Changing specialization info about a type
 // For all:
@@ -107,6 +107,7 @@
 //                               notFilterNamespace:  undefined -- Define namespaces that do not need to be filtered
 //                               localOnly: False -- Hide this type from the nav/search bar on downstream clusters
 //                               custom: any - Custom options for a given type
+//                               product: string - If set, this type's options will only apply when that product is active
 //                           }
 // )
 // ignoreGroup(group):        Never show group or any types in it
@@ -373,6 +374,18 @@ export function DSL(store, product, module = 'type-map') {
   };
 }
 
+const LEGACY_COMPATIBILITY_BUCKET = 'legacyCompatibilityProdRegistration';
+
+function validateProductName(product) {
+  if (!product || typeof product !== 'string' || product.trim() === '') {
+    throw new Error(`Product name must be a non-empty string, got: ${ JSON.stringify(product) }`);
+  }
+
+  if (product === LEGACY_COMPATIBILITY_BUCKET) {
+    throw new Error(`Product name cannot be "${ LEGACY_COMPATIBILITY_BUCKET }" as it is reserved for backward compatibility`);
+  }
+}
+
 let called = false;
 
 export async function applyProducts(store, $extension) {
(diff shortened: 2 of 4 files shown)


aalves08:
FYI @nwmac @richard-cox

shell/store/type-map.js
```diff
@@ -412,7 +414,7 @@ export const state = function() {
     typeMappings:            [],
     typeMoveMappings:        [],
     typeToComponentMappings: [],
-    typeOptions:             [],
+    typeOptions:             {},
     groupBy:                 {},
```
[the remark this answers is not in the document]
aalves08: productDebugger fixed

shell/store/type-map.js
```diff
@@ -1784,20 +1792,35 @@ export const mutations = {
...
-    const match = regexToString(ensureRegex(options.match));
+    const { product, ...typeOptions } = options;
+
+    if (!product) {
+      // eslint-disable-next-line no-console
+      console.error('configureType called without product parameter');
+
+      return;
+    }
+
```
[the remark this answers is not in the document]
aalves08: The configureType action (line 1862) didn't handle the case where callers don't provide a product parameter. Two call sites exist without it:

shell/list/helm.cattle.io.projecthelmchart.vue:38 -> VERY BAD USECASE!!!
shell/pages/c/_cluster/monitoring/route-receiver/index.vue:41, 44 -> VERY BAD USECASE!!!

Solution: Modified the action to default product to the current product (rootGetters['productId']) when not provided. This maintains backward compatibility while fixing the silent failure.

Change:
configureType({ commit, rootGetters }, options) {
  const product = options.product || rootGetters['productId'];
  commit('configureType', { ...options, product });
}
Now those component call sites work without changes and use the appropriate product context.

shell/store/type-map.js
```diff
@@ -557,7 +559,9 @@ export const getters = {
       }
 
       const type = (typeof schemaOrType === 'object' ? schemaOrType.id : schemaOrType);
-      const found = state.typeOptions.find((entry) => {
+      const currentProduct = rootGetters['productId'];
+      const productTypeOptions = state.typeOptions[currentProduct] || [];
+      const found = productTypeOptions.find((entry) => {
```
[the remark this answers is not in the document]
aalves08: When SideNav builds navigation for multiple products simultaneously, `optionsFor` was always using the currently viewed product's configuration instead of the product being queried, which could cause wrong types to be shown/hidden based on incorrect `localOnly`, `ifRancherCluster`, and other product-specific settings

Example: PSA type configured with localOnly: true in explorer would be treated as localOnly: false when viewing from another product

The fix applied ensures each product's configuration is properly scoped and used when building navigation trees.

aalves08:
@rak-phillip @eva-vashkevich @codyrancher @mantis-toboggan-md ready for a re-review! 🙏 🙇 🙇 🙇 🙇 🙇 🙇

shell/store/type-map.js
```diff
@@ -1784,20 +1792,35 @@ export const mutations = {
...
 
   configureType(state, options) {
-    const match = regexToString(ensureRegex(options.match));
+    const { product, ...typeOptions } = options;
 
-    const idx = state.typeOptions.findIndex((obj) => obj.match === match);
-    let obj = { ...options, match };
+    if (!product) {
+      // eslint-disable-next-line no-console
+      console.error('configureType called without product parameter');
```
codyrancher: Is there anything more we can include in the error message so it's easier to debug? The `type` or `match` maybe?

Also, this is going to be a breaking change for extensions won't it? I didn't see that notated in the PR summary.

shell/config/product/manager.js
```diff
@@ -73,6 +73,7 @@ export function init(store) {
   ]);
 
   configureType(SNAPSHOT, { depaginate: true });
+  configureType(CATALOG.CLUSTER_REPO, { listCreateButtonLabelKey: 'catalog.repo.add' });
```
codyrancher, line 76: Is this just tidying up for another issue? It didn't seem immediately related to me.

shell/store/type-map.js
```diff
@@ -549,15 +551,17 @@ export const getters = {
       subTypes:                 [],
     };
 
-    return (schemaOrType, pagination) => {
+    return (schemaOrType, pagination, product) => {
```
codyrancher, line 567: I don't see the product in the docs at the top of this file for optionsFor.

shell/store/type-map.js
```diff
@@ -1784,20 +1792,35 @@ export const mutations = {
   },
 
   configureType(state, options) {
-    const match = regexToString(ensureRegex(options.match));
+    const { product, ...typeOptions } = options;
```
codyrancher, line 1828: I think we're going to need to add product to the docs at the top as well.

codyrancher (commented):
From testing this does appear to work. My biggest concern is that I think this is going to be a breaking change for extensions because of how we're changing the function signatures. If that's intended and we're just documenting it then that's fine.

Here's some recordings for the tests:

**1. Sidebar shows Product One extension**
https://github.com/user-attachments/assets/e02d1e6c-2280-4852-ac45-4a2d865d0650

**2. Explorer Pods list retains cluster context**
https://github.com/user-attachments/assets/6b9c7064-cfdd-45c2-88d7-911f87294b54

**3. Explorer Pods create/edit/detail links**
https://github.com/user-attachments/assets/4864a3e2-cfd3-4baa-9802-b5ae7e7bd1f6

**4. Product One Pods list URL structure**
https://github.com/user-attachments/assets/fbcfc8c4-46e5-4bed-88cc-44c0743bd237

**5. Product One Pods create/edit/detail links**
https://github.com/user-attachments/assets/518c92bc-0b67-4bfd-be2e-57a5c2479a52

shell/config/product/manager.js
aalves08: Since DSL `typeOptions` is now a keyed object, direct configuration for `configureType` is not shared anymore, so options passed directly in `configureType` aren't used in both places anymore, hence why I had to add the `listCreateButtonLabelKey` here (was failing an e2e test)

shell/store/type-map.js
aalves08: we were missing pagination as well (bool). Added both, plus a brief explanation on the usefulness of the `product` param there

shell/store/type-map.js
aalves08: added

shell/store/type-map.js
aalves08: You've got me thinking about a lot of this stuff.... Glad you did it. I think I was able to devise a solution where we ensure compatibility on extensions in all scenarios.

Please have a good look at the code and see if you agree with the implementation

codyrancher: I think this gets us there, I've only identified two scenarios where it's not backwards compatible and I don't think either will happen in practice.

### One: calling `commit` directly
If an extension were to do something like `store.commit('type-map/configureType', { match: 'pod', isCreatable: false });` we could still get an exception thrown. This doesn't seem very likely and I haven't found any cases.

### Two: calling the action without a product when a product isn't currently globally set.
I've attempted to trigger this but I haven't been able to so I think we're fine.

codyrancher (approved):

Issue #11338, opened 2024-07-02
Reduce clickable area of ‘user retention settings’ button


yonasberhe23:
**Setup**
- Rancher version: v2.9-f188faae847b960d2f30cc89f4befc8ce5944f2c-head
- Rancher UI Extensions:
- Browser type & version: Any

**Describe the bug**
Reduce the clickable area of the ‘user retention settings’ button. Currently, the button is clickable anywhere inline, on the right side of the button so if a user were to misclick the 'Create' button they would land on the user retention page inadvertently.

**To Reproduce**

**Result**
The ‘user retention settings’ button is clickable anywhere inline, on the right side of the button.

**Expected Result**
Only the ‘user retention settings’ button is only clickable.

**Screenshots**

https://github.com/rancher/dashboard/assets/127343932/f8d47147-3a60-46ed-8828-8a841b298ada

**Additional context**

PR #17574, opened 2026-05-08
Replace anchor tags styled as buttons with button component

### LINKED ISSUE #13174: Refactor anchor tags that behave as buttons

rak-phillip:
### Summary
This replaces all link button instances with the `RcButton` implementation. #17178 introduced a change to allow `RcButton` to render link buttons when the `to` prop is supplied. The benefit of this is that link buttons can now be handled in a controlled and consistent way. 

Fixes #13174 

- Replace router-link tags with buttons
- Trigger click when space is pressed for link buttons
- Replace anchor tag buttons with button component
- Clean up template comments in home.vue
- Retain link semantics for skip to main content
- Replace `n-link` button instances
- Add `data-testid` to cluster member add button

### Technical notes summary

For accessibility and consistency, link buttons should register a "click" on pressing both "space" and "enter" via the keyboard; this is an effort we addressed during the accessibility epic. While testing I noted that "space" was not registering the click properly when rendering a link button through RcButton, so this has been fixed via 4fa5db820f665574eef94819e202d787467a269a. All link buttons should now have a consistent mode of keyboard interaction throughout Dashboard with this change in place. 

46a87294fbc7dae2f54c4fbec4d5f40ad9649930 contains a change to preserve the "skip to main content" link semantics. It should be a minor change to `vue-router`, but has a global effect that can impact router behavior. More details about this change can be found in the vue-router docs[^1]. 

[^1]: https://router.vuejs.org/guide/advanced/scroll-behavior.html

Items to note:

- `btn-sm` closely maps to the "medium" size for RcButton. This is the default, so the `size` prop has been omitted in these areas.
- Link buttons without a size (`btn-`) class have been assumed to be the "large" size for RcButton. This is being done to maintain consistency with the existing button heights.
- `role-primary` maps to the "primary" variant for RcButton. This is the default, so the `variant` prop has been omitted in these areas. 

- RcButton with `to` prop renders as router-link (<a>), space bar triggers navigation in addition to the enter key
- RcButton with `to` prop carries role="link" so screen readers announce it as a link, not a button
- `n-link` button replacements navigate to the correct routes in Harvester cluster list
- Replaced anchor tag buttons in SAML auth page still perform their actions
- Replaced elements in: 
   - fail-whale page still work (navigate home, reload, etc.)
   - Support index page still navigate or open the correct destinations
- Skip to main content retains `<a>` semantics via the router scroll behavior; clicking scrolls to `#content`

- The new global `scrollBehavior` in `router/index.js` affects all hash-anchor navigation app-wide, not just skip-to-content
- SAML/support/fail-whale pages: if any of those replaced anchors had `target="_blank"`, that attribute needs to be forwarded through RcButton

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/pkg/rancher-components/src/components/RcButton/RcButton.vue
+++ b/pkg/rancher-components/src/components/RcButton/RcButton.vue
@@ -110,6 +110,10 @@ const focus = () => {
   RcFocusTarget?.value?.focus();
 };
 
+const handleSpace = (event: KeyboardEvent) => {
+  (event.target as HTMLElement).click();
+};
+
 defineExpose({ focus });
 </script>
 
@@ -120,6 +124,7 @@ defineExpose({ focus });
     :role="role"
     :to="to"
     :class="{ ...buttonClass }"
+    @keydown.space.prevent="handleSpace"
   >
     <slot
       v-if="$slots.before || props.leftIcon"
--- a/pkg/rancher-components/src/components/RcButton/RcButton.vue
+++ b/pkg/rancher-components/src/components/RcButton/RcButton.vue
@@ -7,8 +7,7 @@
  *
  * <rc-button variant="primary" @click="doAction">Perform an Action</rc-button>
  */
-import { computed, ref, resolveComponent } from 'vue';
-import type { RouteLocationRaw } from 'vue-router';
+import { computed, ref, resolveComponent, watchEffect } from 'vue';
 import {
   ButtonVariantProps,
   ButtonSizeProps,
@@ -16,6 +15,7 @@ import {
   ButtonSizeNewProps,
   ButtonSize,
   IconProps,
+  NavigationProps,
 } from './types';
 import RcIcon from '@components/RcIcon/RcIcon.vue';
 
@@ -44,16 +44,48 @@ const props = withDefaults(
         ButtonSizeProps &
         ButtonVariantNewProps &
         ButtonSizeNewProps &
-        IconProps & { to?: RouteLocationRaw }
+        IconProps &
+        NavigationProps
     >(),
   {
     size: 'medium',
     to:   undefined,
+    href: undefined,
   }
 );
 
-const tag = computed(() => (props.to ? resolveComponent('RouterLink') : 'button'));
-const role = computed(() => (props.to ? 'link' : 'button'));
+if (process.env.NODE_ENV !== 'production') {
+  watchEffect(() => {
+    if (props.to && props.href) {
+      console.warn('[RcButton] "to" and "href" are mutually exclusive. Provide only one.'); // eslint-disable-line no-console
+    }
+  });
+}
+
+const tag = computed(() => {
+  if (props.to) {
+    return resolveComponent('RouterLink');
+  }
+  if (props.href) {
+    return 'a';
+  }
+
+  return 'button';
+});
+
+const role = computed(() => (props.to || props.href ? 'link' : 'button'));
+
+const linkProps = computed(() => {
+  if (props.to) {
+    return { to: props.to };
+  }
+
+  if (props.href) {
+    return { href: props.href };
+  }
+
+  return {};
+});
 
 const activeVariantClassName = computed(() => {
   if (props.variant === 'multiAction' || props.multiAction) {
@@ -122,7 +154,7 @@ defineExpose({ focus });
     :is="tag"
     ref="RcFocusTarget"
     :role="role"
-    :to="to"
+    v-bind="linkProps"
     :class="{ ...buttonClass }"
     @keydown.space.prevent="handleSpace"
   >
@@ -292,7 +324,7 @@ defineExpose({ focus });
       font-size: 12px;
       min-height: 24px;
 
-      padding: 0 8px;
+      padding: var(--rc-button-padding, 0 8px);
       gap: 8px;
     }
   }
@@ -304,7 +336,7 @@ defineExpose({ focus });
       font-size: 14px;
       min-height: 32px;
 
-      padding: 0 12px;
+      padding: var(--rc-button-padding, 0 12px);
       gap: 8px;
     }
   }
@@ -316,7 +348,7 @@ defineExpose({ focus });
       font-size: 16px;
       min-height: 40px;
 
-      padding: 0 16px;
+      padding: var(--rc-button-padding, 0 16px);
       gap: 12px;
     }
   }
--- a/pkg/rancher-components/src/components/RcButton/RcButton.vue
+++ b/pkg/rancher-components/src/components/RcButton/RcButton.vue
@@ -142,7 +142,19 @@ const focus = () => {
   RcFocusTarget?.value?.focus();
 };
 
+const preventScroll = (event: KeyboardEvent) => {
+  if (tag.value === 'button') {
+    return;
+  }
+
+  event.preventDefault();
+};
+
 const handleSpace = (event: KeyboardEvent) => {
+  if (tag.value === 'button') {
+    return;
+  }
+
   (event.target as HTMLElement).click();
 };
 
@@ -156,7 +168,8 @@ defineExpose({ focus });
     :role="role"
     v-bind="linkProps"
     :class="{ ...buttonClass }"
-    @keydown.space.prevent="handleSpace"
+    @keydown.space="preventScroll"
+    @keyup.space="handleSpace"
   >
     <slot
       v-if="$slots.before || props.leftIcon"
--- a/shell/config/router/index.js
+++ b/shell/config/router/index.js
@@ -9,6 +9,23 @@ export const routerOptions = {
   base:     process.env.routerBase || '/',
   routes:   Routes,
   fallback: false,
+  scrollBehavior(to, from, savedPosition) {
+    // Returning the savedPosition will result in a native-like behavior when
+    // navigating with back/forward buttons
+    if (savedPosition) {
+      return savedPosition;
+    }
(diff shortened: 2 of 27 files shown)


rak-phillip:
This currently covers all of the router-links and anchor tags that I could locate. There could be more to uncover, so I'm keeping this as a draft until I'm confident that I've located all entries. 

With that said, I think that the changes represented in this PR are a good starting point for review.

shell/pages/fail-whale.vue
```diff
@@ -100,25 +101,26 @@ export default {
                 {{ displayError }}
               </h2>
               <p class="mt-20">
-                <a
-                  :href="home"
-                  class="btn role-primary"
+                <rc-button
+                  size="large"
+                  :to="home"
```
codyrancher: The old `<a :href="home">` triggered a full browser reload. This `<rc-button :to="home">` renders as a `RouterLink`, doing SPA navigation instead.

On an error page, a full reload is typically intentional to clear corrupted Vuex/app state. SPA navigation may leave the error state intact and trap the user on this page.

Maybe:  `<rc-button @click="() => window.location.href = home">`

shell/list/management.cattle.io.user.vue
```diff
@@ -159,10 +164,8 @@ export default {
   </div>
 </template>
 
-<style lang="scss">
-  .btn-user-retention {
-    display: flex;
-    gap: 0.25rem;
-    padding: 0;
+<style lang="scss" scoped>
+  .btn.rc-button.btn-medium.btn-user-retention:not(.btn-sm) {
```
codyrancher: This selector (`.btn.rc-button.btn-medium.btn-user-retention:not(.btn-sm)`) is tightly coupled to RcButton's internal class structure. If we want to keep it maybe we can target it using the btn-user-retention class.

There are some slight visual differences between the two which makes me think we may just want to get rid of this styling here. And the hightlight box is not an error it really used to be that large.

**Normal state:**
comparison-users-list

**Hover state:**
comparison-users-list-hover

codyrancher (commented):
Only noticed a couple items. There were a couple that I thought needed to be set to small but I'm realizing that btn-sm is closer to our new medium.

momesgin:
@rak-phillip I found a regression where there's extra space between the icon and the link (it's actually a button) on the footer section of app cards:

Screenshot 2026-05-13 at 8 56 51 AM


I found [this PR](https://github.com/rancher/dashboard/pull/17178) to be the regression cause, an easy solution could be making the CSS selector more specific in `shell/pages/c/_cluster/apps/charts/AppChartCardFooter.vue`:
```
     198 - button.variant-ghost.app-chart-card-footer-button {
      198 + .rc-button.btn.variant-ghost.app-chart-card-footer-button {
```

after:
image

I thought it might be good to include it in this PR, but let me know if you prefer a separate issue and a separate PR.

shell/list/management.cattle.io.user.vue
rak-phillip: > This selector (.btn.rc-button.btn-medium.btn-user-retention:not(.btn-sm)) is tightly coupled to RcButton's internal class structure. If we want to keep it maybe we can target it using the btn-user-retention class.

I think that I have an idea for how we can better support this without requiring increased specificity or `!important`, but it might prompt some larger design discussions. 8e79e7a84159daa92673970765065e932a04179e contains an example. 

> There are some slight visual differences between the two which makes me think we may just want to get rid of this styling here. 

Are the visual differences meaningful? The `padding: 0` is used to retain the left-alignment of the button contents. Maybe we'll get more flexibility with the ghost variant.. The hover states reflect the existing `link` variant for the buttons.. We might want to sync with UX on this one real quick

> And the hightlight box is not an error it really used to be that large.

It's not supposed to be that large. It looks like this regressed at some point. We had fixed this in the past: https://github.com/rancher/dashboard/issues/11338

oboc-sts (commented):
Would love to review, but I need some previews if that's the case. cc @rak-phillip

codyrancher (dismissed):
Other than my one remaining comment it lgtm. I can do another pass if we want to make changes after @oboc-sts takes a look.

shell/list/management.cattle.io.user.vue
rak-phillip: After discussing with @oboc-sts, we came to the conclusion that we are missing a few tokens. Most notably, we should have an `RcLink` component to handle regular text links consistently throughout Rancher Dashboard: #17902

Second, we will need to follow up with updates to the RcButton link variant. I think that we will be able to reduce the amount of overrides if we get both of these into place in the near future.

rak-phillip:
@codyrancher this is ready for one more review. I tried to simplify the style overrides as much as possible, but we will need follow-up work in #17902 to better address the issue.

shell/list/provisioning.cattle.io.cluster.vue
```diff
@@ -289,13 +294,14 @@ export default {
           v-if="canImport"
           #extraActions
         >
-          <router-link
+          <rc-button
+            size="large"
+            class="mr10"
```
codyrancher: Looks like a typo for `mr-10`.

I just forced `isExplorer` to be `false` to look 

Image

shell/pages/support/index.vue
```diff
@@ -155,13 +163,14 @@ export default {
...
-              <a
+              <rc-button
                 v-if="hasAWSSupport"
-                class="mr-5 btn role-secondary btn-sm"
-                :href="supportConfigLink"
+                variant="secondary"
+                class="mr-5"
+                @click="downloadSupportConfig"
               >
                 {{ t('support.suse.access.aws.generateConfig') }}
```
codyrancher, line 166: I think we can switch over to the `href` prop vs having the click handler.

pkg/rancher-components/src/components/RcButton/RcButton.vue
```diff
@@ -118,8 +154,9 @@ defineExpose({ focus });
     :is="tag"
     ref="RcFocusTarget"
     :role="role"
-    :to="to"
+    v-bind="linkProps"
     :class="{ ...buttonClass }"
+    @keydown.space.prevent="handleSpace"
```
codyrancher: This looks like it's broken existing behavior but I may be miss-diagnosing the cause.

Typically it's keyup that triggers the button press. While testing this on `Show Configuration` buttons I notice that if I use space to open the drawer the drawer will open when spacebar is `keydown` but immediately closes when I release spacebar.

https://github.com/user-attachments/assets/32919ed6-0cd8-4c12-99de-510e06eedf22

shell/config/router/index.js
```diff
@@ -9,6 +9,23 @@ export const routerOptions = {
...
+    if (savedPosition) {
+      return savedPosition;
+    }
+
+    // Simulate the "scroll to anchor" behavior when the hash changes.
+    // For example, the skip to content link
+    if (to.hash) {
+      const el = document.getElementById(to.hash.slice(1));
+
+      el?.focus();
```
codyrancher, line 23: I haven't found a case where this breaks yet but I was originally concerned that this was going to cause focus to get thrown to tabs when we link to tabs on detail pages.

Did you intend for focus to run for all route changes or did we want to narrow the scope?

codyrancher (changes requested):

pkg/rancher-components/src/components/RcButton/RcButton.vue
rak-phillip: Right, this has been fixed now. Recent iterations of this had guards for dealing with buttons/links, but that might've been lost in the rebase somehow.. Either way, the original implementation exclusively used `keydown.space`, which is inconsistent with native button behavior. This should be fixed now.

shell/pages/support/index.vue
rak-phillip: I think you are correct. This has been fixed.

shell/list/provisioning.cattle.io.cluster.vue
rak-phillip: Good catch! This has been fixed.

shell/config/router/index.js
rak-phillip: I introduced this change via 73fcd6d081d897979ad002a9762ec2642941e465 to preserve the link semantics for the "Skip to main content" link. In reviewing the tab implementation in `shell/components/Tabbed/index.vue`: 

https://github.com/rancher/dashboard/blob/8eed0d0b42f6dc14e9b4423b8dc3609503b9572b/shell/components/Tabbed/index.vue#L235-L267

I'm finding there's quite a bit of hash change logic contained within the component that I was not expecting. Rather than narrow scope, I'm thinking that this change could have some unintended behavior changes in other areas of Dashboard and should be reverted. I'll take another approach at this. If I can't preserve the link semantics, it's not _too_ big of a deal, but I would like maintain expectations when keyboard users encounter the "Skip to main content" link.

rak-phillip: I still want to preserve the semantics, if possible. a1f22209d60cbf8eda3f59b4396d41694ba7e4d7 attempts to limit this behavior to "Skip to main content" links for the time being. The primary goal is to reduce the chance for changing existing behavior elsewhere. 

I'm willing to concede if and find another route if we find that this approach is overkill for what we are trying to achieve.

codyrancher: Yeah, I'm fine with that. This limits any potential blast radius.

codyrancher (approved):
lgtm

PR #17620, opened 2026-05-12
Remove unused deps: `cookie`, `js-cookie`, `set-cookie-parser`

### LINKED ISSUE #17624: Remove unused cookie dependencies

rak-phillip:
### Summary
This removes unused dependencies from the Dashboard repo.

fixes #17624

- Remove unused deps: `cookie`, `js-cookie`, `set-cookie-parser`

### Technical notes summary

These dependencies are declared in `package.json` (both root and `shell/`), but there aren't any imports in source.

CI should pass all gates.

Perform a smoke test with a quick validation of basic dashboard functionality. 

Without any explicit imports in the Dashboard repo, this should be low risk. If any extensions import from transitive dependencies of shell, they will need to declare these as their own dependencies. 

### Screenshot/Video

NA

### Checklist
- [x] The PR has a Milestone 


nwmac (approved):

nwmac (commented):
@rak-phillip We might need to be careful with Extensions - I notice we have the `novnc` package in our package.json, but we don't use it in dashboard, but it is used in Harvester, so I am wondering if we have some dependencies in our package file that are included for other extensions/UIs? If I look at the Harvester codebase, it does not specify the novnc package, so it would seem it relies on it being in dashboard. We might have the same for these dependencies.

rak-phillip:
@nwmac indeed, to follow up from our conversation earlier - including packages like `novnc` that are unused in dashboard increases bloat and the maintenance burden for our repository; there's also the risk that we bump a package like `novnc`, unknowingly breaking extensions like Harvester that may depend on it. We need to encourage extension authors to list their direct dependencies in the extension's `package.json` and to shift away from relying on transitive dependencies.

PR #17631, opened 2026-05-13
Fix Ingress form editor to handle backend.service.port.name

### LINKED ISSUE #17105: Rancher 2.12.2 Ingress form editor cannot properly edit backend.service.port.name
**Setup**
- Rancher version: 2.12.2
- Rancher UI Extensions: 
- Browser type & version: firefox 

**Describe the bug**
We found an issue in Rancher UI when editing an existing Ingress resource.

The Ingress itself is valid and can be edited/saved successfully in YAML mode.
However, in the Rancher form-based editor, the backend Service port is not correctly rendered when the Ingress uses:

spec.rules[].http.paths[].backend.service.port.name

As a result, the port field appears empty in the form UI, and the Save button is disabled or the form cannot be saved normally.

This looks like a UI/form rendering issue rather than a Kubernetes API issue.

**To Reproduce**
1. Create or apply an Ingress resource with backend.service.port.name, for example:
   port:
     name: http

2. Open the Ingress in Rancher UI.

3. Switch to the form-based edit page.

4. Observe that the backend service port field is not properly populated/rendered.

5. Try to save the form.

**Result**
- The backend service port field is blank or not correctly populated in the form editor.
- The form cannot be saved correctly, or the Save button is disabled.
- YAML mode can still save the same resource successfully.
- 
**Expected Result**
Rancher UI should correctly render and allow editing of Ingress resources using:

backend.service.port.name

The form editor should treat service port name and service port number as equally valid backend definitions.

**Screenshots**

**Additional context**

codyrancher:
### Summary
Fixes #17105

The Ingress form editor only read/wrote `backend.service.port.number`, so Ingresses using `backend.service.port.name` showed a blank port field. This adds support for both across reading, writing, validation, and the port dropdown.

### Technical notes summary
- Added `servicePortNamePath` getter to the Ingress model
- `findAndMapServiceTargets` now includes port names in dropdown options
- `RulePath.vue` and `DefaultBackend.vue` read from both `port.number` and `port.name`, write to the correct path based on type
- Replaced single `portNumberOrName` validator with `portRequired` (presence check) and `portRange` (1-65535), both handling scalar and object inputs
- Fixed `Number.parseInt` fallthrough when port value is falsy (e.g. 0)
- Restored required asterisk on DefaultBackend port fields
- List/detail view falls back to port name when port number is not set

### Reproduction resources

Apply these to create a Service with named ports and two Ingresses (one using `port.name`, one using `port.number`):

```yaml
apiVersion: v1
kind: Service
metadata:
  name: test-web-service
  namespace: default
spec:
  ports:
    - name: http
      port: 80
      targetPort: 8080
    - name: https
      port: 443
      targetPort: 8443
  selector:
    app: test-web
```

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: ingress-port-name
  namespace: default
spec:
  rules:
    - host: name.example.com
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: test-web-service
                port:
                  name: http
```

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: ingress-port-number
  namespace: default
spec:
  rules:
    - host: number.example.com
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: test-web-service
                port:
                  number: 80
```

The bug: editing `ingress-port-name` in the form editor shows a blank port field. Editing `ingress-port-number` works fine.

- Edit `ingress-port-name` and verify the port field shows "http"
- Edit `ingress-port-number` and verify the port field shows "80"
- Create a new Ingress, select `test-web-service`, and verify both port numbers and names appear in the dropdown
- Save with a port name and verify the YAML uses `port.name`; same for `port.number`
- Verify the Default Backend tab works with both formats
- Verify the form blocks saving without a port specified
- Verify port range validation rejects values outside 1-65535

- Ingress form creation and editing (rules and default backend)
- Ingress list page port display
- Form validation for port fields

### Screenshot/Video

**Before (bug):** YAML shows `port.name: http` is defined, but the form editor's port field is empty (red highlight)

https://github.com/user-attachments/assets/12e5e558-c02c-4981-aa8a-4606a8c17cd6

**After (fix):** port name "http" is correctly displayed in the form (green highlight), save succeeds

https://github.com/user-attachments/assets/6a226fa9-4fde-40dd-a12a-87ac2f6a77ba

### Checklist
- [x] The PR has a Milestone

### DIFF
--- a/shell/edit/networking.k8s.io.ingress/DefaultBackend.vue
+++ b/shell/edit/networking.k8s.io.ingress/DefaultBackend.vue
@@ -77,6 +77,7 @@ export default {
   },
   methods: {
     update() {
+      // Fresh object so the old port path (name vs number) doesn't linger.
       const backend = {};
       const parsed = Number.parseInt(this.servicePort);
       const servicePort = Number.isNaN(parsed) ? this.servicePort : parsed;
@@ -124,6 +125,7 @@ export default {
         class="col span-3"
         :style="{'margin-right': '0px'}"
       >
+        <!-- :required drives the asterisk; portRequired doesn't have .name === 'required' -->
         <LabeledInput
           v-if="portOptions.length === 0 || isView"
           v-model:value="servicePort"
--- a/shell/edit/networking.k8s.io.ingress/DefaultBackend.vue
+++ b/shell/edit/networking.k8s.io.ingress/DefaultBackend.vue
@@ -78,7 +78,8 @@ export default {
   methods: {
     update() {
       const backend = {};
-      const servicePort = Number.parseInt(this.servicePort) || this.servicePort;
+      const parsed = Number.parseInt(this.servicePort);
+      const servicePort = Number.isNaN(parsed) ? this.servicePort : parsed;
       const portPath = typeof servicePort === 'number' ? this.value.servicePortPath : this.value.servicePortNamePath;
 
       set(backend, this.value.serviceNamePath, this.serviceName);
@@ -127,6 +128,7 @@ export default {
           v-if="portOptions.length === 0 || isView"
           v-model:value="servicePort"
           :mode="mode"
+          :required="true"
           :label="t('ingress.defaultBackend.port.label')"
           :placeholder="t('ingress.defaultBackend.port.placeholder')"
           :rules="rules.port"
@@ -136,6 +138,7 @@ export default {
           v-else
           v-model:value="servicePort"
           :mode="mode"
+          :required="true"
           :options="portOptions"
           :label="t('ingress.defaultBackend.port.label')"
           :placeholder="t('ingress.defaultBackend.port.placeholder')"
--- a/shell/edit/networking.k8s.io.ingress/DefaultBackend.vue
+++ b/shell/edit/networking.k8s.io.ingress/DefaultBackend.vue
@@ -43,7 +43,9 @@ export default {
     const backend = get(this.value.spec, this.value.defaultBackendPath);
 
     this.serviceName = get(backend, this.value.serviceNamePath) || '';
-    this.servicePort = get(backend, this.value.servicePortPath) || '';
+    this.servicePort = get(backend, this.value.servicePortPath) ||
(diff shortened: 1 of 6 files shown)


aalves08 (changes requested):
There's a scenario where this is failing @codyrancher :

- User edits an Ingress, goes to the Default Backend tab
- Selects a service and picks port http (a named port) from the dropdown
- The DefaultBackend.vue update() method correctly detects that "http" is a string and writes it to service.port.name — leaving service.port.number empty
- User clicks Save
- willSave() fires. It reads service.port.number — which is empty/undefined because the port was stored under service.port.name
- !servicePort evaluates to true
- The hook concludes "the default backend is incomplete" and runs set(this.value.spec, path, null), which deletes the entire default backend
- The Ingress is saved without the default backend the user just configured

codyrancher:
> There's a scenario where this is failing @codyrancher :
> 
> * User edits an Ingress, goes to the Default Backend tab
> * Selects a service and picks port http (a named port) from the dropdown
> * The DefaultBackend.vue update() method correctly detects that "http" is a string and writes it to service.port.name — leaving service.port.number empty
> * User clicks Save
> * willSave() fires. It reads service.port.number — which is empty/undefined because the port was stored under service.port.name
> * !servicePort evaluates to true
> * The hook concludes "the default backend is incomplete" and runs set(this.value.spec, path, null), which deletes the entire default backend
> * The Ingress is saved without the default backend the user just configured

That was a good find. I had to make some changes to validation to get the field requirements correct but it should be fixed now.

aalves08 (approved):
Sorry for the delay and thanks for the fix! LGTM!

codyrancher:
> Sorry for the delay and thanks for the fix! LGTM!

@aalves08  No worries, you're way more on top of things than I am. Thanks for the review!

PR #17632, opened 2026-05-13
CA cert bundle validation

### LINKED ISSUE #16003: [BUG] CA Cert Bundle field in Rancher UI should specify expected content format
Interal Reference: SURE-10984

**Rancher Server Setup**
- Rancher version: v2.6, v2.7
- Installation option (Docker install/Helm Chart): This does not matter.
- Proxy/Cert Details:

- Kubernetes version: N/A
- Cluster Type (Local/Downstream): Downstream K3s/RKE2 (v2prov)

**User Information**
Admin, but this doesn't matter. Anyone that can create/manage a v2prov cluster.

**Describe the bug**
The Rancher UI displays a field for "CA Cert Bundle" when specifying a private registry when creating a new v2prov cluster, but does not specify the expected content type. It should be a base64 encoded copy of the CA PEM. While the UI gives a proper error as the webhook rejects creation due to the incorrect content format, the UI should give a hint that it is expecting a base64 encoded version of the CA cert bundle.

**To Reproduce**
1. Create a new v2prov cluster, and navigate to the UI option for `Registries`, click `Configure advanced containerd mirroring and registry authentication options`, click `Add Registry` and observe that `CA Cert Bundle` is simply a blank field with no hints as to the expected content type. 

**Result**

**Expected Result**
CA Cert Bundle should indicate that it expects a base64 encoded version of the CA.

**Screenshots**

Screenshot 2023-03-23 at 12 13 34 PM

**Additional context**

momesgin:
### Summary
Fixes #16003

Adds validation to the CA Cert Bundle input in the Registry Configuration form. The field now accepts either a PEM certificate or a base64-encoded PEM, shows a tooltip describing the expected format, and blocks form submission when the input is invalid.

- Added validation rules to the CA Cert Bundle `LabeledInput` in `RegistryConfigs.vue` that accept both PEM text (`-----BEGIN ...`) and base64-encoded PEM values.
- Added a tooltip to the CA Cert Bundle input describing the accepted formats.
- Validation state is emitted from `RegistryConfigs` through `Registries/index.vue` up to `rke2.vue`, where it is included in `overallFormValidationPassed` to block form submission.
- The Registries tab now shows an error indicator when validation fails.
- On save, the component detects whether the input is already base64-encoded or plain PEM text and only encodes when necessary, preventing double-encoding.
- Added translation keys for the tooltip and validation error messages.

### Technical notes summary
- A new `isBase64EncodedCert` helper was added to `shell/utils/string.js` alongside the existing `isBase64`. The existing `isBase64` helper has two issues that make it unsuitable for CA bundle validation: (1) it does not handle multiline base64, which is common in certificate output (e.g. `openssl base64` wraps at 76 characters), and (2) it produces false positives on short strings like `"test"` which are technically valid base64. The new helper strips whitespace before checking and enforces a minimum length of 16 characters. Both helpers now share an extracted `BASE64_REGEX` constant.
- The validation event follows the same pattern used by `etcdConfigValid` and `basicsValid` — a boolean emitted from the child component and included in the parent's `overallFormValidationPassed` computed property.

- Create a new cluster and add a registry config with a valid PEM certificate — should pass validation and save successfully.
- Create a new cluster and add a registry config with a valid base64-encoded PEM — should pass validation and save as-is without double-encoding.
- Create a new cluster and add a registry config with multiline base64 (line-wrapped) — should pass validation.
- Enter invalid text (e.g. `not-valid-base64!`) — should show a validation error and block form submission.
- Leave the CA Cert Bundle field empty — should pass validation (field is optional).
- Edit an existing cluster that already has a base64-encoded CA bundle — should load, display, and save without issues.
- Verify the tooltip appears on hover over the CA Cert Bundle input.
- Verify the Registries tab shows an error indicator when validation fails.

- Registry configuration save/load — the save logic now conditionally encodes based on whether the value is already base64. Existing clusters with base64-encoded CA bundles should be unaffected.

### Screenshot/Video

https://github.com/user-attachments/assets/e95ad2a5-9ee4-485a-822b-b95718fb1fb8

### Checklist
- [x] The PR has a Milestone 


shell/edit/provisioning.cattle.io.cluster/tabs/registries/RegistryConfigs.vue
```diff
@@ -192,6 +215,10 @@ export default {
               type="multiline"
               label="CA Cert Bundle"
               :mode="mode"
+              :rules="caBundleRules"
+              :require-dirty="false"
+              :tooltip="t('registryConfig.caBundle.tooltip')"
+              @update:validation="onCaBundleValidation(i, $event)"
```
rak-phillip: Future improvement (out of scope): we typically include a button to upload a file when dealing with certs. I think that we should sync with UX to see if we can benefit from adding something similar to this area in the future.

shell/edit/provisioning.cattle.io.cluster/tabs/registries/RegistryConfigs.vue
```diff
@@ -113,6 +129,13 @@ export default {
       this.$emit('updateConfigs', configs);
     },
 
+    onCaBundleValidation(idx, valid) {
+      this.caBundleValidation[idx] = valid;
+      const allValid = Object.values(this.caBundleValidation).every((v) => v !== false);
+
+      this.$emit('validation-changed', allValid);
+    },
```
rak-phillip: hmmm.. I think that I encountered a validation issue while testing. Steps to reproduce:

1. Create 3 separate registries, each with a valid CA Cert Bundle
2. Choose any given registry (I chose the third), cause CA Cert Bundle Validation to fail
3. Remove the registry with the failing validation

Actual behavior:

Validation remains in a failure state, the save button is disabled.

Expected behavior:

Validation state updates with the removal of the entry, the save button is enabled.

Notes:

While doing a quick debug, I noticed the `caBundleValidation` does not get updated when an item is removed, so the validation remains in a failing state.

rak-phillip (changes requested):

shell/edit/provisioning.cattle.io.cluster/tabs/registries/RegistryConfigs.vue
momesgin: thanks for catching this, it's fixed now

rak-phillip (approved):
LGTM

PR #17663, opened 2026-05-14
Hide "View YAML" when "Show configuration" is present

### LINKED ISSUE #16464: Confusing `View YAML` option given new `Show Configuration` option
**Setup**
- Rancher version: v2.14.0

**Describe the bug**
- Installed Apps entries show two menu items
  - Show Configuration (user can view the yaml for the app in slide in)
  - View Yaml (user can view the yaml for the app in a new screen)
- The two options are confusing

**To Reproduce**
- Apps --> Installed Apps --> click on three dot menu for any

**Result**
- two options that do the same thing

**Expected Result**
- one option to do the thing

**Screenshots**

Image

**Additional context**
- normally we show edit config and edit yaml
- we remove edit config given the special edit/upgrade button
- we replace the edit yaml button with a view yaml button given the resource is not editable (not update link, see canUpdate in resource.class)
- think the solution is now if the Show Configuration option is shown don't fall back on view yaml if the resource cannot be edited
- fwiw showConfiguration came in with https://github.com/rancher/dashboard/pull/14825

momesgin:
### Summary
Fixes #16464 
Remove the redundant "View YAML" action from the action menu when "Show Configuration" is already available and the resource cannot be edited.

- On the Installed Apps page, the action menu previously showed both "Show Configuration" and "View YAML", which were functionally identical for read-only resources since the configuration drawer already includes YAML viewing.
- The "View YAML" action is now hidden when "Show Configuration" is enabled and the resource cannot edit YAML. "Edit YAML" remains unaffected.

### Technical notes summary
- The change is in the base `resource-class.js` `_availableActions` getter, making it generic for all resource types rather than specific to catalog apps.
- The `enabled` condition for the YAML action was updated from `this.canYaml` to `this.canYaml && (this.canEditYaml || !showConfigEnabled)`.
- `showConfigEnabled` was extracted into a variable to avoid duplicating the condition between the "Show Configuration" and YAML actions.
- Unit tests were added covering three scenarios: View YAML hidden when Show Configuration is enabled, Edit YAML still visible when Show Configuration is enabled, and View YAML visible when Show Configuration is disabled.

- Installed Apps page: verify the action menu shows "Show Configuration" but not "View YAML" for installed apps.
- Installed Apps page: verify "Show Configuration" opens the drawer and displays YAML content correctly.
- Resources with edit permissions: verify "Edit YAML" still appears alongside "Show Configuration".
- Resources with `disableResourceDetailDrawer: true`: verify "View YAML" still appears since "Show Configuration" is disabled.

- Action menus across all resource list pages, since the change is in the base `resource-class.js`.
- Any resource type where users relied on the "View YAML" action as a shortcut — it will now be hidden if "Show Configuration" is available and the resource is read-only.

### Checklist
- [x] The PR has a Milestone 


codyrancher (approved):
Simple enough, thanks for finding and fixing the issue.

PR #17725, opened 2026-05-19
Bump read-vault-secrets action, fix flaky edit-fake-cluster e2e test

richard-cox:
- Bump read-vault-secrets action
- This was pinned to a version that did not pin it's own actions

Additionally fix a really flakey edit-fake-cluster test (given changes to listing clusters)

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/cypress/e2e/po/components/action-menu-shell.po.ts
+++ b/cypress/e2e/po/components/action-menu-shell.po.ts
@@ -9,8 +9,24 @@ export default class ActionMenuPo extends ComponentPo {
     return this.self().find('[dropdown-menu-item]').eq(index).click();
   }
 
+  checkNoActionsAvailable(present = true) {
+    if (present) {
+      return this.self().should('contain', 'No actions available');
+    } else {
+      return this.self().should('not.contain', 'No actions available');
+    }
+  }
+
+  atLeastOneActiveMenuItem() {
+    return this.self().find('[dropdown-menu-item]:not([disabled])').should('exist');
+  }
+
   getMenuItem(name: string) {
-    return this.self().get('[dropdown-menu-item]').contains(name);
+    return this.self().contains('[dropdown-menu-item]', name);
+  }
+
+  waitForMenuItem(name: string) {
+    return this.getMenuItem(name).should('be.visible');
   }
 
   menuItemNames() {
--- a/cypress/e2e/po/components/sortable-table.po.ts
+++ b/cypress/e2e/po/components/sortable-table.po.ts
@@ -191,7 +191,7 @@ export default class SortableTablePo extends ComponentPo {
 
   rowActionMenu() {
     // Get the visible dropdown menu - this ensures we only interact with a menu that's actually open
-    return new ActionMenuPo(cy.get('[dropdown-menu-collection]:visible'));
+    return new ActionMenuPo('[dropdown-menu-collection]:visible');
   }
 
   noRowsShouldNotExist() {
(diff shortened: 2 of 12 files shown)


cypress/e2e/tests/pages/manager/edit-fake-cluster.spec.ts
```diff
@@ -32,7 +32,13 @@ describe('Cluster Edit', { tags: ['@manager', '@adminUser'] }, () => {
...
 
       clusterList.waitForPage();
-      cy.wait('@provClusters'); // Wait for request which will populate the drop down
+      cy.intercept({
+        method:   'GET',
+        pathname: '/v1/provisioning.cattle.io.clusters',
+        query:    { pagesize: '100000' }
+      }).as('waitForDropDownInfo'); // Wait for request which will populate the drop down
+
+      cy.wait('@waitForDropDownInfo');
```
mantis-toboggan-md: I am seeing some flakiness in the other test block in this file that also appears tied to menu item availability; could this fix be applied to both tests?

I'm also wondering if it should be applied to other tests that interact with the cluster list's action menu (presumably in a separate PR to avoid more scope creep here). From a quick search it looks like there are a number of references to that action menu, though these are the only two tests that are notably flaky atm

mantis-toboggan-md (changes requested):

cypress/e2e/tests/pages/manager/edit-fake-cluster.spec.ts
richard-cox: sorry, made the pr cover different parts.

thought this would fix the flakiness but it didn't. will be have other goes. if it takes a while i'll spin off into another pr.

ive moved into draft, will mark ready and request re-review when i've actually confirmed the fix

cypress/e2e/po/components/sortable-table.po.ts
```diff
@@ -191,7 +191,7 @@ export default class SortableTablePo extends ComponentPo {
 
   rowActionMenu() {
     // Get the visible dropdown menu - this ensures we only interact with a menu that's actually open
-    return new ActionMenuPo(cy.get('[dropdown-menu-collection]:visible'));
+    return new ActionMenuPo('[dropdown-menu-collection]:visible');
```
richard-cox, line 194: this is one of the bigger fixes.

if the entries of the action menu are updated the dropdown-menu-colllection element is re-rendered.

if we pass through an explicit reference to the element it will be lost on action menu change

so calling self() on it would reference the stale version, not a new version

cypress/e2e/po/pages/cluster-manager/cluster-manager-list.po.ts
```diff
@@ -112,7 +112,9 @@ export default class ClusterManagerListPagePo extends BaseListPagePo {
   }
 
   editCluster(name: string) {
-    this.sortableTable().rowActionMenuOpen(name).getMenuItem('Edit Config').click();
+    const rowMenu = this.sortableTable().rowActionMenuOpen(name);
+
+    return rowMenu.waitForMenuItem('Edit Config').click();
```
richard-cox, line 117: be a bit more defensive

cypress/e2e/po/components/action-menu-shell.po.ts
```diff
@@ -9,8 +9,24 @@ export default class ActionMenuPo extends ComponentPo {
...
+    }
+  }
+
+  atLeastOneActiveMenuItem() {
+    return this.self().find('[dropdown-menu-item]:not([disabled])').should('exist');
+  }
+
   getMenuItem(name: string) {
-    return this.self().get('[dropdown-menu-item]').contains(name);
+    return this.self().contains('[dropdown-menu-item]', name);
```
richard-cox, line 25: this was updated as part of the many attempts at a fix. there's little semantic difference between the two versions.. but decided to keep given it works

richard-cox:
@mantis-toboggan-md ok, ready to review. the e2e tests have run three times and the dodgy edit-fake-cluster test passed each time.

mantis-toboggan-md (approved):
LGTM

PR #17641, opened 2026-05-13
Fix delay when navigating to a cluster

richard-cox:
### Occurred changes and/or fixed issues
- `loadCluster` --> `await dispatch('cluster/unsubscribe');` --> `cleanupTasks.push(waitFor(() => !this.$workers[getters.storeName], 'Worker is destroyed'));`
- the waitFor was not immediately resolved, so waited the default 500ms before checking again
- for some reason we're now not in the immediate world so wait the 500ms
- resolution is to reduce the delay in waitFor to 50ms

### Technical notes summary
- i haven't managed to work out why this is now an issue, so this is almost a work around (but a logical one)

### Areas or cases that should be tested
- home page --> cluster list --> cluster (quicker than before, at least 450ms)

### Areas which could experience regressions
- cluster --> home page --> cluster list (this is as quick as before)

### Checklist
- [x] The PR is linked to an issue and the linked issue has a Milestone, or no issue is needed
- [x] The PR has a Milestone 
- [x] The PR template has been filled out
- [x] The PR has been self reviewed 
- [x] The PR has a reviewer assigned
- [x] The PR has automated tests or clear instructions for manual tests and the linked issue has appropriate QA labels, or tests are not needed
- [x] The PR has reviewed with UX and tested in light and dark mode, or there are no UX changes
- [x] The PR has been reviewed in terms of Accessibility
- [x] The PR has considered, and if applicable tested with, the three Global Roles `Admin`, `Standard User` and `User Base`

PR #17733, opened 2026-05-19
Fix delay when navigating to a cluster

### LINKED ISSUE #17634: Noticeable delay when clicking into a cluster from the home page
From the home page, click on a cluster in the list to explore it. Notice a long delay before the view changes.

This is a regression and the longer delay makes you feel its not worked.

This happens when you change cluster - if you click on a cluster and go back to the home page and chose the same cluster again, it is nice and quick, but selecting another one shows the delay.

Note that changing clusters via the side bar "feels" okay and I don't perceive the same delay I see here.

richard-cox:
### Summary
Fixes #17634

- issue 1
  - `loadCluster` --> `await dispatch('cluster/unsubscribe');` --> `cleanupTasks.push(waitFor(() => !this.$workers[getters.storeName], 'Worker is destroyed'));`
  - the waitFor was not immediately resolved, so waited the default 500ms before checking again
  - for some reason we're now not in the immediate world so wait the 500ms
  - resolution is to reduce the delay in waitFor to 50ms
  - i haven't managed to work out why this is now an issue, so this is almost a work around (but a logical one)
- issue 2
  - race condition for `loadCluster` and feature flag
  - we check if vai enabled (via FF), and if not wait for all mgmt clusters. with vai disabled this is fetched later on
  - however if FF aren't fetched by the time we decided to wait for mgmt clusters.. we wait... and if FF are later fetched and vai is enabled we won't fetch all clusters.... so the UI blocks on blue spinner until the wait fails

### Technical notes summary
- Superseded
  - https://github.com/rancher/dashboard/pull/17641
  - https://github.com/rancher/dashboard/pull/17703

- home page --> cluster list --> cluster (quicker than before, at least 450ms)
- nav to any cluster --> Workloads --> Deployment page --> refresh ~5 times (should always load quicker than 10 seconds...)

- cluster --> home page --> cluster list (this is as quick as before)

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/nav/__tests__/TopLevelMenu.test.ts
+++ b/shell/components/nav/__tests__/TopLevelMenu.test.ts
@@ -727,13 +727,13 @@ describe('topLevelMenu', () => {
                     nameDisplay: 'cluster1',
                     id:          'an-id1',
                     mgmt:        { id: 'an-id1' },
-                    isReady:     true
+                    canExplore:  true
                   },
                   {
                     nameDisplay: 'cluster2',
                     id:          'an-id2',
                     mgmt:        { id: 'an-id2' },
-                    isReady:     true
+                    canExplore:  true
                   }
                 ])
               }
@@ -786,7 +786,7 @@ describe('topLevelMenu', () => {
             nameDisplay: 'cluster1',
             id:          'an-id1',
             mgmt:        { id: 'an-id1' },
-            isReady:     true
+            canExplore:  true
           }
         ]);
 
@@ -814,7 +814,7 @@ describe('topLevelMenu', () => {
             nameDisplay: 'cluster1',
             id:          'an-id1',
             mgmt:        { id: 'an-id1' },
-            isReady:     true
+            canExplore:  true
           }
         ]);
 
--- a/shell/plugins/steve/subscribe.js
+++ b/shell/plugins/steve/subscribe.js
@@ -469,8 +469,10 @@ const sharedActions = {
     const worker = (this.$workers || {})[getters.storeName];
 
     if (worker) {
+      const storeName = getters.storeName;
+
       worker.postMessage({ destroyWorker: true }); // we're only passing the boolean here because the key needs to be something truthy to ensure it's passed on the object.
-      cleanupTasks.push(waitFor(() => !this.$workers[getters.storeName], 'Worker is destroyed'));
+      cleanupTasks.push(waitFor(() => !this.$workers?.[storeName], 'Worker to be destroyed', 30000, 10, true));
     }
 
     if ( socket ) {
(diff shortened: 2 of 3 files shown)


richard-cox:
I re-ran the whole suite of tests 10 times. No single run passed all tests. Many failed on different tests each time

I've dived into them all and i don't _think_ any are now related to this PR. The number of same tests sequentially failing is low and i expect are failing in other PRs but pass on the automatic second run

Ones of note
- manager 
  - edit-fake-cluster - failed 7/10. not related. fixed in another PR
- explorer2
  - cron jobs - failed 2/10. i don't think this is related. looks like a timing issue (failing to fetch / watch a resource when going from cronjob list to detail page
- explorer
  - cluster-dashboard - failed 4/10. i don't think this is related. looks like a permission error given the admin user session isn't correctly restored causing failures in an after block deleting resource
- various
  - rancher-setup - failed 3/10. i don't think this is related. looks like a failure when create the user (could be a timing one, we create and then fetch and it's somewhere around there). test though doesn't handle re-runs well


From run 10 (re-running failed suites)
- 11 - 1 failure - manager / repositories(new!)
- 12 - 1 failure - manager / edit-fake-cluster
- 13 - 1 failure - manager / edit-fake-cluster

shell/components/nav/__tests__/TopLevelMenu.test.ts
```diff
@@ -727,13 +727,13 @@ describe('topLevelMenu', () => {
                     nameDisplay: 'cluster1',
                     id:          'an-id1',
                     mgmt:        { id: 'an-id1' },
-                    isReady:     true
```
richard-cox, line 730: these are changes already in via https://github.com/rancher/dashboard/pull/17739

shell/plugins/steve/subscribe.js
```diff
@@ -469,8 +469,10 @@ const sharedActions = {
     const worker = (this.$workers || {})[getters.storeName];
 
     if (worker) {
+      const storeName = getters.storeName;
+
       worker.postMessage({ destroyWorker: true }); // we're only passing the boolean here because the key needs to be something truthy to ensure it's passed on the object.
-      cleanupTasks.push(waitFor(() => !this.$workers[getters.storeName], 'Worker is destroyed'));
+      cleanupTasks.push(waitFor(() => !this.$workers?.[storeName], 'Worker to be destroyed', 30000, 10, true));
```
richard-cox, line 475: i've made the check
- safer given $workers and getter might get nuked
- limited to 30 seconds, it really shouldn't take that long
- check every 10ms
- log activity (which will help in future)

nwmac (approved):

PR #17736, opened 2026-05-19
tweak kontainer tests to reduce flake

### LINKED ISSUE #17735: kontainer-drivers.spec.ts is flaky
There is a race condition in kontainer-drivers-spec.ts see [run](https://github.com/rancher/dashboard/actions/runs/26046571450/job/76582304762) or [run](https://github.com/rancher/dashboard/actions/runs/26103749027/job/76769183339) - the test activates two drivers using a bulk action, so they are downloaded and activated in parallel, but assertions against the state of the drivers happens sequentially.

mantis-toboggan-md:
### Summary
Fixes #17735

This PR updates kontainer driver tests to remove a race condition. Instead of waiting for an Activating state in one driver before asserting an Activating state in another, which may have already progressed to Active, only the Active state is asserted in the bulk activation step.

The Activating -> Active status updates are still validated in the 'can create new driver' test

also in this pr:
- fixed invalid timeout in 'create new driver' test
- updated the delete driver test to use the action menu 'delete' option instead of re-testing the bulk action delete that is covered by another test
- updated the assertion in the delete driver test to fail when the driver is still present
- update activate driver test to ensure the example driver is actually inactive before the rest of the test runs
- update the deactivate driver test to assert that the driver is actually reported as deactivated

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/cypress/e2e/tests/pages/manager/kontainer-drivers.spec.ts
+++ b/cypress/e2e/tests/pages/manager/kontainer-drivers.spec.ts
@@ -5,7 +5,7 @@ import DeactivateDriverDialogPo from '@/cypress/e2e/po/prompts/deactivateDriverD
 import ClusterManagerListPagePo from '@/cypress/e2e/po/pages/cluster-manager/cluster-manager-list.po';
 import ClusterManagerCreatePagePo from '@/cypress/e2e/po/edit/provisioning.cattle.io.cluster/create/cluster-create.po';
 import PromptRemove from '@/cypress/e2e/po/prompts/promptRemove.po';
-import { MEDIUM_TIMEOUT_OPT } from '@/cypress/support/utils/timeouts';
+import { LONG_TIMEOUT_OPT, MEDIUM_TIMEOUT_OPT } from '@/cypress/support/utils/timeouts';
 
 describe('Kontainer Drivers', { testIsolation: 'off', tags: ['@manager', '@adminUser'] }, () => {
   const driversPage = new KontainerDriversPagePo();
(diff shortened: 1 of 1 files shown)


cypress/e2e/tests/pages/manager/kontainer-drivers.spec.ts
```diff
@@ -231,6 +245,9 @@ describe('Kontainer Drivers', { testIsolation: 'off', tags: ['@manager', '@admin
       expect(isMatch(request.body, requestData)).to.equal(true);
     });
 
+    // wait for driver to be active
+    driversPage.list().details(exampleDriver, 1).should('contain', 'Active');
```
mantis-toboggan-md, line 249: No check for Activating->Active status here because the driver may be updated to Active too quickly, introducing some flakiness

cypress/e2e/tests/pages/manager/kontainer-drivers.spec.ts
```diff
@@ -85,7 +85,7 @@ describe('Kontainer Drivers', { testIsolation: 'off', tags: ['@manager', '@admin
     });
 
     driversPage.list().details(exampleDriver, 1).should('contain', 'Activating');
-    driversPage.list().details(exampleDriver, 1).contains('Active', { timeout: 60000 });
+    driversPage.list().details(exampleDriver, 1).should('contain', 'Active', LONG_TIMEOUT_OPT );
```
marcelofukumoto: **Question**
I think the LONG_TIMEOUT_OPT cannot be used on the .should.
That is for most of the cases here. Are you sure it is actually applying the timeout?
I faced this problem before. (And I think I still have similar codes not working)

I usually change to something like this.
```
    driversPage.list().details(exampleDriver, 1).contains('Active', LONG_TIMEOUT_OPT);
```

I am happy to discuss if you are not sure. I did some tests but we can check together.

mantis-toboggan-md: You're right, thanks for catching this! I got .should('contain' and .contains(... mixed up at some point. I updated the PR to use .should('contain'... instead. Given that the tests pass anyway maybe [these two](https://github.com/rancher/dashboard/pull/17736/changes#diff-7bb266ac4a08da841be33431f8450b92364035a513a6550fa1146db35037358fR148-R149) timeouts are not needed at all? I'm kind of inclined to keep them just to be cautious

marcelofukumoto (approved):
- Removed Activating will reduce the Flakiness but reduce one test as well (ideally). It is usually short lived so I can imagine how that improve flakiness.
- Verifying the state on the UI additionally makes it less flaky as well.
- Added timeout.

LGTM

PR #17737, opened 2026-05-19
Enable resource cards on workload detail pages

### LINKED ISSUE #16375: Resource Detail: Add the Resources Card to the detail pages
The detail pages should reflect the changes outlined in `@confluence /display/CU/Resource+page+redesign`

Image

Waiting on https://github.com/rancher/rancher/issues/51337 to implement.

codyrancher:
### Summary
Fixes #16375

Adds resource cards (Pods, Resources, Insights) to workload detail pages. State counts come from the Steve summary API instead of fetching full resource collections. Each workload type only shows rows matching its tabs (e.g. CronJobs show Jobs only, Jobs show relationships only). Pod detail pages get a Containers row and Insights card. All resource types get relationship-based cards via the base Resource class.

### Technical notes summary

- `fetchResourceSummary` is VAI-only. It returns `undefined` with a console warning when VAI is disabled, so callers must handle that gracefully.
- Relationship rows `_resourcesCardRows` filter out entries without `fromId`/`toId` to avoid phantom "missing" rows from generic type-only references.

- Deployment, DaemonSet, StatefulSet detail pages: should show Pods, Resources (Services, Ingresses, relationships), and Insights cards
- CronJob detail page: Resources card should only show Jobs (no Services/Ingresses)
- Job detail page: Resources card should show relationships only (no Services/Ingresses)
- Pod detail page: Resources card should show Containers with state colors, plus relationship rows; Insights card should appear
- Non-workload resource detail pages: should show relationship-based Resources card

- Workload detail page load (parallel summary requests added during fetch)
- Resource card rendering on types that override `resourcesCardRows`

#### Demo (live update of resource counts)

https://github.com/user-attachments/assets/faa203aa-0741-4be0-9359-de1e663ce97a

#### Deployment
deployment

#### DaemonSet (with Services and Ingresses)
daemonset-resources

#### StatefulSet
statefulset

#### CronJob (with Jobs)
cronjob-with-jobs

#### Job
job

#### Pod
pod

### Checklist
- [x] The PR has a Milestone
- [x] The PR has considered, and if applicable tested with, the three Global Roles Admin, Standard User and User Base


shell/plugins/steve/actions.js
```diff
@@ -221,6 +221,58 @@ export default {
...
+   * Fetch summary counts from the Steve API without loading full resource objects.
+   *
+   * @param {string} type - Resource type (e.g. 'pod', 'service')
+   * @param {Object} opt - Options
+   * @param {string} opt.summaryField - Field to summarize (default: 'metadata.state.name')
+   * @param {Object} opt.filter - Key/value pairs for filter params (e.g. { 'metadata.namespace': 'default' })
+   * @param {string} opt.namespaced - Namespace to scope the query to
+   * @returns {{ count: number, summary: Array<{ property: string, counts: Record<string, number> }> }}
+   */
+  async fetchSummary({ getters, dispatch }, { type, opt = {} }) {
```
richard-cox: we should sync with @marcelofukumoto and make sure we're not duplicating methods of calling the summary endpoint.

think we should also make use of the typing PaginationParamFilter and to construct the url steve-pagination-utils.ts  createParamsForPagination  (convertPaginationParams)

this would also only work if the vai feature is enabled. in theory this will always be on from 2.16 so can choose to omit check?

codyrancher: Yeah, all fair points.

@marcelofukumoto it does look like you're [fetching the summaries](https://github.com/rancher/dashboard/pull/17728/changes#diff-101c008dcf749507255b9b7cfef121f037770aee0c9337641892e502984c41dcR371) on the specific page . 

By any chance is there any changes I could make to this action to make it useful for the workload pages? If you already have something more generic in the works I'd be happy to use that as well.

shell/plugins/steve/actions.js
```diff
@@ -221,6 +223,23 @@ export default {
     }
   },
 
+  async fetchSummary({ getters, dispatch, rootGetters }, { type, opt = {} }) {
```
codyrancher: I am debating whether or not to have this fetch locally as a fallback.

If we want to have a  non-performant fallback I think it's better to have it here so that when we're in a vai always on world we won't have to update all of the call sites. 

However, at least in the scenario I'm using it in it would be perfectly acceptable to just not show the cards if vai isn't on.

shell/components/Resource/Detail/Card/StateCard/composables.ts
```diff
@@ -140,3 +265,43 @@ export function useDefaultWorkloadInsightsCardProps(): StateCardProps {
...
+  filters?: PaginationParamFilter[];
+}
+
+/**
+ * Composable that fetches a resource summary and keeps it updated by watching for resource changes.
+ * Automatically stops watching when the component is unmounted or the scope is disposed.
+ *
+ * Must be called during component setup.
+ */
+export function useResourceSummary(type: string, opt: ResourceSummaryOpt) {
```
codyrancher: This is primarily here to facilitate live updates. It appears that the summary api doesn't work with the existing watching mechanism and this allows us to fairly performantly watch the summary counts.

shell/plugins/steve/actions.js
marcelofukumoto: That is great. I think we can just use your function! Mine is just on the composable of my page.
Not sure who will merge first, but as soon as you get yours merged I will update to use your function.

shell/models/workload.js
```diff
@@ -763,11 +812,69 @@ export default class Workload extends WorkloadService {
...
+            for (const service of services) {
+              if (ingress.metadata?.namespace === this.metadata?.namespace && service?.metadata?.name === targetServiceName) {
+                return true;
+              }
+            }
+          }
+        }
+      } catch (err) {
+        return false;
+      }
```
rak-phillip, line 837: Should we surface the error message in any way before returning false?

rak-phillip (changes requested):
I'm able to put the pods card into an error state with the following deployment manifest:

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-with-service
  namespace: default
spec:
  replicas: 1
  selector:
    matchLabels:
      app: nginx-with-service
  template:
    metadata:
      labels:
        app: nginx-with-service
    spec:
      containers:
        - name: nginx
          image: nginx:stable
    
apiVersion: v1
kind: Service
metadata:
  name: nginx-with-service-svc
  namespace: default
spec:
  selector:
    app: nginx-with-service
  ports:
    - port: 80
      targetPort: 80
```

The pods card enters a state that displays a small green tick for the progress bar and `NaN%`:

Image

This also generates a warning in the console:

```
[Vue warn]: Invalid prop: type check failed for prop "count". Expected Number with value NaN, got String with value "0[object Object]". 
  at <RcCounterBadge count="0[object Object]" type="inactive" > 
  at <StatusRow key=0 color="success" label="Running"  ... > 
  at <Index title="Pods" data-testid="resource-detail-status-card" > 
  at <Index key=0 ref_for=true title="Pods"  ... > 
  at <AsyncComponentWrapper key=0 ref_for=true title="Pods"  ... > 
  at <SpacedRow key=0 class="cards-section" > 
  at <Cards key=1 class="cards-section" resource= 
Object { id: "default/nginx-with-service", type: "apps.deployment", links: {…}, apiVersion: "apps/v1", kind: "Deployment", metadata: {…}, spec: {…}, status: {…}, _summaries: {…}, … }
 
> 
  at <Latest key=0 value= 
Object { id: "default/nginx-with-service", type: "apps.deployment", links: {…}, apiVersion: "apps/v1", kind: "Deployment", metadata: {…}, spec: {…}, status: {…}, _summaries: {…}, … }
 
resourceSubtype=null  ... > 
  at <Index key=0 resource="apps.deployment" value= 
Object { id: "default/nginx-with-service", type: "apps.deployment", links: {…}, apiVersion: "apps/v1", kind: "Deployment", metadata: {…}, spec: {…}, status: {…}, _summaries: {…}, … }
 
 ... > 
  at <Index class="outlet" > 
  at <ClusterResourceNamespaceId class="outlet" onVnodeUnmounted=fn<onVnodeUnmounted> ref=Ref< 
 
> > 
  at <RouterView key="/c/local/explorer/apps.deployment/default/nginx-with-service" class="outlet" > 
 
> > 
  at <RouterView> 
  at <App>
```

shell/models/workload.js
codyrancher: Yeah, I think it's probably reasonable to surface it at least as a console warning so I'll go ahead and add that.

codyrancher:
> I'm able to put the pods card into an error state with the following deployment manifest:

Just adding context from our DMs if others read this. This was caused because the backend changed the counts response format from `{ "running": 16 }` to `{ "running": { "total": 16 } }`.

Fixed in https://github.com/rancher/dashboard/pull/17737/commits/217e88a65b32a98f20c7d50991445abc93b04916

codyrancher:
Hold off on reviewing again. It looks like I need to add support for more backend changes that occurred.

codyrancher:
Alright, we're good again for a review.

shell/models/workload.js
```diff
@@ -586,6 +587,54 @@ export default class Workload extends WorkloadService {
...
+      this.$dispatch('fetchResourceSummary', {
+        type: INGRESS,
+        opt:  { summaryField, filters: [nsFilter] }
+      }),
+    ]);
+
+    summaries.services = services;
+    summaries.ingresses = ingresses;
+
+    set(this, '_summaries', summaries);
```
rak-phillip: I might be missing something, but where is this used?

I initially began to ask this question while reviewing `shell/detail/workload/index.vue`, where we initially invoke `fetchSummaries()`, which brought me to here. I can see further down on line 891 that we access the pods: `const summaryData = this._summaries?.pods || null;`, but what about `services` and `ingresses`?

shell/components/Resource/Detail/Card/StateCard/composables.ts
```diff
@@ -49,15 +68,132 @@ export function useResourceCardRow(label: string, resources: any[], stateColorKe
   };
 }
 
+/**
+ * Builds a ResourceRowProps from summary API response data.
+ * The summary API returns state counts as { property: 'metadata.state.name', counts: { running: 2, error: 1 } }.
+ * This maps those state names to display labels and colors using the same logic as resource models.
+ */
+export function useResourceCardRowFromSummary(label: string, summaryResult: SummaryResult | null | undefined, to?: RouteLocationRaw): ResourceRowProps {
```
rak-phillip: Is this used anywhere?

shell/components/Resource/Detail/Card/StateCard/composables.ts
```diff
@@ -140,3 +276,54 @@ export function useDefaultWorkloadInsightsCardProps(): StateCardProps {
...
+  filters?: PaginationParamFilter[];
+}
+
+/**
+ * Composable that fetches a resource summary and keeps it updated by watching for resource changes.
+ * Automatically stops watching when the component is unmounted or the scope is disposed.
+ *
+ * Must be called during component setup.
+ */
+export function useResourceSummary(type: string, opt: ResourceSummaryOpt) {
```
rak-phillip: Similar question: where is this used?

shell/models/workload.js
codyrancher: No, you're not missing anything. All three comments were valid, during the refactoring the usages got removed.

shell/models/workload.js
```diff
@@ -586,6 +587,38 @@ export default class Workload extends WorkloadService {
...
+
+    const podSelectorStr = this.podSelector;
+
+    summaries.pods = null;
+
+    if (podSelectorStr) {
+      const filters = [nsFilter];
+
+      podSelectorStr.split(',').forEach((part) => {
+        const eqIdx = part.indexOf('=');
```
rak-phillip: I think this can break. It looks like the `!=` operator can also be present:

https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#equality-based-requirement

Do we also need to take set-based requirements into account (`in`, `notin`, `exists`)?

https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#set-based-requirement

richard-cox: We convert the podSelector's labelSelector to a matchExpression and then use that to find matching pods 

```
  async fetchPods() {
    if (this.podMatchExpression) {
      return this.$dispatch('findLabelSelector', {
        type:     POD,
        matching: {
          namespace:     this.metadata.namespace,
          labelSelector: { matchExpressions: this.podMatchExpression },
        },
      });
    }

    return undefined;
  }
```

could try using steve-pagination-utils createParamsForPagination with a mock pagination object, or make `convertLabelSelectorPaginationParams` public and use that

shell/components/Resource/Detail/Card/StateCard/composables.ts
```diff
@@ -49,94 +60,55 @@ export function useResourceCardRow(label: string, resources: any[], stateColorKe
...
-      to:        { hash: '#related' }
-    } : undefined
-  ]);
-
-  const remainingPairs = computed(() => pairs.value.filter((p: (Pairs | undefined)): p is Pairs => p !== undefined));
-
-  const rows = useDefaultResources(remainingPairs);
+  interface Tuple extends Count {
+    color: StateColor;
+  }
```
rak-phillip: nit: This Tuple interface is already declared on line 34.

rak-phillip (dismissed):
I only have a minor nitpick that can be considered optional.

LGTM

shell/components/Resource/Detail/Card/StateCard/composables.ts
codyrancher: Good find. I hoisted it up to the module scope.

rak-phillip (approved):

PR #17738, opened 2026-05-19
Migrate the OIDC auth provider form to use vee-validate

### LINKED ISSUE #17754: Migrate OIDC auth provider form to vee-validate
The OIDC auth provider form validates user input through a single `validationPassed` computed prop that silently prevents form submission. Users receive no indication of which field is blocking submission. The validation method also requires that users guess or read the tooltip in the Scopes section to learn that openid, profile, and email are required.

Because this form does not closely integrate with the existing form-validation mixin, it is a good candidate for migrating to a pure vee-validate/zod implementation.

## Expected behavior

- Each required field should display an inline error tooltip when the user leaves the field blank or enters an invalid value (e.g. a non-URL in the URL field)
- The Scopes list should identify the specific missing scope(s) by name rather than showing a generic "required" message
- The overall form validation should be driven by the same logic that produces the inline errors, so the button state and the error hints are always consistent

## Steps to reproduce

- Navigate to Users & Authentication => select an OIDC auth provider (Keycloak/Generic OIDC/Cognito)
- Tab through the form 
- The save button is disabled but no error message appears on any fields
- Remove the email scope from the Scopes list
- The save button is still disabled but the Scopes section gives no indication of what is wrong

## Acceptance criteria
 
- [x] Blurring an empty required field shows an inline error tooltip identifying the field
- [ ] Entering a non-URL value in the URL, Auth Endpoint, or End Session Endpoint fields shows a format error on blur
- [x] Removing a required scope shows a message naming the missing scope(s)
- [x] The save button's disabled state is consistent with the inline error state
- [ ] All tests continue to pass as a result of this change

rak-phillip:
### Summary
This migrates the custom validation implementation for the OIDC auth provider to a pure vee-validate/zod implementation. 

Fixes #17754

- Migrate computed props to composition api for oidc
- Validate oidc auth provider form with vee-validate/zod
- Add vee-validate to ArrayList
- Fix unit tests

### Technical notes summary

In order to get this to work with `useForm` from vee-validate, we need to find a way to sync values from the form's mixins with the composition API. This is currently accomplished via watchers that exist to sync values with the refs (99078ad9f823b10aeb1b9ac17686355044903a44). I think that the proper, long-term solution will be to migrate away from mixins to the composition API. Take care in reviewing this pattern, there's a higher risk of errors with the current approach.

The validation approach using vee-validate/zod is used to demonstrate validation without the form-validation mixin or bridge in use. More information about the patterns represented in this PR can be found in the vee-validation documentation: https://vee-validate.logaretm.com/v4/guide/composition-api/typed-schema/#zod

Test configuring the Keycloak auth provider. Instructions for getting started can be found in: https://github.com/rancher/ui-internal-tools/tree/main/keycloak

There are risks that the authentication logic has been unintentionally altered with this change. Edge-cases could need additional consideration. For example, what happens if the form passes validation in the UI, but the server rejects the form? 

#### Inline validation errors
image

#### URL validation
image

#### ArrayList validation
image

### Checklist
- [x] The PR has a Milestone 

### DIFF
--- a/shell/components/form/ArrayList.vue
+++ b/shell/components/form/ArrayList.vue
@@ -151,6 +151,15 @@ export default {
       validationMessage: ref(null),
     });
 
+    // Only structural changes (add, delete, paste) mark trigger validation
+    // when invoking update()
+    const isStructuralChange = ref(false);
+
+    const validate = () => {
+      veeHandleBlur(undefined, false);
+      veeValidate();
+    };
+
     /**
      * Cleanup rows and emit input
      */
@@ -169,8 +178,10 @@ export default {
         }
       }
       emit('update:value', out);
-      veeHandleBlur(undefined, false);
-      veeValidate();
+      if (isStructuralChange.value) {
+        validate();
+        isStructuralChange.value = false;
+      }
     };
 
     const lastUpdateWasFromValue = ref(false);
@@ -205,6 +216,8 @@ export default {
       isView,
       update,
       effectiveValidationMessage,
+      isStructuralChange,
+      validate,
     };
   },
 
@@ -239,6 +252,7 @@ export default {
   },
   methods: {
     add() {
+      this.isStructuralChange = true;
       this.rows.push({ value: clone(this.defaultAddValue) });
       if (this.defaultAddValue) {
         this.queueUpdate();
@@ -257,6 +271,7 @@ export default {
      */
     remove(row, index) {
       this.$emit('remove', { row, index });
+      this.isStructuralChange = true;
       removeAt(this.rows, index);
       this.queueUpdate();
     },
@@ -268,6 +283,7 @@ export default {
       event.preventDefault();
       const text = event.clipboardData.getData('text/plain');
 
+      this.isStructuralChange = true;
       if (this.valueMultiline) {
         // Allow to paste multiple lines
         this.rows[index].value = text;
@@ -363,6 +379,7 @@ export default {
                   :aria-label="a11yLabel ? `${a11yLabel} ${t('generic.ariaLabel.genericRow', {index: idx+1})}` : undefined"
                   @paste="onPaste(idx, $event)"
                   @update:value="queueUpdate"
+                  @blur="validate"
                 />
                 <LabeledInput
                   v-else-if="rules.length > 0"
@@ -376,6 +393,7 @@ export default {
                   :aria-label="a11yLabel ? `${a11yLabel} ${t('generic.ariaLabel.genericRow', {index: idx+1})}` : undefined"
                   @paste="onPaste(idx, $event)"
                   @update:value="queueUpdate"
+                  @blur="validate"
                 />
                 <input
                   v-else
@@ -386,6 +404,7 @@ export default {
                   :disabled="isView || disabled"
                   :aria-label="a11yLabel ? `${a11yLabel} ${t('generic.ariaLabel.genericRow', {index: idx+1})}` : undefined"
                   @paste="onPaste(idx, $event)"
+                  @blur="validate"
                 >
               </slot>
             </div>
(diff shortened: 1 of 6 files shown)


shell/edit/auth/oidc.vue
```diff
@@ -180,17 +214,31 @@ export default {
...
+    'customEndpoint.value'(v) {
+      this.customEndpointEnabled = v;
+    },
+
+    requiredScopes: {
+      handler(v) {
+        this.requiredScopesRef = v;
+      },
+      immediate: true,
+    },
```
codyrancher: It looks like `requiredScopes` and `requiredScopesRef` hold the same data. `modelId` is already in setup could we derive scopes directly from `BASE_SCOPES[modelId.value]` and drop both this watcher and the options computed?

shell/components/form/ArrayList.vue
```diff
@@ -144,6 +169,8 @@ export default {
         }
       }
       emit('update:value', out);
+      veeHandleBlur(undefined, false);
```
codyrancher: Won't this fire on every add/remove/keystroke, not blur. Is this intentional, just might violate some expectations down the line.

shell/edit/auth/oidc.vue
rak-phillip: Right you are.. I think this was some leftover logic after I migrated the arrayList to use vee-validate. This should be fixed now.

I also refactored the logic so that it's hopefully easier to understand.

shell/components/form/ArrayList.vue
rak-phillip: I'll look at this a little more closely. The validation behavior was difficult to get correct because `update()` (`queueUpdate()`) is deeply intertwined with all aspects of working with members of the array list.

rak-phillip: The validation pattern should behave similarly to the rest of the inputs on the form with 213eeba8b696eb33581e2251720b45ab676966c3

codyrancher (dismissed):
Good with me on the technical side. 

I think you still wanted to get the opinion of UX, if not I think I'd want to talk about the array list error indicator.

rak-phillip:
> I think you still wanted to get the opinion of UX, if not I think I'd want to talk about the array list error indicator.

Yes, I am scheduling time to discuss this with @oboc-sts next week, as well as a few other design changes that have come through with recent PRs.

rak-phillip:
@codyrancher @oboc-sts we've updated the validation to use a banner instead of the tooltip in the array list component. It now looks like this:

image

codyrancher (approved):
I'm a little surprised we're doing that for all array lists but it works for me. It can always be updated pretty easily.

shell/components/form/ArrayList.vue
```diff
@@ -1,18 +1,22 @@
...
+import { ref, watch, computed, toRef } from 'vue';
 import debounce from 'lodash/debounce';
 import { _EDIT, _VIEW } from '@shell/config/query-params';
 import { removeAt } from '@shell/utils/array';
 import { TextAreaAutoGrow } from '@components/Form/TextArea';
 import { clone } from '@shell/utils/object';
 import { LabeledInput } from '@components/Form/LabeledInput';
+import Banner from '@components/Banner/Banner.vue';
+import { useVeeValidateField } from '@shell/composables/useVeeValidateField';
 const DEFAULT_PROTIP = 'Tip: Paste lines into any list field for easy bulk entry';
```
codyrancher, line 11: Not relevant to your changes but I just noticed we had this hardcoded string.

Edit: I don't expect you to update that in this pr.

oboc-sts:
> I'm a little surprised we're doing that for all array lists but it works for me. It can always be updated pretty easily.

@codyrancher Yeah, considering the fact that we're not 100% controlling the way these array lists are used (different configurations, widths, states, etc)... this approach is the safest one I'd say. Otherwise, in many cases, the "error" icon (previous approach) won't be noticeable or it's going to look like it's floating around without any visual anchor.

oboc-sts (approved):
LGTM! 🚀

